PR #12, opened 2020-05-29
CLI: Improve User Experience

feedmeapples:
- Origin issue: https://github.com/temporalio/proposals/issues

# Summary
Refactor and improve a number of items in CLI, such as semantics, commands discoverability, integration with bash/zsh, rethink global vs command options, improve UI, support config file for default option values

# Motivation
The current CLI was designed seemingly for the internal use and the good UX wasn't among the top priorities in the initial implementation.
The UX lacks in the number of areas such as consistency, semantics, commands discoverability, no integration with bash/zsh, few options are global when they arguably shouldn't be, no default configs file support, UI that could do a better job helping users to "scan" though the `help` output

### DIFF
--- a/cli/0000-cli-improve-ux.md
+++ b/cli/0000-cli-improve-ux.md
@@ -13,7 +13,7 @@ The UX lacks in the number of areas such as consistency, semantics, commands dis
 - Integrate with Bash/zsh
 - Support default configs file
 - Standardize wording
-- Reimplement the short versions of commands
+- Rework the short versions of commands
 - Bring consistent schema for operations on multiple entities
 - Use hyphens instead of underscores
 - Lower priority: Improve help command output/UI
--- a/cli/0000-cli-improve-ux.md
+++ b/cli/0000-cli-improve-ux.md
@@ -1,0 +1,138 @@
+- Start Date: 2020-05-29
+- RFC PR:
+- Issue:
+
+# Summary
+
+The current CLI was designed seemingly for the internal use and the good UX wasn't among the top priorities in the initial implementation.
+The UX lacks in the number of areas such as consistency, semantics, commands discoverability, no integration with bash/zsh, few options are global when they arguably shouldn't be, no default configs file support, UI that could do a better job helping users to "scan" though the `help` output  
+
+### Goals
+
+- Improve commands discoverability
+- Integrate with Bash/zsh
+- Support default configs file
+- Standardize wording
+- Reimplement the short versions of commands
+- Bring consistent schema for operations on multiple entities
+- Use hyphens instead of underscores
+- Lower priority: Improve help command output/UI
+
+### Non-goals
+
+- Creating any kind of new CLI business logic
+- Refactoring or improving CLI code base
+
+# Detailed design
+
+### Improve commands discoverability
+
+Problem: As a user i want to easily find commands i'm interested in. Less nesting leads to easier discovery.
+
+Commands discoverability can be improved by limiting CLI commands to a maximum of 2 levels nesting.
+
+Before 
+``` bash
+tctl workflow activity complete {options}
+```
+After 
+``` bash
+tctl activity complete {options}
+```
+
+### Integrate with Bash/zsh
+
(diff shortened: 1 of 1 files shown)


cli/0000-cli-improve-ux.md
```diff
@@ -0,0 +1,138 @@
...
+Before 
+``` bash
+tctl workflow activity complete {options}
+```
+After 
+``` bash
+tctl activity complete {options}
+```
+
+### Integrate with Bash/zsh
```
rylandg, line 43: This isn't expected to work unless you're compiling the CLI yourself right? Even then, won't it require a script which injects completions in `*_profile`?

cli/0000-cli-improve-ux.md
```diff
@@ -0,0 +1,138 @@
...
+
+- Making a typo in a command/option should output candidates for what user possibly meant to type
+
+### Support default configs file
+
+Problem: As a user i want to be able to change the defaults of some options, such as the default namespace
+Create a default config file that CLI will look into to read them and use in commands.
+If the user passes an option explicitly, ignore the defaults from the config file
+For start, support the following default options through the file:
+ - `--namespace`
```
rylandg, line 65: It's not clear what will actually be added and what you're suggesting is nice to have in the future. I assume the only actual change proposed is 

> For start, support the following default options through the file:


But that's hard to grok.

cli/0000-cli-improve-ux.md
```diff
@@ -0,0 +1,138 @@
...
+
+### Standardize wording
```
rylandg, line 67: Why are some of these missing meanings?

cli/0000-cli-improve-ux.md
```diff
@@ -0,0 +1,138 @@
...
+|remove (rm)	|mark resource as removed (if applicable)	|delete, uninstall, clear	|	|	|
+|uri	|Uri value	|address, link, resource	|	|	|
+
+### Rethink the short versions of commands
+
+Short versions are meant for advanced users. No need to preserve semantic meaning by keepig few extra letters. Shortening the commands/options should make them actually short: 1-2 characters.
+
+### Bring consistent schema for operations on multiple entities
+
+Allow operations to run on multiple entities by providing multiple identifiers or by filtering entities by a query
```
rylandg, line 95: What happened to the user stories?

cli/0000-cli-improve-ux.md
```diff
@@ -0,0 +1,138 @@
...
+
+In the help command output, make the command and option names stand out from the rest of text. This can be achieved by either adding color or bold effect
+
+# Drawbacks
+
+The proposal changes the interaction with Temporal CLI. Existing users who have already learned the current CLI will need to check out the changes
+
+# Adoption strategy
+
+Make annoucements about all of the CLI changes when we release them. Update the CLI documentation. Expect some users to learn from the CLI itself if they miss the annoucements/checking docs
```
rylandg, line 138: Migration guide maybe is a good idea

sw-yx:
to be clear is this something we are only discussing or are we actually working on this right now?

Sushisource:
This looks really great to me. `tctl` does feel a bit amateurish for the sophistication of our product.

Shell completions would be huge, and consistency of passing more things as positional arguments rather than "required options" would also be a massive improvement.

Also have the same question as Shawn. I know @wxing1292 showed some interest in potentially Rustifying tctl. I'm not necessarily advocating for that as just tuning up the existing implementation is almost certainly cheaper, but it's something to consider. The argument parsing libs for Rust are excellent. 

Also I know this seems hilariously trivial but in my experience writing CLIs people really, really love colored output. We should figure out where and how we'd like to work that in. It's super cheap to do and people go bananas for it.

sw-yx:
i made colors the most prominent part of my CLI workshop 😹

https://github.com/sw-yx/egghead-cli-workshop#cli-workshop

feedmeapples:
@Sushisource since Rust has generics this makes customizing print outputs much easier than in Go 👍

PR #13, opened 2020-06-12
PHP SDK proposal

wolfy-j:
# Summary
Describes the PHP SDK for Temporal based on [RoadRunner](https://github.com/spiral/roadrunner)'s long-running workers.

# Status
The draft proposal is ready but at this stage, I require Temporal team and community feedback in order to continue.

# Basic example
```php
class ServiceUsageWorkflow extends Workflow
{
    private int $points = 0;

    /** @Workflow\QueryMethod(name="points") */
    public function getPoints(): int
    {
        return $this->points;
    }

    /** @Workflow\SignalMethod(name="points") */
    public function addPoints(int $points): void
    {
        $this->points += $points;
    }

    /** @Workflow(name="serviceUsage") */
    public function run(string $customerID)
    {
        try {
            while(true) {            
                yield Workflow::newTimer(Time::DAY); 
                
                if ($this->points > 0) {
                    $this->points--;
                } else {           
                    yield $this->activities->chargeDailyUsage($customerID);
                    yield $this->activities->prolongAccess($customerID);
                }
            } 
        } catch (CancellationException $e) {
            yield $this->activities->cancelAccess($customerID);
            yield $this->activities->cancelSubscription($customerID);
        }
    }
}
```

> Annotations format TBD.

# Motivation
Not applicable.

# Detailed design
Please see the design in the attached [MD document](https://github.com/wolfy-j/proposals/blob/master/php/php-rr-sdk.md).

> Please note the RR based approach will make it possible to implement IPC based clients in any other language while utilizing native Golang API.

# Drawbacks
IPC based solution, slower than in-memory communication. However, the local pipe latency should not cause any issues since all the events pipeв from a remote server over TCP anyway.

# Alternatives
None. 

# Adoption strategy
It can be handled by the PHP community.

# How we teach this
We can convert proposals to docs at a later stage.

# Unresolved questions
Syntax approach.

### DIFF
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -303,7 +303,7 @@ $b = Workflow::executeActivity('b', 'values', B::class, [/* options */]);
 yield Workflow::waitAll($a, $b);
 
 // or 
-yield new WaitAny($a, $b);
+yield Workflow::waitAny($a, $b);
 ```
 
 > More atomic blocks can be added down the road.
@@ -321,7 +321,7 @@ yield $timer;
 
 // alternative
 
-$timer = Workflow::timer(1 * Timer::DAY);
+$timer = Workflow::newTimer(1 * Timer::DAY);
 ```
 
 Blocking:
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -315,7 +315,7 @@ public function run(string $input)
 }
 ```  
 
-> The `StalledException` or `StopProcessingException` used to notify the workflow that code must be offloaded from memory
+> The `StalledException` or `PauseException` (TBD) used to notify the workflow that code must be offloaded from memory
 > (cache) without cancelling any activity.
 
 ## Examples
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -291,26 +291,20 @@ public function run(string $input)
 {
     $step = 0;
     try {
-        $result1 = $this->activities->doSomething($input)->get();
+        $resultA = yield $this->activities->doA($input)->get();
         $step++;
         
-        $result2 = $this->activities->doSomethingElse($result1)->get();
+        $resultB = yield $this->activities->doB($resultA)->get();
         $step++;
 
-        $result3 = $this->activities->doSomethingElse2($result2)->get();
-        $step++;
-
-        return $result3;
+        return yield $this->activities->doC($resultB)->get();
     } catch (CancellationException $e) {
         switch ($step) {
-            case 3:
-                yield $this->activities->rollbackSomethingElse2($result3);
-                // no break
             case 2:
-                yield $this->activities->rollbackSomethingElse($result2);
+                yield $this->activities->rollbackB($resultB);
                 // no break
             case 1:
-                yield $this->activities->rollbackSomething($result1);
+                yield $this->activities->rollbackA($resultA);
         }
     
         throw $e;   
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -94,15 +94,6 @@ $worker->run(new RoadRunner\Worker(...));
 > It might be beneficial to provide underlying registry abstraction to simplify the coding of workflows. See the
 > example below.
 
-It should be possible to provide third argument to define custom payload serializer/deserializer:
-
-```php
-$worker->register('my.activity', 'functionName', new JsonMarshaller());
-```
-
-The marshaller must operate on method basis, instead of parameter basis in order to ensure effective reflection caching
-and parameter type validation.
-
 ### Using Annotations
 In a more sophisticated frameworks, some activities can be detected and registered automatically using static analysis
 of codebase and annotations. No common interface required.
@@ -122,18 +113,6 @@ class FileController
 
     // ...
 }
-```
-
-Annotation can be used to configure marshaller/mapper as well:
-
-```php
-/**
- * @Workflow\Activity(
- *    "name"   = "file.download",
- *    "mapper" = ProtocMapper::class 
- * ) 
- */         
-public function processFile(string $file): string; 
 ```
 
 > PHP 8 annotations or alternative registration methods possible as well.
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -14,7 +14,7 @@ which can perfectly align with the workflow execution model.
 - [Workflows](#workflows)
     - [Registration](#registration-1)
     - [Using Annotations](#using-annotations-1)
-    - Syntax and Atomic Blocks
+    - [Syntax and Atomic Blocks](#syntax-and-atomic-blocks)
     - [Queries](#queries)
     - [Signals](#signals)
     - [Sessions](#sessions)
@@ -275,7 +275,51 @@ class Workflow
 > The actual annotation format is TBD as it irrelevant to the implementation and might vary from framework to framework. 
 
 ### Syntax and Atomic Blocks
---
+The building blocks of workflow is `yield` commands which might return actual value or a promise for this value. Every
+command wrapped via dedicated class:
+
+```php
+function run()
+{
+    yield new Activity('name', 'values', [/* options */]);
+}
+```
+
+Some commands might automatically return sub-commands. For example blocking activity execution:
+
+```php
+$result = yield (new Activity('name', 'values', [/* options */]))->get();
+```
+
+To wait for multiple activities at the same time we can introduce higher level promises such as:
+
+```php
+$a = yield new Activity('a', 'values', [/* options */]);
+$b = yield new Activity('b', 'values', [/* options */]);
+
+yield new WaitAll($a, $b);
+
+// or 
+yield new WaitAny($a, $b);
+```
+
+> More atomic blocks can be added down the road.
+
+#### Syntax Sugar
+It should be possible to mock `new Activity` and other commands with shorted methods from abstract workflow implementation:
+
+```php
+class MyWorkflow extends Workflow 
+{
+    public function run()
+    {
+        $result = yield $this->activities->doSomething()->get();
+        yield $this->sleep(1);
+
+        // ...
+    }
+}
+``` 
 
 ### Queries
 Workflow queries can be easily implemented using dedicated query method of workflow. The state of the workflow can be 
@@ -333,12 +377,50 @@ public function querySomething(string $world): string
 Unlike queries the signals are fist-class citizen of actual workflow code. Unlike activities the signal can be triggered
 at any moment, without prior expectation by workflow. 
 
+Signal method can be defined explicitly:
 
-EXAMPLE
+```php
+class UploadWorkflow extends Workflow\Workflow
+{
+    /** @Temporal\SignalMethod(name="signal") */
+    public function handleSignal($input):void
+    {
+    
+    }
 
-SELECT EXAMPLE
+    public function run(string $input): string
+    {
+        
+    }
+}
+```
 
-TODO ALTERNATIVE DEFINITION USING METHOD?
+Or at runtime:
+
+```php
+public function run(string $input): string
+{
+    yield new SignalHandler('name', function($input){
+    
+    });       
+}
+```
+
+> Need Temporal authors feedback.
+
+#### Handle Signals
+Incoming signals can be handled during any workflow `yield` call. Workflow should allow to explicitly wait for signal:
+
+```php
+public function run(string $input): string
+{
+    yield new SignalHandler('name', function($input){
+        // do something
+    });       
+
+    yield new WaitSignal('name');
+}
+```
 
 ### Sessions
 Sessions can be created as sub-activity group, using similar approach as `withTimeout` and other methods. However,
@@ -363,6 +445,18 @@ Workflows must avoid calling SPL functions `time()` and `date()`. Context method
 
 ```php
 $ctx->getNow(); //DateTimeImmutable object.
+```
+
+#### Sleeps
+Workflow can sleep using simple command returning promise:
+
+```php
+yield $this->sleep(1)->wait();
+
+// or as promise
+$t = yield $this->sleep(1);
+
+yield new AnyOf($t, $this->activities->doSomething());
 ```
 
 ### Side Effects
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -12,9 +12,12 @@ which can perfectly align with the workflow execution model.
     - Process Isolation
 - Workflows
     - Registration
-    - Syntax
+    - Using Annotations
+    - Syntax and Atomic Blocks
+    - Syntax Sugar
     - Queries
     - Sessions
+    - Deterministic Time
 - Service RPC
 - Implementation
     - Features
@@ -174,27 +177,53 @@ Each yield command will return Promise/Future object which can later be referenc
 look as following:
 
 ```php
+use Temporal\Workflow;
+
 class UploadWorkflow
 {
-    public function run($in) // context to get time and etc?
+    public function run(Workflow\ContextInterface $ctx, string $input): string
     {
-        $a = yield $this->downloadURL('x', 'a');
-        $b = yield $this->process($in);
+        $promise = yield new Workflow\ExecuteActivy(
+            'activityName',
+            $input, 
+            100 // startToFinishTimeout, etc. 
+        );        
+    
+        $result = yield $promise->get(Obj::class);
 
-        // syntax sugar?
-        yield new Workflow\WaitAll($a, $b);
-
-        // blocking Get atomic?
-        return yield ($this->sendEmail($result))->get(new Obj());
+        return $result->getValue();
     }
 }
 ```
 
-> No common interface is required.
+Read more about proposed syntax approach and syntax sugar down below. The given example can be simplified into:
+
+```php
+use Temporal\Workflow;
+
+class UploadWorkflow extends Workflow\Workflow
+{
+    public function run(string $input): string
+    {
+        $result = yield $this->activities
+            ->withStartToFinish(100)
+            ->activityName($input)
+            ->get(Obj::class);
+
+        return $result->getValue();
+    }
+}
+```
+
+> No common interface required.
 
 ### Registration
 
-### Syntax
+### Using Annotations
+
+### Syntax and Atomic Blocks
+
+### Syntax Sugar
 
 ### Queries
 
@@ -202,6 +231,12 @@ class UploadWorkflow
 
 ### Sessions
 
+### Deterministic Time
+Workflows must avoid calling SPL functions `time()` and `date()`. Context method must be used instead:
+
+```php
+$ctx->getNow(); //DateTimeImmutable object.
+```
 
 ### Service RPC
 The Temporal service SDK in PHP can be written as simple RPC bridge to Golang SDK. This section is pretty straight forward
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -17,17 +17,6 @@ which can perfectly align with workflow execution model.
     - Current Milestones
 - Development
 
-## TODO
-- rpc
-- external activities
-- sessions
-- how to schedule activities
-- signals 
-- queries
-
-## Service RPC
-
-
 ## Activities
 Similar to other SDKs, the Activity implementation in PHP won't have any feasible limitations. Is it possible to invoke 
 any callable function, object method or anonymous function. The invocation arguments can be hydrated based on provided
@@ -190,6 +179,12 @@ The system must limit to only one concurrent execution per worker, which makes p
 libraries without changes. 
 
 ## Workflows
+
+- rpc
+- sessions
+- how to schedule activities
+- signals 
+- queries
 :)
 
 ```php
@@ -225,6 +220,8 @@ public function doSomething(Service $service)
 
 // ...
 ```
+
+> Same client can be used to control the state and heartbeats of external activities. 
 
 ## Implementation Details
 We propose to implement the SDK as module of [RoadRunner](https://github.com/spiral/roadrunner) application server. The server
--- a/php/php-rr-sdk.md
+++ b/php/php-rr-sdk.md
@@ -456,7 +456,10 @@ class ServiceUsageWorkflow extends Workflow
                 );
 
                 // adding points request                
-                if ($this->isSignal($case) && $case->getName() === "points"){
+                if (
+                    $this->isSignal($case) 
+                    && $case->getName() === "points"
+                ){
                     $this->points += $case->get();
                     continue;
                 } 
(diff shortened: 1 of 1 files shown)


php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+}
+```
+
+Heartbeat example:
+
+```php
+use Temporal\Workflow\ContextInterface;
+use Temporal\Workflow\Activity;
+
+// parameter order does not matter
```
mfateev, line 55: Is it possible to use some sort of global/thread local variable to avoid passing context explicitly or it is PHP pattern [like in Go](https://blog.golang.org/context)?

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+    $downloader->onProgress(function($progress) use($ctx){
+        // subject of discussion
+        Activity::heartbeat($ctx, $progress);
+    });
+
+    return $downloader->download($url);
+}
+```
+
+Alternatively we can embed the heartbeat method directly into `ContextInterface` or `ActivityContextInterface`:
```
mfateev, line 68: This is the path we took in Java

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+
+```php
+$logger = new ActivityLogger($ctx);
+$logger->debug('Hello!');
+
+// or directly
+$ctx->log('debug', 'message', ['context']);
+```
+
+### Payloads
```
mfateev, line 144: Is it possible to serialize them to JSON seamlessly? Or make serialization pluggable as it is done in Go and Java?

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+one method (query, primary method) meta declation is required. 
+
+```php
+$worker = new WorkflowWorker();
+
+$workflow = (new WorkflowDeclaration('name'))
+                ->setClass(WofkflowClass::class)
+                ->setHandlerMethod('run')
+                ->addQueryHandler('name', 'method-name')
+                ->addSignalHandler('name', 'method-name');
```
mfateev, line 236: This looks weird as which methods are which is property of the workflow and not how it is registered. What about doing dynamic registration of query and signal handlers from the workflow code?

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+#### Syntax Sugar
+It should be possible to mock `new Activity` and other commands with shorted methods from abstract workflow implementation:
+
+```php
+class MyWorkflow extends Workflow 
+{
+    public function run()
+    {
+        $result = yield $this->activities->doSomething()->get();
+        yield $this->sleep(1);
```
mfateev, line 317: What about adding sleep to the workflow context or use similar approach as in activities:

`yield Workflow::sleep(ctx, 1);`

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+
+class UploadWorkflow
+{
+    public function run(Workflow\ContextInterface $ctx, string $input): string
+    {
+        $promise = yield new Workflow\ExecuteActivy(
+            'activityName',
+            $input, 
+            100 // startToFinishTimeout, etc. 
+        );
```
mfateev, line 195: I believe any activity invocation should get ctx as a parameter.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+
+An alternative approach can be based on registering the callback function during the workflow execution:
+
+```php
+yield new QueryHandler('something', function() use (&$step) {
+    return $step;
+});
+```
+
+> Such an approach is more suitable for dynamic workflows. The principle of work is identical to pre-registered queries.
```
mfateev, line 393: In Java we support both. Dynamic approach for both query and signals is needed for libraries which don't have access to the top level workflow object.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+Or at runtime:
+
+```php
+public function run(string $input): string
+{
+    yield new SignalHandler('name', function($input){
+    
+    });       
+}
+```
```
mfateev, line 440: Open question is what happens if you register more than one handler for the given signal type. Will all of them be called? Or double registration is rejected?

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+        // do something
+    });       
+
+    yield new WaitSignal('name');
+}
+```
+
+### Sessions
+Sessions can be created as sub-activity group, using similar approach as `withTimeout` and other methods. However,
+the explicit yield required.
```
mfateev, line 460: An alternative approach is used by Go client is to associate a session with ctx object. Then all other operations that reference the session context are considered part of a session.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+    
+    return $result;
+}
+```
+
+### Deterministic Time
+Workflows must avoid calling SPL functions `time()` and `date()`. Context method must be used instead:
+
+```php
+$ctx->getNow(); //DateTimeImmutable object.
```
mfateev, line 480: why not ctx->time() and ctx->date()?

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+Exception Type | Parent | Comment 
+--- | --- | ---
+EngineException | - | Low level exception, usually local.
+StalledException | EngineException | Triggered to indicate the workflow processing should stop (see below).
+WorkflowException | EngineException | Generic workflow error (like activity can not be created and etc)
+TimeoutException | WorkflowException | Workflow timed out (must contain information about actual timeout type). 
+TerminateException | WorkflowException | Workflow needs to be terminated.
+ActivityException | WorkflowException | Processing errors.
+ActivityTimeoutException | ActivityException | Activity processing timeouted (must contain information about actual timeout type). 
+
```
mfateev, line 532: We recently redefined the way failures are defined in the service proto APIs to allow exception chaining across process boundaries in different languages. This deserves a separate discussion.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,692 @@
...
+> The exception tree is a subject of discussion. 
+
+### Termination and Cancel Requests
+The termination, cancel and stalled requested will be supplied to the workflow in a form of exceptions on `yield` call.
+Similar to activity errors workflow can handle these exceptions and perform a proper rollback.
```
mfateev, line 537: Termination is like "kill -9" it doesn't let workflow code to handle it.
Cancellation requires a separate discussion. What is the standard way in PHP to cancel things?

wolfy-j: There are not standard to cancel things, the process just dies. But since the workflow coordinator sits in another process we have to properly pass some signal to the underlying PHP process to clean up as well. 

TerminateException, in this case, allows us to gracefully stop the processing. It's not intended for handing, but rather to stop the coroutine without affecting other workflows in memory.

php/php-rr-sdk.md
wolfy-j: I'm fine leaving it out of this proposal for now. But I have to outline that `yield` can cause the exceptions of different types.

php/php-rr-sdk.md
wolfy-j: The time() function in PHP returns UNIX timestamp (int), date() returns formatted date. None of them are useful for the time manipulation. 

The go-to solution is to use `DateTime` interfaces which allow safe date/time transformations. So `getNow()` or `now()` are less ambiguous for developers since it does not create a bridge for native functions. However, we can add them as well without any issue, it could be:

```php
$ctx->now(); // date time
$ctx->time(); // Unix timestamp
$ctx->date('D-Y-m') // formatted date

// last two equals to 
$ctx->now()->getTimestamp();
$ctx->now()->format('D-Y-m');
```

php/php-rr-sdk.md
wolfy-j: Yes and no. Practically we are going to run all workflows without one thread, so some _hidden_ context switching will be required. However, we can pass the context as part of the workflow constructor and mock it using easier abstraction.

To be honest, context is not really needed here since each of the workflows will be wrapped with a `yield` handler at the top (which manages the context).

So, technically, we can replace context with something like that:

```php
// was
$ctx->now();

// can be 
$now = yield new Now();

// or 
$now = yield $this->now();
```

wolfy-j: However, I think passing context is easier for 2 reasons:

1) It's more explicit
2) We can handle context passing in the constructor (since it's called only once) and hide the actual context from user code with base class implementation.

php/php-rr-sdk.md
wolfy-j: I like it too.

php/php-rr-sdk.md
wolfy-j: Absolutely, it will be controlled on the coordinator end so we can change serialization mechanisms at any time. Payload reqs are the same as in other languages - be serializable, do not carry resources and etc.

php/php-rr-sdk.md
wolfy-j: Dynamic registration is possible as well. This is just an example of explicit registration outside of the workflow. This is needed since PHP (at the moment) does not provide native support for annotations and attributes, so _something_ needs to parse and provide them to the Temporal engine. 

Consider this as deep internal logic which most users will never see. It's only needed to make the actual workflow code to as bootstrap-less and verbose as possible.

php/php-rr-sdk.md
wolfy-j: Sure. 

Under the hood, each yield command should be handled by the local workflow coordinator so the actual syntax to generate such command does not matter:

```php
yield $this->sleep(1); // same
yield new Sleep(1); // same
yield Workflow::sleep(1); // same
```

We do not need to pass $ctx since the coordinator can only process one workflow command at once (there are no threads in PHP for that). This means that the coordinator always knows what workflow yield command and what is the context. So, essentially, it's always passed inexplicably.

php/php-rr-sdk.md
wolfy-j: See above, the $ctx always passed inexplicably due to the model of how cooperative multitasking works in PHP. So we can make this DSL smaller and easier to understand.

php/php-rr-sdk.md
wolfy-j: Yep, both should be possible at the same time. Not allowing dynamic workflows in PHP would be a weird decision.

php/php-rr-sdk.md
wolfy-j: That's a good question... practically I'm seeing now issues handling query in multiple handlers. Conceptually it might cause some side effects. However, it can also be used to build some more complex things...

php/php-rr-sdk.md
wolfy-j: See above. All `yield` commands associated with $ctx by default. So in this case we are altering the default values of generated (`yield` ed) commands.  The problem with sessions is that we have to explicitly open it via command, so we can't use `with` but rather `yield`.

php/php-rr-sdk.md
mfateev: I see. I would use a different name than Terminate just because Temporal service already uses Terminate for killing workflows.

wolfy-j: It can be variations of: 
PauseException
OffloadException

Since user don't need to handle it we can name a lot of freedom in naming it.

wolfy-j:
+ Time via static global variable which switches when context changes

wolfy-j:
unit-testing is out of scope

wolfy-j:
+ REMOVE LOGGING

wolfy-j:
- replace yield for DECLARATION (non-blocking) with local queue via global context

wolfy-j:
- support previous exception in php

wolfy-j:
- out of cache ex: destroy exeption

rylandg:
I'm going to do my best to look this over thoroughly this week (def not fluent in PHP). Looked things over at a high level already and it looks really high quality work. Seriously appreciate the amount of time and effort you put in here.

wolfy-j:
Thank you. I left comments after talking to Max, will address them next week.

wolfy-j:
@mfateev @rylandg I did some adjustments. Let me know if I missed something.

php/php-rr-sdk.md
```diff
@@ -186,41 +169,46 @@ use Temporal\Workflow;
...
+    public function run(string $input): string
     {
-        $promise = yield new Workflow\ExecuteActivy(
+        $promise = new Workflow\ExecuteActivy(
             'activityName',
             $input, 
-            100 // startToFinishTimeout, etc. 
+            [
+                ExecuteActivy::START_TO_FINISH_TIMEOUT   => 100,        
+                ExecuteActivy::SCHEDULE_TO_START_TIMEOUT => 10
```
mfateev, line 179: I would stick to the same names the service uses:

SCHEDULE_TO_START
START_TO_CLOSE
SCHEDULE_TO_CLOSE

php/php-rr-sdk.md
```diff
@@ -186,41 +169,46 @@ use Temporal\Workflow;
...
+            ]       
         );        
     
+        // blocking operation to get the activity result
```
mfateev, line 183: It is kind of weird to have both get and yield. May be pass object type as a parameter to the ExecuteActivity and then just yield on the result? Something like:

```
$promise = new Workflow\ExecuteActivy(
            'activityName',
            $input,
            Obj::class, //result type
             [
                ExecuteActivy::START_TO_FINISH_TIMEOUT   => 100,        
                ExecuteActivy::SCHEDULE_TO_START_TIMEOUT => 10
              ]       
        ); 

  $result = yield $promise; // No get
  return $result; // not clear why getValue was needed
```

php/php-rr-sdk.md
```diff
@@ -186,41 +169,46 @@ use Temporal\Workflow;
...
+> construction. 
+
+### Context
+In order to ensure easy and reliable access to workflow context (including library implementations) we introduce static
+methods via `Workflow` class. Since only one workflow can be executed at one moment of time we can safely change the
+context before passing control to the workflow:
 
 ```php
-use Temporal\Workflow;
+// explicit
```
mfateev, line 201: ???

php/php-rr-sdk.md
```diff
@@ -186,41 +169,46 @@ use Temporal\Workflow;
...
-            ->get(Obj::class);
+// implicit 
+Workflow::ctx()->now();
+Workflow::ctx()->isReplying();
 
-        return $result->getValue();
-    }
-}
+// sugar
+Workflow::executeActivity();
```
mfateev, line 208: I like more

Workflow::now();
Workflow::isReplaying();

php/php-rr-sdk.md
```diff
@@ -238,6 +226,8 @@ $workflow = (new WorkflowDeclaration('name'))
 $worker->addWorkflow($workflow);
 ```
 
+> Is it only possible to set one query and signal handler per channel per workflow.
```
mfateev, line 229: Nit: one query per query type and one signal handle per signal name

php/php-rr-sdk.md
```diff
@@ -281,11 +271,13 @@ command wrapped via dedicated class:
 ```php
 function run()
 {
-    yield new Activity('name', 'values', [/* options */]);
+    $activity = new Activity('name', 'values', [/* options */]);
```
mfateev, line 274: Is "new Activity" required by PHP? Why not 

$activity = Workflow::executeActivity('name', ...);
$result = yield $activity;

php/php-rr-sdk.md
```diff
@@ -294,40 +286,25 @@ $result = yield (new Activity('name', 'values', [/* options */]))->get();
...
-
 #### Timers
 Workflow timers can operate as promise or blocking operation:
 
 ```php
-$timer = yield $this->timer(1 * Time::DAY);
+$timer = new Timer(1 * Timer::DAY);
+
+// or
+$timer = $this->timer(1 * Timer::DAY);
```
mfateev, line 307: Workflow::newTimer

php/php-rr-sdk.md
```diff
@@ -294,40 +286,25 @@ $result = yield (new Activity('name', 'values', [/* options */]))->get();
...
 
 yield $timer->wait();
```
mfateev, line 309: yield $timer;

php/php-rr-sdk.md
```diff
@@ -337,15 +314,15 @@ Blocking:
...
 ```
 
-Can be comfined with Promises:
+Can be combined with Promises:
 
 ```php
-$a = yield $this->activities->doSomething(...);
-$t = yield $this->timer(1 * Time::MINUTE);
+$a = $this->activities->doSomething(...);
+$t = $this->timer(1 * Time::MINUTE);
```
mfateev, line 325: i"m not sure what $this is in this case.

php/php-rr-sdk.md
```diff
@@ -372,7 +351,7 @@ class DemoWorkflow extends Workflow\Workflow
 
     public function run(string $input): string
     {
-        yield $this->activities->doSomething($input);
+        yield $this->activities->doSomething($input)->get();
```
mfateev, line 354: remote ->get

php/php-rr-sdk.md
wolfy-j: I like this approach, we can always add more sugar later via activity builder

php/php-rr-sdk.md
wolfy-j: agreed since we moved to static context per thread

php/php-rr-sdk.md
wolfy-j: Sure. It's still new Activity behind the hood.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,680 @@
...
+    yield new WaitSignal('name');
+}
+```
+
+### Logging
+The workflow logging does not differ from any other parts of the application, the only exception is the need to exclude
+logs triggered while workflow is replaying. We propose to use mocked PSR-3 logger for this purposes:
+
+```php
+$logger = new NonReplayingLogger(Workflow::ctx(), $parentLogger); // it is possible to omit ::ctx()
```
mfateev, line 459: I would omit the `ctx` here as the NonReplayingLogger constructor can access it directly.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,680 @@
...
+```
+
+### Sessions
```
mfateev, line 462: I think we will need to make another pass on session API after all other features are implemented

php/php-rr-sdk.md
```diff
@@ -0,0 +1,680 @@
...
+                yield $this->activities->rollbackB($resultB);
+                // no break
+            case 1:
+                yield $this->activities->rollbackA($resultA);
+        }
+    
+        throw $e;   
+    }
+}
+```
```
mfateev, line 554: There is a need to give workflow authors an ability to cancel parts of a workflow. This would require introduction of some form of CancellationScope. See the [Java SDK version of it](https://github.com/temporalio/java-sdk/blob/112dd5d31cd7e184730e471f3908faa71644da3c/src/main/java/io/temporal/workflow/Workflow.java#L580).

php/php-rr-sdk.md
```diff
@@ -0,0 +1,680 @@
...
+## Examples
+Following examples demonstrates features described above in a more realistic scenarios.
+
+### Simple Subscription Workflow
+```php
+class SubscriptionWorkflow extends Workflow
+{
+    public function subscriptionWorkflow(string $customerID)
+    {
+        yield $this->activities->onboardFreeTrial($customerID);
```
mfateev, line 565: Add code that initializes `activities` field.

php/php-rr-sdk.md
```diff
@@ -0,0 +1,680 @@
...
+        
+        try {
+            yield Wofkflow::sleep(60 * Time::DAY);
```
mfateev, line 568: replace to Workflow::newTimer

php/php-rr-sdk.md
wolfy-j: See About Activities Stub

php/php-rr-sdk.md
```diff
@@ -456,7 +436,7 @@ The workflow logging does not differ from any other parts of the application, th
 logs triggered while workflow is replaying. We propose to use mocked PSR-3 logger for this purposes:
 
 ```php
-$logger = new NonReplayingLogger(Workflow::ctx(), $parentLogger); // it is possible to omit ::ctx()
+$logger = Workflow::nonReplyingLogger($parentLogger);
```
mfateev, line 439: nonReplayingLogger

Maybe Workflow::getLogger?

php/php-rr-sdk.md
```diff
@@ -466,13 +446,13 @@ the explicit yield required.
 ```php
 public function run(string $input): string
 {
-    $session = new Session([
-        Session::START_TO_FINISH_TIMEOUT => 300,
+    $session = Workflow::newSession([
+        Session::START_TO_FINISH => 300,
```
mfateev, line 450: I would use the same names as the already existing Go impl: 	SessionOptions struct {
		ExecutionTimeout time.Duration
		CreationTimeout  time.Duration
		HeartbeatTimeout time.Duration
	}

php/php-rr-sdk.md
```diff
@@ -466,13 +446,13 @@ the explicit yield required.
...
         // ... 
     ]);
     
-    $a1 = $session->executeActivity(new ExecuteActivity('activity1'));
-    $a2 = $session->executeActivity(new ExecuteActivity('activity2'));
+    $a1 = $session->executeActivity(Workflow::executeActivity('activity1', ...));
+    $a2 = $session->executeActivity(Workflow::executeActivity('activity2', ...));
```
mfateev, line 455: $session->executeActivity('activity2', ...);

php/php-rr-sdk.md
```diff
@@ -556,6 +538,25 @@ public function run(string $input): string
...
+method using (no context binding required).
+
+```php
+class SubscriptionWorkflow extends Workflow
+{
+    private ActivitiesInterface $activities;
+
+    public function __construct() 
+    {
+        $this->activities = new ActivityBuilder();
```
mfateev, line 553: Consider using Java syntax:

`$this->activities = Workflow::newActivityStub(options);`

php/php-rr-sdk.md
wolfy-j: These names are not consistent with Java implementation.

php/php-rr-sdk.md
wolfy-j: Both actually should work in my opinion.

php/php-rr-sdk.md
wolfy-j: No, cos user must supply his own psr3 logger. Workflow does not create logger, it only proxies needed calls to the outer one.

php/php-rr-sdk.md
wolfy-j: Hmmm, this one I need more clarification on.

PR #15, opened 2020-12-13
NodeJS SDK proposal

bergundy:
# Summary

Proposal implementation for the TypeScript NodeJS SDK.


node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+import { ActivityContext } from 'temporal-sdk';
+import axios from 'axios';
+
+export async function httpGet(ctx: ActivityContext, url: string) {
+  const response = await axios.get(url, {
+    onDownloadProgress: (progressEvent) => {
+      const total = parseFloat(progressEvent.currentTarget.responseHeaders['Content-Length']);
+      const current = progressEvent.currentTarget.response.length;
+      const progress = Math.floor(current / total * 100);
+      ctx.heatbeat(progress);
```
sw-yx, line 65:
```suggestion
      ctx.heartbeat(progress);
```

VladimirMilenko:
Hey @bergundy. 
Correct me if I'm wrong, but I don't remember a way to use NodeJS with isolates, as it was removed years ago.
The documentation about isolates leads to Node 0.8 and V8 v3.1, while the current node LTS is 14.5 and V8 is above v6.

The limitation part is super-worrying, not being able to run a regular NodeJS process is a pure complication of running the workers in the cloud. With temporal going to the cloud, other APIs won't need to run in custom containers and could be deployed anywhere. While with this proposal on the Node side, it will be required to have a custom setup. Keep in mind, that it would also harm developer experience.
Also, given that a lot of dependencies are relying on the NodeJS APIs, it doesn't seem good to have that limitation.

Have you considered other options, which are not relying on Isolates?
Generator functions are one option, they can allow you to run, pause, and stop execution at any given point in time.
This also provides the ability to query/signal workflow.

rylandg:
@VladimirMilenko hey there. I will do my best to explain why the problem was approached this way and then @bergundy can correct me where I'm wrong.

I'll start by talking a bit about NodeJS and Isolates. As you clearly know, NodeJS is a server side JavaScript runtime which gives you access to subsystems (filesystem) that are otherwise unavailable. While NodeJS is a JavaScript runtime it is not a JavaScript engine and directly relies on V8 under the hood. As for isolates, they are a mechanism provided by V8 which allows you essentially run a separate instance of the V8. To be clear, isolates by default provide only the core JavaScript runtime but by using contexts its completely possible to surface any other functionality (aka the entire NodeJS feature set) you desire into an isolate. This is all to say, isolates really just describe an instance of v8. 

So why are we going the direct isolate route and not using the higher level NodeJS? The direct answer is determinism. Workflows in Temporal are required to be fully deterministic which in turns means all workflow code needs to be deterministic. Unfortunately, the architecture of NodeJS inherently interferes with this requirement. The main reason this is the case is because of the way NodeJS handles libraries via `require`. Require is problematic because the rules and semantics are incredibly flexible. As a JS library author, I can write code such that it will run when the user imports my library. Furthermore, I can require dynamic variables, the results of http requests and basically anything else. You can also require at runtime. This is all to say, require is not deterministic. This presents a huge problem, as it basically means that a workflow will be non-deterministic the moment you `require` the wrong package. For the record, these are not NodeJS problems we uniquely encounter. These deficits in NodeJS are so profound that the author himself created a new JavaScript runtime "Deno" https://en.wikipedia.org/wiki/Deno_(software) which explicitly targets these issues with Node.

As for requiring a custom setup for NodeJS, AFAIK this won't be the case. V8 is the basis of NodeJS which means isolates should be available natively. Activities will still be written in vanilla NodeJS and will have access to require. Even workflows will have the ability to use libraries but only those that adhere to the more recent module specification. These modules do not allow arbitrary code execution and therefore are safe to import in workflows.

Hope that clears things up a bit.

VladimirMilenko:
@rylandg Thanks for explanation of the isolates.

I'm however a little worried, that developers would need to write workflows and activities, which would work in two different environments.

Also, knowing the required functionality by Temporal, like querying the stack trace, having the workflows instance cache, ability to pause the execution, until result is retrieved and etc - it seems that it would be a mock-rich environment.

And I still don't see a way to abort execution in case of query.

At the end, it looks like you'd need to write pretty much complete v8 environment to support debug and etc.

rylandg:
While separating workflows and activities is suboptimal, there's really very little room for us to maneuver here due to the inherent architecture.  

As for things like stack trace etc, I see no reason why these would be hard to support. AFAIK, the event loop is not actually provided by the v8 model at all. In fact IIRC the ECMAScript language definition of JavaScript doesn't even define an event loop. As long as we can control the event loop (in NodeJS today), all of these features should be possible.

bergundy:
@VladimirMilenko I understand your concern.
Creating an isolated runtime for workflows is not a trivial task.
In the [POC](https://github.com/temporalio/sdk-node) (still a WIP) I needed to mock out the V8 `Promise` implementation and convert all async functions into generators (using the typescript compiler es6 target), and of course implement `setTimeout` and other non-deterministic builtins.
With this approach, the scheduler has complete control over what's executing in the isolate and execution can be aborted in case of query.
Since the solution is based on generators it is basically what you suggested in your comment above.

As for your comment of V8 isolates not being available in NodeJS, they require a native extension in order to be exposed, I used [isolated-vm](https://github.com/laverdet/isolated-vm).

I agree that it'll take a lot of work to get the isolated environment stable and feature complete but I believe that this will pay off because it'll allow our users to write JS the way they already know and to have assurance that if their code runs it is also 100% deterministic.

It seems like you have a good grasp of JS and it'd be great if you could help review some of the work I've done in the POC because I could use some good input.

VladimirMilenko:
I think POC is marked as a private repo, but I'd like to take a look

sw-yx:
@bergundy can you mark out a couple things you specifically would like input on?

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+    - [Logging](#logging)
+    - [Sessions](#sessions)
+    - [Deterministic Time](#deterministic-time)
+    - [Error Handling](#error-handling)
+    - [Cancellation](#cancellation)
+    - [Long Histories](#long-histories)
+
+
+## Activities
+Activity implementations are just JS functions, they run on the default NodeJS runtime, along with the Temporal SDK, and can import arbitrary npm packages.
```
mfateev, line 31: I don't know JS, but in Go and Java we encourage use of object methods for activities. This way any external dependencies can be injected into the object fields during the worker initialization phase. Is it antipattern in JS?

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+  });
+  return response.data;
+}
+```
+
+### Registration
+Activities should be registered in their corresponding workers.
+
+```typescript
+worker.registerActivity('httpGet', httpGet, { passContext: true });
```
mfateev, line 76: Does TS support some form of reflection? In other languages both name and type of the first parameter can be determined from the function signature.

bergundy:
@VladimirMilenko let me check when we can open that up, it's still unusable at this point.
@sw-yx would be nice to have somebody review the scheduling logic.

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+```
+
+### Payloads
```
mfateev, line 79: We should support the pluggable DataConverter from the beginning. See [Java](https://github.com/temporalio/sdk-java/tree/master/temporal-sdk/src/main/java/io/temporal/common/converter) and [Go](https://github.com/temporalio/sdk-go/tree/master/converter) implementations of it.

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+Initially we'll support any JSON serializable object as payload, binary data support can be added if required.
+
+TBD: consider supporting [typed arrays][typed-arrays].
+
+## Workflows
+Worflow code looks like vanilla TypeScript.
+Workflows run in the same NodeJS process as the Temporal SDK and the activities.
+Workflow code must export a `main` function which will be used as the handler method for workflow invocations.
```
mfateev, line 87: Does it mean that we will not be able to define a workflow interface the way it is done in Java SDK?

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+    export var add: Activity<Fn<[number, number], number>>;
+  }
+}
+```
+
+### Registration
+```typescript
+async function main() {
+  const example = 'dist/workflow.js'; // Workflow file generated using the typescript compiler
+  await worker.registerWorkflow(example);
```
mfateev, line 136: I'm confused. In the example above the workflow code was defined directly in the main method. But here we use the main method to initialize the worker.

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Signals
+TBD
+
+### Queries
+TBD
+
+### Logging
+Logging from the workflow can be implemented with `console.log`.
+TBD log format and context
```
mfateev, line 162: Does V8 expose the ability to intercept the console log calls? We need it for deduping logging on replay.

node/node-sdk.md
```diff
@@ -0,0 +1,187 @@
...
+
+### Deterministic Time
+In a worflow isolate, all date and time methods, e.g `new Date()` and `Date.now`, are overridden by the SDK and replaced by deterministic versions.
+
+### Error Handling
+In workflows, errors are handled by catching exceptions and `Promise.catch` handlers.
+
+### Cancellation
+In a workflow, it's possible to handle an activity cancallation by catching a `CancellationError`.
+Activities cannot be cancelled at the moment.
```
mfateev, line 175: Activity cancellation is supported by Temporal service and should be supported by the JS SDK.

mfateev (commented):
Looks very promising!

VladimirMilenko:
@bergundy not looking for a production ready code, but scheduling indeed deserves a review :)

node/node-sdk.md
VladimirMilenko: Yes, normally you'd have a function, which would take a context argument if something needs to be injected.

node/node-sdk.md
VladimirMilenko: No, the reflection can only be done at a static analysis level, however, compiled code doesn't have any signatures and the arguments for function call are just an array.

node/node-sdk.md
sw-yx: i dont know about "normally", but es6 objects and classes handle this fine

node/node-sdk.md
bergundy: It's easy to add the the ability to cancel an activity in a worker but on the activities side we'd probably want to allow standard async functions which don't support cancelation, the best solution I can think of is injecting an awaitable cancellation token into the activity via the context which the activity function will have to utilize.

We can pass an [`AbortController`](https://developer.mozilla.org/en-US/docs/Web/API/AbortController/abort) for that.

``` typescript
async function httpGetJson(ctx: ActivityContext, url: string) {
  const request = await fetch(url, { signal: ctx.abortController.signal });
  return request.json();
}
```

Additionally as a Promise:

```typescript
async function httpGetJson(ctx: ActivityContext, url: string) {
  try {
  const request = await Promise.all(ctx.cancelled, nonCancellableRequest(url));
  } catch (err) {
    if (err instanceof CancellationError) {
      // cleanup
    } 
    throw err;
  }
  return request.json();
}
```

node/node-sdk.md
bergundy: We can still define the interface's functions and let the workflow author decide which methods to implement.
I am missing some information on the workflow interface in Java, I'll read more about it.
We can switch to an interface if we find it more suitable.

bergundy (commented):
Thanks for all the comments everyone.

node/node-sdk.md
```diff
@@ -0,0 +1,290 @@
...
+import * as declarations from '../../activities';
+
+declare global {
+  namespace activities {
+    // Use the signature of an activity function
+    export var httpGet: Activity<typeof declarations.httpGet>;
+    // In case you can't import you function definition (implmented in another language or simply unavailable in this context),
+    // define the signature manually
+    export var add: Activity<Fn<[number, number], number>>;
+  }
```
sw-yx, line 129: this was a surprise to me - are activities really going to be globals? are you planning to inject them in scope in the isolate?

sw-yx:
@bergundy i have [superficially reviewed the scheduler code](https://github.com/temporalio/sdk-node/issues/1) - but i think i need a diagram of what each event is supposed to do to give you substantive feedback/help you catch bugs. any chance this is documented somewhere? if no, want to get on a call and make these docs (if that would help)?

node/node-sdk.md
bergundy: @sw-yx I'm exploring 2 options ATM, we will either infer types from workflows / activities or explicitly define the workflow and activity interfaces and import them.

I have a feeling we'll end up with the latter, I'll update the proposal soon.

node/node-sdk.md
bergundy: After taking another look I understand the confusion.
This `main` is in the entry point of the worker process not the workflow, I'll fix the example.

bergundy:
@sw-yx I'll add docs and a diagram explaining how workflows are executed.

There are 2 options for injecting activities into the worker.

1. Injecting into the global object (which is what we currently do, we put them all in `globalThis.activities`.
2. Intercepting `import` calls and injecting our stubs from the worker's module loader.

With (1) activities can be defined anywhere while with (2) we'd have to come up with a convention for activity module names so we know what to intercept.

Here's a possible solution for (2) using [path aliases](https://www.typescriptlang.org/tsconfig#paths).

```typescript
import { httpGet } from '@activities/http';
```

sw-yx:
cool, will wait for your updated proposal as chatted on slack

bergundy:
@sw-yx I've thought about your suggestion to use `import` statements instead of globals.
Currently I've found these issues:

1. There needs to be a convention of which imports are activities and which are workflow modules.
   e.g. `import { httpGet } from '@activities'` or passing a list of activity module paths to the worker.
2. It allows the user to import non activities in which case the workflow code will break.
3. It requires analyzing the imported code in order to resolve which activities have been referenced.
   For instance to make the examples below work we'd need to understand that the user imported `httpGet` in order to inject a stub in its place.
   ```typescript
   import { httpGet } from '@activities'
   ```
   ```typescript
   import * as activities from '@activities'
   activities.httpGet(...)
   ```
   We could get around this by utilizing the `default` export and injecting a `Proxy` in its place which resolves any attribute to an activity stub:
   ```typescript
   import activities from '@activities'
   activities.httpGet(...) // -> stub
   ```

I feel like it would be simpler to ask the user to define the activity types at the global scope.
The fact that's it's so unconventional signals to the user that this isn't any normal code and provides a reminder that there are limitations to what you can import and declare.
We're going to need a `global.d.ts` file regardless of the way we inject activities because we need to define things like `console`, `Map`, `Set`, `setTimeout` since these should not be available with the limited library in the workflow environment.

The `global.d.ts` file contains all of the activity definitions used by the workflow and is used to explicitly define the activity interface similarly to the way it's done in the Java SDK.

Here's another version of the `global.d.ts` where we explicitly define an `Activities` interface:
```typescript
/// <reference types="@temporal-sdk/workflow-global" />

import * as declarations from '../../activities';

interface Activities = ActivityInterface<{
  httpGet: typeof declarations.httpGet,
  externalActivity: (a: number, b: number) => number,
}>;

declare global {
  export var activities: Activities;
}
```

sw-yx:
- 1 & 2 can be resolved by checking instanceof and throwing error. as a bonus this becomes a runtime check not just a type check
- for 3, assuming we use the proposed class api instead of a function, if we know what context we are running in, we can call a different method based on the context. no proxies needed.


> We're going to need a global.d.ts file regardless 

we can use [tsconfig bases](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#tsconfig-bases) as well. 

eliminating `global.d.ts` is not the goal. it's eliminating unnecessary work when writing activities.

bergundy:
@sw-yx I don't understand how we can check instanceof during the instantiation of isolated module, we don't know what was imported without analyzing the code.

The [new approach](https://github.com/temporalio/proposals/pull/15/commits/354e086848b438d34fc7448a246faaa72ef7a32f) allows us to add type safety for activities implemented in other workers and eliminates the need for `global.d.ts`.

Thanks again for the valuable input.

mfateev (approved):
I'm not qualified to review the Typescript syntax and language choices. But from the Temporal  point of view, the proposal is fine.

node/node-sdk.md
```diff
@@ -7,193 +7,216 @@ The SDK steers developers to write their workflows and activities in TypeScript
...
+      const current = progressEvent.currentTarget.response.length;
+      const progress = Math.floor(current / total * 100);
+      ActivityContext.current().heartbeat(progress);
+    },
+  });
+  return response.data;
 }
 ```
 
 ### Registration
```
sw-yx, line 80: i had a new thought: should we have a concept of "local" vs "global" registration? i.e. if i register a workflow we may not want it to always be accessible by all other teams unless i want it to. somewhat of a security concern, but mostly i just believe that having good isolation between teams helps companies scale.

node/node-sdk.md
```diff
@@ -299,24 +344,49 @@ The `ActivityContext` exposes 2 ways of subscribing to cancellations.
...
+  return Context.ContinueAsNew(x + 1);
 }
 ```
 
+## Project Structure
+A typical project consists of 4 sub-projects with typescript [project references][ts-project-references].
+
+```
+worker/ -> worker code
+intefaces/ -> intefaces (mostly for workflows)
```
sw-yx, line 371:
```suggestion
interfaces/ -> interfaces (mostly for workflows)
```

node/node-sdk.md
```diff
@@ -299,24 +344,49 @@ The `ActivityContext` exposes 2 ways of subscribing to cancellations.
...
+
+Since the project skeleton is complex we'll provide an initializer package to be used with [npm init][npm-init].
+
+### References
+![](./project-structure.svg)
+
+Each of the referenced projects can be imported using [typescript path aliases][tsconfig-paths].
+
+The following aliases are included in the initializer package:
+- `@intefaces`
```
sw-yx, line 386:
```suggestion
- `@interfaces`
```

node/node-sdk.md
```diff
@@ -270,12 +314,13 @@ export async function main(urls: string[]) {
...
-```typescript
-export async function main(urls: string[]) {
+#### `CancellationScope.run()`
+```ts
+// workflow/example.ts
+async function main(urls: string[]) {
   let promises: Array<Promise<string>>;
   const scope = CancellationScope.run(() => {
-    promises = urls.map(activities.httpGet);
+    promises = urls.map(httpGet);
```
sw-yx, line 323: is this example missing an import or is `httpGet` global?

node/node-sdk.md
```diff
@@ -249,11 +291,13 @@ In workflows, errors are handled by catching exceptions and `Promise.catch` hand
 In a workflow, handle activity cancallation by catching a `CancellationError` or cancel activities using a `CancellationScope`.
 
 #### Example
-```typescript
-export async function main(urls: string[]) {
+```ts
+// workflow/example.ts
+async function main(urls: string[]) {
   const scope = new CancellationScope();
+  const cancelableHttpGet = Context.activity(httpGet, { scope });
```
sw-yx, line 298: is this example missing an import or is `httpGet` global?

node/node-sdk.md
```diff
@@ -0,0 +1,401 @@
...
+### Registration
+Activities are automatically registered by path when using the project initializer.
+
+If for any reason you need custom registration, you may do so using `worker.registerActivities`:
+
+```ts
+// worker/run.ts
+import * as impl from './implementation/greeter'; // Non-standard path
+
+worker.registerActivities({ '@activities/greeter': impl });
```
sw-yx, line 89: its not clear what are the rules around this naming schema `@activities/greeter`. 

1. can i name it something different? like `greeter`? what do we do when there are 2 activities with the same name?

2. using object syntax - do we plan to handle this?

```ts
import * as impl from './implementation/greeter'; // Non-standard path
import * as impl2 from './implementation/greeter2'; // Non-standard path
worker.registerActivities({ 
	'@activities/greeter': impl,
	'@activities/greeter2': impl2,
});
```

node/node-sdk.md
```diff
@@ -7,193 +7,216 @@ The SDK steers developers to write their workflows and activities in TypeScript
...
-Each workflow instance runs in a separate [V8 isolate][v8-isolate], which means it has a completely isolated execution context including global variables.
+Workflows run in the same NodeJS process as the worker and the activities.
+Each workflow instance runs in a separate [V8 isolate][v8-isolate], a completely isolated execution context including global variables.
 The Temporal SDK injects and removes references from the global object to ensure that code running in the isolate is fully deterministic.
 
-Activities can be referenced by using `Workflow.newActivityStub` and invoked as a method or with options (see below).
+### Interface
+A workflow's interface is used for validating the implementation and generating a type safe `WorkflowClient` and `ChildWorkflow`.
+
+Workflow interfaces are inteded for the implementation and are written in sync or async form
```
Sushisource, line 170:
```suggestion
Workflow interfaces are intended for the implementation and are written in sync or async form
```

node/node-sdk.md
```diff
@@ -0,0 +1,610 @@
...
+  return `Hello, ${name}!`;
+}
+```
+
+#### `interfaces/workflows.ts`
+```ts
+import { Workflow } from '@temporal-sdk/interfaces';
+
+export interface Example extends Workflow {
+  greet(name: string): Promise<void>;
```
Sushisource, line 22: On my first read-through this confused me for a sec - I thought for some reason we were re-declaring activity signatures. Perhaps calling this `main` a-la the later example would be more clear. (or maybe even something super explicit like `workflowEntrypoint`

node/node-sdk.md
```diff
@@ -0,0 +1,610 @@
...
+A workflow file must export a `workflow` object which can be type checked using a pre-defined inteface.
+
+```ts
+// workflows/example.ts
+import '@temporal-sdk/workflow'; // inject global type declarations
+import { Example } from '@interfaces/workflows';
+
+async function main(name: string) {
+  // setTimeout and Math.random are replaced with deterministic versions
+  await new Promise((resolve) => setTimeout(resolve, Math.random()));
```
Sushisource, line 209: This feels maybe a bit too magical to me. My initial inclination would be to change anything that'd be nondeterministic in the JS builtins to simply throw an error with a message saying "Please use `import { setTimeout } from '@temporal-sdk/determinism'` instead" (or similar) so that we have a bit more explicit control here, and don't have to explain to users why the built-ins work differently in a perhaps less-than-obvious way if they didn't fully read the docs.

Curious to hear everyone's thoughts on that topic.

Sushisource (commented):
This looks really great to me! Looking forward to figuring out design of & integration w/ core sdk.

node/node-sdk.md
bergundy: That's a mistake, it should be called `main`.

bergundy, line 22:
```suggestion
  main(name: string): Promise<void>;
```

node/node-sdk.md
bergundy: You have a point but I tend to think the opposite.
One of the great things about this SDK is that it really feels magical, you write code that looks like JS even though it runs differently.
I'd rather make the workflow runtime deterministic by default instead of letting users run anything non-deterministic or fail at runtime.

mfateev: I agree with @bergundy on this one. Our goal is to minimize nonstandard API surface. And standard random and time should work deterministically without the user caring about it.

node/node-sdk.md
mfateev: The registration is always local in the sense that it only registered implementations with the local process. Currently, there is no concept of registration that the service understands. 

The cross-service communication is a big area that we should add later. It would include public/private interfaces as well as authorization primitives.

Sushisource (approved):

node/node-sdk.md
sw-yx: that didn't fully answer my question - what are the rules of naming the key?  Is it any string goes?

node/node-sdk.md
sw-yx: OK. will mark as resolved for now knowing you're thinking about it

node/node-sdk.md
bergundy: The way it works is that during import from a workflow we:
1. Check if an activities implementation is registered for this import specifier
2. If so inject implementation
3. Otherwise load normal ES module

So basically, any string goes, you directly impact the import system.

node/node-sdk.md
```diff
@@ -0,0 +1,614 @@
...
+export const workflow: Example = { main, signals };
+```
+
+### Limitations
+* The [NodeJS API][nodejs-api] is not available in workflow code as it breaks determinism.
+* Only ES modules can be imported out of the box with V8 isolates while most npm package use the CommonJS module implementation.
+    * We can bypass this by implementing `CommonJS` `require` ([reference](https://nodejs.org/api/modules.html#modules_all_together)).
+    * 3rd party imported modules may contain async / await keywords - this can be overcome by transpiling each required file at runtime.
+* Submitted workflow code must not use async / await syntax.
+    Using async / await bypasses the injected Promise object in the workflow context and hinders our ability to gain full control over scheduling.
```
psigen, line 372: Is there a side-discussion around why it is necessary to inject a customized Promise object vs potentially greening some of the IO operations in the node API.  I'm just curious as I was recently looking into this for a discrete-event simulation system.

bergundy: @psigen I just recently figured out that we don't need to inject a customized Promise object.
Since we work directly with V8 and not the entire NodeJS runtime we can ensure that our workflow runtime is deterministic by controlling all async completions (e.g. activities and timers) and running all of the v8 microtasks each time we call back into the isolate.
I'll update the proposal soon in another PR.
Hope that answers your question.

psigen: Thanks, that's exactly what I was hoping!

PR #22, opened 2021-05-23
Add proposal for user code logging and metrics in the Node SDK

bergundy:
- Origin issue: https://github.com/temporalio/sdk-node/issues/53 (related)

# Summary

Generic "external dependencies" solution to allow users to inject their own metrics and logging solutions into Workflow isolates.

# Motivation

This is a crucial feature for a Temporal SDK.

# Drawbacks

Nothing comes to mind

# Alternatives

In the doc

# Adoption strategy

Not relevant

# How we teach this

Standard documentation on the docs site

# Unresolved questions

Stated in the doc

### DIFF
--- a/node/logging-and-metrics-for-user-code.md
+++ b/node/logging-and-metrics-for-user-code.md
@@ -25,7 +25,7 @@ There are a few options for doing IO inside a Workflow's isolate:
 - When isolate runs out of memory but in that case logs will probably not work anyways.
 - When WF code gets stuck e.g goes into an infinite loop but we can mitigate this problem by enforcing a time limit of execution in the isolate, in case execution times out we will be able to extract the logs.
 
-(3) has the lowest overhead but also means there's an inherent delay from logs and metrics generation to processing - should be reliablility minimal assuming Workflow code is not CPU bound.<br/>
+(3) has the lowest overhead but also means there's an inherent delay from logs and metrics generation to processing - should be relatively minimal assuming Workflow code is not CPU bound - activations are typically short.<br/>
 (3) is limited to either ignored and asynchronous functions since because it does **not** block the isolate.
 
 ### Proposed solution
@@ -33,7 +33,7 @@ There are a few options for doing IO inside a Workflow's isolate:
 Allow users to expose their logging and metrics interfaces to Workflow code as external dependencies.
 We'll support techniques (2) and (3) stated above depending on the type of injected function.
 
-**Extra care should be taken when using values returned from external dependencies because it can easily break deterministic Workflow execution**
+**Extra care should be taken when using values returned from external dependencies because it can easily break deterministic Workflow execution.**
 
 #### Definitions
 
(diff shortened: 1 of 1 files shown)


node/logging-and-metrics-for-user-code.md
```diff
@@ -0,0 +1,184 @@
...
+A hard requirement for our logging and metrics solution for user code is that we do not enforce a specific tool or convention and instead let our users use their existing tools.<br/>
+For Activities, there is not an issue because users can run anything in an Activity. In Workflows OTOH, only deterministic code can run and metrics and logs should usually be dropped when a Workflow is replaying. Additionally the fact that the Node SDK runs Workflows in v8 isolates, mean that there's no way to do IO or anything that interacts with the world without explicit support from the SDK.
+
+### IO in Workflow code
+
+There are a few options for doing IO inside a Workflow's isolate:
+
+1. Load a binary (c++) module into the isolate, the module is not isolated the way JS code is
+2. Inject a function reference from the main NodeJS isolate into a Workflow isolate
+3. Accumulate data in the isolate (e.g when collecting commands to complete an activation) and pass it into a function in the main NodeJS isolate
```
jmoseley, line 12: Is there any concern about reliability with this option? Such as a situation where the logs/metrics get collected but don't have an opportunity to flush. This could make it very challenging to debug unexpected errors that are killing a thread.

bergundy: Good point, this is something that might happen in case the isolate runs out of memory but in that case logs will probably not work anyways, I think the isolate will die.
I thought there might be an issue of WF code getting stuck e.g goes into an infinite loop but we can mitigate this problem by enforcing a time limit of execution in the isolate, in case execution times out we will be able to extract the logs.

node/logging-and-metrics-for-user-code.md
```diff
@@ -0,0 +1,184 @@
...
+### Logger injection example
+
+#### Variant A - `Context.dependency` + `Worker.inject`
+
+`interfaces/logger.ts`
+
+```ts
+export interface Logger {
+  info(message: string): void;
+  error(message: string): void;
```
Sushisource, line 54: These functions signatures should probably more closely match the `console.log` versions (even if we aren't explicitly aping console.log) where you pass a variable amount of objects or string with substitutions

node/logging-and-metrics-for-user-code.md
```diff
@@ -0,0 +1,184 @@
...
+```
+
+#### Comparison
+
+|                      | Variant A                                                        | Variant B                   |
+| -------------------- | ---------------------------------------------------------------- | --------------------------- |
+| Missing dependencies | ❌ Hard to enforce all dependencies are injected into the worker | ✅ Enforced by type checker |
+| Dependency naming    | ❌ Prone to typos                                                | ✅ Enforced by type checker |
+
+I'm strongly leaning towards variant B.
```
Sushisource, line 156: +1

node/logging-and-metrics-for-user-code.md
bergundy: I've stated these concerns in the document, resolving.

bergundy:
@Sushisource @jmoseley I've made a POC implementation of this spec and found out that we have some limitations that did not come to mind when I first created this document.
I think we've come up with a good design and I plan to merge it unless anyone objects in the next few days.

PR #23, opened 2021-05-24
Add Node+Core instrumentation proposal

bergundy:
# Summary

Complete solution for logging, tracing, and metrics for the Node SDK (including Core).
Most of this spec could be applied to any Core based SDK.

# Basic example

N/A

# Motivation

We need the ability to instrument the Node SDK and Core based SDKs in general.
Instrumentation should be able to integrate with users' existing solutions.

# Detailed design

In doc

# Drawbacks

In doc

# Alternatives

In doc

# Adoption strategy

We should document how to use this and provide example for integration with major services such as DataDog and Jaeger.
Documentation should be geared toward operators.

# How we teach this

Via documentation

# Unresolved questions

Core side is still TBD, will go over it with @Sushisource

### DIFF
--- a/node/tracing-logging-and-metrics-for-sdk-and-core.md
+++ b/node/tracing-logging-and-metrics-for-sdk-and-core.md
@@ -1,0 +1,96 @@
+## Tracing, Logging, and Metrics in the NodeJS SDK and Core
+
+### What's already implemented in the Alpha release?
+
+Core [uses opentelemetry](https://github.com/temporalio/sdk-core#debugging) for generating logs and traces, it does not emit any metrics.
+
+The Node SDK uses `WorkerOptions.logger` directly for logging and the opentelemetry API for tracing with no output by default, it does not emit any metrics either.
+
+There's an example of how to export the traces to jaeger in [`bench.ts`](https://github.com/temporalio/sdk-node/blob/a25c0feca791a195eb5e50cf8c0fd9e0b5a1db83/packages/test/src/bench.ts#L105).
+
+See the [documentation](https://docs.temporal.io/docs/node/logging/) on how to customize the Worker logger.
+
+There's currently not a way to log from Workflow internals (inside the isolate).
+
+### Requirements
+
+- Link traces between Node and Core for a more complete picture
+- Core traces should be propagated through NodeJS
+- Core metrics should be propagated through NodeJS
+- Core logs should integrate with the Worker logger
+- Support logging from Workflow internals
+- Support logging in Worker either within an otel span or without context (e.g. Worker state changed)
+- Metrics should only be used when no trace context is available
+
+#### Metrics
+
+- Gauge of cached WFs - should this be exported from both Core and Lang to find discrepencies?
+- Gauge of cached WF memory stats - Lang
+- Counter of WF cache insertions - Lang
+- Counter of WF cache evictions - Lang
+- Gauge of in-flight Activities - Lang
+- Gauge of in-flight WF activations - Lang
+- GRPC call latency and count per status for each route - Core only (can we just use otel spans?)
+
+> PS: More metrics should be considered here, this is just a list off the top of my head.<br/>
+> PPS: We might not need metrics from Core if we can rely on otel spans since those can be converted into metrics using a custom propagator.
+
+#### Tracing
+
+- Starting from Lang
+  - Poll start to completion (WFs and Activities)
+    - Lang passes span context to Core
+  - Connect to Server?
+- Starting from Core
+  - TODO
+
+#### Logging
+
+- Worker logger supports 4 log levels: `DEBUG | INFO | WARN | ERROR`.
+- Logging in trace context can be done with otel span events or by extracting the span context into a log event, it's unclear to me if otel spans are a good fit here, see [this discussion](https://github.com/open-telemetry/opentelemetry-specification/issues/67) for info on the subject.
+
+| Level              | Component | Trace context | Event                                                 |
+| ------------------ | --------- | ------------- | ----------------------------------------------------- |
+| `INFO`             | Worker    | None          | Worker state change                                   |
+| `DEBUG`            | Worker    | Activation    | Activation received with list of jobs and WF context  |
+| `ERROR`            | Worker    | Activation    | WF activation failed                                  |
+| `DEBUG`            | Workflow  | Activation    | WF activation concluded with list of commands         |
+| `DEBUG`            | Worker    | Activity Task | Activity task received (start or cancel) with context |
+| `DEBUG` or `ERROR` | Worker    | Activity Task | Activity task resolved                                |
+| `ERROR`            | Worker    | Any           | Unexpected error occurred                             |
+| `ERROR`            | Workflow  | Activation    | Unexpected error occurred                             |
+| `DEBUG`            | Worker    | None          | Got activity heartbeat                                |
+
+> TODO: Fill in this table with required Core logs
+
+### Solution
+
+- Add context to logs emitted from Worker where applicable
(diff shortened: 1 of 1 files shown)


node/tracing-logging-and-metrics-for-sdk-and-core.md
```diff
@@ -0,0 +1,96 @@
...
+
+See the [documentation](https://docs.temporal.io/docs/node/logging/) on how to customize the Worker logger.
+
+There's currently not a way to log from Workflow internals (inside the isolate).
+
+### Requirements
+
+- Link traces between Node and Core for a more complete picture
+- Core traces should be propagated through NodeJS
+- Core metrics should be propagated through NodeJS
```
Sushisource, line 19: Do they actually need to be? I want to say as long as we use the same tags for metrics they'll get aggregated by the thing that receives them just fine.

node/tracing-logging-and-metrics-for-sdk-and-core.md
```diff
@@ -0,0 +1,96 @@
...
+- Core logs should integrate with the Worker logger
+- Support logging from Workflow internals
+- Support logging in Worker either within an otel span or without context (e.g. Worker state changed)
+- Metrics should only be used when no trace context is available
```
Sushisource, line 23: Why is that?

node/tracing-logging-and-metrics-for-sdk-and-core.md
```diff
@@ -0,0 +1,96 @@
...
+
+#### Metrics
+
+- Gauge of cached WFs - should this be exported from both Core and Lang to find discrepencies?
```
Sushisource, line 27: Definitely

node/tracing-logging-and-metrics-for-sdk-and-core.md
```diff
@@ -0,0 +1,96 @@
...
+
+### Solution
+
+- Add context to logs emitted from Worker where applicable
+- Pass span context as from Lang to Core in the `poll_activity_task` and `poll_workflow_activation` APIs
+- Add methods to the Core API for exporting logs, metrics and traces
+  - Special care should be taken when defining the types here, logs should be structured to support custom formatting on the Lang side
+  - Hook up instrumentation into Lang opentelemetry tracer and Worker logger
+- Inject the Worker logger interface into the isolate for internal use - make sure activation context is baked into the WF logger
+- Worker logger should add a `component` attribute (e.g. `worker | core | workflow`) in the log context so we know each log's origin
```
Sushisource, line 74: If `workflow` is a type then `worker` should probably just be `activity` or something

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: I mean don't do things that you can already get with tracing with metrics e.g: https://github.com/temporalio/samples-go/blob/master/metrics/metrics.go

node/tracing-logging-and-metrics-for-sdk-and-core.md
```diff
@@ -0,0 +1,96 @@
...
+- Propagating traces, logs and metrics from Core to Lang implies buffering since they run in different threads and communicate using a queue
+  - Could result in event loss
+    - Is this tollerable?
+    - Should we also support logging directly to console from Core to mitigate this issue?
+  - Logs from different sources are received out of order
+    - This might not be that big of an issue in production where log aggregators can reorder based on timestamp
+    - During development, we can stall log output in order to merge and chronologically sort the different logs sources
+- Propagating logs from the WF isolate has similar implications
+
+### Alternatives
```
Sushisource, line 87: Other options we discussed on the call:

I wrote the paragraphs below this one first, but after writing them, I have the feeling that the "right" thing to do here is actually always have lang be the initiator of a span (which makes sense) and then core simply attaches it's additional spans/events/metrics/etc into the "context" provided by lang, and that context can be returned by each API call. Unfortunately this makes our APIs super ugly purely for the sake of opentelemetry which I don't like at all. So, I'd still prefer a solution that allows us to include the span/context in our protos and then propagate results out-of-band.

One option would be to have core and lang share a socket (in memory, or not, whatever) which actually uses the gRPC protocol in streaming mode to have core constantly pump tracing data to the lang side. Probably using https://docs.rs/opentelemetry/0.14.0/opentelemetry/propagation/index.html or alternatively https://docs.rs/opentelemetry/0.14.0/opentelemetry/sdk/export/trace/trait.SpanExporter.html -- it's not entirely clear from the docs what makes more sense. Will need to experiment. The span exporter seems slightly more obvious, buuuut, the propagators could potentially better suit our needs if we can somehow just pass them along with every API call, core can use them to inject everything it's doing and then lang will end up with everything on it's side.

Another would be to simply define the lang side as an opentelemetry "processor" (I think? Their concepts are not super well explained) https://opentelemetry.io/docs/concepts/data-collection/ -- but essentially just make lang a receiver of exported data from core, where it would then combine it with it's own data and export everything to whatever backend(s) the user has registered. Problem is, after a little research, exporting the data this way using OTLP for example looks pretty easy https://docs.rs/opentelemetry-otlp/0.7.0/opentelemetry_otlp/ but receiving it looks much harder as there doesn't appear to be much in the way of libraries for that part which isn't terribly surprising. We could implement it ourselves, but at that point the other options are probably easier.

bergundy: Another thing we had considered is that core respects the sampling flags in the trace context https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/trace/sdk.md#sampling 
This should probably come out of the box if we use the library in a standard way.

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: They have to be so users can export them using their current metrics solution (via lang)

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: `workflow` refers to the WF isolate runtime library, for which logs should collected by the worker.

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: See also this comment about missing unique IDs for activations https://github.com/temporalio/sdk-core/pull/125/files#r638239225

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: Will update the doc

node/tracing-logging-and-metrics-for-sdk-and-core.md
bergundy: I'm strongly leaning towards writing a custom propagator or exporter that will push spans from Core to Lang.

Sushisource: Yeah I think that's certainly the first avenue to explore. And yeah I suppose there are some circumstances where it would make sense for core to initiate a span, but they could always be child spans of something initiated by lang. Heartbeats for example initiated by HB call in lang and then have child spans for every individual actual heartbeat request, but that possibly ends up being unnecessary in practice.

Sushisource:
See https://github.com/temporalio/sdk-core/issues/85 for context on exploring actual implementation. Propagation of metrics from core->lang is a very large amount of work and doesn't fit into OTel's design. For now, having users run the OTel collector and exporting metrics to it from core/lang is the best approach.

PR #24, opened 2021-05-24
Revise NodeJS Workflow cancellation design

bergundy:
- Origin issue: https://github.com/temporalio/sdk-node/issues/6

# Summary

Improvements to the NodeJS in-Workflow cancellation design

# Basic example

In doc

# Motivation

In doc

# Detailed design

In doc

# Drawbacks

In doc

# Alternatives

In doc

# Adoption strategy

N/A

# How we teach this

Documentation with a lot of examples

# Unresolved questions

In doc

### DIFF
--- a/node/cancellation-v2.md
+++ b/node/cancellation-v2.md
@@ -57,7 +57,6 @@ cancel -> member CancellationScope.cancel
 ### Behavior
 
 - `CancellationScope` no longer throws `CancellationError`, it simply cancels timers and activities which in turn should throw `CancellationError`
-- Activities and timers started in a `CancellationScope` after it's been cancelled will not start and instead will throw `CancellationError`
 - `Context.cancelled` is moved to `CancellationScope.cancelRequested` and changed to return a promise to assist writing Workflows which are driven solely by signals
 
   ```ts
@@ -261,10 +260,7 @@ export async function main(url: string, data: any): Promise<void> {
   } catch (err) {
     if (err instanceof CancellationError) {
       console.log('Workflow cancelled');
-      // Cleanup logic goes in a nonCancellable scope
-      // If we'd run cleanup outside of a nonCancellable scope it would've been cancelled
-      // before being started because the Workflow's root scope is cancelled.
-      await CancellationScope.nonCancellable(() => cleanup(url));
+      await cleanup(url);
     }
     throw err; // <-- Fail the Workflow
   }
@@ -284,7 +280,7 @@ export async function main(url: string) {
       await CancellationScope.withTimeout(1000, () => httpPostJSON(url));
     } catch (err) {
       if (err instanceof CancellationError) {
-        await CancellationScope.nonCancellable(() => cleanup(url));
+        await cleanup(url);
       }
       throw err;
     }
--- a/node/cancellation-v2.md
+++ b/node/cancellation-v2.md
@@ -49,19 +49,20 @@ Issue (2.2) is a bug but it raises the point that the docs should clarify which 
 ### Naming and structure
 
 ```
-shield -> CancellationScope.shield
-cancel -> CancellationScope.cancel
+cancellationScope -> static CancellationScope.cancellable
+shield -> static CancellationScope.nonCancellable
+cancel -> member CancellationScope.cancel
 ```
 
 ### Behavior
 
-- `cancellationScope` no longer throws `CancellationError`, it simply cancels timers and activities which in turn should throw `CancellationError`
-- Activities and timers started in a `cancellationScope` after it's been cancelled will not start and instead will throw `CancellationError`
-- `Context.cancelled` changes to return a promise to assist writing Workflows which are driven solely by signals
+- `CancellationScope` no longer throws `CancellationError`, it simply cancels timers and activities which in turn should throw `CancellationError`
+- Activities and timers started in a `CancellationScope` after it's been cancelled will not start and instead will throw `CancellationError`
+- `Context.cancelled` is moved to `CancellationScope.cancelRequested` and changed to return a promise to assist writing Workflows which are driven solely by signals
 
   ```ts
   async function main() {
-    await Context.cancelled;
+    await CancellationScope.current().cancelRequested;
   }
 
   let counter = 0;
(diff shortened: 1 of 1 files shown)


bergundy:
See also: https://docs.temporal.io/docs/node/workflow-scopes-and-cancellation

node/cancellation-v2.md
```diff
@@ -0,0 +1,139 @@
...
+2. What happens when a timer or activity was created in a non-cancellable context and awaited in a cancellable context?
+
+```ts
+let p: Promise<void> | undefined = undefined;
+await Context.nonCancellable(async () => {
+  p = sleep(3);
+});
+// p is non cancellable in a cancellable context
+await p;
+```
```
sw-yx, line 96: this is a nice improvement, but still feel like we are not quite there yet. 

can we add a section with some example common use cases, the type that we would show in the docs or JS Samples repo? then happy to work on api design with those usecases in mind.

bergundy: I started going over it and decided to do another iteration, I'll submit it tomorrow, it's already way too late for me.

node/cancellation-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+If the Workflow is cancelled, the user might be surprised if `await p` throws.
+
+2. What happens when a timer or activity was created in a non-cancellable context and awaited in a cancellable context?
+
+```ts
+let p: Promise<void> | undefined = undefined;
+await Context.nonCancellable(async () => {
+  p = sleep(3);
+});
```
Sushisource, line 93: I'm a little confused about this example here. Is it missing an `await p;` at the end of this closure too, since we're talking about it being awaited in both places?

node/cancellation-v2.md
```diff
@@ -0,0 +1,139 @@
...
+
+If the Workflow is cancelled, the user might be surprised if `await p` does not throw.
+
+One approach would be to say that once a `ContextualPromise` is awaited in a non-cancellable context it becomes non-cancellable in all contexts. Another approach is the exact opposite where the promise becomes cancellable once awaited in a cancellable context.
+
+We should be able to support different cancellation policies for a single promise.
+With this approach, when a promise is cancelled while it is awaited in a both cancellable and non-cancellable contexts, an exception is thrown immediately in all cancellable contexts (assuming TRY_CANCEL cancellation type) but the activity does not get cancelled.
+
+The first 2 approaches are still surprising although they're predictable if they are well documented.
+The latter approach seems preferable to me and the least surprising. Not sure how feasable this is and if there are limitations with different cancellation types.
```
Sushisource, line 106: Another possible (but probably crappy) option would be to simply blow up if a promise is awaited in different contexts, just explicitly prohibit the problem all-together, but not very javascript-y and would be much better solved at the type level if we could... :man_shrugging:

node/cancellation-v2.md
bergundy: I thought about this a bit more and one of the problems with this design is that promises receive their context on creation instead of when awaited on.
I'm going to change this so promises are only associated with a context once awaited, that should clear up the confusion with this example.

node/cancellation-v2.md
bergundy: Says the rust programmer :)
I don't think we can implement something like a borrow checker with Typescript.
I'd like to keep the ability to await on a context in multiple contexts if possible but it's definitely an option to just drop support for that.

bergundy (commented):
Thanks for the feedback

node/cancellation-v2.md
bergundy: Added the approach you've mentioned to the doc.

node/cancellation-v2.md
```diff
@@ -0,0 +1,308 @@
...
+  try {
+    await promise;
+  } catch (e) {
+    if (e instanceof CancellationError) {
+      console.log('Exception was propagated 👍');
+    } else {
+      throw e;
+    }
+  }
+}
```
sw-yx, line 161: what is the developer experience if i forget to use try/catch?


```
export async function main() {
  const promise = sleep(1);
  promise.cancel();
  await promise; // what happens here
}
```

bergundy: The promise will throw a `CancellationError` causing the workflow to fail.

node/cancellation-v2.md
Sushisource: This new version makes more sense to me :+1:

node/cancellation-v2.md
```diff
@@ -0,0 +1,343 @@
...
+
+#### `CancellationScope`s can be shielded, preventing cancellation from propagating to created cancellables
+
+```ts
+import { cancellationScope } from '@temporalio/workflow';
+import { httpGetJSON } from '@activities';
+
+export async function main(url: string) {
+  // Prevent Activity from being cancelled and await completion.
+  // Note that the Workflow is completely oblivious to cancellation in this example.
```
Sushisource, line 193: Oblivious, and also impervious, right? Like it just can't be cancelled effectively? Or is what's going to happen here is the activity won't be cancelled, but the workflow will be?

node/cancellation-v2.md
```diff
@@ -0,0 +1,343 @@
...
+export async function main(url: string, data: any): Promise<void> {
+  try {
+    await httpPostJSON(url, data);
+  } catch (e) {
+    if (e instanceof CancellationError) {
+      console.log('Workflow cancelled');
+      // Cleanup logic goes in a shielded scope here
+      // If we'd run cleanup outside of a shielded scope it would've been cancelled
+      // before being started because the Workflow's root scope is cancelled.
+      await cancellationScope({ shield: true }, () => cleanup(url));
```
Sushisource, line 239: This is slightly funky. I get why, but it seems like a really easy mistake to make. Is there any way we could somehow check that `CancellationError` has already been thrown in this scope? Also a bit nasty but if it can work reliably saves users some debugging sessions.

node/cancellation-v2.md
```diff
@@ -0,0 +1,343 @@
...
+import { activity1, activity2 } from '@activities';
+
+export async function main(): Promise<void> {
+  const p1 = activity1(); // <-- Start activity1 in root scope
+  const p2 = activity2(); // <-- Start activity2 in root scope
+
+  const scopePromise = cancellationScope(async (scope) => {
+    const first = await Promise.race([p1, p2]);
+    scope.cancel(); // Does not cancel activity1 or activity2 as they're tied to the root scope
+    return first;
```
Sushisource, line 281: But `scopePromise` will be resolved as cancelled?

node/cancellation-v2.md
bergundy: Yes, impervious as well.

node/cancellation-v2.md
bergundy: Those were my thoughts initially, @mfateev thought otherwise.
This behavior also matches the cancel token example in the [trio blog post](https://vorpus.org/blog/timeouts-and-cancellation-for-humans/#cancel-tokens-are-level-triggered-and-can-be-scoped-to-match-your-program-s-needs).

I also feel like devs will likely get this wrong though, and I'm not sure this behavior is suitable for activities.
Let's go over this again with @mfateev.

node/cancellation-v2.md
bergundy: No, `scopePromise` will be resolved with `first`.

PR #26, opened 2021-06-14
tctl: Output piping [p0]

feedmeapples:
- Origin issue: 

# Summary

- Adds built-in support for pagers `less` and `more` or provide your own favorite for example `bat`
- Optimizes performance because of lazy fetching (in contrast to pipe syntax)
- Improves piping support into tools such as `wc`, `grep`, `jq` (by using `cat` as the pager)

# Basic example

Print or pipe the first page of data
``` bash
tctl workflow list --no-pager --output json | jq "."
```
Print or pipe all pages
```bash
tctl workflow list --pager cat | wc -l
```

Page over items with `less`. That's also the default behavior for Table view `--format table`
```bash
tctl workflow list --pager less # default for Table view
```
Page over items with `more`. That's also the default behavior for JSON and Card views `--format <json|card>`
```bash
tctl workflow list --pager more # default for JSON and Card views
```

Provide your own favorite pager
```bash
tctl workflow list --pager myFavoritePager
```

Use environment variable
``` bash
export PAGER=more && tctl workflow list # use env variable
```

# Motivation
Gives ultimate flexibility over piping and pagination.
Optimizes performance because of lazy fetching

# Detailed design

in doc

# Drawbacks

- since this changes how output of almost all commands work, it adds a risk to break/worsen UX of some commands. Requires thorough testing 

# Alternatives

- 

# Adoption strategy

- this is a breaking change. Release the change together with the rest of tctl UX changes. Create a blog post on the tctl changes, update documentation https://docs.temporal.io/docs/system-tools/tctl

# How we teach this

Terminology: "cli output piping", also "Input/Output Redirection in the Shell"
Documentation for tctl will need to be updated 

# Unresolved questions


cli/001-cli-improve-piping-support.md
```diff
@@ -0,0 +1,33 @@
...
+
+Structure outputs so it would be easy to pipe it into common utilities. 
+
+To achieve that:
+
+ - `wc`, `more` support: integrate output with `less` cmd by default if it's available in the OS. If `less` is not found, create and use builtin functiniality to output all available rows immediately (similar to how `cat` behaves).
+ 
+ Examples: 
+ ``` bash
+ tctl workflow list --open --all | wc
```
robholland, line 16: I'm inferring from this that the default format for non-TTY would not be table (otherwise `wc` would count the header/divider rows). If that's the case it should be specified here or a link given to the proposal which changes that behaviour. Also I think this should be `wc -l`?

bergundy: I wouldn't change the output for tty vs non-tty, it provides a strange experience.
Think of what would happen to a user that runs `tctl wf list ...`, realizes they want to pipe the results and get different output.

cli/001-cli-improve-piping-support.md
```diff
@@ -0,0 +1,33 @@
...
+ ```
+ ``` bash
+ tctl workflow list --all | more
+ ```
+ 
+ - `grep` support: add a flag to output data as cards, where each row will follow format `{field name}: {field value}`
```
bergundy, line 22: I'm not sure I like the card output option.
What if we just support formatting with go templates like docker does? https://docs.docker.com/config/formatting/
I also like the `-q` option for docker commands which will only output IDs.

cli/001-cli-improve-piping-support.md
```diff
@@ -0,0 +1,33 @@
...
+ 
+ Examples:
+ ``` bash
+ tctl workflow list --card --all | grep WorkflowId
+ ```
+ - `jq` support: json output (`--json` flag) should be possible to process using jq
+
+Examples:
+``` bash
+tctl workflow list --json | jq '.execution'
```
bergundy, line 32: How about `--output json` vs `--output table`?
It's more generic this way.

cli/001-cli-improve-piping-support.md
feedmeapples:
```suggestion
 tctl workflow list --open --all | wc -l
```

cli/001-cli-improve-piping-support.md
feedmeapples: i like that go-template: 
 - gives flexibility and power to users
 - makes it easy for us to develop (+less error prone)
 
On the bad side:
 - it's hardly user friendly since it will require majority of users to learn the template formatting if they want to output something similar to the proposed "card" format
 
What's your opinion on the following:
we add `--format` (or --output as you've mentioned [here](https://github.com/temporalio/proposals/pull/26/files#r653123514)) flag that accepts
 - table
 - json
 - card
 - <go-template> - everything that is not above 3 is considered go-template

cli/001-cli-improve-piping-support.md
```diff
@@ -0,0 +1,33 @@
...
+- Issue:
+
+### Support better piping of output 
+Problem: currently piping the output into other tools is not well supported. For example utilities such as `grep`, `wc`, `jq` may work with very limited set of commands / in limited functionality. 
+
+Structure outputs so it would be easy to pipe it into common utilities. 
+
+To achieve that:
+
+ - `wc`, `more` support: integrate output with `less` cmd by default if it's available in the OS. If `less` is not found, create and use builtin functiniality to output all available rows immediately (similar to how `cat` behaves).
```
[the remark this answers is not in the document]
feedmeapples: > TBH I'm not sure we want to automatically pipe to pager

do you mean by default to use a simple builtin pager implementation? Then if people want to use another pager like `less` they would use pipe syntax (ie. ` | less`)? 

In case of piping ` | less` is this actually possible to have lazy pagination over data? (and not first prefetch all data)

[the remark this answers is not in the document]
feedmeapples: my understanding was that it's not possible to lazy load with piping syntax (at least quick google search didn't show anything), and then going over other modern CLIs i've found that they mostly all use internal piping into these tools 

ex. 
aws - https://docs.aws.amazon.com/cli/latest/userguide/cli-usage-pagination.html
bat - https://github.com/sharkdp/bat/blob/91a347bf6d549ba543f1b4bfdc2a05f269c975c8/src/less.rs#L7

feedmeapples:
Have updated the proposal, 

@bergundy @robholland @mfateev created a PR https://github.com/temporalio/tctl/pull/27

cli/001-cli-improve-piping-support.md
feedmeapples: done in this proposal related https://github.com/temporalio/proposals/pull/31

bergundy:
no pager should print all pages IMO

feedmeapples:
> no pager should print all pages IMO

could you elaborate what is meant. Does this refer to default behavior and that it shouldn't print all pages? What if it's a pager such as less and allows users to scroll down and print all pages?

bergundy:
> > no pager should print all pages IMO
> 
> could you elaborate what is meant. Does this refer to default behavior and that it shouldn't print all pages? What if it's a pager such as less and allows users to scroll down and print all pages?

I mean that If I run `tctl --no-pager ...` it should print out all of the available data instead of only the first page.
We should have a different flag for limiting the number of results.

feedmeapples:
> > > no pager should print all pages IMO
> > 
> > 
> > could you elaborate what is meant. Does this refer to default behavior and that it shouldn't print all pages? What if it's a pager such as less and allows users to scroll down and print all pages?
> 
> I mean that If I run `tctl --no-pager ...` it should print out all of the available data instead of only the first page.
> We should have a different flag for limiting the number of results.

this is done & merged

bergundy:
> > > > no pager should print all pages IMO
> > > 
> > > 
> > > could you elaborate what is meant. Does this refer to default behavior and that it shouldn't print all pages? What if it's a pager such as less and allows users to scroll down and print all pages?
> > 
> > 
> > I mean that If I run `tctl --no-pager ...` it should print out all of the available data instead of only the first page.
> > We should have a different flag for limiting the number of results.
> 
> this is implemented & merged

The description in the proposal should be updated to reflect this change in behavior.

bergundy (approved):

PR #27, opened 2021-06-14
tctl: Changing the defaults through configuration file [p1]

feedmeapples:
- Origin issue: 

# Summary

Allows changing tctl defaults through a new `config` command. Stores the defaults in configuration file  

# Basic example

``` bash
tctl config set namespace custom-namespace
```

equivalent with explicit flags:
``` bash
tctl config set --property namespace --value custom-namespace
```

# Motivation

Allows to set some of the flags once and not have to specify them every-time. This is especially relevant for the flags that often remain the same through time, are required in many commands such as --namespace

# Detailed design

in doc

# Drawbacks

# Alternatives

- ensuring that all such configs can be set through env variables

# Adoption strategy

Not a breaking change. Provide good `--help` for the new `config` command. Document in docs 

# Unresolved questions

### DIFF
--- a/cli/002-cli-config-file.md
+++ b/cli/002-cli-config-file.md
@@ -3,7 +3,6 @@
 - Issue:
 
 ### Config file for tctl defaults
-As a user i want to be able to change the defaults of some flags, such as the default namespace.
 Support configuring some of the tctl defaults through a configuration file. 
 
 Add `config` top level command that should allow setting and reading specific config options
--- a/cli/002-cli-config-file.md
+++ b/cli/002-cli-config-file.md
@@ -1,0 +1,19 @@
+- Start Date: 2021-06-11
+- RFC PR:
+- Issue:
+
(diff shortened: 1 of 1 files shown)


robholland (approved):
I don't see the value in the flagged version of config set, but otherwise looks good.

bergundy:
Agree with @robholland, I'd drop the flagged version.

bergundy:
You've mentioned a `--config` command, by which I think you meant `tctl config`, which makes me think that we should have a global `--config` option to specify the config location.

feedmeapples:
> You've mentioned a `--config` command, by which I think you meant `tctl config`, which makes me think that we should have a global `--config` option to specify the config location.

yes, typo. Meant `config`

feedmeapples:
> I don't see the value in the flagged version of config set, but othe

good catch, meant `config`

PR #31, opened 2021-06-15
tctl: Consistent Output formatting as tables, json, cards [p0]

feedmeapples:
- Origin issue: 

# Summary

Allows to customize what columns are printed in Table view.

Reduces noise when rendering Tables.

Standardizes control over printing options through `--output` flag

Adds a new printing format option `card` to render objects as `{field name}: {field value}`
 
# Basic example

``` bash
tctl workflow list --output table # default for most commands
tctl workflow list --output json
tctl workflow list --output card # new view
```

``` bash
tctl workflow list --columns Execution.RunId,SearchAttributes # prints table with mentioned columns
tctl workflow list --columns long # automatically add more columns
```
# Motivation

Improves and standardizes UX related to printing of data 

# Detailed design

in doc

# Drawbacks

One important breaking change is the change of `--print_json` flag to `--output json`. The rest is addition

# Alternatives

# Adoption strategy

Notify community about tctl the breaking change

# How we teach this

`--help` and docs

# Unresolved questions
Is backwards compatibility needed?

### DIFF
--- a/cli/006-cli-output-options.md
+++ b/cli/006-cli-output-options.md
@@ -4,7 +4,7 @@
 
 ### Improve commands output UX and rendering options
 
-##### many commands do not support switching between Table and JSON views.
+#### Many commands do not support switching between Table and JSON views.
 
 In places where we retrieve and print entity objects, support changing how the objects are printed through `--json` and `--table` flags. For arrays of objects, usually prefer Table view as the default. Implement the proposal by creating a unified rendering utility.
 
--- a/cli/006-cli-output-options.md
+++ b/cli/006-cli-output-options.md
@@ -8,17 +8,18 @@ Current Issues and improvement proposals:
 
 #### - Many commands do not support switching between Table and JSON views.
 
-In places where we retrieve and print entity objects, support changing how the objects are printed through `--json` and `--table` flags. For arrays of objects, usually prefer Table view as the default. Implement the proposal by creating a unified rendering utility.
-``` bash
-tctl workflow list --table # default for most commands
-tctl workflow list --json
-```
+In places where we retrieve and print entity objects, support changing how the objects are printed through `--format` flag that accepts the following values:
+- `--format table` # default for most commands
+- `--format json`
+- `--format card` # new view
 
-#### - All commands do not support printing data in a form of cards (`{field name} {field value}` per row), which is especially useful in combination with `grep` or when describing a single object.
+For arrays of objects, usually prefer Table view as the default. Implement the proposal by creating a unified rendering utility.
 
-Add `--card` flag that will print entity object in a format `{field name} {field value}`.
+#### - All commands do not support printing data in a form of cards (`{field name}: {field value}` per row), which is especially useful in combination with `grep` or when describing a single object.
+
+Add `--format card` key that will print entity object in a format `{field name}: {field value}`.
 ``` bash
-tctl workflow list --card | grep WorkflowId
+tctl workflow list --format card | grep WorkflowId
 ```
 
 #### - Output is often noisy, single entity row may spill to the next rows and break the tables.
--- a/cli/006-cli-output-options.md
+++ b/cli/006-cli-output-options.md
@@ -8,30 +8,30 @@ Current Issues and improvement proposals:
 
 #### - Many commands do not support switching between Table and JSON views.
 
-In places where we retrieve and print entity objects, support changing how the objects are printed through `--format` flag that accepts the following values:
-- `--format table` # default for most commands
-- `--format json`
-- `--format card` # new view
+In places where we retrieve and print entity objects, support changing how the objects are printed through `--output` flag that accepts the following values:
+- `--output table` # default for most commands
+- `--output json`
+- `--output card` # new view
 
 For arrays of objects, usually prefer Table view as the default. Implement the proposal by creating a unified rendering utility.
 
 #### - All commands do not support printing data in a form of cards (`{field name}: {field value}` per row), which is especially useful in combination with `grep` or when describing a single object.
 
-Add `--format card` key that will print entity object in a format `{field name}: {field value}`.
+Add `--output card` key that will print entity object in a format `{field name}: {field value}`.
 ``` bash
-tctl workflow list --format card | grep WorkflowId
-```
-
-#### - Output is often noisy, single entity row may spill to the next rows and break the tables.
-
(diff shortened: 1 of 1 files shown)


robholland (commented):
Can we use go-template for this as `kubectl` does rather than the new `card` format?

bergundy:
> Can we use go-template for this as `kubectl` does rather than the new `card` format?

+1

bergundy:
Miss-clicked, sorry.

feedmeapples:
Opened PR with --format flag implementation. Card view is not implemented, just added a stab https://github.com/temporalio/tctl/pull/25

cli/006-cli-output-options.md
```diff
@@ -0,0 +1,36 @@
...
+### Improve commands output UX and rendering options
+
+Current Issues and improvement proposals:
+
+#### - Many commands do not support switching between Table and JSON views.
+
+In places where we retrieve and print entity objects, support changing how the objects are printed through `--json` and `--table` flags. For arrays of objects, usually prefer Table view as the default. Implement the proposal by creating a unified rendering utility.
+``` bash
+tctl workflow list --table # default for most commands
+tctl workflow list --json
```
mfateev, line 14: Should it be more like --output-format with _table_ being the default?

cli/006-cli-output-options.md
```diff
@@ -0,0 +1,36 @@
...
+Change the list commands to output 3-5 columns/fields by default. Add `--details` flag that will add few more of the ~most important fields. This behavior is similar to `nushell`'s `ls --long` flag
+``` bash
+tctl workflow list --details
+```
+
+#### - Limited support for changing the columns/fields in Table view. Only specific fields are explicitly supported through flags, ex. `--print_memo`.
+
+Proposal: Add `--field` flag so users can customize the fields to output in Table/Card (can be passed ). Add `--fields` flag to examine and print all available field names for --field input. If a field passed with `--field` is not found, then notify the user and print the result of `--fields` as a suggestion. 
+``` bash
+tctl workflow list --field Execution.RunId --field searchAttributes
```
mfateev, line 35: It should be RunId, not Execution.RunId.

cli/006-cli-output-options.md
feedmeapples: right, there was a discussion around this in other comments and already implemented in this way using `--format` flag
https://github.com/temporalio/tctl/pull/25

updating this proposal

feedmeapples:
This proposal is entirely implemented. There is one question i keep getting back to:

What is a nicer name for the flag to switch between printing as Table, JSON and Card.
Options:
-  `--view`
-  `--output`
-  `--format`
-  any other?

i've implemented the flag as `--format`, however thinking that maybe `--view` tells clearer what the flag does? At least in Web development "view" is a well defined term. 
What flag name would you prefer?
cc @robholland @bergundy @mfateev

bergundy:
I like the `--view` option but I've never seen it used in CLI before.
`--output` to me sounds like it would accept `stdout / file`.
I guess I'd stay with `--format`.

Can you find some references so we can see what others did?

feedmeapples:
> I like the `--view` option but I've never seen it used in CLI before.
> `--output` to me sounds like it would accept `stdout / file`.
> I guess I'd stay with `--format`.
> 
> Can you find some references so we can see what others did?

Thinking of what's the difference between "view" and "format", i would say that "view [layer]" is a higher level concept that is supposed to act friendlier than a lower level "format" operation.

Another thing to consider is the short version "-f":
  - i would like to use it for another flag "field". "-v" is currently free 

As for what others do:

docker:
`docker images --format <go-template>` - uses --format, though here it's actually low level unfriendly but powerful formatting. I would say that makes sense since they limit formatting only to that.
**tbu:**

cli/006-cli-output-options.md
```diff
@@ -0,0 +1,37 @@
...
+#### - All commands do not support printing data in a form of cards (`{field name}: {field value}` per row), which is especially useful in combination with `grep` or when describing a single object.
+
+Add `--format card` key that will print entity object in a format `{field name}: {field value}`.
+``` bash
+tctl workflow list --format card | grep WorkflowId
+```
+
+#### - Output is often noisy, single entity row may spill to the next rows and break the tables.
+
+Change the list commands to output 3-5 columns/fields by default. Add `--details` flag that will add few more of the ~most important fields. This behavior is similar to `nushell`'s `ls --long` flag
```
feedmeapples, line 27:
```suggestion
Change the list commands to output 3-5 columns/fields by default. Add `--long` flag that will add few more of the ~most important fields. This behavior is similar to `nushell`'s `ls --long` flag
```

cli/006-cli-output-options.md
```diff
@@ -0,0 +1,37 @@
...
+``` bash
+tctl workflow list --details
```
feedmeapples, line 29:
```suggestion
tctl workflow list --long
```

robholland:
`kubectl` has `--output` which I prefer to `--view` which feels like a very web-centric term to me. When `--output=template` they use `--template` to pass the actual template, which again feels clean and avoids the need to guess if an `--output` argument is actually a template.

feedmeapples:
> `kubectl` has `--output` which I prefer to `--view` which feels like a very web-centric term to me. When `--output=template` they use `--template` to pass the actual template, which again feels clean and avoids the need to guess if an `--output` argument is actually a template.

@bergundy updated the proposal to use --output flag. 

Also updated customizing the columns, now it's a `--columns` flag, plus a special case `--columns long` to simplify outputting more of common flags. The reason why `--columns` is separate from `--output` is to make it possible combine various outputs table/card with also customizing columns

cli/006-cli-output-options.md
feedmeapples: since this dynamically examines data and extracts fields, it has to be a full path to the field

rfwagner:
Card view sounds a bit like a homegrown yaml view. Any thought to just supporting yaml instead? This would be more on par with `kubectl` for whatever that is worth

bergundy:
Very good point @rfwagner, I'm wondering if this holds for nested fields too though, as part of the reason to have the card view is to make the output more `grep`able.

bergundy (approved):
Going over this now, it all LGTM, one thing I'm not certain about is the name of the `--columns` option, it's not really columns if it's not rendered in a table, maybe rename to `--fields` or `--attributes`?

feedmeapples:
> Going over this now, it all LGTM, one thing I'm not certain about is the name of the `--columns` option, it's not really columns if it's not rendered in a table, maybe rename to `--fields` or `--attributes`?

`--fields` sounds good to me! Actually had this naming initially

PR #32, opened 2021-06-15
tctl: Consistent coloring [p1]

feedmeapples:
- Origin issue: 

# Summary

Improves coloring when printing commands outputs.

Allows to disable coloring and increase contrast in the output

# Basic example
``` bash
tctl workflow run --workflow_type wtype --taskqueue queue --color never
```

# Motivation

Brings consistency into what colors are used and when. Allows to disable coloring 

# Detailed design

in doc

# Drawbacks
There are tradeoffs to choosing any path. Attempt to identify them here.

# Alternatives

# Adoption strategy

Not a breaking change. There is going to be an additional flag `--color` visible through `--help`

# How we teach this

Providing a `--help` text for the new flag

# Unresolved questions

### DIFF
--- a/cli/007-cli-color.md
+++ b/cli/007-cli-color.md
@@ -8,6 +8,6 @@ Good usage of coloring can help users to read the output and direct attention to
 
 - Standartize the colors used in printing
 - Work through commands and improve the coloring
-- add a flag `--no-color` to disable coloring alltogether
+- add a flag `--color` that takes the following values: `auto` (enable in tty/ disable in non-tty), `always`, `disabled` 
 - low priority: colorize command and flag names in `--help` output
 - low priority: add built-in coloring of `--json` output. Piping into `jq` can be used by users as an alternative so low priority.
--- a/cli/007-cli-color.md
+++ b/cli/007-cli-color.md
@@ -9,5 +9,5 @@ Good usage of coloring can help users to read the output and direct attention to
 - Standartize the colors used in printing
 - Work through commands and improve the coloring
 - add a flag `--no-color` to disable coloring alltogether
-- low priority: colorize the command and flag names in `--help` output
+- low priority: colorize command and flag names in `--help` output
 - low priority: add built-in coloring of `--json` output. Piping into `jq` can be used by users as an alternative so low priority.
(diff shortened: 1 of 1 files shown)


robholland (changes requested):
Unclear if you are just suggesting a `--no-color` flag to remove colors, or are suggesting making changes to which colors are used without this flag.

robholland (commented):
Missed that in the doc you say improve the colors. `improve` is subjective so can't really comment other than to say that more consistency sounds good, but examples would help to know what kind of thing you think is inconsistent currently.

cli/007-cli-color.md
```diff
@@ -0,0 +1,13 @@
...
+- RFC PR:
+- Issue:
+
+### Better coloring
+
+Good usage of coloring can help users to read the output and direct attention to the most interesting parts.
+
+- Standartize the colors used in printing
+- Work through commands and improve the coloring
+- add a flag `--no-color` to disable coloring alltogether
```
bergundy, line 11: We could have `--color/--no-color` with color being the default for TTY and no color for non-TTY output.
You can override the default via the proposed configuration mechanism.

feedmeapples: yea better control over enabling and disabling coloring in tty/non-tty makes sense
 
i've noticed other modern CLIs achieve this by providing the following single flag with values:

- nushell's `ls --help` gives the following:
      -c, --color <when>                 When to use colors: never, *auto*, always

- same with `bat --help`:
            --color <when>                When to use colors (*auto*, never, always).
            
this kinda makes sense since this is just one flag (doesn't pollute help with many flags) and allows to switch between three modes: auto that checks for tty, always or never coloring

bergundy: Yeah, that's a really good suggestion, it's been standard in unix CLI tools for a while now, `grep` is another example.

feedmeapples:
Created PR https://github.com/temporalio/tctl/pull/26

cli/007-cli-color.md
feedmeapples:
```suggestion
- add a flag `--color` that takes the following values: `auto` (enable in tty/ disable in non-tty), `always`, `disabled` 
```

bergundy (approved):

PR #35, opened 2021-06-16
tctl: Format timestamp fields using a flag [p1]

feedmeapples:
- Origin issue: 

# Summary

Adds control over how timestamp fields are printed 

# Basic example

``` bash
$ tctl workflow list --time-format relative # default in tty
$ tctl workflow list --time-format iso # default in non-tty
$ tctl workflow list --time-format raw
$ tctl workflow list --time-format "Jan 2 15:04:05 2006 MST" # ie. custom format
```

# Motivation

Helps reading timestamp values in Tables and Cards.
Reduces noise in tty Tables by default, since relative format (`6 hrs ago`) is shorter and less-noisy than others

# Detailed design

in doc

# Drawbacks

Why should we *not* do this? Please consider:

- changes the default format

# Alternatives

What other designs have been considered? What is the impact of not doing this?

# Adoption strategy

Not a breaking change, though changes the default timestamp format 

# How we teach this

`--help` for this flag should be clear and show possible values

# Unresolved questions

### DIFF
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -1,0 +1,14 @@
+- Start Date: 2021-06-11
+- RFC PR:
+- Issue:
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow swithing the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--time-format` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`. Use as default in non-tty environment
+- `datetime` flag to print in a format `2021-06-16T00:21:24Z` (ie. time.RFC3339)
+- `date` to print in a format `2021-06-15`
+- `time` to print in a format `00:21:24`
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -9,4 +9,4 @@ When printing data in `--table` or `--card` views, allow switching the format of
 Add `--time-format` flag with possible values:
 - `relative` to print in a format `6 hours ago`. Use as default format in tty environment
 - `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
-- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339)
+- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -9,6 +9,6 @@ When printing data in `--table` or `--card` views, allow switching the format of
 Add `--time-format` flag with possible values:
 - `relative` to print in a format `6 hours ago`. Use as default format in tty environment
 - `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`. Use as default in non-tty environment
-- `datetime` flag to print in a format `2021-06-16T00:21:24Z` (ie. time.RFC3339)
+- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339)
 - `date` to print in a format `2021-06-15`
 - `time` to print in a format `00:21:24`
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -10,5 +10,4 @@ Add `--format-time` flag with possible values:
 - `relative` to print in a format `6 hours ago`. Use as default format in tty environment
 - `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
 - `iso` to print in a format `2021-06-16T00:21:24.667Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
-- `tz=utc` to print for specific timezone
 - `Jan 2 15:04:05 2006 MST` # ie. custom format using https://golang.org/pkg/time/#Time.Format
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -10,4 +10,4 @@ Add `--time-format` flag with possible values:
 - `relative` to print in a format `6 hours ago`. Use as default format in tty environment
 - `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
 - `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
-- `Jan 2 15:04:05 2006 MST` # ie. custom format
+- `Jan 2 15:04:05 2006 MST` # ie. custom format using https://golang.org/pkg/time/#Time.Format
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -9,6 +9,6 @@ When printing data in `--table` or `--card` views, allow switching the format of
 Add `--format-time` flag with possible values:
 - `relative` to print in a format `6 hours ago`. Use as default format in tty environment
 - `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
-- `iso` to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
+- `iso` to print in a format `2021-06-16T00:21:24.667Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
 - `tz=utc` to print for specific timezone
 - `Jan 2 15:04:05 2006 MST` # ie. custom format using https://golang.org/pkg/time/#Time.Format
--- a/cli/010-cli-datetime-format.md
+++ b/cli/010-cli-datetime-format.md
@@ -4,7 +4,7 @@
 
 ### Format timestamp fields using a flag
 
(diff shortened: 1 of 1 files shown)


robholland:
golang has https://golang.org/pkg/time/#Time.Format which takes a formatted date string for Jan 2 15:04:05 2006 MST and copies the same format for the supplied datetime value. Might be nice to present this to the user so they can choose exactly how the timestamp should appear. I'm wary of things like `datetime` having meaning because would that be with seconds? What about timezone? Most date related terms are ambiguous. So perhaps `relative`, `raw` and `custom` or something along those lines? Even `date` could be any of 2006-01-02, 01/02/2006 or 02/01/2006

robholland (commented):
golang has https://golang.org/pkg/time/#Time.Format which takes a formatted date string for Jan 2 15:04:05 2006 MST and copies the same format for the supplied datetime value. Might be nice to present this to the user so they can choose exactly how the timestamp should appear. I'm wary of things like `datetime` having meaning because would that be with seconds? What about timezone? Most date related terms are ambiguous. So perhaps `relative`, `raw` and `custom` or something along those lines? Even `date` could be any of 2006-01-02, 01/02/2006 or 02/01/2006

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
+- Start Date: 2021-06-11
+- RFC PR:
+- Issue:
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow switching the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--time-format` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
```
bergundy, line 10: This one is going to be really useful.

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`. Use as default in non-tty environment
+- `datetime` flag to print in a format `2021-06-16T00:21:24Z` (ie. time.RFC3339)
```
bergundy, line 12: Lets just call this `iso`? (BTW we can show subseconds here too).

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+- RFC PR:
+- Issue:
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow switching the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--time-format` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`. Use as default in non-tty environment
```
bergundy, line 11: I'd default to iso, all of these spaces make it hard to capture.

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+- `datetime` flag to print in a format `2021-06-16T00:21:24Z` (ie. time.RFC3339)
+- `date` to print in a format `2021-06-15`
+- `time` to print in a format `00:21:24`
```
bergundy, line 14: I think date is redundant since you can always grep on an iso datetime string.
Not sure I see that much value in displaying only the time, relative is much more meaningful to humans.

bergundy (commented):
I wonder if some of this could be achieved by having a `--format TEMPLATE_STRING` option.

cli/010-cli-datetime-format.md
feedmeapples: kept it because `workflow list` currently shows just time by default. Agree that this format doesn't seem useful. Removing

cli/010-cli-datetime-format.md
feedmeapples:
```suggestion
- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339)
```

cli/010-cli-datetime-format.md
feedmeapples:
```suggestion
```

cli/010-cli-datetime-format.md
feedmeapples:
```suggestion
- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
```

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,12 @@
...
+- Issue:
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow switching the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--time-format` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
+- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339)
```
feedmeapples, line 12:
```suggestion
- `iso` flag to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
```

feedmeapples:
> golang has https://golang.org/pkg/time/#Time.Format which takes a formatted date string for Jan 2 15:04:05 2006 MST and copies the same format for the supplied datetime value. Might be nice to present this to the user so they can choose exactly how the timestamp should appear. I'm wary of things like `datetime` having meaning because would that be with seconds? What about timezone? Most date related terms are ambiguous. So perhaps `relative`, `raw` and `custom` or something along those lines? Even `date` could be any of 2006-01-02, 01/02/2006 or 02/01/2006

how about the following possible values for the flag then? :
 - relative
 - iso
 - raw
 - "Jan 2 15:04:05 2006 MST" - anything that is not `relative`, `iso` or `raw` is treated as "custom" and passed to Time.Format

feedmeapples:
Opened PR https://github.com/temporalio/tctl/pull/24/files

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow switching the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--format-time` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
+- `iso` to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
+- `tz=utc` to print for specific timezone
```
feedmeapples, line 13: or just `utc`/any other timezone and compare against timezones dictionary to tell to process for timezone?

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+- RFC PR:
+- Issue:
+
+### Format timestamp fields using a flag
+
+When printing data in `--table` or `--card` views, allow switching the format of the timestamp fields (ie. `google.protobuf.Timestamp`) 
+
+Add `--format-time` flag with possible values:
+- `relative` to print in a format `6 hours ago`. Use as default format in tty environment
+- `raw` to print in a format  `2021-06-16 00:21:24.472487365 +0000 UTC`.
```
feedmeapples, line 11: @alexshtin is this format used anywhere? I would just remove this

feedmeapples:
@robholland @bergundy @alexshtin adding support for specific timezones due to user request here https://github.com/temporalio/temporal/issues/1607

feedback is welcomed:
https://github.com/temporalio/proposals/pull/35/files#diff-fd43e76660a0644c8d5027ed26ef9dbb8086737f8145733070445413a18f4133R13

cli/010-cli-datetime-format.md
feedmeapples: having `tz=` makes it easier to provide proper error message and print a list of possible values

cli/010-cli-datetime-format.md
bergundy: We should show at least millisecond accuracy

cli/010-cli-datetime-format.md
```diff
@@ -0,0 +1,14 @@
...
+- `iso` to print in a format `2021-06-16T00:21:24Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
```
bergundy, line 12: Let's print out millisecond resolution here
```suggestion
- `iso` to print in a format `2021-06-16T00:21:24.667Z` (ISO 8601, time.RFC3339). Use as default in non-tty environment
```

cli/010-cli-datetime-format.md
bergundy: How does this work in combination with the other options?
`--format-time iso,tz=utc`?
I'm not sure `tz` fits here since it's not a format.

cli/010-cli-datetime-format.md
feedmeapples: sg

cli/010-cli-datetime-format.md
feedmeapples: i'll track TZ in a separate proposal and remove it from timestamp formatting

bergundy (approved):
Only thing I would maybe change is to rename the flag `--time-format` instead of `--format-time` but other than that LGTM.

feedmeapples:
> Only thing I would maybe change is to rename the flag `--time-format` instead of `--format-time` but other than that LGTM.

sg

PR #38, opened 2021-07-27
Add new versioning proposal

Sushisource:
See doc for details


core/versions_2.0.md
```diff
@@ -0,0 +1,71 @@
...
+
+If it's the case that very few users ever actually need version ranges, which seems likely, then a much simpler API can be presented. In fact, it's still possible to conceptually represent version ranges with it as well, if needed.
+
+Instead of `get_version("change-id", minver, maxver) -> number`, consider: `changed("change-id") -> bool`
+
+If the workflow code is executing for the first time, it will write a marker into history meaning "change change-id happened". If it is replaying history, it will verify such a marker exists and return true. If the marker does not exist, it will return false. Versions of the code without the call will fail (more on that in other section).
+
+Version ranges can still be simulated if needed like:
+```rust
+if changed("change-ver-1") {
```
roselander, line 19: I think you want to swap "change-ver-1" and "change-ver-2" here, otherwise the 2nd clause will never apply

core/versions_2.0.md
```diff
@@ -0,0 +1,71 @@
+# Workflow versioning 2.0
+
+This proposal introduces two key changes to what is called `getVersion` in the existing SDKs, but will likely be renamed in subsequent ones (more on naming later). The key changes are:
+
+* Introduce a different API for versioning that does not involve a version range, but instead simply returns a boolean indicating if a named change has happened or not. Ex: `hasChange("change-id")`. TBD if we would eliminate the range-based API, or leave it for advanced users.
```
roselander, line 5: I think the `changed("change-x")` call is a bit confusing, b/c it looks like "Has 'change-x' changed?".

`hasChange` read a bit more natural to me, but I would also throw `supports("feature-x")` into the ring

core/versions_2.0.md
Sushisource: Ah, yeah that's a good point thanks

sw-yx:
this looks great. `getVersion` always troubled me since we also dont make it easy to view workflow versions in temporal web.

core/versions_2.0.md
robzienert: +1 `hasChange`, but -1 `supports`, the latter one is too ambiguous, I think.

zomble:
From on outsider's perspective I was unsure what problem `version` was solving, that some simple logic or if I was genuinely drastically changing something a new workflow entirely would solve.

I assumed I was missing something. As I couldn't see how temporal / workers was actually running all the different "versions" of a workflow, or a simple way to run all the different versions. (Again, could totally be missing something).

Reading the proposal makes a lot more sense to me on what problem workflow versioning is trying to solve. Feel more inline with what change management is really like in the real world.

👍🏻

core/versions_2.0.md
bergundy: Another option is `usePatch('patch-x')`, I find it conveys the concept better, because it's not about if you have the change or patch but more about if you should use it for a specific branch (yes, if you're not replaying or replaying code that used this patch before).

core/versions_2.0.md
```diff
@@ -0,0 +1,71 @@
...
+fn workflow() {
+   // ...previous code....
+   must_have_change("change_id")
+   activity_b().await;
+}
+```
+
+The above code can now fail properly if it encounters any histories which lack the version marker (were run by the old code), while still allowing the user to eliminate the old code. This new code does not write a marker if it is creating this part of history for the first time, rather than replaying.
+
+This means the user cannot deploy this "phase 2" code until they know there are no more workers out there running the original code (as, if such a worker were to encounter a history generated by this latest code, it would not see any marker and proceed). Hence, it is important to provide a way for users to query workers so they can know this (apparently, already exists with enhanced visibility?). After no more workflow runs exist which still contain the marker, the `must_have_change` call may be removed entirely.
```
bergundy, line 71: Using a different marker as suggested in my comment above means that it's safe to deploy version 3 alongside version 2.
The only thing stopping you from deploying version 3 is WF histories without the marker (produced by version 1), it's still "safe" but those WFs cannot be processed and their tasks will fail.

bergundy:
To clarify how the different marker works, I've made this table:

- `x` - no marker
- `m` - normal marker generated with `has_change('changed-id')`
- `d` - deprecated marker, generated with  e.g `must_have_change` or `has_change('change-id', DEPRECATED)`

| hist | wf  | outcome |
| ---- | --- | ------------------------------------ |
| x    | x   |  WF executes code                 |
| m    | x   |  WF does not support this version |
| d    | x   | WF executes code                 |
| x    | m   | WF executes old code             |
| m    | m   | WF executes new code             |
| d    | m   | WF executes new code             |
| x    | d   | WF executes old code             |
| m    | d   | WF executes new code             |
| d    | d   | WF executes new code             |

The restrictions of using these markers are:
- Deprecated marker version cannot be deployed while there are still **workers** running version 2
- Old code cannot be deleted while there are incomplete WFs with **histories** generated with version 1
- Deprecated marker cannot be removed while there are incomplete WFs with **histories** generated with version 1 or 2

The major downside to this is the last restriction, the only reason you have to deploy version 2 before 3 (with old code included) is to allow workers running version 1 to fail if they see the marker.

core/versions_2.0.md
Sushisource: I think you mean safe to deploy 3 alongside 1, (in my original description 3 w/ 2 is still safe) but yeah good point.

Sushisource:
@bergundy I was originally thinking `x hist / d code` would be a failure, in which case you trade

* Old code cannot be deleted while there are incomplete WFs with histories generated with version 1
for
* Version 1 workers must be gone and there are no longer any markerless histories before you deploy version 3

Which I think is what you were saying too but it ended up wrong in the table?


Also
* Deprecated marker version cannot be deployed while there are still workers running version 2 
Must be wrong - workers running version 2 will write the normal marker which can be processed by v3 workers with deprecated marker command. Otherwise there's no way to make the transition. I think you mean v1 which would be consistent with the above?

core/versions_2.0.md
```diff
@@ -0,0 +1,71 @@
+# Workflow versioning 2.0
+
+This proposal introduces two key changes to what is called `getVersion` in the existing SDKs, but will likely be renamed in subsequent ones (more on naming later). The key changes are:
+
+* Introduce a different API for versioning that does not involve a version range, but instead simply returns a boolean indicating if a named change has happened or not. Ex: `hasChange("change-id")`. TBD if we would eliminate the range-based API, or leave it for advanced users.
+* It will no longer be valid to replay a workflow history containing a version marker against code which does not have a corresponding call to the version API.
```
GreyTeardrop, line 6: I wonder if that could make it harder to evolve workflows for a long period of time, as there would be no way to remove any checks for any versions anymore?

bergundy:
Eventually we got on the same page, the mix-up was due to me assuming it was clear that version 3 might still have the old code, e.g:
```ts
if (changed('change-1', { deprecated: true })) {
  // new code
} else {
  // old code
}
```
vs.
```ts
mustHaveChange('change-1');
// new code
```

Sushisource:
OK - so after yet more discussion we have landed on proceeding with the hasChange/mustHaveChange approach and eschewing deprecated markers entirely. We also discussed a hybrid whole-workflow-versioning / markers approach where we could use entire workflow versioning info to make better decisions about what to do with the code if we *do* re-introduce the concept of deprecated markers.

I will add this as a possible future plan to the proposal, and make the proposal reflect what I'm actually implementing, before merging.

PR #39, opened 2021-08-23
tctl: Support multiple configurations

feedmeapples:
- Origin issue: 

# Summary

Allows to create multiple tctl configurations and switch between them.

# Basic example

```bash
$ tctl config set-context <new>
```

# Motivation

Useful for cases when devs have several Temporal service environments (development, production..) and want to have a quick way to switch between configurations

# Detailed design

in the file

# Drawbacks

...

# Alternatives

1. Use an approach implemented here https://github.com/jlegrone/tctx
It uses json as the config file and has pre-detemined configuration structure

2. A similar behavior can be achieved through alias command, though with some drawbacks such as complicated scripting:
``` bash
tctl config set alias.cluster1 "config set certPath ~/cert1 keyPath ~/key1"
tctl config set alias.cluster2 "config set certPath ~/cert2 keyPath ~/key2"
```

Then enabling cluster2 configs:
``` bash
tctl cluster2
``` 

# Adoption strategy

It's going to be available only for the new tctl UX which is currently experimental. Not a breaking change for old tctl

# How we teach this

Provide a help output and describe in http://docs.temporal.io/ 

# Unresolved questions


cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+Users may have multiple Temporal service environments (development, production, ..) with their own respective mTLS configs, namespaces, address etc. Current implementation of tctl config feature only supports configuring a single environment and doesn't help at quickly switching between multiple.
+
+##### Proposal
+
+To make it easy for users to switch between environments, allow configuring multiple instead of just one and provide command to switch between them.
+
+Syntax to activate environment configs:
+
+```bash
```
bergundy, line 17:
```suggestion
```shellsession
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+Split configuration keys into 1) environment scope - keys residing in a scope of specific environments 2) global scope - out of any specific environment
+
+Global scope keys:
+
+- "active" - name of active environment
+- "environments" - sequence of environments
+- "alias" - sequence of command aliases
+
+Environment scope keys:
```
bergundy, line 38: What's the default environment? How do I use this with a single temporal env?

bergundy (commented):
Overall LGTM.

This seems like a pretty advanced feature, is there high demand for it?

cli/011-cli-multi-config.md
feedmeapples: thinking of `local` as the default env

feedmeapples:
> Overall LGTM.
> 
> This seems like a pretty advanced feature, is there high demand for it?

there is this project that implements this feature for temporal cli, though through json config https://github.com/jlegrone/tctx

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+Syntax to activate environment configs:
+
+```bash
+$ tctl config use-environment <name>
+Active environment: <name>
+```
+
+If the environment doesn't yet exist in config, create it and set as active
+
+```bash
```
feedmeapples, line 24:
```suggestion
```shellsession
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+$ tctl --environment cluster-1 workflow list
+```
+
+##### Example of environment aware configuration file:
+
+```yml
+active: development
+environments:
+  - name: development
```
feedmeapples, line 63: local

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+Example:
+```bash
+$ tctl --environment cluster-1 workflow list
+```
+
+##### Example of environment aware configuration file:
+
+```yml
+active: development
```
feedmeapples, line 61:
```suggestion
active: local
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+```shellsession
+$ tctl config use-environment <name>
+Active environment: <name>
+```
+
+If the environment doesn't yet exist in config, create it and set as active
+
+```shellsession
+$ tctl config use-environment <new>
```
feedmeapples, line 25:
```suggestion
$ tctl config use-environment <name>
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+Environment <new> record doesn't exist, adding..
```
feedmeapples, line 26:
```suggestion
Environment <name> record doesn't exist, adding..
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+- "active" - name of active environment
+- "environments" - sequence of environments
+- "alias" - sequence of command aliases
+
+Environment scope keys:
+
+- namespace
+- address
+- data-converter-plugin
```
jlegrone, line 42: Nit: IMO it's nice if these keys map directly to `tctl` flag names, eg. `data_converter_plugin`. This allows users to look at `tctl -h` as a quick reference for which config options are available.

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+
+```shellsession
+$ tctl config use-environment <name>
+Active environment: <name>
+```
+
+If the environment doesn't yet exist in config, create it and set as active
+
+```shellsession
+$ tctl config use-environment <name>
```
jlegrone, line 25: I suppose this can be added later or configured via an alias, but something short like `tctl use <environment-name>` could be nice to have out of the box.

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+- RFC PR:
+- Issue:
+
+### Configuring multiple environments
+
+##### Problem
+
+Users may have multiple Temporal service environments (development, production, ..) with their own respective mTLS configs, namespaces, address etc. Current implementation of tctl config feature only supports configuring a single environment and doesn't help at quickly switching between multiple.
+
+##### Proposal
```
jlegrone, line 11: During onboarding it's pretty common to have a command to copy/paste or script to run that sets up a user's development environment. Could the proposal include an example of what a script to add or update an environment would look like, if that is a supported use case from the CLI?

In particular I'm wondering if `tctl config use-environment <name>` creates a new environment if none exists -- this could lead to some config file pollution when typos happen (and is why adding and updating are separate commands in `tctx`).

cli/011-cli-multi-config.md
jlegrone: Is a "default" environment needed? I think if there is no active environment present then `tctl` default flag values should apply. To override defaults for a single temporal env without flags for subsequent commands, you can create a single environment and set it as active.

Maybe a special `default` environment could automatically be created though if someone attempts to set a config value without an active environment set?

cli/011-cli-multi-config.md
jlegrone: Ah, nevermind I see there's a proposal for changing the flag naming conventions. https://github.com/temporalio/proposals/blob/master/cli/005-cli-hyphens.md

cli/011-cli-multi-config.md
bergundy: I tend to agree, there should be a separate command for creating a new environment.

cli/011-cli-multi-config.md
bergundy: Good point, +1

cli/011-cli-multi-config.md
bergundy: I'm thinking about the scenario where a user adds the first env.
Does it automatically get set as `active`? If so how do users connect to their "local" cluster again? Do they have to explicitly create another `local` env?
Having a `local` env in the config could be convenient and as a bonus it's now customizable.

So I propose this DX:
```shellsession
$ tctl create-env new-env --address remote-cluster:7223
Environment `new-env` created, available environments:
- local (CURRENT)
- new-env
To switch to environments use `tctl use <ENV_NAME>
```

cli/011-cli-multi-config.md
```diff
@@ -0,0 +1,74 @@
...
+To make it easy for users to switch between environments, allow configuring multiple instead of just one and provide command to switch between them.
+
+Syntax to activate environment configs:
+
+```shellsession
+$ tctl config use-environment <name>
+Active environment: <name>
+```
+
+If the environment doesn't yet exist in config, create it and set as active
```
bergundy, line 22: I'm missing a command to update an environment's configuration.

feedmeapples: i was thinking that `tctl config set <property> <value>` would update the currently active environment. Then if `--environment` [see link](https://github.com/temporalio/proposals/pull/39/files#diff-ac63135de4d849c82417254728c180bed798eeb90b570b5ae3248ccb9d0150bfR49) is passed, update the passed environment

cli/011-cli-multi-config.md
feedmeapples: sg!

feedmeapples:
@bergundy PR [Allow configuring multiple flags at a time](https://github.com/temporalio/tctl/pull/140)

feedmeapples:
merging as the feature is implemented

PR #44, opened 2021-10-04
Revise Typescript Workflow API

bergundy:
- Origin issue: 

# Summary

Another design attempt for the Typescript Workflow API

# Basic example

```ts
import { Signal, Query, condition } from '@temporalio/workflow';

export const unblockSignal = new Signal<[]>('unblock');
export const isBlockedQuery = new Query<[], boolean>('isBlocked');

// Types are inferred without an additional type definition
export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
  let isBlocked = true;
  unblockSignal.handler = () => (isBlocked = false);
  isBlockedQuery.handler = () => isBlocked;
  await condition(() => !isBlocked);
}
```

# Motivation

* Maintain type inference capabilities - Mandatory for the Typescript SDK
* Dynamic signals and queries registration first - A single API for Workflow handler registration
* Componentization of related functionality - increase reusability
* Implicit interface declarations - less boilerplate
* No global variables - to support running Workflows in less isolated runtimes

# Detailed design

See changed files

### DIFF
--- a/node/workflow-api.md
+++ b/node/workflow-api.md
@@ -1,0 +1,322 @@
+# 3rd revision of the Typescript SDK Workflow API
+
+### Current API
+
+```ts
+import { Trigger } from '@temporalio/workflow';
+
+export type MyWorkflow = (
+  arg1: number,
+  arg2: string
+) => {
+  execute(): Promise<void>;
+  signals: {
+    unblock(): void;
+  };
+  queries: {
+    isBlocked(): boolean;
+  };
+};
+
+export const myWorkflow: MyWorkflow = (arg1, arg2) => {
+  const unblocked = new Trigger<void>();
+  let isBlocked = true;
+  return {
+    // Main workflow logic
+    async execute() {
+      await unblocked;
+      isBlocked = false;
+    },
+    signals: {
+      unblock() {
+        unblocked.resolve();
+      },
+    },
+    queries: {
+      isBlocked() {
+        return isBlocked;
+      },
+    },
+  };
+};
+```
+
+#### Issues
+
+- Types must be explicitly declared (To make [eslint happy](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/explicit-module-boundary-types.md))
+- Very verbose
+- Trigger (ResolvablePromise) is a foreign concept to JS developers
+- Trigger doesn't provide all of the functionality for this workflow, an additional `isBlocked` boolean variable is required
+- Signals and queries must be defined at the top level in order for types to be inferred.
+
+#### Advantages
+
+- Types for all Workflow handlers can be inferred at Workflow handle creation
+
+### Goals for the redesign
+
+- Maintain type inference capabilities - Mandatory for the Typescript SDK
+- Dynamic signals and queries registration first - A single API for Workflow handler registration
+- Componentization of related functionality - increase reusability
+- Implicit interface declarations - less boilerplate
+- No global variables - to support running Workflows in less isolated runtimes
+
+### New proposal
+
+#### Remove the `Trigger` API in favor of `condition`
+
+`condition` is similar to `Workflow.await` in Java, it returns a Promise and accepts a closure which resolve the Promise when it returns `true`.
+
+```ts
+import { condition } from '@temporalio/workflow';
+
+let isBlocked = true;
+await condition(() => !isBlocked);
+```
+
+#### Remove the Workflow factory wrapper and always use dynamic handler registration
+
+`workflows.ts`
+
+```ts
+import { Signal, Query, condition } from '@temporalio/workflow';
+
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new Query<boolean /* return type */, [] /* args */>('isBlocked');
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
+  let isBlocked = true;
+  unblockSignal.handler = () => (isBlocked = false);
+  isBlockedQuery.handler = () => isBlocked;
+  await condition(() => !isBlocked);
+}
+```
+
+- Note how all unblocking related functionality is co-located now, we can use this to split our Workflow into components, for example we could have a reusable "unblock" component.
+
+  ```ts
+  export async function unblocked() {
+    let isBlocked = true;
+    unblockSignal.handler = () => (isBlocked = false);
+    isBlockedQuery.handler = () => isBlocked;
+    await condition(() => !isBlocked);
+  }
+
+  export async function myWorkflow1() {
+    await unblocked();
+  }
+
+  export async function myWorkflow2() {
+    await unblocked();
+  }
+  ```
+
+#### WorkflowClient usage
+
+`execute-workflow.ts`
+
+```ts
+import { WorkflowClient } from '@temporalio/client';
+import { myWorkflow, unblockSignal, isBlockedQuery } from './workflows';
+
+const client = new WorkflowClient();
+const handle = client.createWorkflowHandle(myWorkflow);
+await handle.start();
+await handle.query(isBlocked /* no args */); // true
+await handle.signal(unblockSignal /* no args */);
+await handle.query(isBlocked /* no args */); // false
+await handle.query<[number, string], string>('queryWithUnexportedType', 1, 'a'); // should return a string
+await handle.result(); // undefined
+```
+
+### Alternatives considered
+
+#### Signal as Trigger
+
+```ts
+export const unblockWithSignal = async () => {
+  let isBlocked = true; // queryable
+  await new Signal('unblock');
+  isBlocked = false;
+};
+```
+
+##### Checklist
+
+| Goal                                      |                    |
+| ----------------------------------------- | :----------------: |
+| Verbosity                                 | :star::star::star: |
+| Type inference for client                 |        :x:         |
+| Dynamic handler registration              | :white_check_mark: |
+| Componentization of related functionality | :white_check_mark: |
+| Remove interface declarations             | :white_check_mark: |
+| No global variables                       | :white_check_mark: |
+
+##### Additional issues
+
+- Does not support multiple signals of the same type or arguments (can be added with a `Signal(name, handler)` signature).
+- Local variables cannot be automatically queryable - a practice which is questionable since it potentially leaks Workflow implementation details
+
+#### Signal/Trigger with types
+
+```ts
+type UnblockSignal = {
+  // can reuse for type safety when invoking, see below
+  name: 'unblock';
+  kind: void;
+};
+
+export const unblockWithSignal = async () => {
+  let isBlocked = true; // queryable
+  await new Signal<UnblockSignal>('unblock');
+  isBlocked = false;
+};
+```
+
+Same issues as the previous example with the additional issue that type definition requires the user to type the signal name and type in both the Workflow and client code.
+
+#### QueryVariable / Signal as async iterator
+
+```ts
+export const unblockWithSignal = () => ({
+  async execute() {
+    await this.signals.unblock.next();
+    this.queries.isBlocked.value = false;
+  },
+  signals: {
+    unblock: new Signal<void>(), // A channel / async iterator
+  },
+  queries: {
+    isBlocked: new QueryVariable<boolean>(true),
+  },
+});
+```
+
+-- OR --
+
+```ts
+export const signalLooper = () => {
+  let state = 'idle';
+  return {
+    async execute() {
+      for await (state of this.signals.changeState);
+    },
+    signals: {
+      changeState: new Signal<string>(),
+    },
+    queries: {
+      isRunning() {
+        state === 'running';
+      },
+    },
+  };
+};
+```
+
+##### Checklist
+
+| Goal                                      |                    |
+| ----------------------------------------- | :----------------: |
+| Verbosity                                 |    :star::star:    |
+| Type inference for client                 | :white_check_mark: |
+| Dynamic handler registration              |        :x:         |
+| Componentization of related functionality |        :x:         |
+| Remove interface declarations             |        :x:         |
+| No global variables                       | :white_check_mark: |
+
+##### Additional issues
+
+- Using channels complicates writing Workflow code, this model is used in the golang SDK and requires the user to drain all signals from the channel before completing the Workflow to ensure all signals are processed. This is critical for Temporal which aims to support financial use-cases where missing signal delivery could result in $$$ lost.
+
+#### Top level `QueryVariable`
+
+```ts
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new QueryVariable('isBlocked', true);
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
+  unblockSignal.handler = () => (isBlocked.value = false);
+  await condition(() => !isBlocked.value);
+}
+```
+
+##### Checklist
+
+| Goal                                      |                    |
+| ----------------------------------------- | :----------------: |
+| Verbosity                                 | :star::star::star: |
+| Type inference for client                 | :white_check_mark: |
+| Dynamic handler registration              | :white_check_mark: |
+| Componentization of related functionality | :white_check_mark: |
+| Remove interface declarations             | :white_check_mark: |
+| No global variables                       |        :x:         |
+
+#### Handlers as Workflow arguments
+
+1. With factory
+
+```ts
+export const myWorkflow: MyWorkflow = ({ signals, queries }) => (arg1, arg2) => {
+  let isBlocked = true;
+  queries.isBlocked = () => isBlocked;
+  signals.unblock = () => (isBlocked = false);
+  await condition(() => !isBlocked);
+};
+```
+
+- :warning: Eslint doesn't like this and the `MyWorkflow` type must be explicitly declared
+
+2. In first argument
+
+```ts
+export async function myWorkflow({ signals, queries }, arg1, arg2) {
+  // ...
+}
+```
+
+- :warning: Makes calling the client's execute function a bit weird since the first argument must be omitted
+
+3. In last argument
+
+```ts
+export async function myWorkflow(arg1, arg2, { signals, queries }) {
+  // ...
+}
+```
+
+- :warning: Makes calling the client's execute function a bit weird, although not as weird as the avove
+
+4. Args in props
+
+```ts
+export async function myWorkflow({ signals, queries: args: [arg1, arg2] }: MyWorkflowProps) {
+  // ...
+}
+```
+
+- :warning: Client execute function is disconnected from workflow definition / implementation
+
+5. React style props
+
+```ts
+export async function myWorkflow({ signals, queries: arg1, arg2 }: MyWorkflowProps) {
+  // ...
+}
+```
+
+- :warning: Client execute function is disconnected from workflow definition / implementation.
+- :warning: Makes cross SDK work weird since we support positional arguments in other SDKs.
+
+##### Checklist
+
+| Goal                                      |                    |
+| ----------------------------------------- | :----------------: |
+| Verbosity                                 | :star::star::star: |
+| Type inference for client                 | :white_check_mark: |
+| Dynamic handler registration              |        :x:         |
+| Componentization of related functionality |        :x:         |
+| Remove interface declarations             |        :x:         |
+| No global variables                       | :white_check_mark: |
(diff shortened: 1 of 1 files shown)


node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new Query<[] /* args */, boolean /* return type */>('isBlocked');
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
+  let isBlocked = true;
+  unblockSignal.handler = () => (isBlocked = false);
+  isBlockedQuery.handler = () => isBlocked;
+  await condition(() => !isBlocked);
```
cretz, line 93: Curious, how frequently is `condition` invoked internally?

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new Query<[] /* args */, boolean /* return type */>('isBlocked');
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
+  let isBlocked = true;
+  unblockSignal.handler = () => (isBlocked = false);
+  isBlockedQuery.handler = () => isBlocked;
```
cretz, line 92: Are these two handlers mutating a global field on `unblockSignal` and `isBlockedQuery`? Or is there some internal mechanism to have it specific to this workflow? What if `myWorkflow` was called concurrently, any clashes here? Does there need to be a `unblockedSignal.forWorkflow(this).handler = `? (pardon my ignorance on the internals of how global state is managed per workflow).

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+
+#### Remove the Workflow factory wrapper and always use dynamic handler registration
+
+`workflows.ts`
+
+```ts
+import { Signal, Query, condition } from '@temporalio/workflow';
+
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
```
cretz, line 85: A generic default of this value may make sense, and switching around the query to have return type first and defaulting the args may also make sense (haven't tested).

node/workflow-api.md
cretz, line 92: May want to allow/note that handlers can support `Promise` returns (i.e. async callbacks)

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+`execute-workflow.ts`
+
+```ts
+import { WorkflowClient } from '@temporalio/client';
+import { myWorkflow, unblockSignal, isBlockedQuery } from './workflows';
+
+const client = new WorkflowClient();
+const handle = client.createWorkflowHandle(myWorkflow);
+await handle.start();
+await handle.query(isBlocked /* no args */); // true
```
cretz, line 127: *nitpick* `isBlockedQuery`*

node/workflow-api.md
bergundy: 👍

node/workflow-api.md
bergundy: It will be implemented with a setter that is workflow context aware

bergundy: Only signal handlers can return promises, queries cannot.
I'm not sure if we should let condition return a promise too, seems like there's no need for that.

node/workflow-api.md
bergundy: At least once before we conclude a Workflow activation (roughly maps to a single Workflow task assuming no local activities).
I'll use the Java implementation as a reference.

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+await handle.signal(unblockSignal /* no args */);
+await handle.query(isBlocked /* no args */); // false
```
bergundy, line 129:
```suggestion
await handle.query(isBlockedQuery /* no args */); // true
await handle.signal(unblockSignal /* no args */);
await handle.query(isBlockedQuery /* no args */); // false
```

node/workflow-api.md
bergundy: Thanks

bergundy (commented):
Thanks @cretz

node/workflow-api.md
cretz: Thanks! Curious how that setter might look. Mostly curious how one determines the workflow they are running in from unrelated code (no real thread-local concept in nodejs and I would think inspecting the stack may be brittle).

cretz: >  I'm not sure if we should let condition return a promise too, seems like there's no need for that.

I suppose not since a `condition` can be implemented as a promise, e.g. (untested):

```typescript
let unblockResolve;
let unblock = new Promise(resolve => { unblockResolve = resolve })
unblockSignal.handler = unblockResolve
await unblock
```

^ Is that a reasonable alternative for how someone might implement a single unblock (i.e. not needing query or ability to re-block)?

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+#### Remove the Workflow factory wrapper and always use dynamic handler registration
+
+`workflows.ts`
+
+```ts
+import { Signal, Query, condition } from '@temporalio/workflow';
+
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new Query<[] /* args */, boolean /* return type */>('isBlocked');
```
cretz, line 86: If querying a mutable value is a common enough use case, may be worth having a helper like (untested):

```typescript
class SimpleValueQuery<T> extends Query<[], T> {
  constructor(name: string, public readonly initialValue: T) {
    super(name)
  }

  newValueRef(): ({ value: T }) {
    const v = { value: this.initialValue }
    // Somehow the calling workflow is known to this setter?
    this.handle = () => v.value
    return v
  }
}
```

bergundy: I guess we can have a `QueryVariable` or `QueryableState` class in the global scope.
The limitation is that the initial value needs to be initialized before workflow execution and can't be based on workflow type or arguments for example.

```ts
export isBlockedQuery = new QueryableState('isBlocked', true);

export async function myWorkflow1() {
  unblock.handler = (isBlockedQuery.value = false); // value is bound to this workflow
  await condition(() => !isBlockedQuery.value);
}
```

node/workflow-api.md
bergundy: Node.js has https://nodejs.org/api/async_context.html#async_context_class_asynclocalstorage, we implement a simple version of that for the workflow runtime too:
- https://github.com/temporalio/sdk-node/blob/7b80995de755d7ea098a235755892e871a1499b3/packages/workflow/src/async-local-storage.ts
- https://github.com/temporalio/sdk-node/blob/7b80995de755d7ea098a235755892e871a1499b3/packages/workflow/src/promise-hooks.ts
- https://github.com/temporalio/sdk-node/blob/7b80995de755d7ea098a235755892e871a1499b3/packages/worker/workflow-isolate-extension.cc

bergundy: It's sort of like the trigger API we have now https://nodejs.temporal.io/api/classes/workflow.Trigger

cretz: I see, thanks! That is a good bit of magic for the developer. May also be worth exposing to the developer some static call that can inform them what workflow they are executing within if we don't already.

node/workflow-api.md
cretz: That is similar to what I was thinking except I don't see where the query handler is installed (inside the `value` setter on first invocation)? In my head I had the below that was not reliant on behaviors in accessors:

```typescript
export isBlockedQuery = new SimpleValueQuery('isBlocked', true)

export async function myWorkflow1() {
  const blocked = isBlockedQuery.newValueRef()
  unblock.handler = () => { blocked.value = false }
  await condition(() => !blocked.value)
}
```

So `newValueRef` is what hooks up the query handler. Basically the same thing then, cool.

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
```
rupalivohra, line 89: What are arg1 and arg2 expected to be? Can the var names be updated to be a bit more reflective of whatever that is?

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+  unblockSignal.handler = () => (isBlocked = false);
+  isBlockedQuery.handler = () => isBlocked;
+  await condition(() => !isBlocked);
+}
+```
+
+- Note how all unblocking related functionality is co-located now, we can use this to split our Workflow into components, for example we could have a reusable "unblock" component.
+
+  ```ts
+  export async function unblocked() {
```
rupalivohra, line 100: It'd be great to include the return types in these function signatures - makes the code a lot easier to read in GitHub/in docs online

sw-yx:
Related design docs

- swyx https://gist.github.com/sw-yx/333f14872482f6190bda482a275b1f48
- roey https://gist.github.com/bergundy/8f71500f7e7a01bd199c7ae996b1d909
- loren https://gist.github.com/lorensr/7cf21519990d004d2aa889b477f8d352
- slack discussion https://temporaltechnologies.slack.com/archives/C01FT8U10GK/p1632433710183000?thread_ts=1632373667.137900&cid=C01FT8U10GK
- more between me and roey

node/workflow-api.md
sw-yx: i appreciate specific names for some type of documentation but for api design the goal is to be generic.

node/workflow-api.md
sw-yx: i'd love for `isBlockedQuery` to just have a default handler, so i dont have to write this boilerplate, perhaps as part of queryable state.

sw-yx:
overall i love this :) :) :) we still have some unsolved issues around the "queryable state" API - imo it should be the default

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+import { Signal, Query, condition } from '@temporalio/workflow';
+
+// Define the signal / query with the type signature at the top level to share the definition for handle creation
+export const unblockSignal = new Signal<[] /* args */>('unblock');
+export const isBlockedQuery = new Query<[] /* args */, boolean /* return type */>('isBlocked');
+
+// Types are inferred without an additional type definition
+export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
+  let isBlocked = true;
+  unblockSignal.handler = () => (isBlocked = false);
```
sw-yx, line 91: issue raised by @vkarpov15 - what about awaiting for signal?

alternative proposal:

```
await unblockSignal // for oneoff signal
unblockSignal(() => {}) // for handler code/repeated signals
```

lorensr: And if unblockSignal can't be promisey itself, could do:
```js
import { Signal } from '@temporalio/workflow';

export const unblockSignal = new Signal<[]>('unblock');

export async function myWorkflow(arg1: number, arg2: string): Promise<void> {
  await unblockSignal.asPromise()
}
```

Sushisource: I would heavily favor something like what @lorensr wrote. The signal itself being a promise you can await on is I guess neat, but kinda nonobvious. If you say "here's a signal" when you declare it, and "here's what it does with it fires" with `.handler`, then you probably want a "here's a wait to await on it" as an explicit concept as well. Maybe just `.fired()`

Sushisource:
This LGTM. I wonder if we want to take this opportunity to just explicitly restrict most of these things to accepting just one argument. We often give the advice in slack/on the forums anyways that people should prefer using an arg object to passing multiple args. It feels to me like we would probably save some people in pain generally by just enforcing that.

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+  await condition(() => !isBlocked);
+}
+```
+
+- Note how all unblocking related functionality is co-located now, we can use this to split our Workflow into components, for example we could have a reusable "unblock" component.
+
+  ```ts
+  export async function unblocked() {
+    let isBlocked = true;
+    unblockSignal.handler = () => (isBlocked = false);
```
vitor-mariano, line 102: Would it be possible to set and wait multiple signals, and call activities inside them, instead of blocking the workflow with a condition/trigger?

``` ts
const { refresh, shutdown } = createActivityHandle<typeof activities>(...)

...

refreshSignal.handler = () => { refresh() }
shutdownSignal.handler = () => { shutdown() }
```

node/workflow-api.md
```diff
@@ -0,0 +1,322 @@
...
+    await condition(() => !isBlocked);
+  }
+
+  export async function myWorkflow1() {
+    await unblocked();
+  }
+
+  export async function myWorkflow2() {
+    await unblocked();
+  }
```
sw-yx, line 113: Maxim concern: stealing signals from each other like in Go

Sushisource: They are different signals. You are reusing signal handling code, not signals. At the end of the day every signal is uniquely bound to a workflow run (consider: when you send signals they go a specific run).

So anything that would allow these two awaits to be anything other than completely independent is a bug. A user could maybe do something nasty that we can't really prevent like make a program wide static value that the shared signal code mutates, in which case they are now tied together, but that's a big no-no (and is something they could've done from just raw workflow code anyway).

What this does say to me is that perhaps we should call 
```
// Define the signal / query with the type signature at the top level to share the definition for handle creation
export const unblockSignal = new Signal<[] /* args */>('unblock');
export const isBlockedQuery = new Query<[] /* args */, boolean /* return type */>('isBlocked');
```

`SignalSignature` and `QuerySignature` (or `Definition` or `Type`). Because when you define them this way that's sort of more accurately what you're doing.

Even more accurate would be to only allow attaching the signal's *identifier* inside of the workflow. In this case you would have something like:

```
export const unblockSignalDef = new SignalDefinition<[]>();

export async function unblocked() {
  let isBlocked = true;
  unblockSignalDef.handler('unblock') = () => (isBlocked = false);
  await condition(() => !isBlocked);
}
```


Both ways have merit. What I demonstrate I think maybe makes the concepts a bit more clear and makes it easy to re-use the same input/return types for multiple signals with different identifiers which is maybe desirable

node/workflow-api.md
psigen: If the issue we are addressing is that a Signal represents an awaitable with multiple firing, are we just replicating a generator?  I.e. is `yield` an appropriate semantic here?

Alternatively, if we rename `handler()` to `then()`, it suddenly seems a lot more like the normal Promise API, where both of the following are part of the standard ES6 API:
```
await myPromise;
myPromise.then(() => {});
```

bergundy: I think making signal awaitable or an async iterator is a separate discussion.
The basic API should be callback based and is flexible enough to support any use case.
To sum this up, these are the issues with use iterators and promises:
1. With an awaitable, how does waiting on multiple signals look like?
  - Does the promise always resolve on the first signal?
  - Is a new promise created when awaited?
  - Does the runtime buffer incoming signals until awaited?
2. With an async iterable, how does a user drain all signals?

node/workflow-api.md
bergundy: The types were omitted for brevity, it's just `Promise<void>` in this case.

node/workflow-api.md
bergundy: Once the workflow function returns the Workflow completes and Temporal will cancel any unresolved activities.
Technically it is possible to schedule activities like in your example but you still need to prevent the workflow from completing.

node/workflow-api.md
bergundy: Yes, these signals are instantiated per WF.
I also suggested to @sw-yx at some point that the API should be called `defineSignal` to explain that this is just a definition.
Worth thinking about the naming a bit more.

In any case the signal definition should include the signal name for it to be used by both WF and client.

joebowbeer:
@Sushisource 
> This LGTM. I wonder if we want to take this opportunity to just explicitly restrict most of these things to accepting just one argument. We often give the advice in slack/on the forums anyways that people should prefer using an arg object to passing multiple args. It feels to me like we would probably save some people in pain generally by just enforcing that.

Do you have specific recommendations? I'm imagining counterexamples, but I don't really know what you are suggesting 😀

joebowbeer:
I'm catching up with this now. Has the proposal changed since the orignal description? In particular, is the basic example still accurate?

I think I'm OK with callbacks as the underlying mechanism. I'd like to see more examples to increase confidence.

Is the one-shot Trigger gone? It is now just a promisified signal handler, right?

Sushisource:
@joebowbeer 
> @Sushisource
> 
> > This LGTM. I wonder if we want to take this opportunity to just explicitly restrict most of these things to accepting just one argument. We often give the advice in slack/on the forums anyways that people should prefer using an arg object to passing multiple args. It feels to me like we would probably save some people in pain generally by just enforcing that.
> 
> Do you have specific recommendations? I'm imagining counterexamples, but I don't really know what you are suggesting grinning

What I mean is right now you are allowed to define activities, workflows, signals that accept multiple positional arguments. However, that often becomes a pain point for users when they realize they need to add or remove a parameter and now might have to deal with compatibility issues between running instances of the workflow code. IE: Passing a struct/object/whatever with some fields that might be empty/null/etc is easier to deal with in a compat way than actually changing how many things are passed at the callsite.

> I naively wonder about versioning, especially if the code is reused. Do signals and queries have versions? If so, how are they indicated? How do they interact with versioned workflows?

At the end of the day is still workflow code (which this example might obfusace -- being able to have the signal handling code be its own function/reusable is cool but maybe makes it seem like this isn't just workflow code) . So it's handled like it would be in workflow code, with the `patched` API. This works fairly well with the reusable component concept as long as you use sufficiently unique patch names.

Versioning of "components" or "libraries" that workflows use is a wildly complex topic. This proposal doesn't change that situation at all, so I don't think we need to worry about it in relation to that. Coming up with a sane story for `patched` calls deep inside re-usable code might only ever be doable with more control over the whole execution environment (something like WASM workflows)... but that's a long topic.

node/workflow-api.md
bergundy: @sw-yx I'll add `QueryableState` in the next proposal.
I think we might even be able to make it into something like `StateVariable` where we auto generate signal handlers too.

node/workflow-api.md
joebowbeer: @bergundy The new proposal reminds me of EventEmitter, which at first glance seems appropriate.

However, I do think `on` is too concise. _Too cute_ is how I would put it. It makes me wonder whether workflow is an EventEmitter somehow. (Spoiler Alert: It is not.)

I would choose a more descriptive name and let the user alias if desired:

```
import { defineSignal, defineQuery, setListener as on } from '@temporalio/workflow';

// ...
on(unblockSignal, () => (isBlocked = false));
```

On second thought, though, it bothers me that signal and query are conflated in this way.

Fundamentally, I think of **signal** as something that may have a _list of listeners_, like an EventEmitter. The use of `on` or `add` prefixes reinforces this idea -- once the EventEmitter analogy is activated.

Whereas a **query** returns a result, hence one body.

Bottom line: I'd like to see more descriptive names that reinforce the differences between signals and queries, and let the user conflate them if desired.

Sushisource: Having this as a function defined on the workflow module is great. Totally clears up the things I mentioned. Like that a lot.

As for naming, I'm of two minds. The conciseness of `on` is a nice properly IMO. The type system will be able to specialize the second argument based on the first (IE: Signals return void, queries return something), and I think that is pretty much sufficient signal to the user.

OTOH, there's nothing wrong with `onSignal` and `onQuery` to differentiate. Queries *cannot* mutate state, and unfortunately we can't really enforce that, so making that clear in the docstring for `onQuery` is a nice property.

RE: Multiple handlers. I can see why it makes sense, especially in the "JS event handler" context. I think I'd prefer us to keep the underlying primitives as simple as possible and let the user layer on top of that. If the need that kind of multiple dispatch behavior, they can easily implement a top-level handler that calls a bunch of other stuff. My main reason for that is you probably end up with users asking for stuff like "I need a function that clears the event handler list", etc, other stuff related to managing that implicit set of handlers. If lots of users want the dispatch behavior, we can always layer it on by composition (ex: `LayerableSignalDefinition` or `onSignalAdditional` or whatever) rather than baking it in to the base primitive.

joebowbeer:
@sw-yx @bergundy Can you clarify for posterity why signal and query typing was removed from the workflow type?

I assume the answer must be for "Dynamic signals and queries registration" but I don't see any discussion of this or where this requirement is coming from. I notice that several of the strawman proposals retain the signals and queries in the workflow type.

What does dynamic signal and query registration mean? That queries can be added and removed dynamically? I don't see this in any of the TypeScript samples.

In its merged form, the workflow type cannot be used to flag unsupported queries or signals at build time, and they are not currently caught at runtime.

PR #45, opened 2021-10-08
Go SDK Workflow Interceptor Updates and OpenTelemetry Tracing Support

cretz:
Origin issue: https://github.com/temporalio/sdk-go/issues/529 and https://github.com/temporalio/sdk-go/issues/526

# Summary

Update interceptors in the Go SDK and implement OpenTelemetry tracing.

# Motivation

* Go SDK doesn't support "wrapping" all calls
* Interceptors didn't have everything necessary for Java parity wrt tracing
* We want OpenTelemetry tracing and it's a good initial use of the interceptor updates to prove functionality
* Activity interceptors are needed

# Detailed design

See [rendered proposal](https://github.com/cretz/proposals/blob/go-workflow-interceptor-updates/go/workflow-interceptor-updates.md)

### DIFF
--- a/go/workflow-interceptor-updates.md
+++ b/go/workflow-interceptor-updates.md
@@ -131,6 +131,17 @@ GetInfo(ctx Context) *WorkflowInfo
 
 Why: All other calls match `workflow` package, why not this?
 
+### WorkflowOutboundCallsInterceptor.NewContinueAsNewError
+
+Currently this doesn't exist. Proposal:
+
+```go
+NewContinueAsNewError(ctx Context, wfn interface{}, args ...interface{}) error
+```
+
+Why: This is needed for the later tracing piece that adds headers to the context.
+
+
 ### ActivityInterceptor
 
 Proposal:
@@ -199,6 +210,82 @@ Why: In the newer gRPC service interfaces this is done to allow interface additi
 [this](https://github.com/grpc/grpc-go/blob/03ca7b7d00cada2ff8a3ea7348fe6c3a2b2ee4fb/examples/helloworld/helloworld/helloworld_grpc.pb.go#L52)
 as an example.
 
+#### Explanation of Forced Embeds
+
+To explain this concept a little further, imagine there is an interface in a library you want to implement:
+
+```go
+type MyInterface interface {
+  DoSomething() error
+}
+
+type MyInterfaceBase struct{}
+func (MyInterfaceBase) DoSomething() error { return nil }
+```
+
+And you implement it but without embedding the base struct:
+
+```go
+type MyImpl struct{}
+var _ otherlib.MyInterface = MyImpl{}
+func (MyImpl) DoSomething() error { return nil }
+```
+
+Now if the library goes back and changes the interface to:
+
+```go
+type MyInterface interface {
+  DoSomething() error
+  DoAnotherThing() error
+}
+
+type MyInterfaceBase struct{}
+func (MyInterfaceBase) DoSomething() error { return nil }
+func (MyInterfaceBase) DoAnotherThing() error { return nil }
+```
+
+You'd upgrade the dependency and then your compilation would break making the library's interface alteration not as
+compatible as it could be. But if the library had originally defined its interface and base as:
+
+```go
+type MyInterface interface {
+  DoSomething() error
+  mustEmbedBase()
+}
+
+type MyInterfaceBase struct{}
+func (MyInterfaceBase) DoSomething() error { return nil }
+func (MyInterfaceBase) mustEmbedBase() {}
+```
+
+Then you'd have _no choice_ but to embed `MyInterfaceBase`:
+
+```go
+type MyImpl struct{ otherlib.MyInterfaceBase }
+// This will fail to compile if you hadn't done the embed
+var _ otherlib.MyInterface = MyImpl{}
+func (MyImpl) DoSomething() error { return nil }
+```
+
+So now, when the library adds a new method to the interface like:
+
+```go
+type MyInterface interface {
+  DoSomething() error
+  DoAnotherThing() error
+  mustEmbedBase()
+}
+
+type MyInterfaceBase struct{}
+func (MyInterfaceBase) DoSomething() error { return nil }
+func (MyInterfaceBase) DoAnotherThing() error { return nil }
+func (MyInterfaceBase) mustEmbedBase() {}
+```
+
+The library knows all implementations of the interface embedded the struct they had to for forwards compatibility and it
+knows that nobody's compilation will fail because they didn't embed the base. This is what gRPC and other libraries are
+starting to do to force forward compatibility of interfaces.
+
 ### WorkerInterceptor
 
 Current:
@@ -227,47 +314,61 @@ For discussion:
 * Would we want a `WorkerInterceptorBase` with default impls that return instances of `ActivityInboundInterceptorBase`
   and `WorkflowInboundInterceptorBase`? Why doesn't Java have this?
 
-## OpenTelemetry Implementation
+## Tracing and Context Propagation
 
-Proposal in separate `opentelemetry` package:
+Current:
+
+* A concept called a context propagator handles marshalling to/from workflow and activity proto "headers" (a construct
+  in the body not to be confused with actual HTTP headers)
+* Client options accept an OpenTracing tracer
+* An internal context propagator is creating for the OpenTracing tracer that marshals to/from a span context to a
+  special key in the headers
+* The tracer is sent all throughout the contextual work of workflows and activities so that spans can be created at
+  different times
+
+Proposal:
+
+* A new ability to retrieve a mutable set of proto "headers" on the context will be made so that interceptors can set
+  workflow/activity headers in either direction
+* Internally, headers will be obtained from the context by the last interceptor
+  * So headers are set on the context before interceptor invocation for mutation by interceptors
+* A `tracing` package will be created with:
+  * `WorkerInterceptor` implementations for OpenTracing and OpenTelemetry tracers
+    * Wraps the same inbound/outbound calls in spans that are otherwise wrapped by the tracer today
+    * On inbound `ExecuteActivity` and `ExecuteWorkflow` span context is unmarshaled from header
+    * On outbound `ExecuteActivity`, `ExecuteLocalActivity`, `ExecuteChildWorkflow`, and `NewContinueAsNewError` span
+      context is marshaled into header
+  * `Client` implementations for OpenTracing and OpenTelemetry tracers
+    * Wraps the same calls in spans that are otherwise wrapped by the tracer today
+    * On `ExecuteWorkflow` and `SignalWithStartWorkflow` span context is marshaled into header
 
 ```go
-// Added to existing internal.ClientOptions struct to allow the gRPC
-// interceptors to be enabled and allow options to be set for them from
-// https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc
-type ClientOptions struct {
-  OpenTelemetry OpenTelemetryClientOptions
-}
+package interceptors
 
-type OpenTelemetryClientOptions struct {
-  Enabled bool
-  Options []otelgrpc.Option
-}
+// Header returns nil if this context does not have the header on it (meaning
+// they cannot be mutated by an interceptor).
+func Header(ctx interface { Value(interface{}) interface{} }) map[string]*commonpb.Payload
+```
 
-// Wraps only the calls implemented in spans, delegates the rest
-func NewTracingClient(opts) client.Client
+```go
+package tracing
 
-func (cl) ExecuteWorkflow(ctx context.Context, options client.StartWorkflowOptions, workflow interface{}, args ...interface{}) (client.WorkflowRun, error)
-func (cl) SignalWithStartWorkflow(ctx context.Context, workflowID string, signalName string, signalArg interface{},
-	options client.StartWorkflowOptions, workflow interface{}, workflowArgs ...interface{}) (client.WorkflowRun, error)
+func NewOpenTracingClient(c client.Client, tracer opentracing.Tracer) client.Client
+func NewOpenTracingInterceptor(tracer opentracing.Tracer) interceptors.WorkerInterceptor
 
-// Wraps only the calls implemented in spans, delegates the rest
-func NewTracingActivityInterceptor(opts) interceptors.ActivityInterceptor
-func NewTracingWorkflowInterceptor(opts) interceptors.WorkflowInterceptor
-
-func (i) InterceptActivity(ctx context.Context, next interceptors.ActivityInboundInterceptor) interceptors.ActivityInboundInterceptor
-func (i) InterceptWorkflow(ctx workflow.Context, next interceptors.WorkflowInboundInterceptor) interceptors.WorkflowInboundInterceptor
-
-func (aii) ExecuteActivity(ctx context.Context, input *interceptors.ActivityInput) (interface{}, error)
-
-func (wii) ExecuteWorkflow(ctx workflow.Context, input *interceptors.WorkflowInput) (interface{}, error)
-
-func (woi) ExecuteActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
-func (woi) ExecuteLocalActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
-func (woi) ExecuteChildWorkflow(ctx workflow.Context, childWorkflowType string, args ...interface{}) workflow.ChildWorkflowFuture
+func NewOpenTelemetryClient(c client.Client, tracer trace.Tracer) client.Client
+func NewOpenTelemetryInterceptor(tracer trace.Tracer) interceptors.WorkerInterceptor
 ```
 
 For discussion:
 
-* This matches Java feature-wise, is this enough? Does Java impl not trace anything we might want to?
-* Do we want to do anything with the existing `opentracing` Go SDK implementation as part of this?
+* Is the approach of `workflow.Header` and `activity.Header` acceptable or do we want to combine them somehow?
+* This effectively deprecates the `ClientOptions.Tracer` field, shall we remove it completely?
+  * If not, should we gut the existing `opentracing` impl and have it default to this implementation (i.e. if
+    `ClientOptions.Tracer` is present, implicitly wrap the client and add the interceptor)?
+* This effectively deprecates `ContextPropagator`s, shall we remove them completely?
+* Is a `tracing` package ok, or would we rather separate`opentracing` and `opentelemetry` packages?
+* We technically could make this a separate module if the dependencies were too onerous for those not wanting tracing,
+  but it's probably fine in the primary module
+* We could add a `InterceptClient(next client.Client) client.Client` to `WorkerInterceptor` and remove the concept of
+  `NewXClient` above? Thoughts? (note Java separates client call interceptors and worker interceptors currently)
--- a/go/workflow-interceptor-updates.md
+++ b/go/workflow-interceptor-updates.md
@@ -136,7 +136,7 @@ Proposal:
 
 ```go
 type ActivityInterceptor interface {
-	InterceptActivity(ctx context.Context, next ActivityInboundInterceptor) ActivityInboundInterceptor
+  InterceptActivity(ctx context.Context, next ActivityInboundInterceptor) ActivityInboundInterceptor
 }
 ```
 
@@ -152,11 +152,13 @@ type ActivityInput struct {
 }
 
 type ActivityInboundInterceptor interface {
-	Init(outbound ActivityOutboundInterceptor) error
+  Init(outbound ActivityOutboundInterceptor) error
 
-	ExecuteActivity(ctx context.Context, input *ActivityInput) (interface{}, error)
+  ExecuteActivity(ctx context.Context, input *ActivityInput) (interface{}, error)
 }
 ```
+
+Also a `ActivityInboundInterceptorBase` implementation with expected delegation.
 
 Why: Mimics `WorkflowInboundInterceptor` but for activities.
 
@@ -169,7 +171,7 @@ type ActivityOutboundInterceptor interface {
   GetInfo(ctx context.Context) ActivityInfo
   GetLogger(ctx Context) log.Logger
   GetMetricsScope(ctx Context) tally.Scope
-	RecordHeartbeat(ctx context.Context, details ...interface{})
+  RecordHeartbeat(ctx context.Context, details ...interface{})
   HasHeartbeatDetails(ctx context.Context) bool
   GetHeartbeatDetails(ctx context.Context, d ...interface{}) error
   GetWorkerStopChannel(ctx context.Context) <-chan struct{}
@@ -182,8 +184,8 @@ Why:
 
 * Mimics `WorkflowOutboundInterceptor` but for activities.
 
-For discussion `GetInfo` returns expensive copies because `activity.GetInfo` does. Is this ok? This is inconsistent with
-workflow info which is a pointer.
+For discussion: `GetInfo` returns expensive copies because `activity.GetInfo` does. Is this ok? This is inconsistent
+with workflow info which is a pointer.
 
 ## Interceptor Possible Updates
 
@@ -202,7 +204,7 @@ Current:
 
 ```go
 type WorkflowInterceptor interface {
-	InterceptWorkflow(info *WorkflowInfo, next WorkflowInboundCallsInterceptor) WorkflowInboundCallsInterceptor
+  InterceptWorkflow(info *WorkflowInfo, next WorkflowInboundCallsInterceptor) WorkflowInboundCallsInterceptor
 }
 ```
 
@@ -256,7 +258,7 @@ func (aii) ExecuteActivity(ctx context.Context, input *interceptors.ActivityInpu
 
 func (wii) ExecuteWorkflow(ctx workflow.Context, input *interceptors.WorkflowInput) (interface{}, error)
 
-func (woi) ExecuteActivity(ctx workflow.Context,activityType string, args ...interface{}) workflow.Future
+func (woi) ExecuteActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
 func (woi) ExecuteLocalActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
 func (woi) ExecuteChildWorkflow(ctx workflow.Context, childWorkflowType string, args ...interface{}) workflow.ChildWorkflowFuture
 ```
--- a/go/workflow-interceptor-updates.md
+++ b/go/workflow-interceptor-updates.md
@@ -6,6 +6,7 @@ This proposal is for updates to Go workflow interceptors with the following feat
 * Support idiomatic Go interceptor design
 * Provide feature parity with Java SDK and other missing features (see https://github.com/temporalio/sdk-go/issues/529)
 * Add activity interceptors
+* Add client interceptors
 * Make inbound interceptors future-proof for argument addition
 * Support for using workflow/activity proto headers from interceptors
   * One notable use case for this is to support validation (e.g. authz) and erroring out based on workflow/activity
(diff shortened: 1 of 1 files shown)


go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,267 @@
...
+### ActivityOutboundInterceptor
+
+Proposal:
+
+```go
+type ActivityOutboundInterceptor interface {
+  GetInfo(ctx context.Context) ActivityInfo
+  GetLogger(ctx Context) log.Logger
+  GetMetricsScope(ctx Context) tally.Scope
+	RecordHeartbeat(ctx context.Context, details ...interface{})
```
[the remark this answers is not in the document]
cretz: I have a few of those, gonna make a commit soon

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,267 @@
...
+
+```go
+// Added to existing internal.ClientOptions struct
+type ClientOptions struct {
+  OpenTelemetry OpenTelemetryClientOptions
+}
+
+type OpenTelemetryClientOptions struct {
+  Enabled bool
+  Options []otelgrpc.Option
```
Sushisource, line 240: I suspect there will need to be additional / other options here, but I think probably not super important to nail down precisely for this proposal.

cretz: That is why I put it in a separate struct. I can't think of any beyond what https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc wants for the unary and stream interceptors.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
+# Go SDK Workflow Interceptor Updates and OpenTelemetry Tracing Support
+
+This proposal is for updates to Go workflow interceptors with the following features:
+
+* Support for and implementation of OpenTelemetry tracing
```
bergundy, line 5: We should make sure that interceptors can be used for authz too (part of the reason we add an explicit `error` to the return type).

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+InterceptWorkflow(info *WorkflowInfo, next WorkflowInboundCallsInterceptor) WorkflowInboundCallsInterceptor
+```
+
+Proposal:
+
+```go
+InterceptWorkflow(ctx Context, next WorkflowInboundCallsInterceptor) WorkflowInboundCallsInterceptor
+```
+
+Why: Callers can use the context to get workflow info.
```
bergundy, line 29: What other added value does passing `Context` provide?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+### WorkflowXCallsInterceptor
+
+Proposal: Change the name to `WorkflowXInterceptor`. Why: `Calls` just adds noise (more accepted in Java, not Go).
+
+### WorkflowInboundCallsInterceptor.ExecuteWorkflow
```
bergundy, line 35: Isn't this supposed to be `WorkflowInboundInterceptor` according to the line above?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+type WorkflowInput struct {
+  Args []interface{}
+}
+
+ExecuteWorkflow(ctx Context, input *WorkflowInput) (interface{}, error)
+```
+
+Why:
+
+* Workflow type can be derived from the context _or_ the info first given to `InterceptWorkflow`
```
bergundy, line 55: Where can we access the headers? in the context? In the other SDKs it's a part of the input.
Also `workflowType` is part of the input and may be modified by an interceptor (although I'm not sure this is desirable).

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+HandleSignal(ctx Context, input *SignalInput) error
+```
+
+Also, this will be called to handle the signal with the last invocation in the chain passing the signal to the workflow.
+
+Why:
+
+* `HandleSignal` matches Java nomenclature and new behavior better
+* Currently sending is sent separate and async, so the current version doesn't wrap the handling. Instead this will be
+  part of the async call and documentation will make clear it needs to be safe for concurrent invocation.
```
bergundy, line 84: After reading this point a few times, I failed to understand what you're suggesting.
Are you suggesting we wrap the handling of the signal? Is that even possible with channels?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+Why:
+
+* All other calls convert the payload outside of the interceptor call, this can too
+* No need to pass a handler, that is just part of the last item in the interceptor chain (i.e. when calling next)
+
+### WorkflowOutboundCallsInterceptor
+
+We can leave everything the way it is because, while it doesn't make the interceptor parameters future-proof with struct
+wrappers, it mimics the actual calls that aren't going to have backwards-compatible alterations.
```
bergundy, line 115: Maybe worth using structs for consistency with the other interceptors?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+```go
+type ActivityInput struct {
+  Args []interface{}
+}
+
+type ActivityInboundInterceptor interface {
+  Init(outbound ActivityOutboundInterceptor) error
+
+  ExecuteActivity(ctx context.Context, input *ActivityInput) (interface{}, error)
```
bergundy, line 157: Should this just be called `Execute`?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+```
+
+Proposal:
+
+```go
+type WorkflowInput struct {
+  Args []interface{}
+}
+
+ExecuteWorkflow(ctx Context, input *WorkflowInput) (interface{}, error)
```
bergundy, line 50: Should this be called `Execute`?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+Why: Mimics `WorkflowInterceptor` but for activities.
+
+### ActivityInboundInterceptor
+
+Proposal:
+
+```go
+type ActivityInput struct {
+  Args []interface{}
```
bergundy, line 151: I'd put the activity type and headers here so they can be modified by the interceptor.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+}
+```
+
+Then change `WorkerOptions.WorkflowInterceptorChainFactories []WorkflowInterceptor` to
+`WorkerOptions.Interceptors []WorkerInterceptor`.
+
+For discussion:
+
+* This maps to what Java does
+* Would we want a `WorkerInterceptorBase` with default impls? Why doesn't Java have this?
```
bergundy, line 226: What would be the default implementations here?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
+# Go SDK Workflow Interceptor Updates and OpenTelemetry Tracing Support
```
bergundy, line 1: We should mention that we plan to break compatibility for current generation of interceptors here.
Also, this pretty much deprecates context propagators and built-in tracing support.
I'd share this spec with the community via the `#go-sdk` slack channel.

bergundy: Built-in tracing could be implemented as an interceptor.
Should we try to do that instead of going for opentelemetry?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+* Would we want a `WorkerInterceptorBase` with default impls? Why doesn't Java have this?
+
+## OpenTelemetry Implementation
+
+Proposal in separate `opentelemetry` package:
+
+```go
+// Added to existing internal.ClientOptions struct
+type ClientOptions struct {
+  OpenTelemetry OpenTelemetryClientOptions
```
bergundy, line 235: Can you explain why this is needed please?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+}
+
+type OpenTelemetryClientOptions struct {
+  Enabled bool
+  Options []otelgrpc.Option
+}
+
+// Wraps only the calls implemented in spans, delegates the rest
+func NewTracingClient(opts) client.Client
```
bergundy, line 244: The name tracing is confusing, makes me think of opentracing.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+func (wii) ExecuteWorkflow(ctx workflow.Context, input *interceptors.WorkflowInput) (interface{}, error)
+
+func (woi) ExecuteActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
+func (woi) ExecuteLocalActivity(ctx workflow.Context, activityType string, args ...interface{}) workflow.Future
+func (woi) ExecuteChildWorkflow(ctx workflow.Context, childWorkflowType string, args ...interface{}) workflow.ChildWorkflowFuture
+```
+
+For discussion:
+
+* This matches Java feature-wise, is this enough? Does Java impl not trace anything we might want to?
```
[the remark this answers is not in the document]
bergundy: In the future we'll want to be able to trace queries and signals too but that requires server / protocol changes.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+* Do we want to do anything with the existing `opentracing` stuff as part of this?
```
bergundy, line 269: Commented above.

bergundy (commented):
I like where this is going.

Where is the client calls interceptor in this proposal?

go/workflow-interceptor-updates.md
cretz: I don't think so for a couple of reasons. Firstly, in Go it's usually <verb><noun> unless the noun is the subject of the method (i.e. if this were a workflow instead of a workflow interceptor, `Execute` makes more sense, otherwise you're not "executing" the interceptor, you're intercepting the execution). Also this matches the name used in the client which I think makes it clearer. However, if we do want to be consistent with signal and query, I could rename this to `HandleExecution`.

go/workflow-interceptor-updates.md
cretz: > Where can we access the headers? in the context?

Headers are usually part of the context in Go, yes. But I can explicitly confirm that behavior.

> Also workflowType is part of the input and may be modified by an interceptor (although I'm not sure this is desirable).

I am unable to find this behavior in the Java SDK. Do we want this feature here and there?

go/workflow-interceptor-updates.md
cretz: Headers are modified in the context usually and I will show examples of that. As for activity type, that's not a feature I see of the Java SDK. Is this something we need to support in interceptors in both places? My goal was to reach feature parity with Java but I can add more if wanted.

go/workflow-interceptor-updates.md
cretz: I don't think so anymore than `ExecuteWorkflow` should be. You aren't executing this interceptor you are executing the activity (as opposed to `Init` where you are initializing this interceptor not the activity).

go/workflow-interceptor-updates.md
cretz: To allow you to opt-in to opentelemetry gRPC and allow you to set current and future options implemented at https://pkg.go.dev/go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+Why:
+
+* Mimics `WorkflowOutboundInterceptor` but for activities.
+
+For discussion: `GetInfo` returns expensive copies because `activity.GetInfo` does. Is this ok? This is inconsistent
+with workflow info which is a pointer.
+
+## Interceptor Possible Updates
+
+### Forced Base Embedding
```
cretz, line 192: Do we want this?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+Proposal: Force the base struct to be embedded in all interface implementations. This can be done via a private
+interface method that is only implemented by the base struct.
+
+Why: In the newer gRPC service interfaces this is done to allow interface additions to always be forward compatible. See
+[this](https://github.com/grpc/grpc-go/blob/03ca7b7d00cada2ff8a3ea7348fe6c3a2b2ee4fb/examples/helloworld/helloworld/helloworld_grpc.pb.go#L52)
+as an example.
+
+### WorkerInterceptor
```
cretz, line 201: Do we want this?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,269 @@
...
+
+  ExecuteActivity(ctx context.Context, input *ActivityInput) (interface{}, error)
+}
+```
+
+Also a `ActivityInboundInterceptorBase` implementation with expected delegation.
+
+Why: Mimics `WorkflowInboundInterceptor` but for activities.
+
+### ActivityOutboundInterceptor
```
cretz, line 165: Note, there is no analogue for this in the Java SDK

go/workflow-interceptor-updates.md
cretz: I added a note that examples will be added during implementation showing header alteration in interceptors and client werappers...are headers enough?

go/workflow-interceptor-updates.md
cretz: Explained in Godoc above the struct

go/workflow-interceptor-updates.md
cretz: Updated to be clearer

go/workflow-interceptor-updates.md
bergundy: > There is a big bold note that says:
> These updates are known to NOT be backwards-compatible

Missed that :/

bergundy: We currently have 3 different tracing concepts.
1. built-in tracing in golang
2. opentracing interceptors in java
3. opentelemetry interceptors in node

The Java and Node interceptors are incompatible.

Java uses this header key and span context definition:
- https://github.com/temporalio/sdk-java/blob/6f86b160d9b948418d2e49d8e5a2bbb260fad149/temporal-opentracing/src/main/java/io/temporal/opentracing/internal/ContextAccessor.java#L36
- https://github.com/opentracing/opentracing-java/blob/a8557d7c7a56210cbcf6bc518fe24b32923aa66f/opentracing-api/src/main/java/io/opentracing/SpanContext.java

Node uses this header key and span context definition:
- https://github.com/temporalio/sdk-node/blob/4505eee94e7d8a10bc187612977fd72bc6d740a6/packages/common/src/otel.ts#L5
- https://github.com/open-telemetry/opentelemetry-js-api/blob/41109c83a9784d689f319f2c5d953b3874c694a3/src/trace/span_context.ts#L23

I'm not sure what go uses but ideally all of our SDKs will be compatible.
We should for now keep the go built-in tracing support backwards compatible.

3 options I see:
- Re-implement built-in tracing using interceptors
- Create a new opentelemetry / opentracing interceptors package that's compatible with at least Java or Node
- Do both

go/workflow-interceptor-updates.md
bergundy: The use-case I know of is (example for StartWorkflow):
1. Client sends StartWorkflow request with auth token in header to a proxy that sits in front of the server (or modified server)
1. Server translates token to auth claim
2. Worker picks up StartWorkflow task, compares request and claims and either processes or rejects the task

go/workflow-interceptor-updates.md
bergundy: I don't know what the best practices in go are, I'm fine with whatever we choose here.
Let's make sure the names are consistent though.

go/workflow-interceptor-updates.md
bergundy: It doesn't exist in `WorkflowInput` although `SignalInput` does include `signalName`:
https://github.com/temporalio/sdk-java/blob/6f86b160d9b948418d2e49d8e5a2bbb260fad149/temporal-sdk/src/main/java/io/temporal/common/interceptors/WorkflowInboundCallsInterceptor.java#L61

I can imagine a use-case where based on some logic user decided to re-route to a different workflow or signal.


Are you referring to [Value / withValue](https://pkg.go.dev/context#WithValue)?
API headers strictly map from `string` to `Payload`, should we enforce this?

go/workflow-interceptor-updates.md
bergundy: I'm not convinced either way, just thought I'd bring it up.

go/workflow-interceptor-updates.md
bergundy: I just saw that the type in missing in Java and Node, I'm wondering if we should add it or not and if it makes sense for that to be mutable by interceptors.

go/workflow-interceptor-updates.md
bergundy: We can unify this discussion with the one above about `ExecuteWorkflow`.

go/workflow-interceptor-updates.md
bergundy: Neither in Node.

go/workflow-interceptor-updates.md
bergundy: `NewOtelTracingClient`?

bergundy:
> Client is already an interface so one can already wrap it. This is how the tracing client will work.

We might want a tighter interface than working at the gRPC level, not sure though, that's at least what we do in Java / Node:
https://github.com/temporalio/sdk-node/blob/4505eee94e7d8a10bc187612977fd72bc6d740a6/packages/client/src/interceptors.ts#L64

go/workflow-interceptor-updates.md
cretz: This struct supports additional items in the future, I don't think such a decision must be made now.

go/workflow-interceptor-updates.md
cretz: Updated to clarify benefit of context.

go/workflow-interceptor-updates.md
cretz: After further discussion, while I will still be adding `WithHeaders`, the `HeaderProvider` concept cannot be deprecated as that is the primary way to add headers via gRPC interceptor. The client/activity/workflow interceptors do not map 1:1 w/ gRPC requests (we even make requests outside of intercepted calls like for registering namespace).

go/workflow-interceptor-updates.md
Spikhalskiy: Java has both opentelementry and opentracing. Opentelemetry is easily pluggable through opentelementry-shim and fully functional.
https://github.com/temporalio/sdk-java/tree/master/temporal-opentracing#opentelemetry

go/workflow-interceptor-updates.md
Sushisource: I don't totally follow what this does. Can you gimme a little explainer on how this helps with compat?

go/workflow-interceptor-updates.md
Sushisource: This seems reasonable to me

go/workflow-interceptor-updates.md
cretz: This tracing proposal has been removed and replaced after I learned more about how headers and tracing worked inside the SDK.

go/workflow-interceptor-updates.md
cretz: Agreed, and my tracing interceptors like that activity and workflow interception is combined into a single place like Java.

go/workflow-interceptor-updates.md
bergundy: I haven't seen an example just heard about this practice.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,374 @@
...
+  workflow/activity headers in either direction
+* Internally, headers will be obtained from the context by the last interceptor
+  * So headers are set on the context before interceptor invocation for mutation by interceptors
+* A `tracing` package will be created with:
+  * `WorkerInterceptor` implementations for OpenTracing and OpenTelemetry tracers
+    * Wraps the same inbound/outbound calls in spans that are otherwise wrapped by the tracer today
+    * On inbound `ExecuteActivity` and `ExecuteWorkflow` span context is unmarshaled from header
+    * On outbound `ExecuteActivity`, `ExecuteLocalActivity`, `ExecuteChildWorkflow`, and `NewContinueAsNewError` span
+      context is marshaled into header
+  * `Client` implementations for OpenTracing and OpenTelemetry tracers
```
bergundy, line 341: The other SDKs have a concept of `WorkflowClientCallsInterceptor`s.
While the same functionality can be achieved with the current proposal it might be worth considering adding the interceptor for consistency with the other SDKs.

cretz: I have added a section on why this is unnecessary. Open for comment.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,417 @@
...
+func NewOpenTracingInterceptor(tracer opentracing.Tracer) interceptors.WorkerInterceptor
+
+func NewOpenTelemetryClient(c client.Client, tracer trace.Tracer) client.Client
+func NewOpenTelemetryInterceptor(tracer trace.Tracer) interceptors.WorkerInterceptor
+```
+
+For discussion:
+
+* Is the approach of `workflow.Header` and `activity.Header` acceptable or do we want to combine them somehow?
+* This effectively deprecates the `ClientOptions.Tracer` field, shall we remove it completely?
```
Sushisource, line 409: I think so? At a high level: Users need a way to specify things like "the endpoint for where my traces go" or "how I want to batch or sample", whatever, which will be specific to what kind of telem impl they are using. OT or OTel.

Those configuration options / instances of things should probably be passed to the constructors of the interception layers which implement them.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,417 @@
...
+  * If not, should we gut the existing `opentracing` impl and have it default to this implementation (i.e. if
+    `ClientOptions.Tracer` is present, implicitly wrap the client and add the interceptor)?
+* This effectively deprecates `ContextPropagator`s, shall we remove them completely?
```
Sushisource, line 412: I believe so yes, but, I think there's a question about how much incompatible changes we're willing to make. We should get feedback from @mfateev

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,417 @@
...
+* Is a `tracing` package ok, or would we rather separate`opentracing` and `opentelemetry` packages?
+* We technically could make this a separate module if the dependencies were too onerous for those not wanting tracing,
+  but it's probably fine in the primary module
+* We could add a `InterceptClient(next client.Client) client.Client` to `WorkerInterceptor` and remove the concept of
+  `NewXClient` above? Thoughts? (note Java separates client call interceptors and worker interceptors currently)
```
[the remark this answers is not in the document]
cretz: 🤔 makes sense. Had discussed with @vitarb off-PR and we were going to do this so users didn't have to wrap a client separately from the interceptors, but Java keeps them separate this way so we can too.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,417 @@
...
+
+Proposal:
+
+```go
+GetInfo(ctx Context) *WorkflowInfo
+```
+
+Why: All other calls match `workflow` package, why not this?
+
+### WorkflowOutboundCallsInterceptor.NewContinueAsNewError
```
cretz, line 136: Per discussion, removing this since it is not the best way to capture continue as new

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,417 @@
...
+func (MyInterfaceBase) DoSomething() error { return nil }
+func (MyInterfaceBase) DoAnotherThing() error { return nil }
+func (MyInterfaceBase) mustEmbedBase() {}
+```
+
+The library knows all implementations of the interface embedded the struct they had to for forwards compatibility and it
+knows that nobody's compilation will fail because they didn't embed the base. This is what gRPC and other libraries are
+starting to do to force forward compatibility of interfaces.
+
+### WorkerInterceptor
```
cretz, line 291: I am going to do this on first implementation

go/workflow-interceptor-updates.md
cretz: Note, for clarity I am changing these headers to desired names instead of what they were.

go/workflow-interceptor-updates.md
Sushisource: What did we decide here?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+  Args []interface{}
+}
+
+HandleQuery(ctx Context, input *QueryInput) (interface{}, error)
+```
+
+Why:
+
+* All other calls convert the payload outside of the interceptor call, this can too
+* No need to pass a handler, that is just part of the last item in the interceptor chain (i.e. when calling next)
```
jlegrone, line 162: Would this still allow an interceptor to access and potentially modify the handler's return value before bubbling it up?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+
+Why: Mimics `WorkflowInboundInterceptor` but for activities.
+
+### ActivityOutboundInterceptor
+
+Proposal:
+
+```go
+type ActivityOutboundInterceptor interface {
+  GetInfo(ctx context.Context) ActivityInfo
```
cretz, line 225: Are we concerned about a the costs of copying the result on every invocation in the chain? Might we want to return a pointer here instead (even though the last call can dereference)?

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+
+```go
+ExecuteWorkflow(ctx Context, workflowType string, args ...interface{}) []interface{}
+```
+
+Proposal:
+
+```go
+type WorkflowInput struct {
+  Args []interface{}
```
cretz, line 97: Today, the arguments for this execution, due to how our converters work, has a pointer to every non-context-non-byte-slice argument instead of the actual values. I am looking into changing this.

cretz: After much effort and consternation, I have worked through an implementation that uses real values and not pointers to real values. This will be another breaking change from how interceptors work today. Today, if you intercept `ExecuteWorkflow` with a `int` param, the `args` actually has it as `*int` which is nonintuitive and causes confusion when replicating this behavior for the `ExecuteActivity` use cases for local/testsuite activities that use similar code but do not have an argument unmarshaling step. With this implementation, arguments will be represented by their actual types assuming I can solve all test cases.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+* This kinda mimics `WorkerInterceptor` calls, but note it only provides an outbound interceptor because all client
+  calls are outbound by their nature so there is no inbound equivalent
+
+### ClientOutboundInterceptor
+
+Proposal:
+
+```go
+type ClientOutboundInterceptor interface {
+  ExecuteWorkflow(ctx context.Context, options StartWorkflowOptions, workflow interface{}, args ...interface{}) (WorkflowRun, error)
```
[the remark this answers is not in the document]
cretz: I might as well go ahead and revert https://github.com/temporalio/proposals/pull/45/commits/14f4e43c7930b8bd4e1021ea4bcb71def87c8a2d since our params will now not be matching the client call necessarily

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+}
+```
+
+```go
+type WorkerOptions struct {
+  // ...
+
+  // Interceptors to apply to the worker. Any interceptors that also implement Interceptor (meaning they implement
+  // ClientInterceptor in addition to WorkerInterceptor) will be used for client interception as well. When interceptors
+  // are here and in client options, the client interceptors are applied first.
```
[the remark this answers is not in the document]
cretz: Fixed.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+```go
+ProcessSignal(ctx Context, signalName string, arg interface{}) error
+```
+
+Proposal:
+
+```go
+type SignalInput struct {
+  SignalName string
+  Arg interface{}
```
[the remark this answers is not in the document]
cretz: This is still a problem at implementation time, I will bring it up in that PR.

go/workflow-interceptor-updates.md
```diff
@@ -0,0 +1,482 @@
...
+* Make inbound interceptors future-proof for argument addition
+* Support for using workflow/activity proto headers from interceptors
+  * One notable use case for this is to support validation (e.g. authz) and erroring out based on workflow/activity
+    header contents
+
+**These updates are known to NOT be backwards-compatible**
+
+## Interceptor Proposed Updates
+
+### Interceptor
```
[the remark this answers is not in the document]
cretz: Done

go/workflow-interceptor-updates.md
cretz: I don't believe we are concerned with the copy

go/workflow-interceptor-updates.md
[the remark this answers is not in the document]
cretz: I have updated the proposal to reflect this change in thinking.

cretz:
The PR for this has been opened at https://github.com/temporalio/sdk-go/pull/610

cretz:
@bergundy - This proposal has long since completed, shall I close this PR?

PR #46, opened 2021-11-08
Add initial proposal for entire WF versioning

bergundy:


features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+- Users deploy their Workflows directly to the central repository
+- When a worker encounters the first Workflow Task in history it decides which Workflow code it needs to run
+  - New Workflows always use the "latest" version
+  - Workflows that were started with an old code version use that same version by default (unless there's an existing patch version (more on that below))
+
+### How does it work?
+
+- In JavaScript (and probably WASM in the future) we bundle the Workflow code with its dependencies into a single file
+- In Java we may use Jars (pending POC)
+- A user uploads their bundle to a central repository stating wheter it's the "latest" version or if it patches an older version
```
Sushisource, line 23:
```suggestion
- A user uploads their bundle to a central repository stating whether it's the "latest" version or if it patches an older version
```

features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+    - side-effect: No need for `deprecatePatch` in most cases
+  - Example of a version graph (tag versions denote patches)
+
+  [![Version graph](https://mermaid.ink/img/eyJjb2RlIjoiZmxvd2NoYXJ0IExSXG4gICAgVjEgLS0-IFYyIC0tPiBWM1xuICAgIFYxIC0tPiBWMScgLS0-IFYxJydcbiAgICBWMiAtLT4gVjInIC0teCBWMicnXG4gICAgVjInIC0tPiBWMicnYVxuICAgIHN1YmdyYXBoIGxhdGVzdFxuICAgICAgICBWM1xuICAgIGVuZFxuICAgIHN1YmdyYXBoIHBhdGNoZml4XG4gICAgICAgIFYyJydhXG4gICAgZW5kIiwibWVybWFpZCI6eyJ0aGVtZSI6ImRlZmF1bHQifSwidXBkYXRlRWRpdG9yIjpmYWxzZSwiYXV0b1N5bmMiOnRydWUsInVwZGF0ZURpYWdyYW0iOmZhbHNlfQ)](https://mermaid-js.github.io/mermaid-live-editor/edit#eyJjb2RlIjoiZmxvd2NoYXJ0IExSXG4gICAgVjEgLS0-IFYyIC0tPiBWM1xuICAgIFYxIC0tPiBWMScgLS0-IFYxJydcbiAgICBWMiAtLT4gVjInIC0teCBWMicnXG4gICAgVjInIC0tPiBWMicnYVxuICAgIHN1YmdyYXBoIGxhdGVzdFxuICAgICAgICBWM1xuICAgIGVuZFxuICAgIHN1YmdyYXBoIHBhdGNoZml4XG4gICAgICAgIFYyJydhXG4gICAgZW5kIiwibWVybWFpZCI6IntcbiAgXCJ0aGVtZVwiOiBcImRlZmF1bHRcIlxufSIsInVwZGF0ZUVkaXRvciI6ZmFsc2UsImF1dG9TeW5jIjp0cnVlLCJ1cGRhdGVEaWFncmFtIjpmYWxzZX0)
+
+- When processing the first Workflow Task, the SDK inserts a version marker to record which code version was used to process this workflow history.
```
Sushisource, line 31: We should require that workflows which are using this feature keep using this feature. IE: If you start a worker whose code you *haven't* uploaded to the repro, it should fail WFTs in a loop.

Alternatively since it sounds like the worker in this situation is downloading/uploading from the repro automatically, we should require that upload succeeds before any processing happens.

features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+- Some way to determine when old workers can be completely removed
+
+#### Comparison of existing recommendation vs. new versioning scheme
+
+##### Pros (for new proposal)
+
+- No need for user to keep old workers around
+- Better out of the box experience
+- Simplifies versioning for the majority of the userbase
+- Activities on old workers may become outdated
```
Sushisource, line 59: This is a con?

features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+
+##### Cons
+
+- Activities may be removed or upgraded and become incompatible with older Workflows
+  - Could we "bundle" those too?
```
Sushisource, line 64: For Java that's fairly easy. For WASM, not so much.

features/entire-workflow-versioning.md
cretz, line 23: This is very close to serverless/FaaS and I think we should explore the landscape here. Many of them have solved similar versioning issues with uploaded software.

features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+The current method of versioning Workflows is complex and error prone.
+By default, there's a good chance that upgrading a worker will result in workflows being stuck due to non-determinism errors.
+Recovering from this situation is not trivial because a user now has histories generated by mulitple versions.
+
+### Proposal
+
+Entire Workflow versioning is an approach that'll greatly simplify the current Workflow versioning situation.
+
+- We add a concept of a central repository for Workflow code
+- Users deploy their Workflows directly to the central repository
```
cretz, line 14: This is very close to serverless/FaaS and I think we should explore the landscape here. Many of them have solved similar versioning issues with uploaded software.

features/entire-workflow-versioning.md
Spikhalskiy: Bundling activities creates another problem of upgrading activities if external APIs are changing.

features/entire-workflow-versioning.md
bergundy: > We should require that workflows which are using this feature keep using this feature. IE: If you start a worker whose code you haven't uploaded to the repro, it should fail WFTs in a loop.
I don't think it'd be a per-workflow flag, I'd rather it be the default worker behavior. We can consider making it configurable but I believe we should not make this an optional feature this for languages that support it.
We can maintain backwards compatibility with histories generated by old workers by defaulting to using `latest`.

> Alternatively since it sounds like the worker in this situation is downloading/uploading from the repro automatically, we should require that upload succeeds before any processing happens.

I imagined the user would upload the workflow bundle not a worker.
The version would have to be present for the Worker to reference it.

features/entire-workflow-versioning.md
bergundy: I'll fix

features/entire-workflow-versioning.md
bergundy: After our talk yesterday I'm pretty convinced that we shouldn't version activities since workflows require versioning the implementation and activities only require versioning their interface.

For activities, it's also fairly simple to keep old ones around and have a "router" activity that directs to an implementation according to the workflow version (which should be made available in the activity context's info).

features/entire-workflow-versioning.md
bergundy: I agree, and this proposal takes us a step closer to hosting workflows (if we ever go that direction).
As for bundling, we've pretty much got that figured out in TypeScript, we provide our own bundler (which internally uses webpack).

features/entire-workflow-versioning.md
```diff
@@ -0,0 +1,68 @@
...
+- When a worker encounters the first Workflow Task in history it decides which Workflow code it needs to run
```
dnr, line 15: Why not have the server decide which version to run, and include it in the workflowexecutionstarted event? Then there could be additional options to the startworkflowexecution rpc to choose a specific version (or "testing" tag, etc.)  Also it could help ensure that a batch of workflows are all started with the same version, even during a deployent.

bergundy: This touches worker deployment which we decided to avoid for the sake of narrowing the scope of this proposal.
Setting the version from the server is an option, it adds more work for the server while this feature might not be supported by all SDKs.
Having the tag set in the startworkflowexecution request is something we'd have to consider.

I think for now we should assume there's only a `latest` tag.

bergundy:
This is now irrelevant

PR #49, opened 2022-01-25
Python SDK - Phase 1

cretz:
# Summary

This is an initial proposal to start development on the Python SDK.

[See it rendered here](https://github.com/cretz/temporal-proposals/blob/python-phase1/python/phase-1.md) (may be deleted when proposal is done).

Feel free to leave a bunch of comments. If discussion becomes too difficult to follow, I will close this PR and start a new one.

### DIFF
--- a/python/phase-1.md
+++ b/python/phase-1.md
@@ -16,7 +16,7 @@ The Python SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-co
 The high-level goals for the Python SDK are:
 
 * Be as simple as possible from a caller POV
-* Be as lightweight as possible wrt dependencies
+* Be as lightweight as reasonable with regards to dependencies
 * Be reasonably flexible with reasonable defaults
 * Support only Python >= 3.7
 * Be friendly to REPL/interpreter use cases
@@ -36,10 +36,10 @@ An example of a client (intentionally missing _a lot_ of features):
 ```python
 import temporalio
 
-def main():
-  client = temporalio.Client("localhost:7233", namespace="MyNamespace")
-  result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
-  print("Result", result)
+async def main():
+    client = await temporalio.Client.create("localhost:7233", namespace="MyNamespace")
+    result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
+    print("Result", result)
 ```
 
 An example of an activity and registration (also missing a lot of features):
@@ -50,10 +50,11 @@ import temporalio
 def say_hello(name: str) -> str:
   return f"Hello, {name}!"
 
-def main(client: temporalio.Client):
-  worker = temporalio.Worker(client, task_queue="somequeue")
-  worker.register_activity(say_hello)
-  await worker.run()
+async def main():
+    core = await temporalio.Core.create("localhost:7233", namespace="MyNamespace")
+    worker = temporalio.Worker(core, task_queue="somequeue")
+    worker.register_activity(say_hello)
+    await worker.run()
 ```
 
 ## API Definitions
@@ -75,16 +76,27 @@ exported from `temporalio`):
 
 ```python
 class Client:
-    def __init__(
-        self,
+    @staticmethod
+    async def create(
         addr: str,
         *,
         namespace: str = "default",
         default_task_queue: str = None,
         identity: str = None
         # ... additional options omitted for brevity
-    ) -> None:
+    ) -> Client:
         # Actually dials gRPC here
+        pass
+
+    def __init__(
+        self,
+        service: workflowservice.WorkflowService,
+        *,
+        namespace: str = "default",
+        default_task_queue: str = None,
+        identity: str = None
+        # ... additional options omitted for brevity
+    ) -> None:
         pass
 
     # Raw gRPC service
@@ -92,7 +104,8 @@ class Client:
     def service(self) -> workflowservice.WorkflowService:
         pass
 
-    def start_workflow(
+    async def start_workflow(
+        self,
         workflow: str,
         *args,
         id: str = None,
@@ -102,15 +115,27 @@ class Client:
         start_signal: str = None,
         start_signal_args=[],
         # ... additional options omitted for brevity
-    ) -> WorkflowHandle[Any]:
+    ) -> WorkflowHandle[Any]:  # This result defines __await__ among other things
         pass
 
     # In a future phase:
     # @overload
-    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowExecution[T]
+    # async def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowHandle[T]
+    #
+    # @overload
+    # async def start_workflow(workflow: T, *args) -> T
+
+    # Just a wrapper for start workflow that awaits the handle
+    async def execute_workflow(
+        self,
+        workflow: str,
+        *args,
+        # ... additional options omitted for brevity
+    ) -> Any:
+        return await (await self.start_workflow(workflow, *args))
 
     # Does not validate whether a workflow exists
-    def workflow(
+    def get_workflow_handle(
         workflow_id: str,
         run_id: str = None
         # ...
@@ -156,10 +181,47 @@ Notes:
   * We could make an abstract base class if we want for typing/purposes that just has the `service` property
     unimplemented
 * We call it `start_workflow` instead of just `start` to be clearer
+* We have an `execute_workflow` to combine start + result waiting
 * We have embedded `start_signal` and `start_signal_args` as parameters, so if the former is present we will perform a
   signal with start. Is this ok?
   * This is an intentional deviation from other SDKs, but makes logical sense
 * The client supports a default task queue, do we want this?
+
+### Core
+
+Exported from `temporalio` (but probably in `temporalio/core/core.py`, exported from `temporalio.core` then
+exported from `temporalio`):
+
+```python
+class Core:
+    async def create(
+        addr: str,
+        *,
+        namespace: str = "default"
+        # ... All the same options as Client.create
+    ) -> Core:
+        pass
+```
+
+Notes:
+
+* If we could use Core to back our API client, then this would not be needed
+  * Even if we didn't, maybe if the worker feature is enabled we can create a core instance during client creation? Then
+    we can pass a client to worker instead of core which would be much cleaner
+* While we could implicitly create this, say, per address I think each core instance has enough uniqueness to require
+  manual user management
+* We could flip the wording of this a bit and have a "worker" not be task queue specific and be 1:1 with core, then have
+  the concept of a "task queue worker" that then lets you register activities and such.
+  * Probably deviates too much from the norm. Still, it's quite unfortunate to have a two-step process to create a
+    worker.
+* Is there any harm in creating one core instance per worker? We could do away with this if not.
+  * Do we lose some of the shared core benefits by only registering one worker for each core instance?
+  * I assume there is a meaningful difference between 1000 core instances that we poll on each and just 1?
+* Any other ideas on removing Core from explicit creation by users while still allowing multiple workers on different
+  server addresses?
+  * We can't do something like one-core-per-server-addr because there are different core options than just addr
+  * Should we just make the `core` optional in worker and have a global one that they can init (and hope they do so
+    before starting worker)?
 
 ### Worker
 
@@ -167,13 +229,23 @@ Exported from `temporalio` (but probably in `temporalio/worker/worker.py`, expor
 exported from `temporalio`):
 
 ```python
+
 class Worker:
     def __init__(
         self,
         client: temporalio.Client,
         *,
         # Uses default in client if not present (or errors if no default)
-        task_queue: str = None
+        task_queue: str = None,
+        # Can be anything accepted by register_activities
+        activities=None,
+        # An executor to be used for synchronous activities. If set, the worker
+        # does not shut this down. If unset, the worker may use a thread pool
+        # internally.
+        activity_executor=None,
+        # Whether to use the given/defaulted executor for async activities as
+        # well.
+        use_activity_executor_on_async: bool = False,
         # ...
     ) -> None:
         pass
@@ -186,8 +258,8 @@ class Worker:
         *,
         # Set the name override. If not set, uses the @activity.run decorator
         # name. If that's not set, uses the function/method name or class name
-        # if it has __call__. If it's a lambda or nested function, a name must
-        # be provided (or it can be decorated of course).
+        # if it has __call__. If it's a lambda, a name must be provided (or it
+        # can be manually decorated).
         name: str = None
         # ...
     ) -> None:
@@ -216,9 +288,61 @@ Notes:
 * Stopping the worker is done via cancellation of the run task/future
 * Internally, activities will run in the same event loop `run()` is called on
 * By default will fail already-registered activities
-* Should we accept a set of activities to register in `__init__`?
+* We also accept a set of activities to register in `__init__` and caller can choose preferred approach
 * Does exporting `temporalio.Worker` directly cause any packaging issues if we want to separate the client as its own
   published package?
+  * I believe we can catch an import error
+* We allow a custom executor and potentially even default one for sync activities
+  * Should we default an executor for sync?
+  * Is it ok to not default an executor for async? Or should we use it for them too?
+
+### Activity Utilities
+
+These are all top-level calls in `temporalio.activity`:
+
+```python
+# Whether the current code is in an activity
+def in_activity() -> bool:
+    pass
+
+
+def info() -> Info:
+    pass
+
+
+class Info:
+    # Other fields/properties skipped for brevity
+
+    @property
+    def cancelled(self) -> bool:
+        pass
+
+    @property
+    def cancellation() -> asyncio.Task:
+        pass
+
+
+# Adapter that extracts info() only when available in context
+class LoggerAdapter(logging.LoggerAdapter):
+    pass
+
+
+_logger = LoggerAdapter(logging.getLogger(__name__))
+
+# Logger that activities can use w/ activity context info instead of making
+# their own (though making your own is recommended)
+def logger() -> logging.Logger:
+    return _logger
+
+
+def heartbeat(*details) -> None:
+    pass
+```
+
+Notes:
+
+* These all uses `contextvars` to get activity info or throw otherwise
+  * We will make the contextual state open to caller injection to support test mocking. The details are undecided atm.
 
 ### Activity Definition/Registration
 
@@ -251,7 +375,7 @@ async def another_activity(arg1):
     try:
         while True:
             await asyncio.sleep(1)
-            await activity.heartbeat("some details")
+            activity.heartbeat("some details")
     except asyncio.CancelledError as exc:
         raise RuntimeError("someone cancelled me!") from exc
 
@@ -289,13 +413,15 @@ def register_all(worker):
     worker.register_activity(activity.run(name="MyLambda2")(lambda arg1: arg1))
     # Methods work just like top-level methods
     worker.register_activity(MyActivities().not_automatically_registered)
-
-    # Fails because no name provided and it's not a top-level function
+    # Nested function registers as "nested_activity"
     def nested_activity():
         pass
     worker.register_activity(nested_activity)
+
     # Fails because it's not callable
     worker.register_activity(MyActivities())
+    # Fails because it's a lambda w/ no name
+    worker.register_activity(lambda arg1: arg1)
 
     # Registers only "activity1"
     worker.register_activities(MyActivities())
@@ -316,16 +442,18 @@ Notes:
   * When we get to workflows, we will do similar with `@workflow.run` vs `@workflow.signal` vs `@workflow.query`, etc
     and `workflow.info()`
 * Cancellation is task cancellation
-  * Do we need to differentiate worker shutdown from server-side cancellation?
-  * So we need an explicit future they can wait on for activity cancellation?
+  * We need to research whether, in Python 3.7, we can interrupt an async task w/ our own error reliably
+* Heartbeat is a synchronous call
+  * Should we throw a cancellation error if we know the activity is cancelled when they try to heartbeat?
+    * is this something core should be returning from `record_heartbeat` instead of lang?
 
-### Data Converters
+### Payload Converters
 
-Exported from `temporalio` (but probably in `temporalio/converter/converter.py`, exported from `temporalio.converter`
+Exported from `temporalio` (but maybe in `temporalio/converter/converter.py`, exported from `temporalio.converter`
 then exported from `temporalio`):
 
 ```python
-class DataConverter(ABC):
+class PayloadConverter(ABC):
     # This can be sync or async. Always takes a list. Can return a single
     # payload or a list of payloads with the same size as the input list or None
     # if the value cannot be converted with this converter. Any other result
@@ -343,7 +471,7 @@ class DataConverter(ABC):
         pass
 
 
-class CompositeDataConverter(DataConverter):
+class CompositePayloadConverter(PayloadConverter):
     def __init__(self, converters) -> None:
         pass
 
@@ -354,13 +482,13 @@ class CompositeDataConverter(DataConverter):
         pass
 
 
-DataConverter.default = CompositeDataConverter(
+PayloadConverter.default = CompositePayloadConverter(
     [
-        BinaryNullDataConverter(),
-        BinaryPlainDataConverter(),
-        JSONProtoDataConverter(),
-        BinaryProtoDataConverter(),
-        JSONPlainDataConverter(),
+        BinaryNullPayloadConverter(),
+        BinaryPlainPayloadConverter(),
+        JSONProtoPayloadConverter(),
+        BinaryProtoPayloadConverter(),
+        JSONPlainPayloadConverter(),
     ]
 )
 ```
--- a/python/phase-1.md
+++ b/python/phase-1.md
@@ -1,0 +1,380 @@
+# Python SDK - Phase 1
+
+The Python SDK will be implemented in three loosely defined phases:
+
+* Phase 1 - Initial project, activity worker, and client
+* Phase 2 - Workflow worker, local activities, async activity completion, etc
+* Phase 3 - Deterministic workflow sandbox, replayer, test suite (some parts may be combined with phase 2)
+
+This proposal is for phase 1. It is intentionally loosely defined and all aspects are subject to discussion/change.
+
+## Overview
+
+The Python SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-core), utilizing a bridge based on
+[PyO3](https://github.com/PyO3/PyO3) similar to `sdk-typescript`'s `neon`-based bridge.
+
+The high-level goals for the Python SDK are:
+
+* Be as simple as possible from a caller POV
+* Be as lightweight as possible wrt dependencies
+* Be reasonably flexible with reasonable defaults
+* Support only Python >= 3.7
+* Be friendly to REPL/interpreter use cases
+
+The stack/approach looks like:
+
+* Python >= 3.7 (so may need `typing_extension` for things like `ParamSpec`)
+* Poetry packaging, black formatting
+  * Pipenv originally considered
+* PyO3 + sdk-core
+* asyncio direct (i.e. no need for anyio-style wrapper)
+* Type annotations everywhere
+* Decorator-based activity and workflow definitions
+
+An example of a client (intentionally missing _a lot_ of features):
+
+```python
+import temporalio
+
+def main():
+  client = temporalio.Client("localhost:7233", namespace="MyNamespace")
+  result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
+  print("Result", result)
+```
+
+An example of an activity and registration (also missing a lot of features):
+
+```python
+import temporalio
+
+def say_hello(name: str) -> str:
+  return f"Hello, {name}!"
+
+def main(client: temporalio.Client):
+  worker = temporalio.Worker(client, task_queue="somequeue")
+  worker.register_activity(say_hello)
+  await worker.run()
+```
+
+## API Definitions
+
+A lot of loosely-defined API definitions are below. They are intentionally lacking details.
+
+High-level notes:
+
+* For demonstration/simplicity here we are leaving off a lot of proper typings (e.g. `Optional`) that the real version
+  will have
+* We leave off a lot of options on calls for this high-level proposal
+* We are not using proper function documentation, but rather inline comments for easier discussion
+* **NOTE: These will have proper docs and types when developed!**
+
+### Client
+
+Exported from `temporalio` (but probably in `temporalio/client/client.py`, exported from `temporalio.client` then
+exported from `temporalio`):
+
+```python
+class Client:
+    def __init__(
+        self,
+        addr: str,
+        *,
+        namespace: str = "default",
+        default_task_queue: str = None,
+        identity: str = None
+        # ... additional options omitted for brevity
+    ) -> None:
+        # Actually dials gRPC here
+        pass
+
+    # Raw gRPC service
+    @property
+    def service(self) -> workflowservice.WorkflowService:
+        pass
+
+    def start_workflow(
+        workflow: str,
+        *args,
+        id: str = None,
+        task_queue: str = None,
+        retry_policy: RetryPolicy = None,
+        # Note, we are embedding signal-with-start in here
+        start_signal: str = None,
+        start_signal_args=[],
+        # ... additional options omitted for brevity
+    ) -> WorkflowHandle[Any]:
+        pass
+
+    # In a future phase:
+    # @overload
+    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowExecution[T]
+
+    # Does not validate whether a workflow exists
+    def workflow(
+        workflow_id: str,
+        run_id: str = None
+        # ...
+    ) -> WorkflowHandle[Any]:
+        pass
+
+
+class WorkflowHandle(Generic[T]):
+    def __init__(
+        self,
+        client: Client,
+        workflow_id: str,
+        *,
+        run_id: str = None
+        # ...
+    ) -> None:
+        pass
+
+    def __await__(self) -> T:
+        pass
+
+    async def cancel(self):
+        pass
+
+    async def describe(self):
+        pass
+
+    async def query(self, name: str, *args):
+        pass
+
+    async def signal(self, name: str, *args):
+        pass
+
+    async def terminate(self, *, reason: str = None):
+        pass
+```
+
+Notes:
+
+* We are intentionally defining a `Client` instead of a `WorkflowClient` since this will be general purpose
+  * Having a general purpose client keeps things simple, but at least for now I can't see defining any more than the two
+    methods and the one property
+  * We could make an abstract base class if we want for typing/purposes that just has the `service` property
+    unimplemented
+* We call it `start_workflow` instead of just `start` to be clearer
+* We have embedded `start_signal` and `start_signal_args` as parameters, so if the former is present we will perform a
+  signal with start. Is this ok?
+  * This is an intentional deviation from other SDKs, but makes logical sense
+* The client supports a default task queue, do we want this?
+
+### Worker
+
+Exported from `temporalio` (but probably in `temporalio/worker/worker.py`, exported from `temporalio.worker` then
+exported from `temporalio`):
+
+```python
+class Worker:
+    def __init__(
+        self,
+        client: temporalio.Client,
+        *,
+        # Uses default in client if not present (or errors if no default)
+        task_queue: str = None
+        # ...
+    ) -> None:
+        pass
+
+    # Cannot be called after run started. Typings and separate @overload
+    # signatures omitted for brevity.
+    def register_activity(
+        # This must be a callable. Can be a lambda.
+        activity,
+        *,
+        # Set the name override. If not set, uses the @activity.run decorator
+        # name. If that's not set, uses the function/method name or class name
+        # if it has __call__. If it's a lambda or nested function, a name must
+        # be provided (or it can be decorated of course).
+        name: str = None
+        # ...
+    ) -> None:
+        pass
+
+    # Cannot be called after run started. Typings and separate @overload
+    # signatures omitted for brevity.
+    def register_activities(
+        # This must be an object or list or dictionary. If an object, this only
+        # uses methods with @activity.run decorators. Dictionaries and lists
+        # don't require the decorator. If a dictionary, the key is the name.
+        # This is just a helper and is the equivalent of just calling
+        # register_activity with all applicable functions.
+        activities,
+        # ...
+    ) -> None:
+        pass
+
+    async def run() -> None:
+        pass
+```
+
+Notes:
+
+* We have left off workflow registration until a later phase
+* Stopping the worker is done via cancellation of the run task/future
+* Internally, activities will run in the same event loop `run()` is called on
+* By default will fail already-registered activities
+* Should we accept a set of activities to register in `__init__`?
+* Does exporting `temporalio.Worker` directly cause any packaging issues if we want to separate the client as its own
+  published package?
+
+### Activity Definition/Registration
+
+```python
+from temporalio import activity
+
+# This adds activity details if they are present. Maybe we should have a
+# temporalio.LoggerAdapter that works for activities _and_ workflows?
+my_activity_logger = activity.LoggerAdapter(logging.getLogger(__name__))
+
+
+def simple_activity(arg1, arg2):
+    # Uses contextvars internally, which means it'll fail in another context but
+    # we will provide the ability to set this for testing situations
+    attempts = activity.info().attempts
+    # This logger has an adapter to include custom details but still named after
+    # this module which is a common pattern
+    my_activity_logger.debug("Attempt %s", attempts)
+    # This logger, on the activity module, uses the name of the activity module.
+    # So that's why we offer either the adapter or the logger.
+    activity.logger().debug("Attempt %s", attempts)
+
+
+@activity.run(name="MyActivity")
+async def another_activity(arg1):
+    # Showing heartbeat and cancellation. In Python 3.7 we cannot give a
+    # specific error to Task/Future.cancel (only in 3.9). So activity
+    # cancellations appear like any other asyncio cancellations (which may be a
+    # good thing anyways).
+    try:
+        while True:
+            await asyncio.sleep(1)
+            await activity.heartbeat("some details")
+    except asyncio.CancelledError as exc:
+        raise RuntimeError("someone cancelled me!") from exc
+
+class MyCallableClass:
+    def __call__(self):
+        pass
+
+@activity.run(name="ManualClassName")
+class AnotherCallableClass:
+    def __call__(self):
+        pass
+
+class MyActivities:
+    @activity.run
+    def activity1():
+        pass
+
+    def not_automatically_registered():
+        pass
+
+def register_all(worker):
+    # Registers as "simple_activity"
+    worker.register_activity(simple_activity)
+    # Using customized name
+    worker.register_activity(another_activity)
+    # Registers as "MyCallableClass"
+    worker.register_activity(MyCallableClass())
+    # Registers as "ManualClassName"
+    worker.register_activity(AnotherCallableClass())
+    # Manual name
+    worker.register_activity(another_activity, name="ManualActivityName")
+    # Lambda w/ name
+    worker.register_activity(lambda arg1: arg1, name="MyLambda")
+    # Lambda can of course manually apply decorator
+    worker.register_activity(activity.run(name="MyLambda2")(lambda arg1: arg1))
+    # Methods work just like top-level methods
+    worker.register_activity(MyActivities().not_automatically_registered)
+
+    # Fails because no name provided and it's not a top-level function
+    def nested_activity():
+        pass
+    worker.register_activity(nested_activity)
+    # Fails because it's not callable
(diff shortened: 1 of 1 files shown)


python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* Type annotations everywhere
+* Decorator-based activity and workflow definitions
+
+An example of a client (intentionally missing _a lot_ of features):
+
+```python
+import temporalio
+
+def main():
+  client = temporalio.Client("localhost:7233", namespace="MyNamespace")
```
cretz, line 40: We might need to change this to `temporalio.Client.create` so it can be `await`d on.

bergundy: I don't think we want to expose `Client` here, I'd rather stick to what we have in Java and TS where there's a grpc client (with a small wrapper to create it), a Worker and a WorkflowClient.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* asyncio direct (i.e. no need for anyio-style wrapper)
+* Type annotations everywhere
+* Decorator-based activity and workflow definitions
+
+An example of a client (intentionally missing _a lot_ of features):
+
+```python
+import temporalio
+
+def main():
```
bergundy, line 39: nit: this should be an async function or should call `asyncio.run`

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+An example of an activity and registration (also missing a lot of features):
+
+```python
+import temporalio
+
+def say_hello(name: str) -> str:
+  return f"Hello, {name}!"
+
+def main(client: temporalio.Client):
+  worker = temporalio.Worker(client, task_queue="somequeue")
```
bergundy, line 54: We would probably want to expose Core directly here so we can reuse a single instance with multiple workers.

bergundy: Are we going to have a singleton Core in Python too?

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+Notes:
+
+* We use `@activity.run` from `temporalio.activity` instead of just `@activity` from because:
+  * We want `activity` to be a module because we want it usually imported as `from temporalio import activity` so
+    top-level calls like `activity.info()` read well
+  * When we get to workflows, we will do similar with `@workflow.run` vs `@workflow.signal` vs `@workflow.query`, etc
+    and `workflow.info()`
+* Cancellation is task cancellation
+  * Do we need to differentiate worker shutdown from server-side cancellation?
+  * So we need an explicit future they can wait on for activity cancellation?
```
cretz, line 320:
```suggestion
  * Do we need an explicit future they can wait on for activity cancellation?
```

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        # Note, we are embedding signal-with-start in here
+        start_signal: str = None,
+        start_signal_args=[],
+        # ... additional options omitted for brevity
+    ) -> WorkflowHandle[Any]:
+        pass
+
+    # In a future phase:
+    # @overload
+    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowExecution[T]
```
bergundy, line 110:
```suggestion
    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowHandle[T]
```

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+    def start_workflow(
+        workflow: str,
+        *args,
+        id: str = None,
+        task_queue: str = None,
+        retry_policy: RetryPolicy = None,
+        # Note, we are embedding signal-with-start in here
+        start_signal: str = None,
+        start_signal_args=[],
```
bergundy, line 103: Not convinced we should unify these APIs

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        pass
+
+    async def describe(self):
+        pass
+
+    async def query(self, name: str, *args):
+        pass
+
+    async def signal(self, name: str, *args):
+        pass
```
bergundy, line 145: These will probably be generic too eventually

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        client: Client,
+        workflow_id: str,
+        *,
+        run_id: str = None
+        # ...
+    ) -> None:
+        pass
+
+    def __await__(self) -> T:
+        pass
```
bergundy, line 133: I'd really rather make an async function called `result` like we have in Java / TS.

Otherwise you can have weird code like this:
```python
res = await (await client.start_workflow(...))
```

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* May be a bit difficult to allow async and non-async converter w/ same base but we can
+  * Ref https://stackoverflow.com/questions/47555934/how-require-that-an-abstract-method-is-a-coroutine
+* This is a combination of the otherwise separate concepts of "DataConverter" and "PayloadConverter" from other SDKs
+  since the extra abstraction seems unnecessary, is that ok?
+
+### Other API Items
+
+The following API items aren't in the proposal because their use should be straightforward:
+
+* We'll extend exception and properly convert to/from for most errors at
```
cretz, line 379: > convert to/from

I should clarify that we will keep raw protos available which is important for failure details

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        # ... additional options omitted for brevity
+    ) -> WorkflowHandle[Any]:
+        pass
+
+    # In a future phase:
+    # @overload
+    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowExecution[T]
+
+    # Does not validate whether a workflow exists
+    def workflow(
```
bergundy, line 113: 1. I'd rather stick to the name `get_handle`
2. This should become generic too (if possible)

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* We are not using proper function documentation, but rather inline comments for easier discussion
+* **NOTE: These will have proper docs and types when developed!**
+
+### Client
+
+Exported from `temporalio` (but probably in `temporalio/client/client.py`, exported from `temporalio.client` then
+exported from `temporalio`):
+
+```python
+class Client:
```
bergundy, line 77: This should probably just be called `WorkflowClient`

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        default_task_queue: str = None,
+        identity: str = None
+        # ... additional options omitted for brevity
+    ) -> None:
+        # Actually dials gRPC here
+        pass
+
+    # Raw gRPC service
+    @property
+    def service(self) -> workflowservice.WorkflowService:
```
bergundy, line 92: I'd expect the `service` to be provided to via constructor so it can be shared by multiple clients.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    async def signal(self, name: str, *args):
+        pass
+
+    async def terminate(self, *, reason: str = None):
+        pass
+```
+
+Notes:
+
+* We are intentionally defining a `Client` instead of a `WorkflowClient` since this will be general purpose
```
bergundy, line 153: Let's discuss this because I'm not convinced this is the right approach.
What other general purpose methods would this client have?

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+### Worker
+
+Exported from `temporalio` (but probably in `temporalio/worker/worker.py`, exported from `temporalio.worker` then
+exported from `temporalio`):
+
+```python
+class Worker:
+    def __init__(
+        self,
+        client: temporalio.Client,
```
bergundy, line 173: You'll need a Core instance here, not a client.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    # Cannot be called after run started. Typings and separate @overload
+    # signatures omitted for brevity.
+    def register_activity(
+        # This must be a callable. Can be a lambda.
+        activity,
+        *,
+        # Set the name override. If not set, uses the @activity.run decorator
+        # name. If that's not set, uses the function/method name or class name
+        # if it has __call__. If it's a lambda or nested function, a name must
+        # be provided (or it can be decorated of course).
```
bergundy, line 190: nit: you should use python docstrings to document the params

bergundy (commented):
Managed to review most of this, I'll get back to it later.

python/phase-1.md
cretz: True

python/phase-1.md
cretz: This is mostly the `WorkflowClient` from the Java and TS perspective, I am just calling it "Client" for clarity because I think that's clearer and lets us add things like async activity heartbeat/completion to it.

I am trying to have the simplest API possible, which means I am trying to avoid two-step client creation. But if y'all really want, I can make a gRPC client, a workflow client, an async activity client, etc. It just feels bloated/unnecessary.

python/phase-1.md
cretz: I was planning on having a singleton. I can accept a kwarg for `core` that I default to the singleton.

python/phase-1.md
cretz: My thoughts here were to allow other client uses such as async activity heartbeat/cancellation. Is it important that we separate that stuff out and make more classes? Why?

python/phase-1.md
cretz: That adds an unnecessary second step. Creating a client and using it is very reasonable/simple.

However since I am switching to a static `create` method for `await` purposes, so I'll have an overload for those that really want to manually create a gRPC client.

python/phase-1.md
cretz: Yup, it's an open question for discussion. I think it adds great simplicity from an API POV, but I would love to hear reasons why not besides "because we never have". I don't think the explicitness of adding `_with_signal` to the end of the name has any more value than adding `start_signal` to the end of the params.

python/phase-1.md
cretz: > I'd rather stick to the name get_handle

Can you provide justification? Running `await client.workflow("myworkflow")` is very clear software. I am really trying to make the code look great from a caller perspective but I fear that we're more concerned with other things.

> This should become generic too (if possible)

It would definitely be generic later. I provided an overload for the previous one just for thought, I do not provide all possible overloads throughout this review.

python/phase-1.md
cretz: `res = await client.start_workflow(...)` is the expected code here. This is awaiting the future internally.

python/phase-1.md
cretz: Yes, there are a lot that will be. I wanted to make haste with high level discussion points here (I said in the description, the top of the readme, and in other comments that I intentionally do not provide every type).

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* Decorator-based activity and workflow definitions
+
+An example of a client (intentionally missing _a lot_ of features):
+
+```python
+import temporalio
+
+def main():
+  client = temporalio.Client("localhost:7233", namespace="MyNamespace")
+  result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
```
cretz, line 41: I forgot that I changed this to `start_workflow`

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    # Showing heartbeat and cancellation. In Python 3.7 we cannot give a
+    # specific error to Task/Future.cancel (only in 3.9). So activity
+    # cancellations appear like any other asyncio cancellations (which may be a
+    # good thing anyways).
+    try:
+        while True:
+            await asyncio.sleep(1)
+            await activity.heartbeat("some details")
+    except asyncio.CancelledError as exc:
+        raise RuntimeError("someone cancelled me!") from exc
```
Sushisource, line 256: We will also want something like https://typescript.temporal.io/api/classes/activity.context/#cancelled

(recall Local acts can be cancelled/timed-out w/o heartbeating)

Not having our own type is a biiit rough. B/c you can imagine the user cancelling a task they themselves created inside the activity, and the error thrown there is now indistinguishable from a "Your whole activity got cancelled" error, since AFAICT there's not any extra information on the [type](https://docs.python.org/3/library/asyncio-exceptions.html#asyncio.CancelledError) (not the hugest fan of python docs...) -- and clearly we'll want some way to make that distinction. I suppose we can always just cram in some extra field or whatever since python, but that'd be hard to discover.

python/phase-1.md
cretz: >  What other general purpose methods would this client have?

Off the top of my head probably only async activity stuff. From below:

> but at least for now I can't see defining any more than the two methods and the one property

It is a good, simple concept to "create client, call client". We are trying to hard to add words and terms and steps and other barriers to simplicity.

python/phase-1.md
cretz: Derp, that is true! Not sure how I missed that.

python/phase-1.md
cretz: From above:

>  We are not using proper function documentation, but rather inline comments for easier discussion

Also from above:

> **NOTE: These will have proper docs and types when developed!**

I bolded it. If I need to resubmit the proposal with exact types and docs as the real implementation I can.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+class MyCallableClass:
```
Sushisource, line 258: Do we want activity classes? IMO they're a bit of an antipattern as they encourage attaching implicit state to something that we encourage users to think of as function invocation.

I kinda think we should push people towards making their activities take all their input as args. The counterargument here is doing things like injecting mocks, etc... but that might be nicer to get off a context?

Not super firm on this, but do want to throw it out there.

python/phase-1.md
Sushisource, line 320: In my other comment, but yup

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+Notes:
+
+* We use `@activity.run` from `temporalio.activity` instead of just `@activity` from because:
+  * We want `activity` to be a module because we want it usually imported as `from temporalio import activity` so
+    top-level calls like `activity.info()` read well
+  * When we get to workflows, we will do similar with `@workflow.run` vs `@workflow.signal` vs `@workflow.query`, etc
+    and `workflow.info()`
+* Cancellation is task cancellation
+  * Do we need to differentiate worker shutdown from server-side cancellation?
```
Sushisource, line 319: It would be good to include that information somehow

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    ]
+)
+```
+
+Notes:
+
+* May be a bit difficult to allow async and non-async converter w/ same base but we can
+  * Ref https://stackoverflow.com/questions/47555934/how-require-that-an-abstract-method-is-a-coroutine
+* This is a combination of the otherwise separate concepts of "DataConverter" and "PayloadConverter" from other SDKs
+  since the extra abstraction seems unnecessary, is that ok?
```
Sushisource, line 373: I'm not sure what the point of the distinction is either.

I do think we eventually want separate bytes-to-bytes processors from the serializers, but different story.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+```python
+import temporalio
+
+def say_hello(name: str) -> str:
+  return f"Hello, {name}!"
+
+def main(client: temporalio.Client):
+  worker = temporalio.Worker(client, task_queue="somequeue")
+  worker.register_activity(say_hello)
```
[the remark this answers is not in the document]
cretz: To confirm, you think we should require not only the decorator, but also that they explicitly provide a string name? I like the simplicity of the current approach personally.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+    # Cannot be called after run started. Typings and separate @overload
+    # signatures omitted for brevity.
+    def register_activity(
+        # This must be a callable. Can be a lambda.
+        activity,
+        *,
+        # Set the name override. If not set, uses the @activity.run decorator
+        # name. If that's not set, uses the function/method name or class name
+        # if it has __call__. If it's a lambda or nested function, a name must
```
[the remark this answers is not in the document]
cretz: I disagree. I think simplicity is important here, even if it means implicitness by default. I don't think there's more value in a dynamic situation like Python making people type more in this case.

I think TS is unnecessarily complex to do this stuff in. I think we should really optimize for simplicity from the caller POV with defaults that can be overridden.

The lambda use case is just so I don't have to care about the type of callable. Simply put "if it's callable, you can register it as an activity (but you might have to provide a name)".

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+exported from `temporalio`):
+
+```python
+class Worker:
+    def __init__(
+        self,
+        client: temporalio.Client,
+        *,
+        # Uses default in client if not present (or errors if no default)
+        task_queue: str = None
```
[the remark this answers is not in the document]
cretz: From a bullet above:

> The client supports a default task queue, do we want this?

This is an intentional deviation from existing clients, but I am happy to revert if not wanted. But in this case, I derp'd and assumed I could reuse a client for the worker and not a core instance. So we will require a task queue here.

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: Retired in favor of other thread

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: It's a super minimal amount of extra typing that makes things more explicit in a way that prevents a certain type of bug we have seen users run into in the real world. That's a good thing, IMO. If all your worker code is always bundled together this is 0% an issue, but if you're calling activities from other languages, you have to understand it's derived from the fn name (to be fair that's pretty obvious) which means you also have to be smart enough to realize you don't get to change that function name once you start relying on it. That's the footgun.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        pass
+```
+
+Notes:
+
+* We have left off workflow registration until a later phase
+* Stopping the worker is done via cancellation of the run task/future
+* Internally, activities will run in the same event loop `run()` is called on
+* By default will fail already-registered activities
+* Should we accept a set of activities to register in `__init__`?
```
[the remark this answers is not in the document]
cretz: The simplicity to me of "create worker, add activities to worker, run worker" is what drove the current API. There's enough validation (and potential future options) to go with registration that allowing them to be done by method is reasonable. I am not sure a whole separate builder adds anything more than weight/bloat from an API POV.

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: It's also a nice signal to other non-temporal code you might have in the same process that "Hey, this is sometimes used as a temporal activity".

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+* Does exporting `temporalio.Worker` directly cause any packaging issues if we want to separate the client as its own
+  published package?
```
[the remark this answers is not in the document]
cretz: My goal was to have the commonly used pieces available in the top level module. `temporalio.Client`, `temporalio.Worker`, `temporalio.DataConverter`, etc (of course they're defined in their own module, but re-exported here).

But if I have to split them up due to pypi package needs I can. Though I wonder if the common stuff should be in `temporalio` and its worker that should be a separate package? Or do I have to introduce a `temporalio.common` now too? Or is something like `temporalio.DataConverter` get a separate package too like `converter`? I think this is going from simple to sprawl pretty quick

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+### Activity Definition/Registration
+
+```python
+from temporalio import activity
+
+# This adds activity details if they are present. Maybe we should have a
+# temporalio.LoggerAdapter that works for activities _and_ workflows?
+my_activity_logger = activity.LoggerAdapter(logging.getLogger(__name__))
```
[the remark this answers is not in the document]
cretz: I am intentionally making both available to let the caller choose. There are reasons for `activity.logger()` (and especially its analog `workflow.logger()`) and using your own logger w/ just an adapter. And really `activity.logger()` is just an already defined logger using the adapter for context data.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    attempts = activity.info().attempts
+    # This logger has an adapter to include custom details but still named after
+    # this module which is a common pattern
+    my_activity_logger.debug("Attempt %s", attempts)
+    # This logger, on the activity module, uses the name of the activity module.
+    # So that's why we offer either the adapter or the logger.
+    activity.logger().debug("Attempt %s", attempts)
+
+
+@activity.run(name="MyActivity")
```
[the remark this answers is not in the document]
cretz: From below:

> * We use `@activity.run` from `temporalio.activity` instead of just `@activity` from because:
>   * We want `activity` to be a module because we want it usually imported as `from temporalio import activity` so
>     top-level calls like `activity.info()` read well
>   * When we get to workflows, we will do similar with `@workflow.run` vs `@workflow.signal` vs `@workflow.query`, etc
>     and `workflow.info()`

It will match `@workflow.run` a bit better. If we ever get activity signals, we'll be happy we did this, heh.

python/phase-1.md
cretz: > We will also want something like https://typescript.temporal.io/api/classes/activity.context/#cancelled

Can do

> Not having our own type is a biiit rough.

We don't in Go, we use the languages native contextual cancellation process. I agree we need our own version, but from above:

>     # Showing heartbeat and cancellation. In Python 3.7 we cannot give a
>     # specific error to Task/Future.cancel (only in 3.9). So activity
>     # cancellations appear like any other asyncio cancellations (which may be a
>     # good thing anyways).

I admittedly have not done enough research here, but I agree if I can use our own cancellation error I will. But I was just worried reading docs that issuing a cancel in asyncio world in the version of Python we're targeting doesn't allow you to customize the type.

python/phase-1.md
cretz: > Do we want activity classes?

We don't care what type of callable they are using. If it's callable, we can call it.

> IMO they're a bit of an antipattern as they encourage attaching implicit state to something that we encourage users to think of as function invocation.

Not sure it is an antipattern. We often encourage users to use activity structs in Go when you want something like a client shared.

We agree that basic serializable values should be params.

python/phase-1.md
cretz: Yes, in addition to a future that they can wait on for cancellation, I will also provide an "is cancelled" bool (or other similar approach to https://docs.python.org/3/library/asyncio-task.html#asyncio.Task.cancelled).

cretz:
@bergundy and @Sushisource - Do we want to consider using core's gRPC client for _all_ gRPC client calls? Obviously this then makes the client package require core (which makes the separation of published packages unnecessary), so that's a downside. But is there enough upside to connection sharing? Have y'all seen on the TS side annoyances with, say, configuring TLS or interceptors for two separate connections rough?

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: Sure, but, my suggestion applies equally to `workflow.run`. I think the suggested alternatives fit it just as well

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: Sure, you want both - but why ask this question then?

My point is if I can just say `TemporalLogger(mylogger)` and it discovers what context it's in and does the right thing that's an easier experience. If we can't do it, sure, instantiate them differently.

python/phase-1.md
bergundy: I'm fine with unifying the async activity completion client and workflow client into a single `Client` class.

python/phase-1.md
bergundy: This would actually be `execute_workflow` since it returns a result (at least that's what it's called in Java / TS)

python/phase-1.md
bergundy: As mentioned above I'm fine if this is just `WorkflowClient` + `ActivityCompletionClient` combined.
In hindsight I wish we'd just combined those in TS.

python/phase-1.md
bergundy: 👍

Sushisource:
@cretz I mean it's annoying but it's not exactly a gigantic amount of work. It's actually pretty darn straightforward once we have the policies documented. Connection sharing has a few other advantages besides de-dupe as well.

Technically it doesn't need to require all of core either, as there is now a separate client crate. 

As to whether the compiled extensions are a burden... I think less so for Python than JS that's for sure. So honestly it's not a super clear choice to me. My inclination would be to just go for it and if shipping it that way is a burden then just... stop doing it that way. If we know up front some additional good reasons not to, I don't feel strongly about it.

python/phase-1.md
bergundy: `workflow` is confusing, it's not a verb so it's unclear what it actually does.
Additionally this method should not be async since it does not do any IO.

I'm fine with not providing all of the overloads, didn't mean for this to be addressed in this PR.
We'll revisit later when we get to defining the workflow API

python/phase-1.md
bergundy: `await start_workflow` will return a handle not the result.

python/phase-1.md
bergundy: Let's drill down into the actual handle API when we define the workflow part of the spec.
I'm treating this as a general structure which I think you got right.

python/phase-1.md
bergundy: working on the go SDK for too long :)

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+        self,
+        client: temporalio.Client,
+        *,
+        # Uses default in client if not present (or errors if no default)
+        task_queue: str = None
+        # ...
+    ) -> None:
+        pass
+
+    # Cannot be called after run started. Typings and separate @overload
```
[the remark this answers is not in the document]
bergundy: Maybe instead of all of these `register_activity` calls we can just give a mapping of name to implementation to the worker at construction time, then you don't have to worry about when this method can be called.

That way you can just do something like:
```python
import activities

Worker(task_queue, activities=vars(activities))
```

python/phase-1.md
bergundy: This was just a nit, the proposal is fine the way it is for this stage.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: no builders please, they're not a thing in Python.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: We might be able to put the worker under a feature flag in a single package with the client.

python/phase-1.md
bergundy: We don't have to do that for Python, the same interface can be used for both.

bergundy: I think we can unify the 2 concepts into a single class, I think `PayloadConverter` is a better name because payloads are a known Temporal term.

python/phase-1.md
bergundy: Good use for `raise ... from` https://docs.python.org/3/tutorial/errors.html#exception-chaining

bergundy: We'd also need all of the `TemporalFailure` subclasses (https://github.com/temporalio/sdk-typescript/blob/690ea3c1ca0887b957b5b672689489daa6b77f55/packages/common/src/failure.ts#L48)

bergundy (commented):
I'm not 100% sure about the activity registration API.
I would prefer if we could pass all of the activities to the worker on construction and get rid of the registration methods.

python/phase-1.md
cretz: Yay! Thanks.

python/phase-1.md
cretz: > Good use for raise ... from

Yes, the example inside the activity uses `raise` `from`. This is a different deal, this is saying I need to keep the raw proto available inside the exception class

> We'd also need all of the TemporalFailure subclasses

That's what I mean with this line you are commenting on when saying:

> We'll extend exception and properly convert to/from for most errors at https://pkg.go.dev/go.temporal.io/api/serviceerror

python/phase-1.md
cretz: Sure. Because this may have to change if we want a more raw equivalent in order to solve for ambiguities (e.g.what if my workflow class defines a `@workflow.signal` as `signal`).

python/phase-1.md
cretz: Ah, true. And I can't really make `start_workflow` non-async and just return the awaitable because there's a difference between start-wait and completion-wait. I think I will change to follow the TS lead here and have `execute_workflow` that returns the completion result. But I would like to leave the handle as awaitable. They should never do `await (await client.start_workflow(...))` because they should be using `execute_workflow` instead.

python/phase-1.md
cretz: It's half getter half creator (you are creating the handle, not getting it). Basically, you might as well construct the workflow handler via constructor w/ a client and workflow ID but obviously that's uglier. Often when it's just a simple constructor, one expects to leave the `new_` or `create_` or whatever. And in order to disambiguate, I wanted the word `workflow` in there.

How about `get_workflow`? Or would you rather `get_workflow_handle`?

python/phase-1.md
cretz: Yes, I will be providing an `execute_workflow` overload that is basically await start_workflow + await handle.

python/phase-1.md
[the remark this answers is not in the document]
cretz: I brought up this point above with:

> Should we accept a set of activities to register in `__init__`?

Somewhat discussed at https://github.com/temporalio/proposals/pull/49/files/40c1178e1937b223c93334a7cefc75e31afc6bb9#r792182996. I'm now thinking both. I don't want the worker to start on construction anyways (I need it awaitable), so `run()` will always exist. Also, there benefit of the separate `register_` functions with regards to overload typing, documentation, and potential future per-activity options.

python/phase-1.md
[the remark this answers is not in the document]
cretz: To confirm, you are suggesting users should always have to provide a string literal for every activity even if, say, they always want it the function name? I think this footgun is only with some, others recognize the function name will be used as the activity name. It's kinda unfair that Go and Java get this simplicity benefit that this newer SDK will not.

python/phase-1.md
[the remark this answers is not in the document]
cretz: That would be quite nice. I will investigate this during implementation.

python/phase-1.md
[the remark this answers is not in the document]
cretz: > Sure, you want both - but why ask this question then?

Which question?

> Why does the activity need to use our logger?

It doesn't, we just offer one if they want to

> Let's let users log however they want.

Exactly, that's why we provide an adapter onto existing loggers to put contextual activity info if they want

> In TS we have a sample for injecting a logger into the activity context.

In this case, you attach something to the logger, not inject one. It's a more Python approach to contextual logging. We just happen to also have a `activity.logger()` which is essentially the logger inside the activity package + the logger adapter that automatically puts contextual info on it.

python/phase-1.md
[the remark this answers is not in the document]
cretz: Agreed, but I don't like them as much either. The others didn't fit as well:

* `@activity.def` - triggered the highlighter for the reserved `def`
* `@activity.define` - sounds like you're doing something instead of just marking it
* `@activity.definition`/`@activity.defn` - there's no `@workflow` equivalent because I want to mark where the entry is, not the entire workflow
* `@activity.fun`, `@activity.function` - "function" doesn't apply to all callables

But I too am not sure I like `@activity.run`. The main thing I'm thinking about from a workflow POV is marking the entrypoint of the workflow (not to be confused with marking the entire workflow). Then I just wanted to have an activity equivalent. Other options:

* `@activity.entry` - This doesn't fit activity as well as `@workflow.entry` because of course where you call the activity is its entrypoint
* `@activity.exec` - This actually works as well as `@workflow.exec`
* `@activity.call` - More the action than the marker (though can say the same for "run" and "exec", feels uglier here)

I have scoured the thesaurus. I am open to other options.

> Bikeshed territory, but an important one.

Very important IMO

python/phase-1.md
cretz: > Which raises another complication with non async activity functions.

They can't heartbeat anyways, heh. We should "accept" non-async activities since they can be just quick CPU bound tasks, but in all our examples, we should `async def` activities just to make it clear they are the suggested approach.

python/phase-1.md
cretz: > We might be able to use custom tasks so we can throw our own CancelledError subclass.

You have to do more than that. You also have to do your own event loop because you have to be able to intercept https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_future too not just tasks.

python/phase-1.md
cretz: I think "Data Converter" is a known Temporal term too, but I also prefer `PayloadConverter`. I've been looking around, and technically a `PayloadEncoder` + `PayloadDecoder` are more apt terms, but we combine them and there's no settled name in the ecosystem for such a combination. So I will switch to `PayloadConverter`.

python/phase-1.md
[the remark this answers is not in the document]
cretz: Setting the activity name at the same spot as the function name is reasonable (it can be registered to lots of workers). The decorator also serves other purposes such as for people that want to mark their activities clearly (though obv not required) and its required for activity class registration to mark which methods are activities.

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+@activity.run(name="MyActivity")
+async def another_activity(arg1):
+    # Showing heartbeat and cancellation. In Python 3.7 we cannot give a
+    # specific error to Task/Future.cancel (only in 3.9). So activity
+    # cancellations appear like any other asyncio cancellations (which may be a
+    # good thing anyways).
+    try:
+        while True:
+            await asyncio.sleep(1)
+            await activity.heartbeat("some details")
```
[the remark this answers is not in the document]
cretz: Makes sense. And I am also thinking about throwing cancellation from `heartbeat` if it the activity was cancelled, is that ok?

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+    # Manual name
+    worker.register_activity(another_activity, name="ManualActivityName")
+    # Lambda w/ name
+    worker.register_activity(lambda arg1: arg1, name="MyLambda")
+    # Lambda can of course manually apply decorator
+    worker.register_activity(activity.run(name="MyLambda2")(lambda arg1: arg1))
+    # Methods work just like top-level methods
+    worker.register_activity(MyActivities().not_automatically_registered)
+
+    # Fails because no name provided and it's not a top-level function
```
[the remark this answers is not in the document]
cretz: Yeah, I can change this. My original reason was that while `__name__` worked, `__qualname__` is outer + `.<locals>.` + inner and I just didn't want callers to be confused on what might be used.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: I like that we have `Worker.create()` and `Worker.run()` in TS.
Note that `create()` will try to connect to the server.

python/phase-1.md
cretz: Updated

python/phase-1.md
cretz: Updated

python/phase-1.md
cretz: Changed to use core

python/phase-1.md
cretz: Fixed

python/phase-1.md
cretz: I have added `execute_workflow` for those that want to just start and get result.

python/phase-1.md
cretz: I have changed this to `get_workflow_handle`

python/phase-1.md
cretz: Per other conversation, we're good here

python/phase-1.md
cretz: Updated

python/phase-1.md
cretz: Done

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+## Overview
+
+The Python SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-core), utilizing a bridge based on
+[PyO3](https://github.com/PyO3/PyO3) similar to `sdk-typescript`'s `neon`-based bridge.
+
+The high-level goals for the Python SDK are:
+
+* Be as simple as possible from a caller POV
+* Be as lightweight as possible wrt dependencies
```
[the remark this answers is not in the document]
cretz: Fixed

python/phase-1.md
cretz: I have added a section on activity utilities that provides the cancellation options.

python/phase-1.md
cretz: Done

python/phase-1.md
cretz: Added cancelled boolean and a task to wait for cancellation

python/phase-1.md
cretz: Changed to payload converter

python/phase-1.md
```diff
@@ -0,0 +1,380 @@
...
+
+```python
+from temporalio import activity
+
+# This adds activity details if they are present. Maybe we should have a
+# temporalio.LoggerAdapter that works for activities _and_ workflows?
+my_activity_logger = activity.LoggerAdapter(logging.getLogger(__name__))
+
+
+def simple_activity(arg1, arg2):
```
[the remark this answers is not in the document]
cretz: Added executor option and we can decide once we get into implementation if we want to have a default one and/or whether we want to also use the executor for async def's too.

python/phase-1.md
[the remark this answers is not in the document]
cretz: Made heartbeat a sync func. Still need discussion on whether it should raise if the activity is cancelled (obviously it'd be the _next_ heartbeat after the one that reported a cancellation async via core). I just checked Go and it doesn't seem to ever error.

python/phase-1.md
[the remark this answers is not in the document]
cretz: Changed.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: For the `register_*` functions, there are other ways to merge multiple activity sources without them.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: yes

python/phase-1.md
cretz: Fixed

python/phase-1.md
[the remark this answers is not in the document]
bergundy: I guess this makes sense for Python where logging is in the standard library.
We should allow interception of this method so users can control what contextual information to add.

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: Yeah. I really don't think it's meaningfully simpler. It's marginally less _verbose_, absolutely, but I just can't see someone getting frustrated because they had to provide a name. Alternatively like in TS / mentioned here: https://github.com/temporalio/proposals/pull/49#discussion_r792249398 if we require the user to provide an explicit mapping I think that's good too.

But, yes, if I had the chance I would undo what Go and Java do where the name is implicitly derived because I've seen it confuse multiple users.

python/phase-1.md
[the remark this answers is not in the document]
cretz: I added the ability to provide them at creation. Since we are creating core separately from worker, we don't need a static async `create`, class init is fine. The `run` here is like the one in TS.

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: K, then I think at construction is a little easier maybe than separate method but... it's really a super slim difference

python/phase-1.md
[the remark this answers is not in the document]
cretz: They can control what contextual information to add by using their own logger of course and that is suggested.

Having said that, I do agree that an "activity outbound interceptor" will be available for all of these activity utilities.

python/phase-1.md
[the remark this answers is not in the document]
Sushisource: `handler` ? `.code`?

I dunno, `defn` might be best. For a workflow, sure, it's the "entry" and not the "entire workflow", but I don't really feel that's an important distinction here. The intent is pretty clear there.

python/phase-1.md
Sushisource: Seems unpopular so nevermind. For activities it's not a very big deal. I feel more strongly about it for workflows, though. Attaching hidden state to workflows makes it easy for different workers to end up doing different things and violate nondeterminism. So ideally there we try to make them as "pure-function-y" as possible

python/phase-1.md
```diff
@@ -156,24 +181,71 @@ Notes:
...
+    we can pass a client to worker instead of core which would be much cleaner
+* While we could implicitly create this, say, per address I think each core instance has enough uniqueness to require
+  manual user management
+* We could flip the wording of this a bit and have a "worker" not be task queue specific and be 1:1 with core, then have
+  the concept of a "task queue worker" that then lets you register activities and such.
+  * Probably deviates too much from the norm. Still, it's quite unfortunate to have a two-step process to create a
+    worker.
+* Is there any harm in creating one core instance per worker? We could do away with this if not.
+  * Do we lose some of the shared core benefits by only registering one worker for each core instance?
+  * I assume there is a meaningful difference between 1000 core instances that we poll on each and just 1?
```
Sushisource, line 219: There should only be one per namespace. There would be substantially more overhead involved in making many many more Core instances.

python/phase-1.md
```diff
@@ -156,24 +181,71 @@ Notes:
...
+* Any other ideas on removing Core from explicit creation by users while still allowing multiple workers on different
+  server addresses?
+  * We can't do something like one-core-per-server-addr because there are different core options than just addr
+  * Should we just make the `core` optional in worker and have a global one that they can init (and hope they do so
+    before starting worker)?
```
Sushisource, line 224: It its not unreasonable to have a behind-the-scenes mapping of `<addr/namesspace -> core instance>`

Which users could explicitly override by simply `install`ing one explicitly before creating a worker a-la TS.

Sushisource: All that said part of me still kinda thinks you ought to just run multiple processes if you want multiple instantces

python/phase-1.md
```diff
@@ -156,24 +181,71 @@ Notes:
...
 ```python
+
 class Worker:
     def __init__(
         self,
         client: temporalio.Client,
         *,
         # Uses default in client if not present (or errors if no default)
-        task_queue: str = None
+        task_queue: str = None,
```
Sushisource, line 239: Don't forget this guy needs to be required for workers

python/phase-1.md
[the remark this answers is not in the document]
bergundy: If Core is a parameter we don't need an async create method. We will need one to support a singleton Core.

python/phase-1.md
```diff
@@ -316,16 +442,18 @@ Notes:
   * When we get to workflows, we will do similar with `@workflow.run` vs `@workflow.signal` vs `@workflow.query`, etc
     and `workflow.info()`
 * Cancellation is task cancellation
-  * Do we need to differentiate worker shutdown from server-side cancellation?
-  * So we need an explicit future they can wait on for activity cancellation?
+  * We need to research whether, in Python 3.7, we can interrupt an async task w/ our own error reliably
+* Heartbeat is a synchronous call
+  * Should we throw a cancellation error if we know the activity is cancelled when they try to heartbeat?
+    * is this something core should be returning from `record_heartbeat` instead of lang?
```
Sushisource, line 448: Yes and maybe also yes. We didn't return errors from heartbeating in core for reasons that I never thought were super great so we can revisit that decision for sure.

That said, for now, it's simple to throw if you already know you're cancelled

python/phase-1.md
Sushisource: The thing I'm still curious about here is if we can attach discoverable information to `asyncio.CancelledError`? It doesn't seem like we really can. The added ability to get the cancel info helps for sure but if we could it'd be nice to have it on the exception itself somehow

python/phase-1.md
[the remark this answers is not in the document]
bergundy: One thing we can do with the decorator is set the default activity options.
These options can be used in workflows running in the same worker that the activities are registered in.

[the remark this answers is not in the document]
bergundy: Another options for activity class registration is to expose all public async functions by default.

python/phase-1.md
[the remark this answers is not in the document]
bergundy: The default (implicit) executor approach is not ideal because activity authors might not be aware that they need to write thread safe code.

python/phase-1.md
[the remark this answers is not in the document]
cretz: I think TS kinda has to because getting the function name in JS at runtime is not always clear/easy w/ how often arrow functions are used and refs passed around. However in Python it seems pretty common to use function name for this stuff.

I have seen it confuse users too, but am not convinced that it deserves to be disallowed because of those users. If we knew how many users liked the inference, I think we could decide better.

python/phase-1.md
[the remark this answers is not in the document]
cretz: I can get behind `@activity.defn`. One of my secret ideas was to have a concept of a mutually exclusive `@workflow.run` and `@workflow.start` and if the latter was used, it'd just assume an infinite loop and be event based, heh. I think I can still get there with `@workflow.defn` though.

python/phase-1.md
cretz: The answer to that will require experimentation. Task factories and event loop inheritance has been mentioned as possible options.

cretz:
I am closing this draft in favor of another one I am about to open incorporating all feedback from this PR and during off-PR discussions

PR #50, opened 2022-01-26
Python SDK - Phase 1

cretz:
# Summary

This is an initial proposal to start development on the Python SDK.

[See it rendered here](https://github.com/cretz/temporal-proposals/blob/python-phase1/python/phase-1.md) (may be deleted when proposal is done). Also see [the previous draft PR here](https://github.com/temporalio/proposals/pull/49) where several discussions were had.

Feel free to leave a bunch of comments. If discussion becomes too difficult to follow, I will close this PR and start a new one (like I already have once :-)).

### DIFF
--- a/python/phase-1.md
+++ b/python/phase-1.md
@@ -16,7 +16,7 @@ The Python SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-co
 The high-level goals for the Python SDK are:
 
 * Be as simple as possible from a caller POV
-* Be as lightweight as possible wrt dependencies
+* Be as lightweight as reasonable with regards to dependencies
 * Be reasonably flexible with reasonable defaults
 * Support only Python >= 3.7
 * Be friendly to REPL/interpreter use cases
@@ -36,10 +36,10 @@ An example of a client (intentionally missing _a lot_ of features):
 ```python
 import temporalio
 
-def main():
-  client = temporalio.Client("localhost:7233", namespace="MyNamespace")
-  result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
-  print("Result", result)
+async def main():
+    client = await temporalio.Client.create("localhost:7233", namespace="MyNamespace")
+    result = await client.execute_workflow("MyWorkflow", "arg1", id="someid", task_queue="somequeue")
+    print("Result", result)
 ```
 
 An example of an activity and registration (also missing a lot of features):
@@ -50,10 +50,11 @@ import temporalio
 def say_hello(name: str) -> str:
   return f"Hello, {name}!"
 
-def main(client: temporalio.Client):
-  worker = temporalio.Worker(client, task_queue="somequeue")
-  worker.register_activity(say_hello)
-  await worker.run()
+async def main():
+    core = await temporalio.Core.create("localhost:7233", namespace="MyNamespace")
+    worker = temporalio.Worker(core, task_queue="somequeue")
+    worker.register_activity(say_hello)
+    await worker.run()
 ```
 
 ## API Definitions
@@ -75,16 +76,27 @@ exported from `temporalio`):
 
 ```python
 class Client:
-    def __init__(
-        self,
+    @staticmethod
+    async def create(
         addr: str,
         *,
         namespace: str = "default",
         default_task_queue: str = None,
         identity: str = None
         # ... additional options omitted for brevity
-    ) -> None:
+    ) -> Client:
         # Actually dials gRPC here
+        pass
+
+    def __init__(
+        self,
+        service: workflowservice.WorkflowService,
+        *,
+        namespace: str = "default",
+        default_task_queue: str = None,
+        identity: str = None
+        # ... additional options omitted for brevity
+    ) -> None:
         pass
 
     # Raw gRPC service
@@ -92,7 +104,8 @@ class Client:
     def service(self) -> workflowservice.WorkflowService:
         pass
 
-    def start_workflow(
+    async def start_workflow(
+        self,
         workflow: str,
         *args,
         id: str = None,
@@ -102,15 +115,27 @@ class Client:
         start_signal: str = None,
         start_signal_args=[],
         # ... additional options omitted for brevity
-    ) -> WorkflowHandle[Any]:
+    ) -> WorkflowHandle[Any]:  # This result defines __await__ among other things
         pass
 
     # In a future phase:
     # @overload
-    # def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowExecution[T]
+    # async def start_workflow(workflow: Callable[P, T], *args: P.args, ...) -> WorkflowHandle[T]
+    #
+    # @overload
+    # async def start_workflow(workflow: T, *args) -> T
+
+    # Just a wrapper for start workflow that awaits the handle
+    async def execute_workflow(
+        self,
+        workflow: str,
+        *args,
+        # ... additional options omitted for brevity
+    ) -> Any:
+        return await (await self.start_workflow(workflow, *args))
 
     # Does not validate whether a workflow exists
-    def workflow(
+    def get_workflow_handle(
         workflow_id: str,
         run_id: str = None
         # ...
@@ -156,10 +181,47 @@ Notes:
   * We could make an abstract base class if we want for typing/purposes that just has the `service` property
     unimplemented
 * We call it `start_workflow` instead of just `start` to be clearer
+* We have an `execute_workflow` to combine start + result waiting
 * We have embedded `start_signal` and `start_signal_args` as parameters, so if the former is present we will perform a
   signal with start. Is this ok?
   * This is an intentional deviation from other SDKs, but makes logical sense
 * The client supports a default task queue, do we want this?
+
+### Core
+
+Exported from `temporalio` (but probably in `temporalio/core/core.py`, exported from `temporalio.core` then
+exported from `temporalio`):
+
+```python
+class Core:
+    async def create(
+        addr: str,
+        *,
+        namespace: str = "default"
+        # ... All the same options as Client.create
+    ) -> Core:
+        pass
+```
+
+Notes:
+
+* If we could use Core to back our API client, then this would not be needed
+  * Even if we didn't, maybe if the worker feature is enabled we can create a core instance during client creation? Then
+    we can pass a client to worker instead of core which would be much cleaner
+* While we could implicitly create this, say, per address I think each core instance has enough uniqueness to require
+  manual user management
+* We could flip the wording of this a bit and have a "worker" not be task queue specific and be 1:1 with core, then have
+  the concept of a "task queue worker" that then lets you register activities and such.
+  * Probably deviates too much from the norm. Still, it's quite unfortunate to have a two-step process to create a
+    worker.
+* Is there any harm in creating one core instance per worker? We could do away with this if not.
+  * Do we lose some of the shared core benefits by only registering one worker for each core instance?
+  * I assume there is a meaningful difference between 1000 core instances that we poll on each and just 1?
+* Any other ideas on removing Core from explicit creation by users while still allowing multiple workers on different
+  server addresses?
+  * We can't do something like one-core-per-server-addr because there are different core options than just addr
+  * Should we just make the `core` optional in worker and have a global one that they can init (and hope they do so
+    before starting worker)?
 
 ### Worker
 
@@ -167,13 +229,23 @@ Exported from `temporalio` (but probably in `temporalio/worker/worker.py`, expor
 exported from `temporalio`):
 
 ```python
+
 class Worker:
     def __init__(
         self,
         client: temporalio.Client,
         *,
         # Uses default in client if not present (or errors if no default)
-        task_queue: str = None
+        task_queue: str = None,
+        # Can be anything accepted by register_activities
+        activities=None,
+        # An executor to be used for synchronous activities. If set, the worker
+        # does not shut this down. If unset, the worker may use a thread pool
+        # internally.
+        activity_executor=None,
+        # Whether to use the given/defaulted executor for async activities as
+        # well.
+        use_activity_executor_on_async: bool = False,
         # ...
     ) -> None:
         pass
@@ -186,8 +258,8 @@ class Worker:
         *,
         # Set the name override. If not set, uses the @activity.run decorator
         # name. If that's not set, uses the function/method name or class name
-        # if it has __call__. If it's a lambda or nested function, a name must
-        # be provided (or it can be decorated of course).
+        # if it has __call__. If it's a lambda, a name must be provided (or it
+        # can be manually decorated).
         name: str = None
         # ...
     ) -> None:
@@ -216,9 +288,61 @@ Notes:
 * Stopping the worker is done via cancellation of the run task/future
 * Internally, activities will run in the same event loop `run()` is called on
 * By default will fail already-registered activities
-* Should we accept a set of activities to register in `__init__`?
+* We also accept a set of activities to register in `__init__` and caller can choose preferred approach
 * Does exporting `temporalio.Worker` directly cause any packaging issues if we want to separate the client as its own
   published package?
+  * I believe we can catch an import error
+* We allow a custom executor and potentially even default one for sync activities
+  * Should we default an executor for sync?
+  * Is it ok to not default an executor for async? Or should we use it for them too?
+
+### Activity Utilities
+
+These are all top-level calls in `temporalio.activity`:
+
+```python
+# Whether the current code is in an activity
+def in_activity() -> bool:
+    pass
+
+
+def info() -> Info:
+    pass
+
+
+class Info:
+    # Other fields/properties skipped for brevity
+
+    @property
+    def cancelled(self) -> bool:
+        pass
+
+    @property
+    def cancellation() -> asyncio.Task:
+        pass
+
+
+# Adapter that extracts info() only when available in context
+class LoggerAdapter(logging.LoggerAdapter):
+    pass
+
+
+_logger = LoggerAdapter(logging.getLogger(__name__))
+
+# Logger that activities can use w/ activity context info instead of making
+# their own (though making your own is recommended)
+def logger() -> logging.Logger:
+    return _logger
+
+
+def heartbeat(*details) -> None:
+    pass
+```
+
+Notes:
+
+* These all uses `contextvars` to get activity info or throw otherwise
+  * We will make the contextual state open to caller injection to support test mocking. The details are undecided atm.
 
 ### Activity Definition/Registration
 
@@ -251,7 +375,7 @@ async def another_activity(arg1):
     try:
         while True:
             await asyncio.sleep(1)
-            await activity.heartbeat("some details")
+            activity.heartbeat("some details")
     except asyncio.CancelledError as exc:
         raise RuntimeError("someone cancelled me!") from exc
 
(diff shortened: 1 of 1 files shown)


python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+  * Probably
+
+### Client
+
+Exported from `temporalio` (but probably in `temporalio/client/client.py`, exported from `temporalio.client` then
+exported from `temporalio`):
+
+```python
+class Client:
+    def __init__(core_or_service: temporalio.Core) -> None:
```
cretz, line 120:
```suggestion
    def __init__(core: temporalio.Core) -> None:
```

(leftover from an old version that accepted a union of core or pure Python gRPC which we don't want atm)

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+        # internally.
+        activity_executor: Optional[concurrent.futures.Executor]=None,
+        # Whether to use the given/defaulted executor for async activities as
+        # well.
+        use_activity_executor_on_async: bool = False,
+        # ...
+    ) -> None:
+        pass
+
+    async def run() -> None:
```
cretz, line 244: Missing `self`

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+
+### Client
+
+Exported from `temporalio` (but probably in `temporalio/client/client.py`, exported from `temporalio.client` then
+exported from `temporalio`):
+
+```python
+class Client:
+    def __init__(core_or_service: temporalio.Core) -> None:
+        pass
```
Sushisource, line 121: I don't think we want to init clients this way - I think `Core` should have a `get_client` method, and I think a client constructor should accept some client options object (which Core will also accept when created) which it will use to create a core-independent client via the publicly exposed interface here: https://github.com/temporalio/sdk-core/blob/e3f54276f971bcc9b6e1d1cf4d0ecbeb853387ae/client/src/lib.rs#L210

IE: You should be able to have a client tied to core if you want, but you should also be able to get a client without any core at all.

Obviously the rub is telemetry like we talked about a little bit. We can clean up how that works in core to make it easy, but we can start without any changes by simply having clients-not-from-core not do any telemetry.

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+            "activity1_method": MyActivities().activity1,
+            # Can use lambda
+            "lambda": lambda arg1: arg1,
+            # Fails because not callable
+            "not_callable": MyActivities(),
+        },
+    )
+```
+
+Notes:
```
cretz, line 375: Need to document our discussions around why we removed activity decorator and why we require a mapping w/ explicit activity name

python/phase-1.md
cretz: I had originally had similar, but:

* I didn't want to make an implementation of this class with non-core because I couldn't figure out how to return `service` properly as I don't think the core and non-core will have the same signatures for the call. So I am not sure I can have a core-independent client for all gRPC calls unless I maintain all gRPC calls manually. But I haven't dug enough to confirm.
* We don't have `Core.create_worker`, so I figured it was more consistent to pass core to clients/workers than to either be inconsistent or force worker creation to be on core too.

I personally think it's cleaner this way (and any other potential uses of the same core instance could potentially take core in their constructor too). But I can definitely change this if consensus is to do so.

Not too worried about the telemetry stuff, we can figure that out a bit later (I have some ideas that basically moves `tokio::tracing` to lang-side and requires core `init` to accept a spawner unfortunately since I can't see how to do thread-local on tokio runtime threads, but that's for a different setting).

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+
+* After discussion, it has been decided that both clients and workers will be backed by core
+    * Yes that means clients require the Rust library which we accept
+* After discussion, it has been decided that we will not have an implicit local default like TypeScript SDK, but instead
+  an explicit default
+  * `default_local()` memoizes success
+* Should we have a constructor, an `__aenter__`, and an `__aexit__`?
+  * Probably
```
bergundy, line 111: yes please, that's very pythonic.

python/phase-1.md
bergundy: My main issue with tying Core to the Client is that they're 2 completely different entities and it's actually Core that takes a client and not the other way around on the rust side).

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+        client: temporalio.Client,
+        workflow_id: str,
+        *,
+        run_id: Optional[str] = None
+        # ...
+    ) -> None:
+        pass
+
+    def __await__(self) -> T:
+        pass
```
bergundy, line 179: Another issue with this approach is that `await client.get_workflow_handle()` now waits will wait for completion which could be confusing.

bergundy: Let's reconsider adding an explicit `result` method.

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+* We have left off workflow registration until a later phase
+* Stopping the worker is done via cancellation of the run task/future
+* The constructor will fail if the task queue has already been registered with that instance of core
+* Internally, activities will run in the same event loop `run()` is called on
+* We allow a custom executor and potentially even default one for sync activities
+  * Should we default an executor for sync?
+  * Is it ok to not default an executor for async? Or should we use it for them too?
+* Should we have an `__aenter__` and an `__aexit__`?
+  * Probably, and change the examples to use `with` to ensure worker is shut down properly
+  * This means we have to treat those two calls as separate start and shutdown
```
bergundy, line 259: Doesn't look like this would work here.
What would be in the body of the `with` statement? It'd have to block on something.

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+
+Notes:
+
+* We have left off workflow registration until a later phase
+* Stopping the worker is done via cancellation of the run task/future
+* The constructor will fail if the task queue has already been registered with that instance of core
+* Internally, activities will run in the same event loop `run()` is called on
+* We allow a custom executor and potentially even default one for sync activities
+  * Should we default an executor for sync?
+  * Is it ok to not default an executor for async? Or should we use it for them too?
```
bergundy, line 256: I vote against default executors, explicit is better than implicit here too IMHO.
(I've raised my concerns in the previous PR)

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+    )
+```
+
+Notes:
+
+* Cancellation is task cancellation
+  * We need to research whether, in Python 3.7, we can interrupt an async task w/ our own error reliably
+* Heartbeat is a synchronous call
+  * Should we throw a cancellation error if we know the activity is cancelled when they try to heartbeat?
+    * is this something core should be returning from `record_heartbeat` instead of lang?
```
bergundy, line 381: heartbeat should not be tied to cancellation, this separation allows local activities to be cancelled.

python/phase-1.md
```diff
@@ -0,0 +1,441 @@
...
+        pass
+
+    async def encode(self, *values):
+        pass
+
+    async def decode(self, *payloads):
+        pass
+
+
+PayloadConverter.default = CompositePayloadConverter(
```
bergundy, line 418: After thinking about this some more should Python support [pickle](https://docs.python.org/3/library/pickle.html) by default?
A lot of python devs would expect us to support this (for arbitrary object serialization).
One obvious drawback to this is that it's not a polyglot solution.

python/phase-1.md
Sushisource: I suppose it works (to use only a client in your python process) as long as you never register any workers. It's not really how I've thought of what Core is, but I guess it technically works. It's not super clean. You still have this core instance that you can "shut down" but that doesn't really mean anything, and it's got facilities for things you're never going to use if all you want is a client.

And "I couldn't figure out how to return service properly as I don't think the core and non-core will have the same signatures for the call. " is easily solved. It's not exposed right now in core but it's trivial to add, so I wouldn't let that stop you.

Regarding "Core.create_worker" -- why not? I mean I don't care a whole lot either way here but I think that makes sense too. If we want it to accept core that's fine still, and client can either accept core (`Client.from_core`), or can accept options for a core-independent client.

python/phase-1.md
Sushisource: I'm not really sure why you'd need it - unless maybe you're passing already-pickled objects from some datastore directly into activities/wfs (which I can believe).

Otherwise, if you're using our client or package, you're going to get a reasonable form of serialization for free already (with one of the defaults), so why do you care if it's pickle?

At minimum, seems like something we can add later if wanted, doesn't need to be in the proposal.

python/phase-1.md
cretz: > My main issue with tying Core to the Client is that they're 2 completely different entities and it's actually Core that takes a client and not the other way around on the rust side).

Can you clarify this? Are you talking about the off-PR discussion of whether to tie the client to core? I had understood that as settled.

> And "I couldn't figure out how to return service properly as I don't think the core and non-core will have the same signatures for the call. " is easily solved. It's not exposed right now in the rust client crate but it's trivial to add, so I wouldn't let that stop you.

I am not sure it's easily solved. If the Python gRPC client, say, takes contextual call options I can't expect to abstract those properly.

> Regarding "Core.create_worker" -- why not? I mean I don't care a whole lot either way here but I think that makes sense too. If we want it to accept core that's fine still, and client can either accept core (Client.from_core), or can accept options for a core-independent client.

I think it's cleaner to create instances of classes via Python's class creation mechanisms whenever reasonable.

Can I get clarity on what's being asked for here? So we want `get_client` and `create_worker` on core? Or we don't want tying Core to a client? Or we want an abstract client that works on pure Python gRPC and Core client including abstracting all workflowservice calls over both? I am looking for concrete complete suggestions (the piece here, even down to only have a Core version and to have the client and worker accept core in constructor non-defaulted was what was settled on during discussion, but maybe it was unclear and of course we can change).

python/phase-1.md
cretz: > Another issue with this approach is that await client.get_workflow_handle() now waits will wait for completion which could be confusing.

This is only unclear because I was forced to change the name. `await client.workflow("my workflow")` was very clear originally from the previous PR.

python/phase-1.md
cretz: Agreed, so we won't use a default executor for sync.

The question here becomes, and sorry it was worded unclearly, do we want async functions to use the same executor as sync functions? Or do we want a bool like I have here? Or do we want to allow a separate executor to be passed for async functions?

python/phase-1.md
cretz: Ok

python/phase-1.md
cretz: >  One obvious drawback to this is that it's not a polyglot solution.

The most obvious. But I'll provide a pickle-based converter they can choose to use if they want.

python/phase-1.md
cretz: I am marking this resolved as we are having off-PR discussions about how core, clients, and workers interact.

python/phase-1.md
bergundy: I don't think the name change is what's making it unclear.
It's that you get 2 completely different results with an without `await`.

cretz: That's normal I think for awaitable objects that carry other info. I will change to adding a result method though.

python/phase-1.md
bergundy: I don't think we should mess with executors at all.
We should just run async functions (function that returns a Future) on the currently installed event loop in a single threaded fashion.
I'd start with async only activities and explain that if users want to run sync code they need to wrap it in an executor.

python/phase-1.md
bergundy: The problem with JSON is that by default python objects aren't serializable, only primitives are.
We can recommend using dataclasses which can be turned into dicts https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict

python/phase-1.md
cretz: Fixed.

python/phase-1.md
cretz: Fixed

python/phase-1.md
cretz: Documented

python/phase-1.md
cretz: Done

python/phase-1.md
cretz: Added `result()`

python/phase-1.md
cretz: Added `start` and `shutdown`

python/phase-1.md
cretz: I think that's a bit much. I think it's normal to only put `async` on a function if you need to. Also, there could be value in running async functions in executor too. But yes, I accept both as params that are defaulted to none atm.

python/phase-1.md
cretz: Updated

python/phase-1.md
cretz: > The problem with JSON is that by default python objects aren't serializable, only primitives are.

In all SDKs we accept the limitation that the in/out types must be serializable by the converter by default. What happens if I have a function on a TypeScript type? Technically they could subclass `json.JSONEncoder` and impl `default` for out types, but yeah it'd just come back as the raw form of that.

>  We can recommend using dataclasses for args and return types which can be turned into dicts https://docs.python.org/3/library/dataclasses.html#dataclasses.asdict

We can do this for "encode" of the `JSONPlainPayloadConverter`, but decoded values will remain dicts. So our suggestion should really be "use dicts directly or dataclass knowing it will come in as a dict that you'll have to create your class with if needed". But I think this limitation is fine compared to, say, trying to store the class name in the payload by default (but we can definitely support that if needed, same as we support pickle but not by default).

roselander:
In my previous life (not very long ago) I maintained a library that used Python's asyncio, "await" and I thought it was extremely half-baked.

Crucially, the only methods which truly implement async behavior are low-level I/O primitives such as opening a TCP socket. In other words, unless you eventually delegate to one of these "magic" methods, your "async" code will never release the Global Interpreter Lock, which kind of defeats the purpose IMO.

So basically my question is whether someone who understands Python asyncio at a deep level has thought about how this would actually be implemented. I'm kind of assuming that's the case but just wanted to double-check.

bergundy:
@roselander I actually think asyncio is pretty well designed, Guido was the one who lead the effort and he based it on popular async libraries like twisted and tornado.

Coroutines have nothing to do with the GIL, they help you write concurrent code in a single threaded environment.
If you want to use multiple cores in python your only option is to use multiple processes or use C extensions which can release the GIL.
Our Python SDK, is going to use asyncio and be an async first SDK for a couple of reasons:
- You can use synchronous code in an async environment and not the other way around (e.g. the async keyword is contagious).
- For workflow logic we want to be able to reuse known concurrency primitives such as [`gather`](https://docs.python.org/3/library/asyncio-task.html#asyncio.gather) and [`as_completed`](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed)

python/phase-1.md
```diff
@@ -0,0 +1,460 @@
...
+
+* For demonstration/simplicity here we are leaving off a lot of proper typings (e.g. `Optional`) that the real version
+  will have
+* We leave off a lot of options on calls for this high-level proposal
+* We are not using proper function documentation, but rather inline comments for easier discussion
+* **NOTE: These will have proper docs and types when developed!**
+
+### Client
+
+Exported from `temporalio` (but probably in `temporalio/client/client.py`, exported from `temporalio.client` then
```
cretz, line 80: I am thinking this may not be clearest exported from `temporalio` directly, but instead leave only in the `client` module. So you would have to `from temporalio import client` and use `client.Client`. This more aligns with our expectations of using `from temporalio import activity` for e.g. `activity.info()`. Thought I do think all common code such as retries and errors may deserve to be in the `temporalio` package directly.

Thoughts?

python/phase-1.md
```diff
@@ -0,0 +1,460 @@
...
+        self, *values: Any
+    ) -> Union[temporalio.Payload, list[temporalio.Payload]]:
+        pass
+
+    # This can be sync or async. Always takes a list. Can return a single value
+    # or a list of values with the same size as the input list or None if the
+    # value cannot be converted with this converter. Any other result causes an
+    # error. Any error raised is bubbled.
+    @abstractmethod
+    async def decode(self, *payloads: temporalio.Payload) -> Union[Any, list[Any]]:
```
cretz, line 422: There is an (obvious in hindsight) ambiguity issue here. What if your single payload is a list?

Any suggestions on how to combine "decode single payload -> multiple values" and "decode multiple payloads -> multiple values" and "decode multiple payloads -> single value"? Do we have to support the last use case?

I am now thinking about adding extra `encode_multiple` and `decode_multiple` calls that default to just calling these and if someone, say for encryption purposes, wants to convert multiple to single, they have to override those two also. Thoughts?

cretz: Due to additional ambiguities, I'm gonna have to return a `Tuple[Any, bool]` from `decode(self, payload)` with the second value being `True` if it was valid. There is ambiguity if I just return `-> Any` because what if `None` is a real value?

Another approach would be to separate out into data and payload converters again, but that seems unnecesary.

cretz:
@roselander - While this API will be available because it is autogenerated from our API, we're actually going to be delegating to a native extension in Rust to make gRPC invocations since that extension is already handling advanced server communication and event handling for the worker. This is a deviation from TypeScript which has an in-lang gRPC client separate from the Rust client. After discussion of the tradeoffs of separate connections vs requiring a C extension (in Rust), we settled on the latter.

roselander:
@cretz 

> Can you clarify why this is a problem? Our extension will end up calling asyncio functions as well.

There was no asyncio gRPC at the time, and so in practice it wasn't actually possible for us to implement our async library, because there was no existing async primitive to build our library on top of.

python/phase-1.md
```diff
@@ -0,0 +1,460 @@
...
+
+    async def shutdown(self) -> None:
+        pass
+
+    async def __aenter__(self) -> Worker:
+        await self.start()
+        return self
+
+    async def __aexit__(self) -> None:
+        await self.shutdown()
```
bergundy, line 252: I don't think this is required, what is this supposed to block on?

```
async with worker:
  # what goes here?
```

python/phase-1.md
```diff
@@ -0,0 +1,460 @@
...
+
+```python
+class PayloadConverter(ABC):
+    # This can be sync or async. Always takes a list. Can return a single
+    # payload or a list of payloads with the same size as the input list or None
+    # if the value cannot be converted with this converter. Any other result
+    # causes an error. Any error raised is bubbled.
+    @abstractmethod
+    async def encode(
+        self, *values: Any
```
bergundy, line 413: After thinking about this a bit, we might want to use reflection here to get the expected type from the activity signature like we do in Java / Go

python/phase-1.md
bergundy: I thought these methods were meant for multiple payloads, it's okay to have a payload that's an array.
The method can map M to N payloads.

bergundy (approved):
There are still 1-2 unresolved items but I'm already approving

python/phase-1.md
cretz:
```
async with worker:
  # everything you want to do with a worker
```

If you want to use it that way. The point is that it stops when done there. If you don't have your code encapsulated like that, you can just run the worker, you don't have to use `async with`.

python/phase-1.md
cretz: I agree. I can mutate this proposal to include that (of course not require type annotations). I will have to do some research on runtime annotation resolution.

cretz:
I am merging this as most of our discussions are now to implementation specifics which we can do on PRs that are already rolling in to https://github.com/temporalio/sdk-python. The exact APIs of this phase, as predicted, are changing during implementation.

PR #52, opened 2022-03-04
Worker versioning

Sushisource:
Proposal [rendered](https://github.com/Sushisource/proposals/blob/worker-versions/versioning/worker-versions.md)

### DIFF
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -105,3 +105,33 @@ workerOptions := worker.Options{
 ```java
 WorkerOptions.newBuilder().setWorkerVersion("1.0").build()
 ```
+
+## When and how to increment the worker version number
+
+The information here will be added as a documentation page once this feature
+has shipped.
+
+--
+
+The worker version number encompasses changes to all workflow code registered
+with the worker, as well as changes to any code which might affect the path
+of execution within workflow code.
+
+#### Super-Major
+Require a new workflow type name (or new task queue name)
+
+- A backwards-incompatible change to the workflow interface
+    - Changing a required parameter type
+    - Adding or removing a required parameter
+
+
+#### Major
+
+- A backwards-incompatible change to the workflow implementation
+    - Anything that would cause the workflow to fail replay due to nondeterminism, given any possible valid history from the unaltered version of the workflow. 
+    - Includes changes to data converters or interceptors which would result in such incompatibilities.
+
+#### Minor
+
+- Backwards compatible change to workflow implementation
+    - This includes adding properly implemented `getVersion` or `patch` calls, for fixing bugs.
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -49,6 +49,8 @@ At a high level:
   to the latest version of the `foo` queue.
 * Provide good notifications and information about when old workers may be
   decommissioned
+* Ensure histories replayed on older versions fail fast with informative errors
+* Easy visibility into what versions deployed workers are using
 
 Building on this base set of features, we plan to eventually implement
 automatic safe-rollout of changes, and possibly tools like k8s operators
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -1,0 +1,97 @@
+# User-specified Worker versions
+
+Temporal workflow authors experience a number of difficulties when it comes to
+changing workflow code. The current techniques for versioning workflows are:
+
+**GetVersion & patch APIs**
+These APIs allow branching within the same workflow code to handle the presence
+or absence of a certain change. See them for [TS](https://docs.temporal.io/docs/typescript/patching#typescript-sdk-patching-api), [Java](https://docs.temporal.io/docs/java/versioning/), and [Go](https://docs.temporal.io/docs/go/versioning/).
+
+These APIs are confusing and challenging to use correctly, and clutter up code.
+We advise users to generally use them only to fix bugs that would otherwise prevent
+the workflow from completing.
+
+**Versioned task queues**
+Conceptually simpler, but operationally more effort, is to version task queues.
+IE: A worker operating on task queue `foo-v1` is changed, and when redeployed
+points at `foo-v2`.
+
+This avoids complicating the workflow code itself, but means users need to be
+aware of when all workflows on `foo-v1` are complete, allowing them to decomission
+the old workers. It also means potentially deploying larger numbers of workers.
+
+It also requires updating clients who are starting workflows to point at the
+newer task queue name.
+
+**Versioned workflow names**
+This approach is much like versioning task queue names, but trades off some of
+the operational complexity for some code maintenance complexity.
+
+Rather than needing to keep old workers around, users now need to keep the
+old code around until no more workflows using the old type are open.
+
+They still have the same problem with needing to tell clients to use the new
+workflow name.
+
+## Improving the situation
+
+The `patch` type APIs can't be elimiated completely. There will always be 
+situations where a "broken" workflow needs to be fixed to make progress. Such
+fixes are typically less common than just making changes to workflow logic, though.
+
+To make the more common case easier, we will provide a set of features that
+make the "versioned task queues" approach much easier to use.
+
+At a high level:
+* Provide a simple API for versioning entire workers (this proposal)
+* Allow clients to point at task queue `foo`, and automatically route new workflow executions
+  to the latest version of the `foo` queue.
+* Provide good notifications and information about when old workers may be
+  decomissioned
+
+Building on this base set of features, we plan to eventually implement
+automatic safe-rollout of changes, and possibly tools like k8s operators
+which can automatically manage worker deployments.
+
+## SDK API changes
+
+The meat of this proposal is simple:
+**Add a version field to worker options in each SDK**
+
+Because version information will need to be sent to the Temporal Cluster,
+we define it as a protobuf message (which will show up in our API somewhere):
+```protobuf
+message WorkerVersions {
+    string worker_version = 1;
+}
+```
+
+This message is pre-emptively named pluarally in order to accomodate potential
+increases in versioning granularity in the future. For example, we have considered
+versioning Workflows, Activities, and Data Converters/Interceptors all separately.
+They could be added as new fields to this message if/when that happens.
+
+Each language will expose a way to set the worker version to start, supporting
+the possibility other versions later. This will translate into the above message
+to be sent to server.
+
+### TypeScript
+
+```typescript
+Worker.create({ workerVersion: '1.0', ...otherOptions })
+```
+
+### Go
+
+```go
+workerOptions := worker.Options{
+  WorkerVersion: "1.0",
+  // other options
+}
+```
+
+### Java
+
+```java
+WorkerOptions.newBuilder().setWorkerVersion("1.0").build()
+```
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -127,7 +127,7 @@ Require a new workflow type name (or new task queue name)
 
 #### Major
 
-- A backwards-incompatible change to the workflow implementation
+- A backwards-incompatible change to the workflow implementation. See more [here](https://docs.temporal.io/docs/concepts/what-is-a-workflow-definition/#deterministic-constraints)
     - Anything that would cause the workflow to fail replay due to nondeterminism, given any possible valid history from the unaltered version of the workflow. 
     - Includes changes to data converters or interceptors which would result in such incompatibilities.
 
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -63,7 +63,16 @@ Because version information will need to be sent to the Temporal Cluster,
 we define it as a protobuf message (which will show up in our API somewhere):
 ```protobuf
 message WorkerVersions {
-    string worker_version = 1;
+    UserVersion worker_version = 1;
+}
+
+message UserVersion {
+    // The string the user passed as their version representation
+    string original = 1;
+    // Parsed from original string
+    uint32 major = 2;
+    // Parsed from original string
+    uint32 minor = 3;
 }
 ```
 
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -1,4 +1,4 @@
-# User-specified Worker versions
+# User-specified Worker Versions
 
 Temporal workflow authors experience a number of difficulties when it comes to
 changing workflow code. The current techniques for versioning workflows are:
@@ -17,7 +17,7 @@ IE: A worker operating on task queue `foo-v1` is changed, and when redeployed
 points at `foo-v2`.
 
 This avoids complicating the workflow code itself, but means users need to be
-aware of when all workflows on `foo-v1` are complete, allowing them to decomission
+aware of when all workflows on `foo-v1` are complete, allowing them to decommission
 the old workers. It also means potentially deploying larger numbers of workers.
 
 It also requires updating clients who are starting workflows to point at the
@@ -35,7 +35,7 @@ workflow name.
 
 ## Improving the situation
 
-The `patch` type APIs can't be elimiated completely. There will always be 
+The `patch` type APIs can't be eliminated completely. There will always be 
 situations where a "broken" workflow needs to be fixed to make progress. Such
 fixes are typically less common than just making changes to workflow logic, though.
 
@@ -47,7 +47,7 @@ At a high level:
 * Allow clients to point at task queue `foo`, and automatically route new workflow executions
   to the latest version of the `foo` queue.
 * Provide good notifications and information about when old workers may be
-  decomissioned
+  decommissioned
 
 Building on this base set of features, we plan to eventually implement
 automatic safe-rollout of changes, and possibly tools like k8s operators
@@ -66,7 +66,7 @@ message WorkerVersions {
 }
 ```
 
-This message is pre-emptively named pluarally in order to accomodate potential
+This message is pre-emptively named plurally in order to accommodate potential
 increases in versioning granularity in the future. For example, we have considered
 versioning Workflows, Activities, and Data Converters/Interceptors all separately.
 They could be added as new fields to this message if/when that happens.
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -40,7 +40,8 @@ situations where a "broken" workflow needs to be fixed to make progress. Such
 fixes are typically less common than just making changes to workflow logic, though.
 
 To make the more common case easier, we will provide a set of features that
-make the "versioned task queues" approach much easier to use.
+make the "versioned task queues" approach much easier to use for changes to
+workflows which break history compatability (but not interface compatability).
 
 At a high level:
 * Provide a simple API for versioning entire workers (this proposal)
@@ -71,9 +72,9 @@ increases in versioning granularity in the future. For example, we have consider
 versioning Workflows, Activities, and Data Converters/Interceptors all separately.
 They could be added as new fields to this message if/when that happens.
 
-Each language will expose a way to set the worker version to start, supporting
-the possibility other versions later. This will translate into the above message
-to be sent to server.
+Each language will expose a way to (optionally) set the worker version to 
+start, supporting the possibility of other versions later. This will translate 
+into the above message to be sent to server.
 
 ### TypeScript
 


versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+## SDK API changes
+
+The meat of this proposal is simple:
+**Add a version field to worker options in each SDK**
+
+Because version information will need to be sent to the Temporal Cluster,
+we define it as a protobuf message (which will show up in our API somewhere):
+```protobuf
+message WorkerVersions {
+    string worker_version = 1;
```
dnr, line 65: I'm kind of allergic to stuffing structured data into strings.. if we're actually going to enforce semver-like semantics on these, maybe this should be two or three ints instead?

Sushisource: Great idea, will incorporate.

lorensr (approved):

versioning/worker-versions.md
cretz: Maybe the same for the SDKs. Meh, we may not need to expose a typed version per se since it may discourage setting it, but we should at least eagerly validate it at runtime (which of course we have to if putting the individual pieces into proto fields).

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+
+Rather than needing to keep old workers around, users now need to keep the
+old code around until no more workflows using the old type are open.
+
+They still have the same problem with needing to tell clients to use the new
+workflow name.
+
+## Improving the situation
+
+The `patch` type APIs can't be elimiated completely. There will always be
```
cretz, line 38: eliminated*

versioning/worker-versions.md
cretz: May want to have the SDKs have a typed version too, though not required since they'll at least have to parse to fill this proto.

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+
+## SDK API changes
+
+The meat of this proposal is simple:
+**Add a version field to worker options in each SDK**
+
+Because version information will need to be sent to the Temporal Cluster,
+we define it as a protobuf message (which will show up in our API somewhere):
+```protobuf
+message WorkerVersions {
```
cretz, line 64: WorkerVersion*?

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+```
+
+This message is pre-emptively named pluarally in order to accomodate potential
+increases in versioning granularity in the future. For example, we have considered
+versioning Workflows, Activities, and Data Converters/Interceptors all separately.
+They could be added as new fields to this message if/when that happens.
+
+Each language will expose a way to set the worker version to start, supporting
+the possibility other versions later. This will translate into the above message
+to be sent to server.
```
cretz, line 76: I appreciate the intentionally short proposal. Will another proposal be covering the implementation details or maybe it's better to impl in one SDK and the server to gauge challenges?

versioning/worker-versions.md
cretz, line 64: Will the SDK ever be asked to put more information in here? I believe we discussed we want simple versioning now that trusts the user, but one could see in the future having workflow and activity names/signatures in here even if just for visibility. Or maybe just language should be included here. Just future thoughts, no need to apply here.

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+
+```typescript
+Worker.create({ workerVersion: '1.0', ...otherOptions })
+```
+
+### Go
+
+```go
+workerOptions := worker.Options{
+  WorkerVersion: "1.0",
```
cretz, line 88: Implementation detail so maybe doesn't belong here, but I'd like if along with this we said - "In Go, if you use worker versioning, you must register all workflows and activities _before_ you start the workflow" (it's a restriction I wish we could apply always).

cretz (approved):

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+    string worker_version = 1;
+}
+```
+
+This message is pre-emptively named pluarally in order to accomodate potential
+increases in versioning granularity in the future. For example, we have considered
+versioning Workflows, Activities, and Data Converters/Interceptors all separately.
+They could be added as new fields to this message if/when that happens.
+
+Each language will expose a way to set the worker version to start, supporting
```
bergundy, line 74: It's worth mentioning that this is optional

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+The `patch` type APIs can't be elimiated completely. There will always be 
+situations where a "broken" workflow needs to be fixed to make progress. Such
+fixes are typically less common than just making changes to workflow logic, though.
+
+To make the more common case easier, we will provide a set of features that
+make the "versioned task queues" approach much easier to use.
+
+At a high level:
+* Provide a simple API for versioning entire workers (this proposal)
+* Allow clients to point at task queue `foo`, and automatically route new workflow executions
```
bergundy, line 47: We should mention that this doesn't take into account API changes.
Users should still use versioned task queue names for that purpose.

versioning/worker-versions.md
bergundy: We might add more info here but we've also considered implementing worker queries for that purpose.
Still TBD.
+1 to disable late registration.

versioning/worker-versions.md
bergundy: Semver is more than just 3 ints, I'm also for strong types and TBH I'm not convinced this should be semver at all since we don't use the patch version or any other info semver carries (like build info / rc).

versioning/worker-versions.md
```diff
@@ -0,0 +1,97 @@
...
+They could be added as new fields to this message if/when that happens.
+
+Each language will expose a way to set the worker version to start, supporting
+the possibility other versions later. This will translate into the above message
+to be sent to server.
+
+### TypeScript
+
+```typescript
+Worker.create({ workerVersion: '1.0', ...otherOptions })
```
bergundy, line 81: This comment applies to all langs.

The name `workerVersion` doesn't mean much to the user.
It's technically `workflowWorkerVersion` because it's only used for workflow routing.
We should also explain in this doc what changes affect this version number.

bergundy (approved):

versioning/worker-versions.md
lorensr: We've been talking about `foo.bar` version, where `foo` is a backward-incompat change to workflow interface or impl, and `bar` is back-compat change to workflow impl, including `patch` calls. Why do we need the latter? / what do we do with it?

> **What if Tommy wants to make a patch?**
> Same process, except he can use the `patch` API if needed, and increments the minor version number.

versioning/worker-versions.md
Sushisource: Is plural because we may well add more to it - also users will likely never really see this directly so it shouldn't be a problem in that regard.

versioning/worker-versions.md
Sushisource: Yeah I am planning on working through things with one lang and the server changes to get something e2e going. The implementation details SDK side are super trivial though. Really just parsing the version string and packing it into the proto message and sending it off during connection/polling.

versioning/worker-versions.md
Sushisource: Yes, more can certainly be added here. Mappings of workflow/act names to versions/other data certainly a possibility.

versioning/worker-versions.md
Sushisource: So late-registration doesn't really violate anything here - as the author of the workflow code it's up to you to rev this appropriately, and that applies whether you register up-front or later (unless you're dynamically loading code or something in which case this is all broken anyways).

That said I clearly have a preference for registering everything up front, and docs about this will say as much, but it's not strictly speaking incompatible with this kind of versioning.

versioning/worker-versions.md
cretz: Add more versions to it or add more stuff to the singular version?

versioning/worker-versions.md
cretz: > sending it off during connection/polling.

Hrmm. So at least for the initial version each worker should be able to handle all versions?

versioning/worker-versions.md
cretz, line 88: Can you give an idea how one might configure a worker to work with separate versions? Or is that a completely separate worker (and if so, they don't share a polling)? (any lang is fine)

versioning/worker-versions.md
Sushisource: It isn't though. It could be used for a change to an interceptor or data converter which breaks history compatibility.

I will add a section about what changes should affect the number, which will clear up why it's called that.

versioning/worker-versions.md
Sushisource: Fix bugs. 

EX: Your workflow code derefs a null pointer while parsing some data returned by an activity and your open WFs are stuck failing a wf task at that point.

To fix this, you might not _need_ to use `patch`, but say you wanted to remove the activity call entirely to fix this, so you use patch and in the branch w/ the patch you don't call the activity any more.

versioning/worker-versions.md
Sushisource: More versions to this message. EX: Maybe break out workflow/activity versions separately

versioning/worker-versions.md
Sushisource: > Semver is more than just 3 ints, I'm also for strong types and TBH I'm not convinced this should be semver at all since we don't use the patch version or any other info semver carries (like build info / rc).

(see my latest change)

True, and true, but the good news is users aren't sending this API directly. So, they can avoid caring about this, except I think we should ask the strings they write be semver-compatible (and we will then parse them). 99.99% of the time they can just write "4.2" and that'll work, but if they want to write "4.2.3-prerelease+build" then they can too, and we'll only use major/minor but we can show their whole string in UI/queries/whatever.

lorensr: I get that the design is the user increments the minor version to fix bugs, but what do we do with that information? Like with the major number, we change the task queue shadow version.

Sushisource: Same thing - new workflows (and new workflow tasks within the same major version) are started using the most recent version, including the patch number.

versioning/worker-versions.md
bergundy: What do you imagine this version being used for later on?

versioning/worker-versions.md
bergundy: I'm not sure we want to let users specify full semver.
Would they be surprised if they bumped an rc or a patch and we simply ignored that?

versioning/worker-versions.md
lorensr: Either aggregated

image


or just DCs/interceptors?

image

versioning/worker-versions.md
lorensr: > Same thing - new workflows (and new workflow tasks within the same major version) are started using the most recent version, including the patch number.

Is the only benefit to this that when I deploy `2.2` and I leave my `2.0/2.1` workers up, new `2.*` tasks & executions will no longer go to `2.0/2.1` workers? 

If that's the only benefit, then I'd much prefer the current process—when I deploy a back-compat change to code, I send shutdown signal to current workers. (With major versioning, I'd send shutdown to just the existing v2 workers.) Then I don't have to change a version number every time I update my code. And also the API gets simpler—instead of "foo.bar" where I have to explain A) it's not semver, B) when to use which number, I just have one integer `workerVersion: 2`. Or could name it more informative, like:

```
backwardIncompatibleVersion
breakingWorkflowHistoryVersion
breakingHistoryVersion
breakingVersion
```

versioning/worker-versions.md
```diff
@@ -0,0 +1,137 @@
...
+workerOptions := worker.Options{
+  WorkerVersion: "1.0",
+  // other options
+}
+```
+
+### Java
+
+```java
+WorkerOptions.newBuilder().setWorkerVersion("1.0").build()
```
mmcshane, line 106: Do we ever consider the possibility that the author of the worker startup code is distinct from the author of the workflow? I could see an infra team being in charge of the "workflow platform" and an appdev team implementing the workflows themselves. IIUC this proposal has a tacit assumption that the workflow author is always simultaneously in charge of worker startup - is that correct?

versioning/worker-versions.md
```diff
@@ -0,0 +1,137 @@
...
+## When and how to increment the worker version number
+
+The information here will be added as a documentation page once this feature
+has shipped.
+
+--
+
+The worker version number encompasses changes to all workflow code registered
+with the worker, as well as changes to any code which might affect the path
+of execution within workflow code.
```
mmcshane, line 118: Is the implication here that a workflow version update now _must_ entail an update to deployment topology? I.e. the mapping from workfows -> workers and the balancing/tuning of worker counts in a given deployment environment probably needs to change as a downstream effect of an app developer changing this version.

versioning/worker-versions.md
Sushisource: Yeah, that situation argues for the "version every workflow function individually" approach. Honestly in many ways I prefer that but we decided to go with this for it's simplicity and the fact that it looks consistent in every language / is easier to introduce in a nonbreaking way.

The unfortunate result of that is yes you've gotta know about underlying changes in that situation.

Sushisource: Also worth nothing - we could always layer on that approach later if it becomes clear it's desired. You could start using per-wf-type versions and that could get aggregated up into a `workflowsVersions` mapping which is used in a similar way to this version number.

versioning/worker-versions.md
Sushisource: @bergundy 
>  Would they be surprised if they bumped an rc or a patch and we simply ignored that?

I don't think they would. They have to understand the incrementing rules to even make use of this feature at all in the first place. If you understand them, then you understand we only care about major/minor, and anything else is just your personal bookkeeping. 

@lorensr 
So nothing's preventing you from doing it that way still if you think it's easier, just deploy and don't actually bump the minor number. New tasks would still work with the previous minor versions if you do bump. They just won't be routed to them, because presumably you were upgrading for a reason and would prefer to run on the new code.

The version _is_ a semver version. It adheres to the spec, we just happen to only match on / care about major and minor. It's exactly like specifying `^1.0` in npm land, which is a common thing to do anyway.

I suppose just one number is in some sense simpler, but, you still have to understand what constitutes a breaking change and what doesn't which seems to me to be most of the work.

versioning/worker-versions.md
mmcshane: Yeah, I would have preferred the individual approach but I get your reasoning as well.

versioning/worker-versions.md
lorensr: Gotcha. So we could support both ways. And we want to support the including-minor-version way because we expect there will be users who will be leaving up their `2.1` workers for some period of time after deploying `2.2`?

If yes, I'm wondering whether we should recommend / explain first what we think is the better DX path in the docs. Like:

- Breaking change: set `workerVersion: '2'`
- Not breaking:
  - If you can send shutdown signal to current v2 workers at the same time you deploy, do that
  - Otherwise, set `workerVersion: '2.1'` and deploy

Sushisource: Any users who need to maintain availability, we'd expect there is a small period of overlap between 2.1 and 2.2. They need to avoid a stop-the-world-then-restart it situation which would harm availability while that happens.

Using 2.1 and 2.2 explicitly ensures that new tasks go to the new workers, instead of some new tasks potentially still taking the old path. Then once all open workflows have processed at least one task on 2.2, you're notified you know it's OK to retire the 2.1 workers.

Otherwise you could turn on all your next version workers, then once they're up, send the shutdown signal to your prev version ones, but while you are doing that it's likely that some new tasks will go to old workers. Probably most of the time, that's totally fine, but having the sequencing be more explicit and visible is a nice property. It also allows you to understand what workers have what versions without you needing to remember that yourself.

So I'm inclined to recommend that people use the minor version, but we can certainly point out that strictly speaking you don't have to use it if you don't want in the docs.

bergundy: > I don't think they would. They have to understand the incrementing rules to even make use of this feature at all in the first place. If you understand them, then you understand we only care about major/minor, and anything else is just your personal bookkeeping.

I don't think this is good DX, it basically means that either:
1. Our system doesn't distinguish between these 2 different workers and workflows may be routed to either.
2. Worker is rejected when it connects to the system because it uses a pre-registered version and a different binary checksum.

versioning/worker-versions.md
```diff
@@ -0,0 +1,137 @@
...
+
+The `patch` type APIs can't be eliminated completely. There will always be 
+situations where a "broken" workflow needs to be fixed to make progress. Such
+fixes are typically less common than just making changes to workflow logic, though.
+
+To make the more common case easier, we will provide a set of features that
+make the "versioned task queues" approach much easier to use for changes to
+workflows which break history compatability (but not interface compatability).
+
+At a high level:
```
yiminc, line 46: I assume user still need to run old workers as long as there is old version of workflow running. So, from high level this proposal is similar to `Versioned task queues` mentioned above, but with some benefits. Maybe you can list the benefits more explicitly.

versioning/worker-versions.md
Sushisource: It means 1. Are you suggesting it's more clear to explicitly only pass major/minor? I'm not sure that really fixes much. You still have to understand what they mean - and if you only allow that, users have no way to attach something they might want to show up when querying worker versions/etc like a build id. You could make a separate field for that, but then it's the same thing, you could say they might be confused why that field doesn't affect routing. Additionally, you can make the same argument about even that - if the user is just plowing ahead blindly they might expect incrementing minor to do something different than it does, which is sorta the same as having some incorrect expectation about incrementing patch and having that do something.

With the semver string we have:
* It's familiar to users
* Allows inclusion of extra metadata user might care about but system doesn't
* Future proof if we _do_ care about patch/etc later on
* Easy to write
* No change in routing with patch/prerelease/build is at least congruent with reasonable expectations about how they would behave 

TLDR: I don't think there's a version of this API that allows you to never understand anything and use it blindly, successfully, and the string has some other nice advantages

versioning/worker-versions.md
Sushisource: It really is just improvements to versioning task queues.

The benefits are the ones listed here - clients automatically point at the latest version without needing to be updated, and users can get easily notified of when it's time to retire old workers. Those are pretty substantial DX annoyances when doing it by hand right now. I added a couple more bullets too.

versioning/worker-versions.md
bergundy: Okay, I'm convinced this is the right approach

versioning/worker-versions.md
```diff
@@ -0,0 +1,139 @@
...
+* Provide a simple API for versioning entire workers (this proposal)
+* Allow clients to point at task queue `foo`, and automatically route new workflow executions
+  to the latest version of the `foo` queue.
+* Provide good notifications and information about when old workers may be
```
bergundy, line 50: Technically we can't really know this because closed workflows can still be queried.

versioning/worker-versions.md
```diff
@@ -0,0 +1,139 @@
...
+* Easy visibility into what versions deployed workers are using
+
+Building on this base set of features, we plan to eventually implement
+automatic safe-rollout of changes, and possibly tools like k8s operators
+which can automatically manage worker deployments.
+
+## SDK API changes
+
+The meat of this proposal is simple:
+**Add a version field to worker options in each SDK**
```
jlegrone, line 62: _TLDR: I'd like to explore whether it is possible to deal with backwards-incompatibility of workflow code without relying on the human judgement element of semver-like versioning. In this comment I’m also making the assumption that it is possible to detect backwards-incompatibility via workflow replay._

My company runs an internal workflow automation platform built on Temporal. It's just vanilla Temporal with a few more opinionated layers on top like workflow interfaces defined in protobuf and a worker framework that handles authz, observability, deployment config, etc.

Among our customers, a common thread is that regardless of an engineer's seniority level or experience with Temporal, people sometimes forget to use workflow versioning, or may write a bug while attempting to do so. Recovering from a bad code change like this is tricky. So a while back we set out to solve the same issues described in this proposal, but with an added constraint: don't depend on human judgement to make worker deployments safe by default.

### Why not rely on a human specified version?

- The same engineers who forget to add version checks in workflow code can forget to update the worker's version.
- How and when to update the version requires perfect knowledge of both workflow code backwards compatibility rules AND the super-major/major/minor rules defined in this proposal. Failing to increment the version correctly when there has been a breaking change would appear to result in the same "recover from bad code change" complexity we set out to avoid.
- Adding, updating, or removing an interceptor can also break workflow replay. Changes to other library code can do the same. We try to avoid this stuff, but it's hard to ensure compatibility only through code review and workflow history snapshot/replay testing.
- Coordinating to increment a version among a large team of developers can be tricky, to the point that we've moved away from semver for releases of other internal services.

### Automated Breaking Change Detection

I think there are elements of this proposal and our internal approach that can be combined. First I'll describe our current approach to breaking change detection:

Each worker has two task queues, eg. `foo:canary` and `foo:main`. All production traffic flows to the `foo:main` queue. Our worker framework also automatically registers a "reflection" workflow on each task queue that allows querying a list of all workflows, activities, signals and queries supported by the worker, and a “replay” workflow that allows replaying any workflow ID.

During worker deployment, we first roll out the new version on the canary task queue. A generic “breaking change detection” workflow is then executed which:

1. Uses the reflection workflow to query a list of workflow and activity types registered on `foo:main`
2. Does the same to retrieve a list of workflow and activity types registered on `foo:canary`
3. Diffs the two responses to statically check for removed workflow/activity types etc.
4. Queries Temporal for a list of running workflow executions of each type found on `foo:main`, then replays each workflow's history on `foo:canary`.

If issues are surfaced by the breaking change detection workflow, then we block the deployment and issue a manual judgement. If a developer decides to roll out anyway, then we direct them to resources on batch reset or termination of broken workflows.

Note that we don’t view this as the end-all-be-all of developer experience. Most of these issues should be caught by continuous integration tests before a breaking change is ever merged.

### Alternative Proposal: Binary Checksums

There are some very interesting ideas here that have me rethinking the above approach. Some downsides of the canary/main replay testing technique are that:

- In the time we spend replaying workflow histories, more workflows may be started and others may make progress, resulting in a chance of breaking changes being missed.
- We may have thousands or more workflows to replay, in which case sampling may be necessary in order to reduce deployment latency. Again this increases our chances of missing breaking changes.
- We had to make large investments in tooling to make this work, and our solution may be incongruent with the practices of the larger Temporal community.

If this proposal were accepted as-is, then I think we might be able to simplify our internal solution by having our deployment tooling auto-increment workers by a major version for every change and skipping the replay parts of the breaking change detection workflow. This is driven by the assumption that any rollout has the possibility of containing a breaking change, and because we value the ability to recover gracefully over rollout speed (is this is the tradeoff between major and minor/patch version bumps)?

> Building on this base set of features, we plan to eventually implement automatic safe-rollout of changes

So I’m left wondering, is it possible to simplify things further for other users as well? For an alternative approach we could look to [worker binary checksums](https://github.com/temporalio/api/blob/336a14569b47d145300989c830cf50ab8ccad577/temporal/api/workflowservice/v1/request_response.proto#L212-L213), which are an existing method to track worker versions.

When breaking changes to workflow code are rolled out today, worker binary checksums are an important tool for recovering workflows. For example after a bad deployment, an operator can mark a binary as bad to prevent further progress being made at that version:

```bash
tctl namespace update --add_bad_binary sha256:deadbeef
```

 Then roll back the worker deployment and reset any workflows that had made progress at that version:

```bash
tctl workflow reset-batch --query “TaskQueue = ‘foo’ and ExecutionStatus = 'Running'" --only_non_deterministic --reset_type BadBinary --reason "bad worker version" --reset_bad_binary_checksum sha256:deadbeef
```

My thought is that we could combine binary checksums with a new “rollout strategy” worker option. Rollout strategy would default to the existing behavior of “any (non-bad) worker binary can poll for work”, but also include a “versioned cutover” option that matches the behavior described in this proposal for major version bumps. This has two key benefits:

- Keeps semantic versioning of workers optional and orthogonal to rollout strategy. Different rollout strategies may easily be applied in dev/staging/prod without changing versioning strategy between environments.
- Relies on an existing concept which is already key to more advanced workflow batch operations and recovery options. This means very little investment needed from anyone already using binary checksums, and leaves less room for confusion between use cases for semantic versions vs. binary versions.

Note that binary checksums are not sortable; Temporal server would probably need to keep track of the first observed poll time for the last _n_ checksums on a given task queue or within a given namespace so that we know which is the target “latest” version. An API for rollbacks would also be necessary so that an older checksum could be set to the target version. This could also be useful debugging info to surface in the web UI and for suggesting actions for workflow operators in the event of bad deploys.

I think this gets us pretty close to “automatic safe-rollout of changes”, since workflows would be automatically replayed before cutting over between every new worker version. Any non-deterministic code changes, missing workflow/activity/signal types, payload encoding issues, etc should raise errors during replay and thus keep those workflows polling on the previous worker version.

bergundy:
I really like the binary checksum approach @jlegrone.

To provide all of the capabilities listed in this proposal, all the server needs to know is:
1. Which checksum is the latest
2. Which checksum "patches" a previously seen checksum
The semantic version is irrelevant.

It goes back to the experience we were thinking about with "Entire Workflow versioning" (another proposal in this repo that did not go through). I argued that specifying the semantic version in worker config is error prone and suggested that there would be an additional step where the user would upload their workflow code bundle (TypeScript specific) to a "repository" and in that step the user would have to say whether they're patching an existing version or deploying a completely new one.
The default would be not to patch and the act of patching ties well into the `patch` workflow API semantic.

There are different approaches as to how to expose this functionality (via worker config / operator API) that we should consider.

With the checksum approach we could still offer automatic safe rollout and possibly auto "latest" promotion without user intervention.

versioning/worker-versions.md
Sushisource: Thanks for the very thoughtful feedback!

I think what you're getting at is totally reasonable, and the idea of making the cut-over work with just checksums so you can pick that over manual versioning if you prefer makes sense to me. The process you described of replaying existing histories to verify new versions are good is very much part of the automatic safe-rollout stuff we'll build later that I mentioned. So a lot of this is in-line with what we already are thinking of.

The meat of what you're getting at I think is really "what does safe rollout look like" which is the next step after this. I think the only alteration required to my proposal is to allow specifying compatibility more flexibly than just semver. For major changes, things are essentially exactly equivalent assuming you auto-bump major revisions (which, maybe that's the primary reason to _only_ use checksums).

So, if you wanted to do a safe-rollout of a non-major change using only checksums, maybe you have something like:

```typescript
Worker.create({ isCompatWith: ['deadbeef', 'feedbeef'], ...otherOptions })
```

When this worker starts up, it can
1. Tell server "Here's my checksum" so that server knows it is latest and begins queuing new workflow starts in that queue
2. Replays all (or some sample of, probably configurable) workflows on the queues it's compat with
3. If that works, tells server "passed compat check"
4. Begins polling (probably after polling a "have I passed yet?" check)

There are a few complications here, like you probably only want one worker to do all the replaying (or even more advanced, break up work among workers) so that other workers can immediately start processing new workflow starts, since we don't care about compat for these anyways. So you'd either have to explicitly nominate one worker as "leader" by passing in an env var or something, or possibly server just designates the first worker to connect as leader.

I think that could be really nice. Can you elaborate on "... "rollout strategy" worker option ..." and "Keeps semantic versioning of workers optional and orthogonal to rollout strategy. "? 

My feeling is that user-specified semver provides pretty much nothing except a pre-defined ordering of the versions which could be used to sanity check that checksums don't go "backward", and maybe just as a nice human-readable thing in the UI. 

I want to draw up some sequence diagrams of the above (I had done so for the semver-based thing but they aren't included in here) to think through it a bit more and we can use that do decide if we want to do this at all, or replace semver, or compliment semver.


I wrote the above after I wrote the below, but, replied to some specific things down there.

---

> If this proposal were accepted as-is, then I think we might be able to simplify our internal solution by having our deployment tooling auto-increment workers by a major version for every change

This is pretty much exactly what I imagined people who prefer your sort of approach would do.

Regarding:
> we value the ability to recover gracefully over rollout speed (is this is the tradeoff between major and minor/patch version bumps)?

Not even, really. In this proposal minor versions still get their own queue, though rather than only starting brand new workflows, new workflow _tasks_ may be started there. Really the only reason to use a minor version is because you have some bug that's causing workflows to get "stuck" and you want to unblock them, and for whatever reason you can't/don't want to reset instead. If pure deployment speed is your concern, you wouldn't even bump minor version here so you can keep using the same queue, but of course this also allows "old" workers to keep processing new tasks (which I assume generally people would prefer to avoid). You could bump patch purely for your own bookkeeping.

<br/>

One of the sticking points with the checksums is they're surprisingly difficult to reliably calculate for TypeScript and upcoming Python, but in practice it's probably solvable for everything but the most unusual setups.

The lack of a defined ordering is the other issue as you mentioned.
> keep track of the first observed poll time for the last n checksums on a given task queue or within a given namespace so that we know which is the target “latest” version.

I think that works the vast majority of the time, it's slightly error prone in terms of making sure you don't connect workers in some out-of-order fashion but that'd be a pretty odd situation to be in. No worse than some of the other things you could forget, and you need the rollback API anyway to undo that.

Sushisource: I made a diagram of what his might look like
```mermaid
sequenceDiagram
actor Client
participant T as Temporal
participant W11 as v1.1 Worker

Note right of T: A v1 (0xbeef1) Worker is already established as current "latest"<br/> (not shown)
W11 ->> T: Upon connection: "I have bin id 0xbeef1.1 and I am compat with 0xbeef1" <br/> I use rollout strategy "wait for replay to pass"
T -> T: A new compat worker is connected, but has not been approved as compatible

Client ->> T: Start workflow on `foo`
T -> T: Latest for `foo` is 0xbeef11.1, so I will start the WF on that queue
Note right of T: Whether to queue starts & new tasks on the new ver or use the existing<br/> one can be a configurable part of rollout strat.

W11 ->> T: Poll for "Am I approved yet?" 
T -->> W11: No, and you are the leader

W11 ->> T: Fetch workflow histories from 0xbeef1 queue
T -->> W11: Histories
W11 -> W11: Replay histories

W11 ->> T: 0xdeadbeef1.1 has passed compatability check
W11 ->> T: Poll for "Am I approved yet?"  (all 1.1 workers will be doing this)
T -->> W11: Yes
Note left of W11: Now 1.1 workers know it is OK to start pulling real work

W11 ->> T: Poll on `foo-0xbeef1.1` queue
```

bergundy:
I think we should decouple marking a binary checksum as latest or "patches X" from worker registration.
If we add explicit APIs for those, it would allow much more flexibility in implementing different deployment strategies.

The main reason why I think it shouldn't be part of the worker configuration is that it removes the need for users to modify their code just to change the checksum tags (latest / patchFor). That sort of change tends to be overlooked in reviews when switching between different rollout types and might be blindly copied by other engineers who're just looking for reference code for creating a worker.
Another advantage of this decoupling is that it only takes a simple script to modify the tags as opposed to redeployment of the worker (possibly generating a new checksum).
I'd categorize these tags the same way we (should) look at activity task queue rate limits, they're global configuration and not worker specific.
If one desires to couple tagging with Worker registration, they can call the API on worker process startup.

Sushisource:
I've made a new PR [here](https://github.com/temporalio/proposals/pull/54) which focuses on build-id based versioning, since this PR already has many comments. This PR will remain open until we decide which is preferred.

Sushisource:
The other proposal was merged

PR #53, opened 2022-03-08
Propose the update workflow gRPC service interface

mmcshane:
rendered: https://github.com/mmcshane/proposals/blob/mpm/update-rpc-interface/server/update-workflow-interface/update-workflow-interface.md


server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+### Update
+
+An Update is the execution of arbitrary workflow logic (subject to the standard
+workflow code constraints on determinism) triggered by an UpdateRequest and
+eventually resulting in a single UpdateResponse. The processing of an Update is
+performed by workflow code and as such the timeliness of the UpdateResponse is
+dependent on worker resources being available to process the Update. A given
+Update applies to exactly one workflow execution. The lifetime of an updated
+begins at the first UpdateRequest triggering said Update and ends when the
+UpdateResponse has been mad durable. Consequently, the lifetime of an Update may
```
Sushisource, line 29: _mad_ durable, yo 😎 

Lol forgive me, was browsing through this on my phone and just had to get this one in before you fix it.

mmcshane: Do not sleep on this level of durability. I kind of want to keep it.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+### UpdateRequest
+
+An UpdateRequest is a remote call originating either from outside the Temporal
+system entirely or from another Temporal workflow.  UpdateRequests are
+deduplicated such that issuance of the same UpdateRequest multiple times is
+effectively the same as issuing a single request. The UpdateRequest includes an
+indication of the workflow execution at which the Update is targeted.
+UpdateRequests are issued over gRPC or through SDK-specific APIs wrapping the
+same. Both issuing and receiving UpdateRequests becomes a durable part of a
+given workflow's history.
```
cretz, line 45: So obviously I can re-issue a known update to get what the response _was_. Some questions here:

1. Does that work even after the workflow is complete? EDIT: I see below that they may be accessed after workflow complete, though unsure if we'll provide an easy way.
2. Is there anyway to say "only-if-already-issued" so the update isn't an upsert of sorts, but really just a get-result-if-update-present? Maybe there's no use case for this?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+The gRPC API comprises two functions, `UpdateWorkflow` and
+`AwaitUpdateWorkflowResponse` to be added to the `WorkflowService` service
+definition. The former is used to invoke an update and follows the standard
+patterns for identifying a workflow execution and carrying arbitrary metadata
+(Header) and payloads (Payloads). The request type for this function includes a
+`wait_timeout` property indicating how long the client is willing to wait for a
+synchronous response once the request has been accepted by the system. This
+value can be set to zero indicating no such waiting should occur.  The
+`request_id` property uniquely identifies the logical request (_not_ the
+issuance of the request) and is intended for use as an idempotency token.
```
cretz, line 115: Is this `request_id` alone what is used for deduplication? Or is it combined with other values?

If `request_id` is the idempotency token for this purpose, can it not be used as the update ID? It seems much simpler to have a user-provided update ID that is used for deduplication and can be used everywhere to reference the update.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+### UpdateResponse
+
+The output value of an Update. The UpdateResponse may lack a data playload, in
+which case the UpdateResponse only indicates the success or failure of the
+Update operation. The UpdateResponse is accessed via gRPC or SDK-specific APIs
+wrapping the same. UpdateResponses become durable parts of a given workflow's
+history. UpdateResponses can be accessed after the lifecycle of the Update has
+completed, subject to history storage & archival configuration.
+
+## Use Cases
```
cretz, line 56: I can think of three more use cases:

* Fire-and-forget
* Duplicate/reentrant updates (maybe with different payload?)
* UpdateWithStart - SignalWithStart, but with an update (we don't have to support this now, just a thought)
* As a signal replacement - From a caller POV, what is a signal but an update that you don't care the response of? Why would I ever use signals again?
* UpdateExternalWorkflow - From _inside_ a workflow updating another (not part of the client gRPC of course)

None of this needs to be solved now, just some idle thoughts. I haven't really been involved in any of the discussions around update so I am surely missing lots of context.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+message UpdateWorkflowResponse {
+    UpdateResultType result_type = 1;
+    string update_id = 2;
+    temporal.api.common.v1.Payloads output = 3;
+}
+
+enum UpdateResultType {
+    UPDATE_RESULT_TYPE_UNSPECIFIED = 0;
+    UPDATE_RESULT_TYPE_SUCCESS = 1;
+    UPDATE_RESULT_TYPE_ERROR = 2;
```
cretz, line 149: Is this only for workflow-returned errors or does it also represent update timeout, workflow not present, update not handled by workflow, etc?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+    rpc UpdateWorkflow(UpdateWorkflowRequest) returns (UpdateWorkflowResponse){}
+    rpc AwaitUpdateWorkflowResponse(AwaitUpdateWorkflowResponseRequest) returns (UpdateWorkflowResponse){}
+}
+
+message UpdateWorkflowRequest {
+    temporal.api.common.v1.Header header = 1;
+    string request_id = 2;
+    string namespace = 3;
+    temporal.api.common.v1.WorkflowExecution execution = 4;
+    string update_name = 5;
```
cretz, line 135: Can you confirm this value is required?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+Client calls `UpdateWorkflowRequest` with a zero `wait_timeout` and uses the
+returned `update_id` to obtain the UpdateResponse at a later time with
+`AwaitWorkflowResponse`.
+
+### Invalid Update
+
+If the UpdateRequest is rejected by the workflow execution while the RPC client
+is still connected, a suitable gRPC error can be returned synchronously. Note
+however rejected UpdateRequests are _not_ recorded in workflow history.
```
cretz, line 183: > rejected UpdateRequests are _not_ recorded in workflow history.

This is interesting. So what if I send an fire-and-forget update or short update to a "running" workflow that does not have a worker running it at the time? If I don't wait around it may not get in history but if I do wait around it does?

IMO like signals, a successfully received update by the _server_ should be all that is required for history persistence. The state of the worker should have nothing to do with whether the update will be processed in the future. It is every confusing to the caller that something as inexact as whether a worker is actively processing tasks for this workflow to affect some so exact as its history.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+### "Failing" Update
+
+An Update that is passes verification by the workflow code and runs to
```
cretz, line 187: Kinda related to my other comment - why does an update need to pass workflow code validation? Why does workflow code even need to exist for it? I can think of many cons, but what benefits are there of requiring workflow code run to validate a request inline? Do those benefits outweigh the ability to provide updates on a workflow not on a worker at the time?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+attached to the execution of the previously submitted request. In either case,
+the `UpdateWorkflow` RPC call proceeds and returns normally from the client's
+perspective.
+
+
+## Concurrent Semantics
+
+Update processing both within a single workflow execution and across multiple
+workflow executions can procede in parallel however at most one update will be
+processed by any single workflow execution at a time. Updates that overlap in
```
cretz, line 209: Does this mean my update handler may not be async in the SDK? Meaning, can I get an update request, execute an activity, and use the result of that activity as the update response?

server/update-workflow-interface/update-workflow-interface.md
mmcshane: It's to avoid that problem where a bad (possibly unintentionally so) actor floods a workflow's history with invalid updates. The hard requirement we settled on is that invalid requirements (which is a determination that can only be made by workflow code) _must not_ consume history space.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: The combination of an update ID always dereferencing to _some_ result _eventually_ and invalid UpdateRequests not causing history traffic means that it _has_ to be the server generating the update ID. If you want something like "do X if you haven't already done it" the client could use a hash of the UpdateRequest as the request_id.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Update ID
+
+The unique identifier of an Update. Generated by the Temporal server.
+
+### UpdateRequest
+
+An UpdateRequest is a remote call originating either from outside the Temporal
+system entirely or from another Temporal workflow.  UpdateRequests are
+deduplicated such that issuance of the same UpdateRequest multiple times is
```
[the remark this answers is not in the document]
mmcshane: Updates are not deduplicated, only UpdateRequests. I left out the specific mechanism here as this section is for concepts but you'll see the `request_id` field used for this purpose further down.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: I've sort of mentioned this obliquely in some other responses but the idea is that if there is an update ID then the update will (eventually) happen and there will (eventually) be an UpdateResponse. For that to be the case, it has to be generated on the server side _after_ the workflow has validated the UpdateRequest.

mmcshane: P.S. I'm noting that this arrangement (request_id vs update_id and the implications) wasn't clear in the proposal and I'll improve that.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+}
+
+message UpdateWorkflowRequest {
+    temporal.api.common.v1.Header header = 1;
+    string request_id = 2;
+    string namespace = 3;
+    temporal.api.common.v1.WorkflowExecution execution = 4;
+    string update_name = 5;
+    temporal.api.common.v1.Payloads input = 6;
+    google.protobuf.Duration wait_timeout = 7;
```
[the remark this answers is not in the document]
mmcshane: The reason to have this apart from the gRPC timeout is so that the server can return an empty-ish response after `wait_timeout` that includes the `update_id`. From the gRPC perspective that's a successful call/response pair but at a higher layer we understand that there is more to do to see the UpdateResponse. Giving back the `update_id` is necessary so that the client can come look for it later.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+}
+
+message UpdateWorkflowResponse {
+    UpdateResultType result_type = 1;
```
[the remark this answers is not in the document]
mmcshane: Ooo I didn't know about the `Failure` type, thanks.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: IIUC, yes, that's the intent. Seems like good flexibility to have, no? Or have I missed your point?

server/update-workflow-interface/update-workflow-interface.md
cretz: Request ID's are traditionally auto-generated by our SDK for use by the server for their idempotency reasons of the literal gRPC request. Not to tie potentially separate requests together. Are we expecting the user to provide this request ID? I wonder if there is a better name (I can't think of one off-hand).

server/update-workflow-interface/update-workflow-interface.md
cretz: You're good here, I got confused in thinking this was a user-facing construct for them to be idempotent with their updates. But really this is internal idempotency between client and server like out other request IDs.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: > There is no user-facing idempotence

There is though? Issue two UpdateReqests with the same request_id and they get coalesced/deduped

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
cretz: I see.

I wonder then if it's clearer from a code POV to always require wait for response as a separate call and have this call always be fast? Basically we don't allow `StartWorkflowExecution` to wait on its response. So have this function always return the update ID immediately and always make obtaining the result use another call. Internally on the server I figure they share that kind of "awaiting" code anyways, and no need for shortcuts for the SDK's benefit (it's actually easier from an SDK to have only one way to wait on a result).

server/update-workflow-interface/update-workflow-interface.md
cretz: :+1: Bonus if we can change the requirement that workflow code has to run before an update ID is even issued. Mentioned elsewhere, but like signals, I'd like not not care about the worker to get this into history for processing.

server/update-workflow-interface/update-workflow-interface.md
cretz: :+1: I like it and I would like it to be required like signal and query names.

server/update-workflow-interface/update-workflow-interface.md
cretz: > The update won't hit history until it's validated by the workflow so you can wait but until there's a worker running the update won't hit history.

But why? Is there real benefit to running worker code inline on every update before it hits history? I like the idea of sending an update like a signal.

> But the conclusion we came to was that the threat of filling history with invalid updates was a greater concern than keeping that consistency.

There is a threat of invalid updates but not invalid signals? Can we expand on this threat a bit? I don't think this is a real problem anymore than with signals. If updates were just signals with responses, you get so much more freedom. But forcing a worker to be involved in the creation of the update is rough IMO with very limited benefit and preventing a problem I am not sure exists.

server/update-workflow-interface/update-workflow-interface.md
cretz: Do you think if we reimplemented signals from scratch again we'd require worker code to execute before it was in history for fear of this bad actor? What about a bad actor flooding my workflow with signals? Are we saying we wish we hadn't allowed signals to avoid inline worker validation?

Some other related questions:

* Can update handling have side effects?
* Can update validation have side effects?
* Can I have custom update validation?

server/update-workflow-interface/update-workflow-interface.md
cretz: Can you clarify your response - can update handlers be async?

Assuming you meant that update handlers _can_ be async for purposes of flexibility, that makes sense, I was just making sure that "at most one update will be processed by any single workflow execution at a time" didn't mean "processed through to update response" but instead means "processed but the update result can be provided later".

server/update-workflow-interface/update-workflow-interface.md
mmcshane: I have a slight preference not to use presence/absence in proto3 just because _everything_ is optional. It's like protocol design with one hand tied behind your back. That said, doing `oneof { Payloads | Failure | Empty }` or something like that is a cleaner protobuf encapsulation (the mapping of oneof into language constructs though ... I'm not sure I want to inflict that on anyone).

server/update-workflow-interface/update-workflow-interface.md
cretz: :-( I think that choice unfortunately hampers legitimate use of "always process the update regardless of worker state".

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
cretz: > Can't always be fast because we want it to return the UpdateID

It could if the worker wasn't in the way :-) But I understand. I don't have a strong opinion here. To me a send request and get response is a very natural separation for Temporal (especially if sending a request doesn't block until the worker accepts it).

server/update-workflow-interface/update-workflow-interface.md
cretz: I am going to close this thread as it's basically the same as the other.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+## Concepts
+
+### Update
+
+An Update is the execution of arbitrary workflow logic (subject to the standard
+workflow code constraints on determinism) triggered by an UpdateRequest and
+eventually resulting in a single UpdateResponse. The processing of an Update is
+performed by workflow code and as such the timeliness of the UpdateResponse is
+dependent on worker resources being available to process the Update. A given
+Update applies to exactly one workflow execution. The lifetime of an updated
```
bergundy, line 27:
```suggestion
Update applies to exactly one workflow execution. The lifetime of an Update
```

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+UpdateRequests are issued over gRPC or through SDK-specific APIs wrapping the
+same. Both issuing and receiving UpdateRequests becomes a durable part of a
+given workflow's history.
+
+### UpdateResponse
+
+The output value of an Update. The UpdateResponse may lack a data playload, in
+which case the UpdateResponse only indicates the success or failure of the
+Update operation. The UpdateResponse is accessed via gRPC or SDK-specific APIs
+wrapping the same. UpdateResponses become durable parts of a given workflow's
```
bergundy, line 52: They only become durable if the workflow accepts them

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+history. UpdateResponses can be accessed after the lifecycle of the Update has
+completed, subject to history storage & archival configuration.
```
bergundy, line 54: Are we considering mutable state "history"?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+### Update
+
+An Update is the execution of arbitrary workflow logic (subject to the standard
+workflow code constraints on determinism) triggered by an UpdateRequest and
+eventually resulting in a single UpdateResponse. The processing of an Update is
+performed by workflow code and as such the timeliness of the UpdateResponse is
+dependent on worker resources being available to process the Update. A given
+Update applies to exactly one workflow execution. The lifetime of an updated
+begins at the first UpdateRequest triggering said Update and ends when the
```
Sushisource, line 28:
```suggestion
Update applies to exactly one workflow execution. The lifetime of an update
begins at the first UpdateRequest triggering said Update and ends when the
```

server/update-workflow-interface/update-workflow-interface.md
Sushisource: Signal replacement makes a lot of sense to me. Reducing the number of primitives we have is always a nice thing.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+```protobuf
+service WorkflowService {
+    rpc UpdateWorkflow(UpdateWorkflowRequest) returns (UpdateWorkflowResponse){}
+    rpc AwaitUpdateWorkflowResponse(AwaitUpdateWorkflowResponseRequest) returns (UpdateWorkflowResponse){}
+}
+
+message UpdateWorkflowRequest {
+    temporal.api.common.v1.Header header = 1;
+    string request_id = 2;
```
Sushisource, line 132: nit: Might read easier to just move some of the prose above into comments inside the proto message.

server/update-workflow-interface/update-workflow-interface.md
Sushisource: +1 to required

server/update-workflow-interface/update-workflow-interface.md
bergundy: Updates could replace signals in some use cases.
There are tradeoffs to each approach.
Updates can be rejected by the workflow in which case they will not be persisted in history, this is useful when you need to validate the update before applying it and makes update a mechanism that could be exposed directly to clients.
Signals OTOH are better for async processing, they don't require a worker to be up to be processed, they're much cheaper and involve less bookkeeping.

UpdateExternalWorkflow is hinted above in the `UpdateRequest` section.
UpdateWithStart was discussed too, we will definitely have that.

server/update-workflow-interface/update-workflow-interface.md
Spikhalskiy, line 27: typo? updated -> Update?

server/update-workflow-interface/update-workflow-interface.md
paulnpdev: fwiw, I'm in agreement with Chad here ... worker code shouldn't be in the way of writing history under any circumstances.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+The unique identifier of an Update. Generated by the Temporal server.
+
+### UpdateRequest
+
+An UpdateRequest is a remote call originating either from outside the Temporal
+system entirely or from another Temporal workflow.  UpdateRequests are
+deduplicated such that issuance of the same UpdateRequest multiple times is
+effectively the same as issuing a single request. The UpdateRequest includes an
+indication of the workflow execution at which the Update is targeted.
```
Spikhalskiy, line 42: Is this indication a WorkflowID? Or WorkflowID + parentRunId? Would be good to add a clarification on what's a target here. I assume that it's very close to Signal semantic.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+UpdateRequests are issued over gRPC or through SDK-specific APIs wrapping the
+same. Both issuing and receiving UpdateRequests becomes a durable part of a
```
Spikhalskiy, line 44: It's not exactly clear what does it look like. Can we add a piece of history? Will it be the same with the Signal - UpdateRequest event followed by WFT?

server/update-workflow-interface/update-workflow-interface.md
Sushisource: I think having the validation and defaulting to preventing growing history unnecessarily makes good sense.

And at first blush, it also seems valuable to have the ability to fire off updates as fast as possible. However, when you think about "what's the usecase for firing off a signal/update and wanting it to be in history ASAP?" I think it's not immediately clear what that's for. Either you don't care about a result, in which case just use the zero timeout and from a caller's perspective it looks the same, or you do, in which case waiting for the round trip was necessary anyway.  The only situation seemingly solved by the super-quick write is being able to get an ID back as fast as possible, but that pushes you into the "actually I do care about the result" situation :thinking:

server/update-workflow-interface/update-workflow-interface.md
Spikhalskiy, line 52: What's the point of storing an empty UpdateResponse? Just to confirm that it was processed? Isn't it enough to know the fact that the following WFT was completed?

server/update-workflow-interface/update-workflow-interface.md
paulnpdev: Since the issue you are working on here is SO simple and clearly not challenging enough ;-) I just wanted to ask: are y'all considering the tangentially related question of a streaming model where the SDK talks to the backend over a durable connection (in some cases)?  There are lots of use cases where a single Workflow goes through a (potentially quite large) flurry of transitions and we would like to be able to process these in a (dramatically!) more-efficient fashion.  Obviously we're talking about a rather large change here but the payoff is huge (in some cases) so we have to at least discuss it.  Not sure now is the time, but the work you're doing right now brought it to mind ... thought you'd want to know it's on the table.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
+# Proposal: Update Workflow Server Interface
```
Spikhalskiy, line 1: Feel free to ignore, but I don't like the name. Signal is a good name. "Update" is not. This term is so overloaded. I think we should choose a more descriptive name for this abstraction.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+### Short Update
+
+It may be that it is known a priori that an update is expected to execute in an
+amount of time that is reasonable for an RPC request/response interaction. In
+such cases the issuer may want to block the issuing thread of execution for the
+full duration of the update.
+
+### Uncertain Duration Update
+
+It may be the case that there is no expectation on how long an update may take
```
Spikhalskiy, line 67: If we have this, I don't think a separate "Short Update" is a really valid usecase. All blocked wait should be timed. Even if user doesn't see it and the timeout is hidden somewhere in the Options.

server/update-workflow-interface/update-workflow-interface.md
paulnpdev: > I think having the validation and defaulting to preventing growing history unnecessarily makes good sense.

some would argue that an attempted invalid entry IS part of the history.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+history. UpdateResponses can be accessed after the lifecycle of the Update has
+completed, subject to history storage & archival configuration.
+
+## Use Cases
+
+### Short Update
+
+It may be that it is known a priori that an update is expected to execute in an
+amount of time that is reasonable for an RPC request/response interaction. In
+such cases the issuer may want to block the issuing thread of execution for the
```
Spikhalskiy, line 62: issuer as an original issuing client? It looks like the document is intended to discuss the Service interface. Maybe it would be beneficial to discuss what kind of method should be exposed on the client separately?

server/update-workflow-interface/update-workflow-interface.md
mmcshane: @cretz My intent was to say that yes, updated handlers can be async

@paulnpdev Not sure exactly what you're getting at with "durable connection" but most gRPC clients multiplex lots of messages onto "transport" instances (roughly: connections) which have a longer lifecycle.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+implementation details for a deeper design review and instead relate the API,
+its expected usage and the guarantees provided.
+
+## Concepts
+
+### Update
+
+An Update is the execution of arbitrary workflow logic (subject to the standard
+workflow code constraints on determinism) triggered by an UpdateRequest and
+eventually resulting in a single UpdateResponse. The processing of an Update is
```
Spikhalskiy, line 24: Isn't UpdateResponse can be implemented as a marker? It looks like just a marker for me. If it can, can't the whole proposal be implemented and framed as a Signal + a Marker?

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+The second function, `AwaitUpdateWorkflowResponse` is used when a client has
+already issued an `UpdateWorkflow` request and has the resulting `update_id`.
+That `update_id` can be provided along with a `wait_timeout` (same semantics as
+above) to re-attach to an in-flight or completed Update and obtain the
+associated UpdateResult in the form of an `UpdateWorkflowResponse`.
+
+
+```protobuf
+service WorkflowService {
```
bergundy, line 125: We have talked about having separate endpoints for long poll vs. short lived requests.
It makes more sense for latency metrics on these endpoints.

We should consider adding another StartUpdateWorkflow RPC that schedules the update and returns the update_id ASAP.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+enum UpdateResultType {
+    UPDATE_RESULT_TYPE_UNSPECIFIED = 0;
+    UPDATE_RESULT_TYPE_SUCCESS = 1;
+    UPDATE_RESULT_TYPE_ERROR = 2;
+}
+
+message AwaitUpdateResponseRequest {
+    string update_id = 1; 
+    google.protobuf.Duration wait_timeout = 2;
```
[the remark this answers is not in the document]
mmcshane: Ahhhh I think you're right. I can't think of a reason why it's required. Thanks.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
bergundy: I'm foreseeing issues where the request deadline is shorter than wait_timeout, can we just use the deadline for this purpose?

server/update-workflow-interface/update-workflow-interface.md
cretz: For me, I don't want my update request to fail anymore than I would for start, cancel, or signal when the worker is not processing the workflow atm. Maybe I care about the response maybe I don't. Separating the request of an update from its execution where the worker is only required for the latter is a reasonable separation there I think. I don't think this should be phrased like "well if you're gonna have to wait for response anyways, might as well have a two-phase validation portion too", but instead "why are we adding such extreme complexity for this validation use case"? Is our lack of prevalidation today a real problem?

I think once implementation time comes around and we see how the APIs manifest in SDKs for validating vs processing the update as separate things, the complexity will rise. But I will go with the team if we still want to have worker validation before history placement.

A few more questions (sorry, last two kinda unrelated):

1. What validation timeout is leveraged when there is no worker processing the workflow at that point? How long will it wait for a worker to appear to process just the validation?
2. An update execution can be retried, but can an update validation be retried if the worker doesn't respond?
3. Does every update have to be responded to before the workflow completes?
4. I have heard we have no process-update timeout, so the update can take many weeks to be responded to right?

server/update-workflow-interface/update-workflow-interface.md
cretz: >  UpdateWithStart was discussed too, we will definitely have that.

When we do get there, I'd like to propose we instead just add fields to `StartWorkflowExecutionRequest` instead of new RPCs (same for signal really, but of course we already have something there). Can I send a signal _and_ 3 updates to start the workflow with? :-)

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+}
+
+message UpdateWorkflowResponse {
+    UpdateResultType result_type = 1;
+    string update_id = 2;
+    temporal.api.common.v1.Payloads output = 3;
+}
+
+enum UpdateResultType {
+    UPDATE_RESULT_TYPE_UNSPECIFIED = 0;
```
bergundy, line 147: We never use the `UNSPECIFIED` variant in other API enums, we should have a distinct value for update_id.

cretz: I see it used in plenty of places at https://github.com/temporalio/api/tree/master/temporal/api/enums/v1. I appreciate it due to following https://google.aip.dev/126, but in general I can see value in removing this altogether and just relying on presence of an error to determine errored.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: @paulnpdev what it comes down to is whether we are okay allowing invalid updates to consume space on the history. We allow it for signals but at the onsite the consensus was that invalid updates _must not_ eat space in the history. Are you willing to give ground on the history requirement? Or alternatively, do you see a way to identify invalid updates without running workflow code?

cretz: Another question I just thought of:

5. Can I use workflow state in my update validator? What if that state is one value during validation and another value during execution? Are we are going to ask them to re-validate before processing because they can't trust what was validated at validation time is still the case at processing time?

server/update-workflow-interface/update-workflow-interface.md
mfateev: We decided a long time ago not to use "switch structures" to decide which operation is implemented by the given gRPC method. It was done when Thrift was used instead of gRPC which didn't have explicit support for "oneof" types. For consistency I would keep that pattern unless we want to refactor all the other APIs. For example Complete/Fail activity, etc.

server/update-workflow-interface/update-workflow-interface.md
mfateev: We tried it with ExecuteWorkflow and users pushed back. They wanted the ability to start workflow only and do not wait for it. The same is here. Sometimes an RPC timeout is only for update start, not for waiting for its result.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+
+The gRPC API comprises two functions, `UpdateWorkflow` and
+`AwaitUpdateWorkflowResponse` to be added to the `WorkflowService` service
+definition. The former is used to invoke an update and follows the standard
+patterns for identifying a workflow execution and carrying arbitrary metadata
+(Header) and payloads (Payloads). The request type for this function includes a
+`wait_timeout` property indicating how long the client is willing to wait for a
+synchronous response once the request has been accepted by the system. This
+value can be set to zero indicating no such waiting should occur.  The
+`request_id` property uniquely identifies the logical request (_not_ the
```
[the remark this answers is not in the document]
mfateev: This is documentation of the gRPC API. It should document fields even if SDKs hide them.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mfateev: I think service can take min of the deadline and wait_timeout. Is the deadline causes a client to cut the request or it is purely implemented by the server side?

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mfateev: I would introduce UpdateSuccess and UpdateFailure protos to wrap these. We might need to add new fields to them later.

server/update-workflow-interface/update-workflow-interface.md
mfateev: The ability to reject signals was requested multiple times by users. We cannot do it for signals, but I think supporting it for updates is important. 

"StartUpdate" can happen even if a worker is down. The update will be rejected when the worker is back.

server/update-workflow-interface/update-workflow-interface.md
mfateev: @paulnpdev We do consider pipelining all the way to the workflow, but it is out of scope for this particular proposal.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+service WorkflowService {
+    rpc UpdateWorkflow(UpdateWorkflowRequest) returns (UpdateWorkflowResponse){}
+    rpc AwaitUpdateWorkflowResponse(AwaitUpdateWorkflowResponseRequest) returns (UpdateWorkflowResponse){}
+}
+
+message UpdateWorkflowRequest {
+    temporal.api.common.v1.Header header = 1;
+    string request_id = 2;
+    string namespace = 3;
+    temporal.api.common.v1.WorkflowExecution execution = 4;
```
bergundy, line 134: Let's add `first_execution_run_id` here to be able to update the last run in a chain.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
bergundy: Client must throw if deadline exceeds, it might never get a response from the server..

server/update-workflow-interface/update-workflow-interface.md
bergundy: We have the value in all our enum definitions but we never use it, it means what it says: UNSPECIFIED

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
bergundy: The client needs a way to specify the deadline because it might have limited time to wait but I recommend we always use the grpc deadline mechanism.

server/update-workflow-interface/update-workflow-interface.md
bergundy: AFAIR (with limited knowledge of the server architecture) we also mentioned that it'd reduce latency if we didn't have to write to history the way signals must be written.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mmcshane: I wanted to use the `wait_timeout` field to reinforce conceptually that there is a difference between "how long this RPC might run" and "how long I want to wait for an UpdateResponse." However it doesn't seem to be clarifying very much and may be having the opposite effect. I can proceed using just the built-in gRPC timeout and have that _implicitly_ be the UpdateResponse wait interval.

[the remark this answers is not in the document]
cretz: > Having one RPC call to send request and get response is critical for low-latency use case

I think it may be good to clarify the server logic here. Is the update-and-wait code not internally equivalent to calling request update then calling wait? If so, the only latency concern is a separate gRPC invocation? If I want really low latency, can I disable update validation?

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Answers for @cretz questions

1. Same behavior as any other pending task waiting for a Worker
2. Yes, it can and should be replayed
3. Great question and not one which I/we had spend any time considering. My gut instinct is to say that yes, workflows _can_ complete with "running" Updates. But that's an early instinct and the question needs more scrutiny.
4. Correct
5. Yes but as discussed, no mutation of state. The expectation is that many Update implementations will be of the conditional variety, e.g. "Do X if and only if state Y still has value Z" but that needs to be implemented by the workflow code.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Indeed it's exactly the Signal semantic - would be confusing otherwise. The exact fields we use are seen in the probobufs below.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mmcshane: Responded in a separate thread but for anyone browsing - I ended up agreeing with @bergundy here. I'm going to remove the `wait_timeout` from both messages.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+While the issuer of an update may wish to block on an UpdateResponse, network
+conditions or software failures may occur which forcibly disconnects the update
+issuer from the server at the time that the Update completes _and_ before the
+issuer becomes aware of the Update ID. The issuer must be able to re-issue the
+same UpdateRequest to access the server-generated Update ID.
+
+## gRPC API
+
+The gRPC API comprises two functions, `UpdateWorkflow` and
+`AwaitUpdateWorkflowResponse` to be added to the `WorkflowService` service
```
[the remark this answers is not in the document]
mmcshane: Sure, I'll change it

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+take more time than the issuer is willing to block the thread of execution
+invoking the update. In such cases the update issuer would prefer to not wait at
+all for the UpdateResponse, preferring instead to poll for it later.
+
+### Invalid Update
+
+UpdateRequests can be invalid given the state of the receiving workflow. In such
+cases the UpdateResponse will contain an error indication and payload. Invalid
+updates _must not_ be recorded in workflow history to prevent a malicious issuer
+from causing a denial of service on the workflow by issuing repeated invalid
```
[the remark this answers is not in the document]
mmcshane: You're right. I used the term "malicious user" when really a better phrasing would have been "byzantine client." I will update.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+UpdateRequests.
+
+### "Failing" Update
+
+UpdateRequests can be semantically acceptable but still fail due to external
+dependencies. In such cases the UpdateResponse will contain an error indication
```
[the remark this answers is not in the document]
mmcshane: Yes. If the UpdateRequest is accepted then the outcome will always be persisted, even errors. I will clarify

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+and payload.
+
+### Unexpected Disconnect
+
+While the issuer of an update may wish to block on an UpdateResponse, network
+conditions or software failures may occur which forcibly disconnects the update
+issuer from the server at the time that the Update completes _and_ before the
+issuer becomes aware of the Update ID. The issuer must be able to re-issue the
+same UpdateRequest to access the server-generated Update ID.
```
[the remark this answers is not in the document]
mmcshane: Yes.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: When I said "empty" WorkflowResponse I wasn't clear enough. It may be the case that the response contains no payload bytes but it will still carry the success/failure indication.

> some would argue that an attempted invalid entry IS part of the history

I actually agree with this argument but the baggage associated with an the pre-validated history entry is currently too high (latency, space). It's an engineering tradeoff but in principle I think we agree.

server/update-workflow-interface/update-workflow-interface.md
paulnpdev: > what it comes down to is whether we are okay allowing invalid updates to consume space on the history. We allow it for signals but at the onsite the consensus was that invalid updates must not eat space in the history. Are you willing to give ground on the history requirement? Or alternatively, do you see a way to identify invalid updates without running workflow code?
@mmcshane I agree.  I'm saying that there's a tradeoff here.  I consider this and ROI question. If allowing invalid updates to consume history space buys us enough (statistically) in terms of latency and/or throughput (and my intuitions says it does) then that's a cost I'm willing to pay, as long as we maintain correctness.

server/update-workflow-interface/update-workflow-interface.md
paulnpdev: @mfateev Roger that.  FYI, I wasn't trying to co-opt this proposal to include the pipelining feature, just to bring it to the attention of the people involved.  It's entirely possible that the code can be written in a way that leaves the door (more?) open to that if the author is aware of the potential need.  I know that if I was working on this layer I would rather know than not.

paulnpdev: > Not sure exactly what you're getting at with "durable connection" but most gRPC clients multiplex lots of messages onto "transport" instances (roughly: connections) which have a longer lifetime than the request-response of a single RPC.

@mmcshane  As max points out we're not considering pipelining "right away" but just for the record, I'm aware of the underlying mechanism you describe.  The overhead I want to eliminate (someday:) is the VERY expensive stack on top of the underlying TCP connection.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+    string namespace = 3;
+    temporal.api.common.v1.WorkflowExecution execution = 4;
+    string update_name = 5;
+    temporal.api.common.v1.Payloads input = 6;
+    google.protobuf.Duration wait_timeout = 7;
+}
+
+message UpdateWorkflowResponse {
+    UpdateResultType result_type = 1;
+    string update_id = 2;
```
yycptt, line 142: Once the `update_id` is returned to caller, does the sever guarantee it can recognize the ID later, if so for how long?  I guess it's related to if server needs to persist the update and ID before returning the ID.

server/update-workflow-interface/update-workflow-interface.md
cretz: @mmcshane - Thanks. We need to make clear to users that their validation code will need to be reevaluated at update time because validation time and update time can differ by a long time and workflow state may have changed since then. (or maybe we expect SDKs to run validation code twice or whatever...ug, I am not a fan of this special validation step at all)

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Yes. The update_id is only issued after the workflow as accepted the UpdateRequest and is made durable in the workflow history.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,217 @@
...
+### Update
+
+An Update is the execution of arbitrary workflow logic (subject to the standard
+workflow code constraints on determinism) triggered by an UpdateRequest and
+eventually resulting in a single UpdateResponse. The processing of an Update is
+performed by workflow code and as such the timeliness of the UpdateResponse is
+dependent on worker resources being available to process the Update. A given
+Update applies to exactly one workflow execution. The lifetime of an updated
+begins at the first UpdateRequest triggering said Update and ends when the
+UpdateResponse has been made durable. Consequently, the lifetime of an Update
```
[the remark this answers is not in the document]
mmcshane: Good catch. I messed up the wording for the "begin" part. I updated to define the _start_ of an Update being when the workflow accepts/validates the UpdateRequest. With this definition change, the UpdateResponse will always be persisted.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Added UPDATE_WORKFLOW_RESULT_TYPE_IN_PROGRESS to avoid overloading UNSPECIFIED.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: This is a good point that I had neglected to deal with. I went back through the use cases and ended up in a slightly different place. In the latest version, UpdateWorkflow will always be (in a rough sense) fast. The client either wants the response "inline" with the gRPC response, in which case we clamp the timeout to a couple of seconds at max, or the client only waits for update_id and will poll later on. I think/hope this lets us get away with just the one UpdateWorkflow call.

As for metrics, given that UpdateWorkflow will invoke workflow code in all cases, my inclination is to exclude this function from latency metrics as it is out of Temporal's control.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mmcshane: > why not have the client generate the update ID?

The nice thing about server-generated UpdateID is that if the UpdateID exists then it the Update was definitely accepted and there will definitely be a result. So polling on that ID should never get an "I've never heard of that ID" type response. We can also reasonably guarantee uniqueness of IDs on the server side.

That being said, using the request_id as the UpdateID is attractive in that it makes the idempotency filtering easier - we wouldn't need separate request_id tracking or a request_id -> update_id mapping.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,249 @@
...
+
+enum UpdateWorkflowResultType {
+    UPDATE_WORKFLOW_RESULT_TYPE_UNSPECIFIED = 0;
+    UPDATE_WORKFLOW_RESULT_TYPE_IN_PROGRESS = 1;
+    UPDATE_WORKFLOW_RESULT_TYPE_SUCCESS = 2;
+    UPDATE_WORKFLOW_RESULT_TYPE_FAILURE = 3;
+}
+
+message PollUpdateResponseRequest {
+    string update_id = 1;
```
yycptt, line 177: I am not quite sure how server is able to route the poll request to a specific history host, if needed, or find the workflow containing the update response just from the `update_id`. 

If the `update_id` contains extra information like target ns, workflowID, runID, would it be better to return an opaque binary token instead of a string?

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Made this more clear in the wording :+1:

server/update-workflow-interface/update-workflow-interface.md
mmcshane: There's not a huge difference at that point but I can think of two reasons to prefer Update:

1. As you suggest, the user might prefer the Update mechanism for accessing the UpdateResponse
2. The user might prefer Update's ability to avoid spending history space for rejected UpdateRequests

server/update-workflow-interface/update-workflow-interface.md
yycptt: Thanks. Does that mean for each update request, there will be two persistence operation? First when accepting the request, then when the update response is available and both will be persisted in the workflow being updated.  

That also means returned update_id will no longer be valid after the target workflow's retention period (if archival is not enabled)?

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
mmcshane: > Is the update-and-wait code not internally equivalent to calling request update then calling wait?

The intent is for those two approaches to be functionally equivalent. If I've made it seem otherwise then I messed something up.

> If I want really low latency, can I disable update validation?

This is rather a longer conversation but it's the closes thing to having our cake and eating it too. If we were somehow able to inform the history service that a given Update path has no validation for all running versions then we could optimize away the validate/execute two-step. As it stands AFAIK the sever has no understanding of the structure of Queries/Signals supported by a workflow (maybe I'm wrong on this) but in theory we could generate an IDL-like structure at runtime so that the server could be aware of the existence of validation (along with other things like the name bindings of Update functions)

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Ended up with nullable UpdateSuccess and UpdateFailure types/fields with a discriminator enum. While the enum is redundant it seems consistent with existing practices to have it.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Yeah, that's a different discussion. Not sure which aspects of the stack above TCP are of greatest concern to you but personally I'd include TCP itself in the assessment.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: > ... would it be better to return an opaque binary token instead of a string?

I think you're right. I'll switch to `google.protobuf.BytesValue` or a `temporal.api.common.v1.Payload`

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Prompt? Troll?

I'm thinking about it but haven't come up with anything that's obviously better than update yet.

server/update-workflow-interface/update-workflow-interface.md
yycptt: Or simply `bytes`? Looks like most tokens we returned have `bytes` type, so might be good to be consistent? e.g. https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L258

I don't have strong opinion though.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: Ideally the SDKs can do the validate+execute in the same activation so that the response back to the server can persist the UpdateResponse and UpadateID at the same time, resulting in a single I/O hit.

And yes, the UpdateID will be associated with the workflow so anything that causes the workflow data to age-out will also be getting rid of all the associated UpdateResponses.

server/update-workflow-interface/update-workflow-interface.md
mmcshane: I'd prefer the wrapper types but since we're already using raw `bytes` elsewhere I don't think the benefit is enough to justify being inconsistent.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
cretz: To clarify, I don't mind not having such a skip-validation option, but I just thought it was strange we're afraid of two gRPC calls due to low-latency, but we're ok with two worker calls.

server/update-workflow-interface/update-workflow-interface.md
cretz: Marking unresolved just to get my comment seen (feel free to mark resolved after).

While I understand this is a server provided identifier, I wonder if there are other concerns with packing a custom binary encoded value into the identifier. But even if we do, arguably a UI-visible identifier has value anyways. Does one expect a large blob to be base64'd by the frontend here? Or in logs when one is handle the update and wants to uniquely identify it?

I don't mind leaving as is (sounds like we may have to if we're not willing to do state lookup on our side), just want to make sure we consider the other benefits of short, readable, arbitrary identifiers over encoded blob identifiers. (feel free to resolve conversation)

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
cretz: After further discussion, I have been told two important things:

1. We will not require a worker be running a workflow to accept an update from a client
2. We will not separate the update validation and update execution steps across a server call boundary, only worker-local separation if any

Given these, all my concerns go away (can resolve this conversation, leaving open for my response to be visible).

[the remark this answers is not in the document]
cretz: I no longer have a strong opinion on whether this individual call is a long polling call. Resolving.

server/update-workflow-interface/update-workflow-interface.md
cretz: After further discussion, I have been told two important things:

    We will not require a worker be running a workflow to accept an update from a client
    We will not separate the update validation and update execution steps across a server call boundary, only worker-local separation if any

Given these, all my concerns go away (can resolve this conversation, leaving open for my response to be visible).

samarabbas: Although most of the discussion here was around preventing execution history from flood of invalid updates.  But the main thing we need to make sure we can process the update with a single write during `happy path`.  The ability to dispatch the update without any write and only perform the actual write when update is successfully processed is very essential to update use cases which are typically low latency.  Our ability to skip the history write when update is rejected a nice side effect of the above behavior.

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,253 @@
...
+    string update_name = 5;
+    temporal.api.common.v1.Payloads input = 6;
+    UpdateWorkflowResultAccessStyle result_access_style = 7;
+}
+
+message UpdateWorkflowResponse {
+    bytes update_id = 1;
+    UpdateWorkflowResultType result_type = 2;
+    UpdateWorkflowSuccess success = 3;
+    UpdateWorkflowFailure failure = 4;
```
bergundy, line 160: nit: Consider calling this `failed` and replacing `output` below with `failure`.
We've used this pattern in core and it reads pretty well: `reponse.failed.failure`

server/update-workflow-interface/update-workflow-interface.md
```diff
@@ -0,0 +1,253 @@
...
+part of the gRPC respons and if that same gRPC response indicates that the
+Update is still in progress (i.e. `UpdateWorkflowResponse.result_type ==
+UPDATE_WORKFLOW_RESPONSE_TYPE_IN_PROGRESS`), the client can then use
+`PollUpdateWorkflowResponse` to wait for Update completion.
+
+### Invalid Update
+
+If the UpdateRequest is rejected by the workflow execution while the RPC client
+is still connected, a suitable gRPC error can be returned synchronously. Note
+however rejected UpdateRequests are _not_ recorded in workflow history.
```
bergundy, line 219: Arguably they should be recorded in mutable state though in order to deliver the response asynchronously.

server/update-workflow-interface/update-workflow-interface.md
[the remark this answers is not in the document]
lorensr: Having an `update_id` in addition to the idempotency key also seemed weird to me. And seems to me like costs (complexity, both in caller's head and the server implementation) might outweigh the benefits.

> The nice thing about server-generated UpdateID is that if the UpdateID exists then it the Update was definitely accepted and there will definitely be a result. 

The client can also determine "the Update was definitely accepted and there will definitely be a result" based on the received `UpdateWorkflowResponse`. (Assuming they can differentiate between "didn't pass validation" and "passed validation and update handler returned an error"? I wasn't clear on that from the proposal.)

> We can also reasonably guarantee uniqueness of IDs on the server side.

It seems like the most important thing either way is uniqueness of request id? If I generate request_id foo for a request now, and happen to generate the same request_id months later, I will get the result from the first request, regardless of whether update_id exists and its uniqueness. (That's if you just use request_id for idempotency. If you also use other info, like workflow id and input, then in my scenario, assume those match as well, eg updating an infinite workflow.)

PR #54, opened 2022-03-17
Worker build id based versioning

Sushisource:
[Rendered](https://github.com/Sushisource/proposals/blob/worker-build-id-versions/versioning/worker-versions.md)

This is take two on https://github.com/temporalio/proposals/pull/52 which I will close if/when we decide we prefer this.

### DIFF
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -1,4 +1,4 @@
-# User-specified Worker Versions
+# Worker Build ID Based Versioning
 
 Temporal workflow authors experience a number of difficulties when it comes to
 changing workflow code. The current techniques for versioning workflows are:
@@ -41,10 +41,11 @@ fixes are typically less common than just making changes to workflow logic, thou
 
 To make the more common case easier, we will provide a set of features that
 make the "versioned task queues" approach much easier to use for changes to
-workflows which break history compatability (but not interface compatability).
+workflows which break history compatability (but not interface compatability),
+or would like to safely roll out expected-to-be-compatible changes.
 
 At a high level:
-* Provide a simple API for versioning entire workers (this proposal)
+* Provide a simple methodology for versioning entire workers (this proposal)
 * Allow clients to point at task queue `foo`, and automatically route new workflow executions
   to the latest version of the `foo` queue.
 * Provide good notifications and information about when old workers may be
@@ -56,84 +57,118 @@ Building on this base set of features, we plan to eventually implement
 automatic safe-rollout of changes, and possibly tools like k8s operators
 which can automatically manage worker deployments.
 
-## SDK API changes
+## Safe-rollout
 
-The meat of this proposal is simple:
-**Add a version field to worker options in each SDK**
+Intentionally avoiding deep into details here, the foundation laid here should enable
+us, as well as users, to more easily implement automated and safe rollout
+of changes to workers. A process like this one can become the de-facto experience:
 
-Because version information will need to be sent to the Temporal Cluster,
-we define it as a protobuf message (which will show up in our API somewhere):
+* If deploying a known-breaking change, new workflows are only processed by
+  the new version, existing open workflows are processed by old workers.
+* If deploying an expected-to-be-compatible change, new workflow _tasks_ are
+  attempted by new workers, and if they succeed, that workflow will subsequently
+  only be processed by new workers. If they fail, execution goes back to old
+  workers.
+
+# Using opaque whole-worker-identifiers as worker versions
+
+Thanks to @jlegrone for his great feedback encouraging us to revive this approach.
+
+We already have a field called `binaryChecksum` (or similar) in `WorkerOptions`
+today in SDKs. This field's intent is to uniquely identify a specific worker
+build. Since we believe versioning entire workers is the easiest solution, we
+can lean on this existing field.
+
+Deprecating the name `binaryChecksum`and replacing it with something more 
+accurate like `workerBuildId` would be appropriate, since most of our SDKs
+don't even produce a binary.
+
+With this approach, users who wish to make use of the improved versioning 
+features should set the `workerBuildId` during their CI process to a value that
+is unique for that worker's implementation. _Implementation_ rather than 
+_build_ is an important distinction here, since rebuilds of identical
+code should (ideally) produce workers wit the same ID. It is not necessary to
+do this, and it is varyingly difficult depending on the language, but when
+done it enables users to produce workers with the same ID from their CI builds
+if they didn't touch worker code, but something else (ex: The CI definition itself).
+This involves some work on the part of the user, but is more reliable than expecting
+worker authors to increment a version number.
+
+Unfortunately it is infeasible to automatically set an appropriate ID for many languages.
+
+## How is versioning accomplished with the `workerBuildId`?
+
+Every new worker ID will be assumed to be incompatible with the existing id(s)
+unless explicitly specified otherwise. New workers will not process *anything* 
+until explicitly enabled if they opt into build-id based versioning. Users who 
+wish to always have new workers begin processing new workflows can make an API 
+call immediately establishing the new worker as latest as part of their 
+deployment process (see below).
+
+In order to either (safely or not) migrate *open* workflows to a new version,
+users will explicitly indicate that an ID is compatible with some already-extant
+ID.
+
+Because this behavior is a breaking change, SDKs will need to add a
+`enableBuildIdBasedVersioning` boolean flag to their `WorkerOptions`.
+
+## Server-side APIs for `workerBuildId`-based versioning
+
+Importantly, we _already_ send the id (checksum, today) on every poll request.
+It's TBD if the worker will be changed to poll a unique queue per ID, or if
+that will happen on the frontend, or matching.
+
+These APIs could very well end up in a different service, but are shown
+as part of `WorkflowService` here for simplicity.
 ```protobuf
-message WorkerVersions {
-    UserVersion worker_version = 1;
+service WorkflowService {
+    rpc SetWorkerBuildIdOrdering (SetWorkerBuildIdOrderingRequest) returns (SetWorkerBuildIdOrderingResponse) {}
+    // This could / maybe should just be part of `DescribeTaskQueue`, but is broken out here to show easily.
+    rpc GetWorkerBuildIdOrdering (GetWorkerBuildIdOrderingRequest) returns (GetWorkerBuildIdOrderingResponse) {}
 }
 
-message UserVersion {
-    // The string the user passed as their version representation
-    string original = 1;
-    // Parsed from original string
-    uint32 major = 2;
-    // Parsed from original string
-    uint32 minor = 3;
+message SetWorkerBuildIdOrderingRequest {
+    string namespace = 1;
+    // Must be set, the task queue to apply changes to.
+    string task_queue = 2;
+    // The build id we are targeting.
+    string worker_build_id = 3;
+    // When set, indicates that the `workerBuildId` in this message is compatible
+    // with the one specified in this field. Because compatability should form
+    // a DAG, any build id can only be the "next compatible" version for one
+    // other ID at a time, and any setting which would create a cycle is invalid.
+    string is_compatible_with = 4;
+    // When set, establishes the specified `workerBuildId` as the latest
+    // for the queue. Those workers will begin processing new workflow executions.
+    bool latest = 5
 }
-```
+message SetWorkerBuildIdOrderingResponse {}
 
-This message is pre-emptively named plurally in order to accommodate potential
-increases in versioning granularity in the future. For example, we have considered
-versioning Workflows, Activities, and Data Converters/Interceptors all separately.
-They could be added as new fields to this message if/when that happens.
-
-Each language will expose a way to (optionally) set the worker version to 
-start, supporting the possibility of other versions later. This will translate 
-into the above message to be sent to server.
-
-### TypeScript
-
-```typescript
-Worker.create({ workerVersion: '1.0', ...otherOptions })
-```
-
-### Go
-
-```go
-workerOptions := worker.Options{
-  WorkerVersion: "1.0",
-  // other options
+message GetWorkerBuildIdOrderingRequest {
+    string namespace = 1;
+    // Must be set, the task queue to interrogate about worker id ordering
+    string task_queue = 2;
+    // Limits how deep the returned DAG will go. 1 will return only the
+    // latest build id.
+    uint32 max_depth = 3;
 }
+message GetWorkerBuildIdOrderingResponse {
+    // The currently established latest worker build id
+    WorkerBuildIdNode current_latest = 1;
+    // Other current latest-compatible versions who are not the overall latest
+    repeated WorkerBuildIdNode compatible_leaves = 2;
+}
+
+message WorkerBuildIdNode {
+    string worker_build_id = 1;
+    WorkerBuildIdNode compatible_with = 2;
+    WorkerBuildIdNode previous_incompatible = 3;
+}
+
+// Additionally, some way to limit version DAG depth is necessary - likely
+// something settable via the Operator service.
+
 ```
 
-### Java
-
-```java
-WorkerOptions.newBuilder().setWorkerVersion("1.0").build()
-```
-
-## When and how to increment the worker version number
-
-The information here will be added as a documentation page once this feature
-has shipped.
-
---
-
-The worker version number encompasses changes to all workflow code registered
-with the worker, as well as changes to any code which might affect the path
-of execution within workflow code.
-
-#### Super-Major
-Require a new workflow type name (or new task queue name)
-
-- A backwards-incompatible change to the workflow interface
-    - Changing a required parameter type
-    - Adding or removing a required parameter
-
-
-#### Major
-
-- A backwards-incompatible change to the workflow implementation. See more [here](https://docs.temporal.io/docs/concepts/what-is-a-workflow-definition/#deterministic-constraints)
-    - Anything that would cause the workflow to fail replay due to nondeterminism, given any possible valid history from the unaltered version of the workflow. 
-    - Includes changes to data converters or interceptors which would result in such incompatibilities.
-
-#### Minor
-
-- Backwards compatible change to workflow implementation
-    - This includes adding properly implemented `getVersion` or `patch` calls, for fixing bugs.
+---
+This proposal is a second take / alternative to the one [here](https://github.com/temporalio/proposals/pull/52)
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -105,3 +105,33 @@ workerOptions := worker.Options{
 ```java
 WorkerOptions.newBuilder().setWorkerVersion("1.0").build()
 ```
+
+## When and how to increment the worker version number
+
+The information here will be added as a documentation page once this feature
+has shipped.
+
+--
+
+The worker version number encompasses changes to all workflow code registered
+with the worker, as well as changes to any code which might affect the path
+of execution within workflow code.
+
+#### Super-Major
+Require a new workflow type name (or new task queue name)
+
+- A backwards-incompatible change to the workflow interface
+    - Changing a required parameter type
+    - Adding or removing a required parameter
+
+
+#### Major
+
+- A backwards-incompatible change to the workflow implementation
+    - Anything that would cause the workflow to fail replay due to nondeterminism, given any possible valid history from the unaltered version of the workflow. 
+    - Includes changes to data converters or interceptors which would result in such incompatibilities.
+
+#### Minor
+
+- Backwards compatible change to workflow implementation
+    - This includes adding properly implemented `getVersion` or `patch` calls, for fixing bugs.
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -49,6 +49,8 @@ At a high level:
   to the latest version of the `foo` queue.
 * Provide good notifications and information about when old workers may be
   decommissioned
+* Ensure histories replayed on older versions fail fast with informative errors
+* Easy visibility into what versions deployed workers are using
 
 Building on this base set of features, we plan to eventually implement
 automatic safe-rollout of changes, and possibly tools like k8s operators
--- a/versioning/worker-versions.md
+++ b/versioning/worker-versions.md
@@ -1,0 +1,97 @@
+# User-specified Worker versions
+
+Temporal workflow authors experience a number of difficulties when it comes to
+changing workflow code. The current techniques for versioning workflows are:
+
+**GetVersion & patch APIs**
+These APIs allow branching within the same workflow code to handle the presence
+or absence of a certain change. See them for [TS](https://docs.temporal.io/docs/typescript/patching#typescript-sdk-patching-api), [Java](https://docs.temporal.io/docs/java/versioning/), and [Go](https://docs.temporal.io/docs/go/versioning/).
+
+These APIs are confusing and challenging to use correctly, and clutter up code.
+We advise users to generally use them only to fix bugs that would otherwise prevent
+the workflow from completing.
+
+**Versioned task queues**
+Conceptually simpler, but operationally more effort, is to version task queues.
+IE: A worker operating on task queue `foo-v1` is changed, and when redeployed
+points at `foo-v2`.
+
+This avoids complicating the workflow code itself, but means users need to be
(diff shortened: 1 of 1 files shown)


Sushisource:
Tagging everyone who contributed to the last proposal:
@dnr @cretz @lorensr @jlegrone @mmcshane @bergundy @yiminc

bergundy (approved):
Great proposal.
There might be some slight bikeshedding and some enhancements to the proposed API but other than that I think this is the way to go.

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+Unfortunately it is infeasible to automatically set an appropriate ID for many languages.
+
+## How is versioning accomplished with the `workerBuildId`?
+
+Every new worker ID will be assumed to be incompatible with the existing id(s)
+unless explicitly specified otherwise. New workers will not process *anything* 
+until explicitly enabled if they opt into build-id based versioning. Users who 
+wish to always have new workers begin processing new workflows can make an API 
+call immediately establishing the new worker as latest as part of their 
+deployment process (see below).
```
lorensr, line 106: Trying to envision the easiest user flow.

So in TS, my CI would:

1. Deploy worker code
2. Run a script that:
  - Asks TS SDK what the `workerBuildId` is for the same `WorkerOptions` I just deployed
  - Calls `SetWorkerBuildIdOrdering` without `is_compatible_with` (or SDK can do this as well)

And if I know that it's actually a compatible change, I manually call `SetWorkerBuildIdOrdering` to correct CI's call, with an `is_compatible_with` that I find in `tctl tq describe --tq foo`, which has two new columns, `BUILD_ID` and `FIRST_ACCESS_TIME` (so I know when I deployed it).

How does that sound?

We could avoid step 2 and simplify by deploying code that had an option like:

```ts
Worker.create({
  ...opts,
  useAsLatestVersion: true
})
```

bergundy:
We can generalize this proposal to tag based matching.

For example, a client starts a workflow tagged `version=latest`.
A worker registers with `buildId=deadbeaf`.
After worker deployment, a client calls `RewriteTaskQueueTag('version=latest', 'buildId=deadbeaf')` and new tasks for latest get routed to the new deployment.

Similarly, to set a `patchOf` relationship, you would call `RewriteTaskQueueTag('buildId=deadbeaf', 'buildId=eddaebfa')`.

Even workflow bundles would work this way.
A worker could register with `workflowRuntime=1`, and use rewrite rules like:

- `version=latest` -> `bundleId=deadbeaf`
- `bundleId=deadbeaf` -> `bundleId=eddaebfa`
- `bundleId=deadbeaf` -> `workflowRuntime=1`

Since we already have search attributes, we might be able to use those for tags.

versioning/worker-versions.md
Sushisource: Yeah, that's about right

> Asks TS SDK what the workerBuildId is for the same WorkerOptions I just deployed

In principle you may already know what it is, since you're deploying something you built and you set the ID when building, you might just pass along that info to the deploy job - but it's a good point that we want an easy way to "ask" a given worker what it's build ID is.

And yeah, the avoid step 2 part, I think that can pretty easily be done by just having your worker app make that call at the same time it starts up the worker if that's the behavior you want. We can indeed make it a first-class option if people are doing this constantly.

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+
+Importantly, we _already_ send the id (checksum, today) on every poll request.
+It's TBD if the worker will be changed to poll a unique queue per ID, or if
+that will happen on the frontend, or matching.
+
+These APIs could very well end up in a different service, but are shown
+as part of `WorkflowService` here for simplicity.
+```protobuf
+service WorkflowService {
+    rpc SetWorkerBuildIdOrdering (SetWorkerBuildIdOrderingRequest) returns (SetWorkerBuildIdOrderingResponse) {}
```
lorensr, line 125: If I call this with `build_id=new, latest=true`, and I want `new` to no longer be latest, can I call `build_id=new, latest=false` and the previous latest will be used?

Wonder if we can automate certain types of rollback scenarios, like if a build id is failing all tasks, or not polling, auto roll back to previous latest.

Sushisource: You can do it with the existing API by following the link back to the previous version, and then setting that one as latest.

But, perhaps that potentially common situation should be made easier. I don't think we want to do it exactly the way you propose since simply _not setting_ latest and explicitly setting it to false look the same to protobuf and that makes this situation confusing. 

There is already an API for marking a binary checksum as "bad" which you'd probably want to make do what you're talking about. Unfortunately that information lives only on the frontend currently and there's some internal stuff we'd have to work through, but in principle that'd be a sensible place to make do this.

versioning/worker-versions.md
lorensr: > you may already know what it is, since you're deploying something you built and you set the ID when building

Ah okay, I was thinking it was still generated based on code like the checksum.

> And yeah, the avoid step 2 part, I think that can pretty easily be done by just having your worker app make that call at the same time it starts up the worker if that's the behavior you want.

I wouldn't want to just call `SetWorkerBuildIdOrdering` on start, since old deployed workers might get restarted, but I could do:

```
isNewlyDeployed = !GetWorkerBuildIdOrdering()
if isNewlyDeployed
  SetWorkerBuildIdOrdering()
```

versioning/worker-versions.md
cretz, line 125: Can this only be run once per worker build ID or is it expected to be able to be repeatedly called to make changes? Can I unset latest or change compatibility? Or should it error if I try to change anything (but accept same-parameters since this may be called on each worker deploy)?

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+    string namespace = 1;
+    // Must be set, the task queue to apply changes to.
+    string task_queue = 2;
+    // The build id we are targeting.
+    string worker_build_id = 3;
+    // When set, indicates that the `workerBuildId` in this message is compatible
+    // with the one specified in this field. Because compatability should form
+    // a DAG, any build id can only be the "next compatible" version for one
+    // other ID at a time, and any setting which would create a cycle is invalid.
+    string is_compatible_with = 4;
```
cretz, line 140: So each worker ID can only be compatible with one other?

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+    string task_queue = 2;
+    // Limits how deep the returned DAG will go. 1 will return only the
+    // latest build id.
+    uint32 max_depth = 3;
+}
+message GetWorkerBuildIdOrderingResponse {
+    // The currently established latest worker build id
+    WorkerBuildIdNode current_latest = 1;
+    // Other current latest-compatible versions who are not the overall latest
+    repeated WorkerBuildIdNode compatible_leaves = 2;
```
cretz, line 159: Can you clarify this a bit? Say I have the following worker build IDs for a task queue:

* `foo` - compatible with `old_foo1`, latest=true
* `old_foo1` - not compatible with anything
* `bar` - compatble with `old_bar1`
* `old_bar1` - not compatible with anything

How do I retrieve `bar`? Is that what shows up here?

versioning/worker-versions.md
Sushisource: You can call it more than once and it'll reflect the latest write.

versioning/worker-versions.md
Sushisource: Yes. When dispatching, we will always go all the way out to the leaf to run on the "latest compatible" worker

versioning/worker-versions.md
Sushisource: `bar` will be in `compatible_leaves` , and have a pointer back to `old_bar1`

You raise a good point though that if I send an ordering request for something that is not compatible with anything, and also not latest, that should be an error since it's just going to be some free-floating node.

versioning/worker-versions.md
cretz: So if I call it with `latest` unset after calling with `latest=true`, does that mean there is no latest anymore?

Sushisource: No -- it'll be an error since you didn't change anything. Per https://github.com/temporalio/proposals/pull/54#discussion_r830183599 I think we would want to be very explicit about "unsetting" latest by setting it to the previous (or whatever other) version.

Sushisource:
@bergundy As discussed on zoom, I think a more generic relationships graph is a good way to ultimately implement this, so implementation details wise we can make an interface that looks like this and build the functionality on it. Potentially extend usage of it later for other features.

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+wish to always have new workers begin processing new workflows can make an API 
+call immediately establishing the new worker as latest as part of their 
+deployment process (see below).
+
+In order to either (safely or not) migrate *open* workflows to a new version,
+users will explicitly indicate that an ID is compatible with some already-extant
+ID.
+
+Because this behavior is a breaking change, SDKs will need to add a
+`enableBuildIdBasedVersioning` boolean flag to their `WorkerOptions`.
```
yiminc, line 113: Could this be implied by the presence of a non empty workerBuildId?

versioning/worker-versions.md
yiminc, line 125: Should this SetWorkerBuildIdOrdering be part of Operator API? Each worker poll request should contains the workerBuildId info, but the ordering only need to be set by operator once per each workerBuildId.

versioning/worker-versions.md
```diff
@@ -0,0 +1,174 @@
...
+    // The build id we are targeting.
+    string worker_build_id = 3;
+    // When set, indicates that the `workerBuildId` in this message is compatible
+    // with the one specified in this field. Because compatability should form
+    // a DAG, any build id can only be the "next compatible" version for one
+    // other ID at a time, and any setting which would create a cycle is invalid.
+    string is_compatible_with = 4;
+    // When set, establishes the specified `workerBuildId` as the latest
+    // for the queue. Those workers will begin processing new workflow executions.
+    bool latest = 5
```
yiminc, line 143: maybe use the concept of default instead of latest? you may want to use a non-latest as default.

versioning/worker-versions.md
Sushisource: Yeah, it could be. I'm worried though it's maybe a bit non-obvious. Plus, if we deprecate the old checksum, you might want to set this but not actually opt in to the versioning behavior.

versioning/worker-versions.md
Sushisource: I think that's reasonable. I don't have a strong preference one way or another.

Sushisource: There's no way to distinguish between you setting it false explicitly, and it being unset as a protobuf, so like most protobuf bools it should only have an effect when set true

cretz: There are many ways to handle this. You could use `google.protobuf.BoolValue` or you could change the name to `move_to_latest` or anything of the sort.

I don't think it matters that much of course. It's just the concern that happens when sharing a always-important readable value with an only-sometimes-important writable value.

Sushisource:
@bergundy 

I made a generic version per your proposal. I feel it complicates things a bit too much. It's easier to use incorrectly, and the generalization can be realized at a later point if necessary. Supporting workflow bundles in the existing proposal is a pretty simple matter of just transforming the `worker_build_id` fields to a `oneof` which is either build id or bundle id.

The below thing works, but still requires some specific customizations for efficiency/usability. For example `GetWorkerBuildIdOrdering` in the current proposal requires adding some kind of "special query" or filter type to the below, which can return all the build id version leaves. Without that, the caller has to be returned and iterate through the whole graph. It's certainly extensible but feels pretty YAGNI to me right now.

# Task Queue Metadata Graph
Here is a more generic approach which supports the versioning use case as well as many others.
However it must be limited to relatively small numbers of links and nodes unless ported to 
a more appropriate form of persistence.

Right now I would expect this graph to end up inside of [here](https://cs.github.com/temporalio/temporal/blob/8433f9f43709e09d03a6dc91430c2faceb0249e7/proto/internal/temporal/server/api/persistence/v1/tasks.proto?q=TaskQueueInfo#L48)
which would require parsing/repacking the graph on each update. Storing in ES I don't think is
an option since versioning needs consistent updates, and should probably work without advanced
visibility.

```protobuf
syntax = "proto3";

service ServiceTBD {
  rpc UpdateTaskQueueMetadata(UpdateTaskQueueMetadataRequest) returns (UpdateTaskQueueMetadataResponse) {}
  rpc GetTaskQueueMetadata(GetTaskQueueMetadataRequest) returns (GetTaskQueueMetadataResponse) {}
}

message UpdateTaskQueueMetadataRequest {
  string task_queue = 1;
  // Insert new relationships in this list
  repeated Node add_links = 2;
  // Remove any relationships in this list. For items in this list if the node is specified but not
  // links, delete that node and all links from/to it.
  repeated Node remove_links = 3;
}
message UpdateTaskQueueMetadataResponse {

}
message GetTaskQueueMetadataRequest {
  NodeId root = 1;
  uint32 max_depth = 2;
}
message GetTaskQueueMetadataResponse {
  // Root from request is guaranteed to be the first node
  repeated Node nodes = 1;
}

message Node {
  NodeId data = 1;
  repeated OutLink links = 2;
}
message OutLink {
  NodeId to = 1;
  Relationship relationship = 2;
}

message NodeId {
  oneof id {
    SpecialNodeType special = 1;
    string worker_build = 2;
    string workflow_bundle = 3;
  }
}

enum SpecialNodeType {
  UNSPECIFIED = 0;
  LATEST_BUILD_ANCHOR = 1;
}

enum Relationship {
  // Invalid to be unset
  UNSPECIFIED = 0;
  // The from and to nodes are worker build IDs (or workflow bundles) and the from node is newer than &
  // compatible with the to tag
  PREVIOUS_COMPATIBLE = 1;
  // The from and to nodes are worker build IDs (or workflow bundles) and the from node is newer than &
  // *not* compatible with with the to tag
  PREVIOUS_INCOMAPTIBLE = 2;
}
```

cretz:
@Sushisource - I like that API, but I think it can be even simpler (almost what you have in-proposal here). I question whether we really need relationship or more than one type of node identifiers. My thoughts:

```proto
service ServiceTBD {
  rpc UpdateVersions(UpdateVersionsRequest) returns (UpdateVersionsResponse) {}
  rpc GetVersions(GetVersionsRequest) returns (GetVersionsResponse) {}
}

message UpdateVersionsRequest {
  // Add or update
  repeated Version set_versions = 1;
  repeated string remove_versions = 2;
  repeated VersionLink add_compatible_versions = 3;
  repeated VersionLink remove_compatible_versions = 4;

  message VersionLink {
    string version_id = 1;
    string compatible_version_id = 2;
  }
}

message UpdateVersionsResponse {
}

message GetVersionsRequest {
  // Optional
  string compatible_with_version_id = 1;
  uint32 max_depth = 2;
}

message GetVersionsResponse {
  // Latest of each requested version guaranteed (then all compatible with up
  // to max depth)
  repeated Version versions = 1;
}

message Version {
  string id = 1;
  string task_queue = 2;
  // Setting this to true sets latest as false for all others this is compatible
  // with
  bool latest = 3;
  // Bidirectional
  repeated string compatible_with = 2;
}
```

I am not familiar w/ what was discussed on Zoom, but I always favor simplicity :-)

Sushisource:
@cretz Yes, that's very similar to the written proposal. I also prefer it for it's specificity and simplicity over a more general backing store. The more general store can be used to support this and exposed later if it's needed for other features down the line.

bergundy:
FTR a simplified general purpose link API could like like this:

```
service ServiceTBD {
  rpc UpdateLinks(UpdateLinksRequest) returns (UpdateLinksResponse) {}
  rpc GetLinks(GetLinksRequest) returns (GetLinksResponse) {}
}

message Tag {
  string key = 1;
  string value = 2;
}

message Link {
  Tag from = 1;
  Tag to = 2;
}

message UpdateLinksRequest {
  string task_queue = 1;
  // Add or update
  repeated Link set_links = 2;
  repeated Link remove_links = 3;
}

message UpdateLinksResponse {
}

message GetLinksRequest {
  string task_queue = 1;
  // Optional
  Tag starting_tag = 2; // default to latest
  uint32 max_depth = 3;
}

message GetLinksResponse {
  repeated Link links = 1;
}
```

bergundy:
> Right now I would expect this graph to end up inside of [here](https://cs.github.com/temporalio/temporal/blob/8433f9f43709e09d03a6dc91430c2faceb0249e7/proto/internal/temporal/server/api/persistence/v1/tasks.proto?q=TaskQueueInfo#L48)
which would require parsing/repacking the graph on each update. Storing in ES I don't think is
an option since versioning needs consistent updates, and should probably work without advanced
visibility.

We talked about this, it's no different from how you'd store version specific information in the current proposal.
There's no need to store this in ES.


> The below thing works, but still requires some specific customizations for efficiency/usability. For example GetWorkerBuildIdOrdering in the current proposal requires adding some kind of "special query" or filter type to the below, which can return all the build id version leaves. Without that, the caller has to be returned and iterate through the whole graph. It's certainly extensible but feels pretty YAGNI to me right now.

We could add another flag to only resolve the leaves here if there's demand for that.

bergundy:
In the above simplified proposal `Tag` could just refer to search attributes in order to reuse existing concepts.

Sushisource:
> FTR a simplified general purpose link API could like like this:

How do you distinguish previous-incompatible vs. compatible relationships with this API?

> Tag could just refer to search attributes in order to reuse existing concepts.

But now you're back to storing things in ES

Sushisource:
Per discussion w/ @bergundy I've updated the API to account for other future version types like workflow bundles. The way the worker-build-id things would work is effectively unchanged, but this gives us some flexibility in the future.

bergundy:
I'm fine with going with the narrower approach as you've suggested @Sushisource.

To respond to your last comments though:

> But now you're back to storing things in ES

I never said we'd store the links there, just the tags, which you'll end up doing anyways because this data needs to be searchable.

> How do you distinguish previous-incompatible vs. compatible relationships with this API?

I'm not sure I understand your question.
If you don't have a link from one tag to another they're incompatible.

versioning/worker-versions.md
```diff
@@ -0,0 +1,191 @@
...
+    // When set, indicates that the `version_id` in this message is compatible
+    // with the one specified in this field. Because compatability should form
+    // a DAG, any build id can only be the "next compatible" version for one
+    // other ID of a certain type at a time, and any setting which would create a cycle is invalid.
+    VersionId is_compatible_with = 4;
+    // When set, establishes the specified `version_id` as the default of it's type
+    // for the queue. Workers matching it will begin processing new workflow executions.
+    // The existing default will be marked as a previous incompatible version
+    // to this one, assuming it is not also in `is_compatible_with`.
+    bool default = 5;
```
bergundy, line 147: Weird technicality, where are tasks dispatched to before there's a "default"?
Do we assume that task queues with not build ID ordering are "traditional" task queues?

Sushisource: Yeah, pretty much. The tasks will just go in and whatever worker grabs them first will grab them - after that it'll be known which version they were processed with.

versioning/worker-versions.md
```diff
@@ -0,0 +1,191 @@
...
+is unique for that worker's implementation. _Implementation_ rather than 
+_build_ is an important distinction here, since rebuilds of identical
+code should (ideally) produce workers wit the same ID. It is not necessary to
+do this, and it is varyingly difficult depending on the language, but when
+done it enables users to produce workers with the same ID from their CI builds
+if they didn't touch worker code, but something else (ex: The CI definition itself).
+This involves some work on the part of the user, but is more reliable than expecting
+worker authors to increment a version number.
+
+Unfortunately it is infeasible to automatically set an appropriate ID for many languages.
```
alexshtin, line 97: Go 1.18 has support for it and it automatically emmits git SHA to every binary. And git SHA is very good candidate for build Id.

cretz: > And git SHA is very good candidate for build Id.

Not necessarily, especially if the process is used for more than just worker handling. This is not to be confused with what might be good for "worker binary ID" because _this_ ID will actually affect which workflows are scheduled where and for many they may want multiple workers with different build IDs running in the same process (even for the same task queue). Also, you don't necessarily want to change a worker build ID/version on every commit, especially if you don't link the versions together. Imagine having to either have code linking every commit together, or having to keep a process around for every commit just so it can finish workflows started on it.

I am not sure we should assume that binary checksum is a good default value for this as the beginning of this section says. But if we expose this with an easy way for them to set it, we don't really care then and versioning is opt-in anyways. They can choose how unique they want to be.

alexshtin (commented):
I like the fundamental idea of this proposal. It can be very well automated with go 1.18 w/o extra requirements to worker build/CI systems.
Using git SHA as `buildId` has one more advantage: there are many tools to visualize the graph (from `tig` to github.com).

versioning/worker-versions.md
```diff
@@ -0,0 +1,191 @@
...
+    string namespace = 1;
+    // Must be set, the task queue to apply changes to.
+    string task_queue = 2;
+    // The version id we are targeting.
+    VersionId version_id = 3;
+    // When set, indicates that the `version_id` in this message is compatible
+    // with the one specified in this field. Because compatability should form
+    // a DAG, any build id can only be the "next compatible" version for one
+    // other ID of a certain type at a time, and any setting which would create a cycle is invalid.
+    VersionId is_compatible_with = 4;
```
alexshtin, line 142:
```suggestion
    VersionId compatible_with = 4;
```

versioning/worker-versions.md
Sushisource: I don't mean to say we should lean on _the way it is set by default_. I mean to say we can lean on _the fact that it already exists_. This is also why I want to deprecate that old name and rename it. The checksum by itself is effectively useless. It was really always a stand in for "this worker's implementation" and the name conflated the approach Go used by default with the goal of the field.

We certainly should encourage users to set it as part of their CI process. Git SHA is I believe a pretty reasonable thing to use, because it succinctly expresses the intent ("the set of code we used to build the worker"). It is true that for cases like multiple workers, you'll want to use something else (but that could easily be `<sha>-workera` and `<sha>-workerb`.

And the idea here is you really _do_ want to change it on every commit. You're going to lean on the mechanisms we'll build to make sure each worker deployment can take workflows without breaking them. Every commit is not linked together, but every deployment is - by virtue of the fact that it's the next (assumed incompatible) default version, unless you explicitly specify your intent (via API call) that this newly deployed version should be compatible with some extant one.

If you don't want this behavior, just don't opt in to the feature and go wild west like the existing system, but the whole intent here is to have a safe-by-default approach for people who want it (of which we know of many)

versioning/worker-versions.md
```diff
@@ -0,0 +1,191 @@
...
+    repeated VersionIdNode current_defaults = 1;
+    // Other current latest-compatible versions who are not the overall default. These are the
+    // versions that will be used when generating new tasks by following the graph from the
+    // version of the last task out to a leaf.
+    repeated VersionIdNode compatible_leaves = 2;
+}
+
+message VersionIdNode {
+    VersionId version = 1;
+    VersionIdNode compatible_with = 2;
```
alexshtin, line 171: I believe there are some limitation on recursion in proto. Need to double check.

Sushisource: It can't have cycles, but we already explicitly will enforce that.

alexshtin: Even w/o cycles there is performance degradation with deep depth.

Sushisource: Yeah, any large depth is going to be a no-no here too. The perf characteristics will be awful. We will explicitly limit the depth which I mentioned in an easily-missed comment towards the bottom.

versioning/worker-versions.md
cretz: Is a company that will use the Git SHA expected to either: 1) add links on every deployment to the other Git SHA if compatible (manually or via some ugly automated thing), and/or 2) keep a worker around for every Git SHA that may have started a workflow?

I would think most companies would want semver or a more specific way to signal compatibility. They may make hundreds of commits/deployments without a single backwards incompatibility, there's no need to tie the commit to the worker version.

Of course they have the freedom to do what they want, but when we get to documenting suggested approaches, I think a human-approved/changeable version number makes the most sense since they have to make the decision as a human on whether it's compatible anyways.

Sushisource:
Talked with Roey and we're fine moving forward with the more specific approach. Merging this, and we'll begin prototyping and implementation work!

PR #55, opened 2022-03-18
Python SDK - Phase 2 - Workflows

cretz:
# Summary

This is the second proposal for the Python SDK focusing on workflows.

[See it rendered here](https://github.com/cretz/temporal-proposals/blob/python-phase2/python/phase-2.md) (may be deleted when proposal is done). This follows from Phase 1 at https://github.com/temporalio/proposals/pull/50.

Feel free to leave a bunch of comments. If discussion becomes too difficult to follow, I will close this PR and start a new one.

### DIFF
--- a/python/phase-2.md
+++ b/python/phase-2.md
@@ -22,7 +22,7 @@ from temporalio.client import Client
 from temporalio.worker import Worker
 
 @activity.defn
-def say_hello(name: str) -> str:
+async def say_hello(name: str) -> str:
     return f"Hello, {name}!"
 
 @workflow.defn
@@ -61,7 +61,7 @@ from temporalio.client import Client
 from temporalio.worker import Worker
 
 @activity.defn
-def create_greeting_activity(salutation: str, name: str) -> str:
+async def create_greeting_activity(salutation: str, name: str) -> str:
     return f"{salutation}, {name}!"
 
 @workflow.defn
@@ -91,7 +91,7 @@ class GreetingWorkflow:
             )
             if self._salutation_update.is_set():
                 self._salutation_update.clear()
-            if self._complete.is_set():
+            elif self._complete.is_set():
                 return self._current_greeting
 
     @workflow.signal
@@ -557,7 +557,6 @@ Notes:
 * `ChildWorkflowHandle.signal` will have the same typed overloads that `WorkflowHandle.signal` does
 * `ChildWorkflowHandle.cancel` is present to simplify cancellation. See "temporalio.workflow Scoped Cancellation" for
   more discussion.
-  * How does TypeScript return a failure if the child workflow cancel
 * `start_child` was considered instead of `start_child_workflow`, but this seems clearer. Can be convinced otherwise.
 
 #### `temporalio.workflow.execute_child_workflow()`
--- a/python/phase-2.md
+++ b/python/phase-2.md
@@ -145,7 +145,7 @@ Notes:
 * `@workflow.defn`
   * Must be on class and accepts an optional workflow `name` which defaults to class `__name__`
   * This is required to clearly mark the class as a workflow
-  * No base classes can have this value
+  * This is not inherited from any base classes, it must be explicitly present on this final class
 * `@workflow.run`
   * Must be on a single async method (regardless of inheritance level) and accepts no params
   * The params represent the workflow params and the response represents the workflow response
@@ -715,8 +715,10 @@ Notes:
 class Info:
     pass
 
-def is_replaying() -> bool:
-    pass
+class unsafe:
+    @staticmethod
+    def is_replaying() -> bool:
+        pass
 
 def info() -> Info:
     pass
@@ -731,7 +733,7 @@ Notes:
 
 * Fields of `Info` not shown, but notably absent is `is_replaying` so it can remain immutable
   * Instead, like Java and Go, we move that to the package level
-* Should we move `is_replaying` to a place that makes it more clear it's unsafe?
+* We may move `is_replaying` to a separate module altogether instead of just in a class
 * It's clearer to require uses of `temporalio.workflow.now()` than to wait for and require overriding `datetime.now()`
   when a deterministic sandbox appears. It is clearer to disallow deterministic calls than implicitly replace some.
 * `logger` uses a special adapter that is skipped during replay
@@ -828,3 +830,167 @@ Notes:
 * A full workflow execution sandbox may not occur until phase 3
   * Some work on doing this via `eval` are at https://github.com/cretz/python-determinism-runner
 * We will, however, go ahead and abstract the "runner" so that we can make it pluggable/opt-out if the caller chooses
+
+## Additional API Concerns
+
+### Activities as Methods
+
+It is very normal to want to have activities as methods instead of top-level functions. For example, say you have:
+
+```python
+class ActivityClass:
+    def __init__(self, some_init_arg: int) -> None:
+        pass
+    
+    @activity.defn
+    def some_activity(self, arg1: str) -> List[str]:
+        return ["Yay"]
+```
+
+You cannot just `workflow.execute_activity(ActivityClass.some_activity, "foo")` because technically the first parameter
+is `self` there. But sometimes they may want to call
+`workflow.execute_activity(ActivityClass.some_activity, ActivityClass("init-arg1"), "foo")` and who are we to say they
+can't pass in the self arg explicitly?
+
+What do we do? Here are the notes/options (activity options left off of calls for brevity):
+
+* Cannot use something like `functools.partial(ActivityClass.some_activity, None)` because `partial` removes our
+  attributes which are set in the decorator
+* Don't want to init an instance of the class like this:
+
+    ```python
+    workflow.execute_activity(ActivityClass("init-arg").some_activity, "foo")
+    ```
+* Can create the activity class without initializing it like this:
+
+    ```python
+    activities = ActivityClass.__new__(ActivityClass)
+    workflow.execute_activity(activities.some_activity, "foo")
+    ```
+  This is nice because it doesn't require the Python SDK to care (and people have been doing this in class methods for
+  factories for a while now). We can even provide a helper that does basically this like so:
+
+    ```python
+    activities = workflow.proxy_activity_class(ActivityClass)
+    workflow.execute_activity(activities.some_activity, "foo")
+    ```
+* We can make it possible to pass `None` as the `self` parameter like this:
+
+    ```python
+    workflow.execute_activity(ActivityClass.some_activity, None, "foo")
+    ```
+  And then we could check for that being self and ignore it. However, due to Mypy not yet supporting all of `ParamSpec`,
+  that requires us to have a two-parameter overload version of `execute_activity` _and_ to make the first param optional
+  like so:
+  
+    ```python
+    @overload
+    async def execute_activity(
+        activity: Callable[[LocalParamType, LocalParamType2], LocalReturnType],
+        arg: Optional[LocalParamType],
+        arg2: LocalParamType2,
+        /,
+        *,
+    ```
+  This means if you had a real two-param activity, you'd now be able to have an optional first parameter. Also, I am not
+  yet confident how this will look when we remove all of these overloads once Mypy gets their `ParamSpec` support.
+* We could have a completely separate call like so:
+
+    ```python
+    workflow.execute_activity_method(ActivityClass.some_activity, "foo")
+    ```
+* Instead of having `@activity.defn` set an attribute on the function (which is then transitively transferred for others
+  using `functools.wraps`), we could have it return something like `ActivityDefinition[ParamTypes, ReturnType]` (which
+  does not get the transitive benefits). The reason we might want this is then we can have `@activity.method_defn` or
+  similar that is smart enough to convert to an activity definition. I have been trying to avoid actually wrapping
+  functions so far.
+* We could support the following with an overload:
+
+    ```python
+    workflow.execute_activity(ActivityClass.some_activity, "foo", method=True)
+    ```
+  This would work with a special overload like:
+
+    ```python
+    @overload
+    async def execute_activity(
+        activity: Callable[[Any, LocalParamType], LocalReturnType],
+        arg: LocalParamType,
+        /,
+        *,
+        method: Literal[True],
+    ```
+  This seems like a fairly sane way, but requires callers pass a special parameter to call
+* Another outstanding question - is there added benefit _at this time_ to support registering a class full of activity
+  methods instead of individually? I am thinking not (yet).
+
+### Workflow Contracts
+
+Sometimes users may want to separate the workflow into a contract and an impl so that they can be separated. We will
+support this approach.
+
+Here is what it looks like with `typing.Protocol`s:
+
+```python
+@workflow.defn
+class MyWorkflow(Protocol):
+    @workflow.run
+    def run(self, arg: str) -> List[str]:
+        ...
+
+@workflow.defn(name=MyWorkflow.__name__)
+class MyWorkflowImpl:
+    @workflow.run
+    def run(self, arg: str) -> List[str]:
+        return [arg]
+
+async def run_workflow(client: Client) -> List[str]:
+    return await client.execute_workflow(MyWorkflow.run)
+
+def create_worker(client: Client) -> Worker:
+    return Worker(workflows=[MyWorkflowImpl])
+```
+
+And here's what it looks like with abstract classes:
+
+```python
+@workflow.defn
+class MyWorkflow(ABC):
+  @workflow.run
+  @abstractmethod
+  def run(self, arg: str) -> List[str]:
+    raise NotImplementedError
+
+@workflow.defn(name=MyWorkflow.__name__)
+class MyWorkflowImpl(MyWorkflow):
+    @workflow.run
+    def run(self, arg: str) -> List[str]:
+        return [arg]
+
+async def run_workflow(client: Client) -> List[str]:
+    return await client.execute_workflow(MyWorkflow.run)
+
+def create_worker(client: Client) -> Worker:
+    return Worker(workflows=[MyWorkflowImpl])
+```
+
+Notes:
+
+* This has no special logic to handle this approach. The same approach can be used to separate activity class method
+  contracts from their impls.
+* It is unfortunate that we require a `@workflow.defn` on both the base and impl classes.
+  * We want to require this decorator at client-side _and_ worker-side to ensure name sharing
+  * With `Protocol` it is obvious why, there is technically no code shared
+  * With abstract base classes, we maybe could support `@workflow.defn(extends=MyWorkflow)` but that doesn't give you
+    much
+  * I am hesitant to try to implicitly derive the base name because there are other use cases for inheritance besides
+    just contract/impl separation
+* It is unfortunate that we require a `@workflow.run` on both the base and impl classes.
+  * I can't necessarily assume the base-class's run is overridden the exact same way in the child class
+* I suppose I could do this: Always inherit decorators and if there are decorators on the same method/class in the tree
+  but they don't match, raise an error. But allow `inherit=False` to work around.
+  * Yes, this might be a cleaner approach. I see no reason why people need to have different decorator values for the
+    same item.
+  * But this means that it may no longer be obvious something is a workflow or a signal or whatever since the decorator
+    may only be on the base class.
+  * Looking for feedback here and if we decide otherwise, I'll remove the no-decorator-inheritance rule.
--- a/python/phase-2.md
+++ b/python/phase-2.md
@@ -4,7 +4,8 @@ The Python SDK will be implemented in three loosely defined phases:
 
 * Phase 1 - Initial project, activity worker, and client
 * Phase 2 - Workflow worker, local activities, async activity completion, etc
-* Phase 3 - Deterministic workflow sandbox, replayer, test suite (some parts may be combined with phase 2)
+* Phase 3 - Deterministic workflow sandbox, replayer, test suite, versioning, helpers (some parts may be combined with
+  phase 2)
 
 This proposal is for phase 2. It is intentionally loosely defined and all aspects are subject to discussion/change.
 
@@ -923,6 +924,47 @@ What do we do? Here are the notes/options (activity options left off of calls fo
   This seems like a fairly sane way, but requires callers pass a special parameter to call
 * Another outstanding question - is there added benefit _at this time_ to support registering a class full of activity
   methods instead of individually? I am thinking not (yet).
+* Another option that was requested to be shown is to disallow `@activity.defn` on methods and make an easy way to
+  externally define those as callable. Here's the easiest way I can think of making that look:
+
+    ```python
+    class ActivityClass:
+        def __init__(self, some_init_arg: int) -> None:
+            pass
+
+        def some_activity1(self, arg: str) -> List[str]:
+            ...
+
+        def some_activity2(self, arg: int) -> List[int]:
+            ...
+
+    some_activity1 = workflow.activity_method_reference(ActivityClass.some_activity1)
+    some_activity2 = workflow.activity_method_reference(ActivityClass.some_activity2, name="some_other_name")
+
+    def new_worker(client: Client) -> Worker:
+        return Worker(client, activities=[ActivityClass.some_activity1, ActivityClass.some_activity2])
+
+    @workflow.defn
+    class MyWorkflow:
+        @workflow.run
+        async def run(self):
+            res1 = await workflow.execute_activity(some_activity1, "arg")
+            res2 = await workflow.execute_activity(some_activity2, 123)
+    ```
+    It works but it's a bit ugly compared to `execute_activity_method`. At that point, technically they could just
+    hand-write stubs like:
+
+    ```python
+    @activity.defn
+    def some_activity1(arg: str) -> List[str]:
+        raise RuntimeError("Cannot call stub directly!")
+    ```
+    Ant it has the same effect from inside the workflow.
+
+#### Result
+
+After discussion we have decided to use the `execute_activity_method` approach. Of course people can still do the
+`ActivityClass.__new__(ActivityClass)` approach if they want, but that's now how we'll me suggesting to do it.
 
 ### Workflow Contracts
 
@@ -994,3 +1036,191 @@ Notes:
   * But this means that it may no longer be obvious something is a workflow or a signal or whatever since the decorator
     may only be on the base class.
   * Looking for feedback here and if we decide otherwise, I'll remove the no-decorator-inheritance rule.
+
+#### Result
+
+After discussion, we have decided it is this for `@workflow.defn`:
+
+```python
+ClassType = TypeVar("ClassType", bound=Type)
+
+@overload
+def defn(*, implements: ClassType) -> DefnDecorator[ClassType]:
+    pass
+
+@overload
+def defn(*, name: Optional[str] = None) -> DefnDecorator[Any]:
+    pass
+
+@overload
+def defn(cls: ClassType) -> ClassType:
+    pass
+
+def defn(cls: Optional[ClassType] = None, *, name: Optional[str] = None, implements: Optional[ClassType] = None):
+    if cls is None:
+        # TODO(cretz): Do stuff with name/implements
+        return defn
+    return cls
+
+class DefnDecorator(Protocol[ClassType]):
+    def __call__(self, cls: ClassType) -> ClassType:
+        raise NotImplemented
+```
+
+Notes:
+
+* Every worker-registered or called-called workflow must have `@workflow.defn` of course
+* `@workflow.defn` will have a parameter called `implements` that accepts either a `Protocol` or a base class. This
+  parameter is mostly only used for aligning names and not for general inheritance-based code reuse
+  * The implements does enforce that the class actually does implement that
+  * Cannot provide a name if you provide `implements=` (that base one provides the name)
+* Rules for general inheritance:
+  * There can be only one `@workflow.run` (but it can be overridden)
+  * Every method override of a `@workflow.<x>` _must_ have the same decorator and decorator args and method args
+    * So overridding, say, run or a signal handler or whatever must have the same
+    * We probably can't easily confirm the arg type annotations are equivalent due to so many possibilities with
+      generics and such
+* This means that classes marked as workflows will be typed as their `implements`, _not_ their actual class type, which
+  should be ok
+
+### Dynamic Signals/Queries
+
+There is a concern that we won't allow easy runtime signal/query handler management from inside the workflow. Currently,
+we allow catch-all signal and query handlers like so:
+
+```python
+@workflow.defn
+class MyWorkflow:
+    @workflow.run
+    async def run(self):
+        ...
+
+    @workflow.signal(dynamic=True)
+    async def default_signal_handler(self, signal_name: str, *args: Any) -> None:
+        ...
+
+    @workflow.query(dynamic=True)
+    def default_query_handler(self, query_name: str, *args: Any) -> Any:
+        ...
+```
+
+So technically you could have a runtime mixin helper like so:
+
+```python
+class DynamicHandlerMixin:
+    def __init__(self, *args, **kwargs) -> None:
+        ...
+
+    def set_signal_handler(self, signal_name: str, callback: Callable) -> None:
+        ...
+
+    @workflow.signal(dynamic=True)
+    async def default_signal_handler(self, signal_name: str, *args: Any) -> None:
+        ...
+
+    def set_query_handler(self, signal_name: str, callback: Callable) -> None:
+        ...
+
+    @workflow.query(dynamic=True)
+    def default_query_handler(self, query_name: str, *args: Any) -> Any:
+        ...
+```
+
+Which could be used like:
+
+```python
+@workflow.defn
+class MyWorkflow(DynamicHandlerMixin):
+    def __init__(self) -> None:
+        ...
+
+    @workflow.run
+    async def run(self):
+        if some_conditional():
+            self.set_signal_handler("my signal name", self.some_signal_handler)
+
+    def some_signal_handler(s: str) -> None:
+        ...
+```
+
+But there are downsides, namely:
+
+* This would require user-land to buffer the signals instead of us
+* This would require user-land to throw on missing query instead of our error
+  which includes known registered names
+* You have to opt-in to this
+
+Instead, we can offer this in the `temporalio.workflow` package:
+
+```python
+def set_signal_handler(name: str, handler: Callable) -> None:
+    ...
+
+def set_query_handler(name: str, handler: Callable) -> None:
+    ...
+```
+
+Notes:
+
+* This will be in addition to the catch-all handler, but we will suggest use of this less than the explicit
+  `@workflow.signal`/`@workflow.query` approach
+* Like other SDKs, we'll buffer unhandled signals and return query error on unhandled queries
+* What should we do if there is already a handler for that item's name?
+  * I'm thinking latest wins
+* Should we offer a `set_default_signal_handler`/`set_default_query_handler` that is runtime equivalent of
+  `dynamic=True`?
+  * I'm thinking yes
+
+### Dynamic Signal Args
+
+A request for a use case has been made to show "A RPC-like system that accepts different arg types for the same signal
+name". Here's how such a thing may look in Python:
+
+```python
+@workflow.defn
+class MyWorkflow:
+
+    @workflow.run
+    async def run(self) -> None:
+        ...
+
+    @overload
+    async def do_call(self, arg: SomeClass1) -> None:
+        ...
+
+    @overload
+    async def do_call(self, arg: SomeClass2) -> None:
+        ...
+
+    @workflow.signal
+    async def do_call(self, arg: Any) -> None:
+        # Python 3.10+ for switches
+        match type(arg):
+            case SomeClass1:
+                self.do_some_class1_call()
+            case SomeClass2:
+                self.do_some_class1_call()
+
+async def make_some_calls(client: Client):
+    handle = client.get_workflow_handle_for(MyWorkflow, "some id")
+    # Passes type checker
+    await handle.signal(MyWorkflow.signal, SomeClass1())
+    # Passes type checker
+    await handle.signal(MyWorkflow.signal, SomeClass2())
+    # Fails type checker
+    await handle.signal(MyWorkflow.signal, SomeClass3())
+```
+
+Notes:
+
+* All of this is supported out of the box.
+* You can do the same thing for queries, activities, or even the primary `@workflow.run` entrypoint meaning you can have
+  a workflow accept different param types but have the type checker still keep it limited
+* You could get the same thing with just a function like:
+
+    ```python
+    @workflow.signal
+    async def do_call(self, arg: Union[SomeClass1, SomeClass2]) -> None:
+        ...
+    ```
+  But for some maybe this is easier
--- a/python/phase-2.md
+++ b/python/phase-2.md
@@ -179,7 +179,7 @@ Notes:
 
 ### Client
 
-#### WorkflowHandle
+#### `WorkflowHandle`
 
 Updated with the following generics:
 
@@ -189,7 +189,7 @@ class WorkflowHandle(Generic[WorkflowType, WorkflowReturnType]):
 
 The workflow type is used for typing the "self" of signal/query methods and the return type is for result
 
-#### Client.start_workflow
+#### `Client.start_workflow`
 
 The following overload _would_ be added to client:
 
@@ -236,7 +236,7 @@ Where `LocalParamType` is a `ParamSpec`. But since `Concatenate` is
 Where `LocalParamType` is a `TypeVar`. This means that multi-param workflows are not currently type-able which is
 probably an acceptable limitation for now.
 
-#### Client.get_workflow_handle_for
+#### `Client.get_workflow_handle_for`
 
 Due to the same no-mypy-`Concatenate` problem as `start_workflow`, this has a union for the param:
 
@@ -268,7 +268,7 @@ Notes:
   param names or order of params? Currently as described is the clearest.
 
 
-#### WorkflowHandle.result
+#### `WorkflowHandle.result`
 
 This is already typed and doesn't really change, but the type name changed so it looks like:
 
@@ -277,7 +277,7 @@ This is already typed and doesn't really change, but the type name changed so it
       pass
 ```
 
-#### WorkflowHandle.signal
+#### `WorkflowHandle.signal`
 
 Due to the same no-mypy-`Concatenate` problem as `start_workflow`, this has two overloads:
 
@@ -304,7 +304,7 @@ Due to the same no-mypy-`Concatenate` problem as `start_workflow`, this has two 
         pass
 ```
 
-#### WorkflowHandle.query
+#### `WorkflowHandle.query`
 
 Due to the same no-mypy-`Concatenate` problem as `start_workflow`, this has two overloads:
 
@@ -337,31 +337,31 @@ Due to the same no-mypy-`Concatenate` problem as `start_workflow`, this has two 
 
 ### Worker
 
-#### Worker.__init__.activities
+#### `Worker.__init__.activities`
 
 Currently `activities` are accepted as `Mapping[str, Callable]`. Due to `@activity.defn` we will change this to
 `Iterable[Callable]` and require all activities have `@activity.defn`. See the overview for notes/discussion on this.
 
-#### Worker.__init__.workflows
+#### `Worker.__init__.workflows`
 
 Workers will accept a `workflows` param which is `Iterable[Type]`. These will be classes that have `@workflow.defn` and
 they are validated at runtime.
 
-#### Worker.__init__.workflow_runner
+#### `Worker.__init__.workflow_runner`
 
 Optional class (not shown) that can be used to invoke workflows. Hopefully, the default for this will be a deterministic
 sandbox in phase 3. But in the meantime this is just an abstraction in preparation. It's not called "executor" because
 `activity_executor` already exists which uses the literal "concurrent executors" from Python.
 
-#### temporalio.worker.Interceptor
+#### `temporalio.worker.Interceptor`
 
 Will add `intercept_workflow` to this and support all inbound and outbound interception.
 
-### temporalio.workflow
+### `temporalio.workflow`
 
 The workflow package will contain functions that can only be called from inside a workflow.
 
-#### temporalio.workflow.start_activity()
+#### `temporalio.workflow.start_activity()`
 
 ```python
 
@@ -450,28 +450,28 @@ Notes:
 * Why does does mypy only work here when we separate out async from sync, but allows the use of `Union` for
   `WorkflowHandle.query`?
 
-#### temporalio.workflow.execute_activity()
+#### `temporalio.workflow.execute_activity()`
 
 Async shortcut for `temporalio.workflow.start_activity()` + `handle.result()`.
 
-#### temporalio.workflow.ActivityConfig
+#### `temporalio.workflow.ActivityConfig`
 
 `TypedDict` of `start_activity`/`execute_activity` configuration options.
 
-#### temporalio.workflow.start_local_activity()
+#### `temporalio.workflow.start_local_activity()`
 
 Same as `start_activity`, except no `heartbeat_timeout` param. For now, we will reuse `ActivityHandle` for the result
 since there would be nothing unique about a `LocalActivityHandle`.
 
-#### temporalio.workflow.execute_local_activity()
+#### `temporalio.workflow.execute_local_activity()`
 
 Shortcut for `temporalio.workflow.start_local_activity()` + `handle.result()`.
 
-#### temporalio.workflow.LocalActivityConfig
+#### `temporalio.workflow.LocalActivityConfig`
 
 `TypedDict` of `start_local_activity`/`execute_local_activity` configuration options.
 
-#### temporalio.workflow.start_child_workflow()
+#### `temporalio.workflow.start_child_workflow()`
 
 ```python
 class ChildWorkflowHandle(Generic[WorkflowType, WorkflowReturnType], Cancellable):
@@ -560,15 +560,15 @@ Notes:
   * How does TypeScript return a failure if the child workflow cancel
 * `start_child` was considered instead of `start_child_workflow`, but this seems clearer. Can be convinced otherwise.
 
-#### temporalio.workflow.execute_child_workflow()
+#### `temporalio.workflow.execute_child_workflow()`
 
 Shortcut for `temporalio.workflow.start_child_workflow()` + `handle.result()`
 
-#### temporalio.workflow.ChildWorkflowConfig
+#### `temporalio.workflow.ChildWorkflowConfig`
 
 `TypedDict` of `start_child_workflow`/`execute_child_workflow` configuration options.
 
-#### temporalio.workflow.continue_as_new()
+#### `temporalio.workflow.continue_as_new()`
 
 ```python
 # Overload for self (cannot type args)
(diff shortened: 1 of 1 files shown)


python/phase-2.md
```diff
@@ -0,0 +1,831 @@
...
+A simple example of a workflow definition and worker run:
+
+```python
+from datetime import timedelta
+from temporalio import activity, workflow
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+@activity.defn
+def say_hello(name: str) -> str:
```
bergundy, line 25: nit: doesn't this require registering an executor with the worker?

cretz: It does, I forgot `async` on here, will fix.

python/phase-2.md
```diff
@@ -0,0 +1,831 @@
...
+    return f"Hello, {name}!"
+
+@workflow.defn
+class MyWorkflow:
+    @workflow.run
+    async def run(arg1: str) -> str:
+        return await workflow.execute_local_activity(
```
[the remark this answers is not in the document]
cretz: We can't support named parameters in workflows or activities today because our payload system has no such concept (only a collection of payloads). I already enforce this only-positional-parameter limitation today.

python/phase-2.md
cretz: Fixed

python/phase-2.md
```diff
@@ -0,0 +1,831 @@
...
+Notes:
+
+* Workflows are classes
+  * Consideration was given to making workflows be functions (in addition to classes or having the class equivalent sit
+    on top) with several possibilities for defining signals and queries, but all were too ugly/complicated
+* `@activity.defn`
+  * Must be on function and accepts an optional activity `name` which defaults to function `__name__`
+  * This was originally not wanted in the activity impl (phase 1) due to complaints of the name not being obvious
+  * We need to define the activity name on the activity for execution by the workflow
+  * Is it ok to now make this required? Is it ok to now make the name implicit? Is it ok to now use a list instead of a
```
bergundy, line 143: I'm not a fan of this decorator.
It'd be much more practical to have some collection class where you could accumulate these grouped activities:

```python
coll = ActivityCollection()

@coll.activity
def my_activity():
    ...

Worker(activities=coll)
```

The above is a common pattern in Python frameworks.
That being said, I wouldn't want to limit to only using this collection concept and I'd rather let users decide how they want to register their activities.

If we want to keep this decorator for allowing the name override I'd be fine with that but I wouldn't force this convention.

python/phase-2.md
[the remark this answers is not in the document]
bergundy: But if you know the activity signature why would you want to limit calling with named params?

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+      client=client,
+      activities=[create_greeting_activity],
+      workflows=[GreetingWorkflow],
+    )
+```
+
+Notes:
+
+* Workflows are classes
+  * Consideration was given to making workflows be functions (in addition to classes or having the class equivalent sit
```
bergundy, line 137: I've raised my concerns with this approach already but I'll just put them down here FTR:

1. With classes it's much harder to have reusable components because that requires multiple inheritance
2. With classes, because the start params are only provided to the `run` method, you cannot initialize your class in the constructor which in some cases requires making some attributes optional where they really are not resulting in redundant type assertions throughout the implementation.

python/phase-2.md
```diff
@@ -0,0 +1,831 @@
...
+    async def complete_with_greeting(self) -> None:
+        self._complete.set()
+
+    @workflow.query
+    async def current_greeting(self) -> str:
+        return self._current_greeting
+
+def test_workflow(client: Client) -> None:
+    # Start the workflow
+    handle = await client.start_workflow(GreetingWorkflow.run, "John")
```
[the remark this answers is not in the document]
cretz: I am intentionally avoiding that proxy approach. I suspect if Java had better function references, it wouldn't have done things that way either.

There are several problems with that approach:

1. It's hard to separate "start" from "execute"
2. It's hard to separate run entrypoint from signals from queries from sync updates from just normal methods
3. It's hard for readers to know what's being done ("is that a query or a signal?")
4. You end up either down an ugly "untyped stub" path or an ugly "completely different ways to start a workflow" path

There are other reasons too. After lots of thought, I think function references are the way to go here.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+* `@activity.defn`
+  * Must be on function and accepts an optional activity `name` which defaults to function `__name__`
+  * This was originally not wanted in the activity impl (phase 1) due to complaints of the name not being obvious
+  * We need to define the activity name on the activity for execution by the workflow
+  * Is it ok to now make this required? Is it ok to now make the name implicit? Is it ok to now use a list instead of a
+    dict to specify activities for the worker?
+* `@workflow.defn`
+  * Must be on class and accepts an optional workflow `name` which defaults to class `__name__`
+  * This is required to clearly mark the class as a workflow
+  * No base classes can have this value
```
bergundy, line 148: Can you elaborate here? Can I not extend a workflow class? Why?

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    @workflow.query
+    async def current_greeting(self) -> str:
+        return self._current_greeting
+
+def test_workflow(client: Client) -> None:
+    # Start the workflow
+    handle = await client.start_workflow(GreetingWorkflow.run, "John")
+
+    # Confirm the current greeting
+    assert "Hello, John!" == await handle.query(GreetingWorkflow.current_greeting)
```
bergundy, line 115: What about type safe calls in situations where the caller does not have access to the implementation?
Are you recommending using abstract classes or protocols for those?
Can you add a sample to this proposal?

python/phase-2.md
cretz: Why do we have to force global state on people? I just need a simple marker.

> I'd rather let users decide how they want to register their activities.

This is the opposite conclusion that was previously reached which was that we don't want a bunch of different ways to register

> If we want to keep this decorator for allowing the name override I'd be fine with that but I wouldn't force this convention.

Why not require the decorator on every activity and on every workflow? What benefit is there to allowing activities be not marked as such?

This middleground approach was chosen because I was repeatedly told we want to be explicit on the name and this is the only way I can think of having the name be somewhat explicit (you know it's named what's beneath the decorator unless set).

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+  * Must be on function and accepts an optional activity `name` which defaults to function `__name__`
+  * This was originally not wanted in the activity impl (phase 1) due to complaints of the name not being obvious
+  * We need to define the activity name on the activity for execution by the workflow
+  * Is it ok to now make this required? Is it ok to now make the name implicit? Is it ok to now use a list instead of a
+    dict to specify activities for the worker?
+* `@workflow.defn`
+  * Must be on class and accepts an optional workflow `name` which defaults to class `__name__`
+  * This is required to clearly mark the class as a workflow
+  * No base classes can have this value
+* `@workflow.run`
```
bergundy, line 149: You might want to call this `execute` to match the client API

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+        self,
+        workflow: Callable[[WorkflowType, LocalParamType], Awaitable[WorkflowReturnType]],
+        arg: LocalParamType,
+        /,
+        *,
+        # ...
+    ) -> WorkflowHandle[WorkflowType, WorkflowReturnType]:
+        pass
+```
+Where `LocalParamType` is a `TypeVar`. This means that multi-param workflows are not currently type-able which is
```
bergundy, line 236: That fine as long as they're invokable.

python/phase-2.md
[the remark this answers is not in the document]
cretz: Because we can't decode the param names on the other side. Similar occurs with defaults, but we let those slide as local concerns. We also want to encourage single param functions too I believe.

Someone is gonna switch the order of the params and not know that's a breaking change because they always use named params. Temporal works with positional parameters only across all SDKs, I don't think we need to rebuild the positions from the given named params.

Also, I don't want to encourage different calling styles from string name to typed. I think there's value in how similar they are today.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+) -> NoReturn:
+    # Do stuff
+```
+
+Notes:
+
+* Like all other SDK languages so far, we can't reasonably provide types for args on the "self" continue as new
+* More investigation is needed on whether I can stop the event loop inside the continue as new without it returning
+  * I probably can't so I probably need to make this call `async`, but I probably don't have to throw
+  * While I thought I could just stop the event loop and then return, that doesn't prevent the next line of code from
```
bergundy, line 629: That should be fine as long as it's deterministic

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    potentially setting local state that could change a query result
+  * This will be subject to implementation-time research. The goal is to try to never return from the function (if it's
```
bergundy, line 631: Note that you still have to support querying the workflow after it has completed so you can't entirely kill its runtime.

python/phase-2.md
cretz: I do not have explicit support for that today and I'm hesitant to do so at least at first.

The protocol would have to match the implementation exactly, down to the proper decorators. But technically it'd work, so you could have:

```python
@workflow.defn
class GreetingWorkflow(Protocol):
    @workflow.run
    async def run(self, name: str) -> str:
        ...

    @workflow.signal
    async def update_salutation(self, salutation: str) -> None:
        ...

    @workflow.signal
    async def complete_with_greeting(self) -> None:
        ...

    @workflow.query
    async def current_greeting(self) -> str:
        ...
```

And then `handle.query(GreetingWorkflow.current_greeting)`. Same with abstract base classes. But all you've really said is that you know how to use common Python interface/inheritance features, it's not specific to Temporal.

My concern is that I am not sure we want to, at least for Python, encourage completely separate pieces of code for the contract. If they want to use inheritance or `@runtime_checkable`, no prob, we support it.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+* Originally there was concern that I couldn't report data conversion failure on search attributes, but since custom
+  converters are not allowed, I can just make sync calls to the JSON data converter
+
+#### `temporalio.workflow` Async Utilities
+
+```python
+async def sleep(delay: float, result: T = None) -> T:
+  pass
+
+async def wait_condition(fn: Callable[[], Awaitable[bool]], *, timeout: Optional[float] = None):
```
bergundy, line 700: Still not a fan of `await wait_...`

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+
+#### `temporalio.workflow` Miscellaneous
+
+```python
+
+@dataclass(frozen=True)
+class Info:
+    pass
+
+def is_replaying() -> bool:
```
bergundy, line 718: Let's put this under `unsafe` like TS and Java

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    pass
+
+logger = LoggerAdapter(logging.getLogger(__name__), None)
+```
+
+Notes:
+
+* Fields of `Info` not shown, but notably absent is `is_replaying` so it can remain immutable
+  * Instead, like Java and Go, we move that to the package level
+* Should we move `is_replaying` to a place that makes it more clear it's unsafe?
```
bergundy, line 734: Of course this is mentioned here

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+Notes:
+
+* Unlike TypeScript, Python still allows you to explicitly cancel individual activities and child workflows if you
+  started them with `start_`
+  * The consistency of having everything that has a `start_` be cancellable has value. I still expect most will use the
+    `execute_` form though
+* The workflow will have a root scope
+* This will work using `with CancellationScope()` and similar and is powered by `contextvars`
+  * `run` is async because the callable may have an async result, but `with` is the preferred approach for explicit use
+* `start_activity`/`execute_activity`/`start_child_workflow`/`execute_child_workflow` will implicitly use the current
```
bergundy, line 792: FTR sleep and condition and possibly other methods will need to use it too

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+  cancellation scope
+* Do we need extra capabilities such as:
+  * Ability to explicitly add cancellables?
+  * Ability to unset the current instead of just forcing a detached one each time you want to opt out?
+* Should we disallow the `parent` to even be set in the constructor?
+  * Thinking we probably shouldn't allow you to set that
```
bergundy, line 798: I've never seen this used in TS (or any instantiation of CancellationScopes as a matter of fact)

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+
+* We are gonna use `asyncio`, but whether we have to use our own event loop remains to be seen
+  * Have shown at https://github.com/cretz/python-determinism-runner basically using the approach from
+    https://stackoverflow.com/a/29797709 that we can run a single step and enforce that we have yielded only at
+    workflow-acceptable spots
+* Do want to prevent use of `asyncio.sleep`, `asyncio.wait_for`, and similar that are based on system time?
+  * This is much clearer than replacing their implementations so that people aren't confused about why a 50ms wait does
+    not work
+  * We may not be able to block everything until we use a sandbox (see next section)
+  * Making alternatives is a tad ugly, e.g. one can use `asyncio.gather()` without timeout but they can't with timeout,
```
bergundy, line 821: Is this still true if we implement our own event loop?

python/phase-2.md
cretz: 1. Yes, this is a limitation of class
2. There are ways we can make it a bit easier

Here are the concerns with the function approach:

1. You have to remember to manually set handlers everywhere - no good
2. The definition for your query/signal is divorced from the implementation, but your workflow gets to happily have the definition and implementation together - no good
3. You either don't get the benefits of class state or we have to support multiple ways of writing a workflow
4. For purposes of running (not yet determined how needed), functions referencing other functions are difficult to isolate without isolating the entire runtime. It can be hard to reason about isolated state when it's a loosely connected set of functions instead of a class we can ask they keep state within.
5. If workflows are functions, how do you define which queries/signals belong to which workflows at compile time? TS forces you mostly to file-per-workflow if you want to avoid such confusion which I don't think is a good model.

I can of course go back on this, but I might need to see some examples to justify it. For instance, two workflows with separate signals and queries and what their client calls look like.

bergundy (approved):
There are some missing Workflow APIs in this proposal but it's okay, it doesn't need to be complete.

python/phase-2.md
cretz: The primary reason is to avoid the "Diamond Problem" when resolving attribute ambiguity. I think it needs to be clear that the workflow is sealed/final so the reader is not surprised by the actual implementation.

Having said that, with how I mentioned abstract classes/protocols could be used, I think we can relax this constraint to say that "Every registered workflow class must have the decorator" so even if you extend one, you must put the decorator on yourself since it's essentially a marker for name.

python/phase-2.md
cretz: Why not `@workflow.start` which matches the client API? I considered both and figured this was the shortest and clearest to define what it is. I also considered `@workflow.entry` but I don't think that makes it as clear that the workflow only lasts for the life of the function.

python/phase-2.md
cretz: Even if it is deterministic, I want to stop execution immediately on continue-as-new which I believe requires it being async. More research is required here.

python/phase-2.md
bergundy: > 1. You have to remember to manually set handlers everywhere - no good

True for your class based proposal too

> 2. The definition for your query/signal is divorced from the implementation, but your workflow gets to happily have the definition and implementation together - no good

Some might say this is a good thing, especially since there are plenty of cases where the caller does not have the implementation available

> 3. You either don't get the benefits of class state or we have to support multiple ways of writing a workflow

It might not be too bad, I'm thinking of *also* adding class support to TS.

> 4. For purposes of running (not yet determined how needed), functions referencing other functions are difficult to isolate without isolating the entire runtime. It can be hard to reason about isolated state when it's a loosely connected set of functions instead of a class we can ask they keep state within.

It's easy enough create a class or access any variable in scope.
With functions one can test any part of their code where with classes there's extra setup involved.

> 5. If workflows are functions, how do you define which queries/signals belong to which workflows at compile time? TS forces you mostly to file-per-workflow if you want to avoid such confusion which I don't think is a good model.

TS is flexible and doesn't impose any restrictions on file structure, I've seen cases where signals are all defined in a central location and reuse in different workflows.

All of that being said, working with non local variables in Python is a bit inconvenient and classes might be a better fit here.

python/phase-2.md
cretz: Right. I think I can kill the event loop but still invoke the not-allowed-to-be-async query method.

python/phase-2.md
cretz: It must be tough to write Python code for you today then, heh. Besides the obvious `await wait` and `await wait_for`, you have `Event.wait` and `Condition.wait` which this construct is very similar to.

python/phase-2.md
cretz: :+1:

python/phase-2.md
cretz: :+1:

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+        pass
+
+    @property
+    def cancelled(self) -> bool:
+        pass
+```
+
+Notes:
+
+* Unlike TypeScript, Python still allows you to explicitly cancel individual activities and child workflows if you
```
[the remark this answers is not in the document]
cretz: Yeah, basically I still have the `execute_activity` I expect most to use and I still have cancellation scopes that work just like in TS. I just _also_ offer a `start_activity` if you did want to explicitly cancel. This also has the added benefit of, when we have other ways of interacting activities such as activity signals, I can just add them on the handle here.

It also makes the `start_` and `execute_` APIs very consistent across the codebase.

python/phase-2.md
cretz: :+1:

python/phase-2.md
cretz: It seems harmless to have and provides some clarity. Technically I made up this use case for TS:

```typescript
parent = CancellationScope.current()
scope1 = new CancellationScope()
scope2 = new CancellationScope({ parent })
f1 = scope1.run(activityA())
f2 = scope1.run(activityB())
f3 = scope2.run(activityC())
f4 = scope1.run(activityD())
f5 = scope2.run(activityE())
```

But I admit that's rare.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+
+### Workflow Stepped Execution
+
+Notes:
+
+* We are gonna use `asyncio`, but whether we have to use our own event loop remains to be seen
+  * Have shown at https://github.com/cretz/python-determinism-runner basically using the approach from
+    https://stackoverflow.com/a/29797709 that we can run a single step and enforce that we have yielded only at
+    workflow-acceptable spots
+* Do want to prevent use of `asyncio.sleep`, `asyncio.wait_for`, and similar that are based on system time?
```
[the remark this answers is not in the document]
cretz: > If we're supporting multiple Python workflow runtimes, I'd disallow those and provide our own sleep method.

Yes, that was the goal. Temporal sleep does not work like system sleep (especially at a low granularity) so I think it's worth the separation. But I can be convinced either way...and I can have whichever way we want only be default with a way to opt-out.

python/phase-2.md
cretz: My fear is what may seem like a local operation can increase history with timer creation. But I am on the fence.

python/phase-2.md
cretz: > True for your class based proposal too

Not true, the definition and the handler are married. You can't create a definition without the handler like you can in TS.

> Some might say this is a good thing, especially since there are plenty of cases where the caller does not have the implementation available

I don't think this will be that common. And of course as mentioned before, having to keep two contracts in sync is rough.

> It might not be too bad, I'm thinking of also adding class support to TS.

During the last phase, I was constantly told not to have multiple definition/registration approaches.

> It's easy enough create a class or access any variable in scope. With functions one can test any part of their code where with classes there's extra setup involved.

I think this is an acceptable cost. If that setup is a real problem, I can only imagine the problems with the other things you have to setup for tests.

> TS is flexible and doesn't impose any restrictions on file structure, I've seen cases where signals are all defined in a central location and reuse in different workflows.

What this also means is you can have a signal that is not even on a workflow and no way in the type system to prevent using it. There is nothing at compile time in TS that puts signals on a workflow, and I think it's valuable to have such a thing easy to see for a dev. One could only imagine a complex workflow where I'd have to walk a bunch of code to see what signals are even on a workflow.

> All of that being said, working with non local variables in Python is a bit inconvenient and classes might be a better fit here.

In the same vein, functions are the Python code organization construct of choice and other similar Python solutions do let you decorate functions. Simply stated, "does the clarity of predefining multiple things belonging to a workflow outweigh the annoyance of a class?"

Sushisource: I quite like the class based approach. I'm honestly not sure re-usable definitions/components actually net out to be a good thing in the long run. It seems very easy to break compatibility for some workflow you didn't realize your code was being used in. Stated differently: I think you want to keep the call depth of workflows shallow, because you want their execution path to be obvious. 

The way I might imagine having the function based approach work is by defining a workflow as a collection of signal/query definitions (which may come with implementations, or point to them) along with it's entry point signature/implementation, but I think as I brought that up in Slack there was no way to make that work statically with Python type checking.

python/phase-2.md
Sushisource: > I'd rather let users decide how they want to register their activities.

Why? I just don't at all follow why allowing for a bunch of different ways to do something required is a good thing for users. This should be a zero-thought operation. You register your activities this way, do it once, forget about it. Choice just increases the user's cognitive load and I don't follow what the benefit is to allowing bunches of different ways to do it.

Explicit and obvious for required, only-once operations seems to me a very good thing. Set it and forget it.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    it became too confusing
+  * The reference to this function is used for starting a workflow with a typed handle
+* `@workflow.signal`
+  * Can be on any number of methods (async or not, can be inherited)
+  * Can have an optional `name` which defaults to function `__name__`
+  * Params are signal params (only positional params allowed), cannot return value
+  * Can have optional `dynamic` boolean which, if set to true, means the signal is a catch-all for otherwise unhandled
+    signals. `name` cannot be present and there must be two params: a name param and a varargs param
+    * An alternative approach that is similarly Pythonic would be to have a `__temporal_signal__` function that could be
+      defined to catch them, but the decorator is clearer
```
Sushisource, line 164: The special function has the benefit that there can only be one, which should be the case. It's uglier, though. Probably we just want to enforce at init time that you've only defined one catch-all handler.

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+#### `temporalio.workflow.continue_as_new()`
+
+```python
+# Overload for self (cannot type args)
+@overload
+def continue_as_new(
+    *args: Any,
+    # ... (excludes "workflow" kwarg)
+) -> NoReturn:
+    pass
```
Sushisource, line 579: It's just too bad this isn't typed.

Could the workflow.run decorator shadow this definition to use the appropriate types so that it works out when called within workflow implementations?

python/phase-2.md
cretz: Agreed

python/phase-2.md
cretz: I can't think of a clean way to do this at compile time. TS hasn't solve this either, they also don't implicitly type the continue-as-new for self.

The only somewhat-reasonable way I could think of doing this would be to require the function passed in for every continue-as-new, even the self one.

This is one of the few costs of not having an explicitly generic typed context that we pass around everywhere and put all of these methods on (but of course the benefits of not having a Python user manually type and pass around some generic context everywhere are much larger).

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+  * Ability to unset the current instead of just forcing a detached one each time you want to opt out?
+* Should we disallow the `parent` to even be set in the constructor?
+  * Thinking we probably shouldn't allow you to set that
+* Cancellation of a workflow or a scope causes all things started with that to be issued a cancel
+  * Originally we weren't gonna use cancellation scopes but instead reuse the `asyncio` cancel feature but that throws
+    inside of awaited things and we want to cancel whether awaited or not
+  * This means `asyncio` cancellation (and `shield` etc) affects only in-workflow tasks awaited on, _not_ the actual
+    underlying activity/child being executed
+    * We could disable these features of `asyncio` so people aren't confused by their concept of cancellation and ours,
+      but there is still value in cancelling just the tasks
```
Sushisource, line 805: What's the value in cancelling the tasks? I'd be inclined to say just remove this potential bit of confusion, but, since it will break at runtime maybe that's of dubious value anyway. Seems like a footgun people will run into but I dunno that there's any easy fix

python/phase-2.md
Sushisource: I would be fine with just eliminating this and requiring you to pass the workflow type. Don't feel super strongly about it though.

python/phase-2.md
bergundy: My main point is that we should be flexible enough to support different registration methods.
If users want to use a "collection" the API should support it, I never said we should provide a collection primitive.

If the worker accepts a mapping or an iterable that should be flexible enough.

If y'all want a mandatory decorator on activity functions I'm fine with that although I don't see the value apart from overriding the activity name TBH.

python/phase-2.md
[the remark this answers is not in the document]
bergundy: I do think that it's a very reasonable expectation that we could translate kwargs to positionals given a signature but I could live with this limitation for the sake of having kwargs reserved for activity and workflow options.

python/phase-2.md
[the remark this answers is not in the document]
bergundy: FTR, I'd be fine with the proposed solution but to address your concerns:

> It's hard to separate "start" from "execute"

From the type system perspective?
Are you saying this isn't doable?
 
> It's hard to separate run entrypoint from signals from queries from sync updates from just normal methods

You could solve that with a convention.

> It's hard for readers to know what's being done ("is that a query or a signal?")

I haven't tried this is in Python but with TS you could infer the possible signals given a class definition and still preserve the `handle.signal('name')` signature (in TS `name` would be `keyof T['signals']`).

> You end up either down an ugly "untyped stub" path or an ugly "completely different ways to start a workflow" path

In TS this wasn't the case, I haven't looked deeply enough into the Python type system to see if that'd be possible.

python/phase-2.md
bergundy: 👍

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    async def run(arg1: str) -> str:
+        return await workflow.execute_local_activity(
+            say_hello, arg1, start_to_close_timeout=timedelta(seconds=5)
+        )
+
+
+async def main(stop_event: asyncio.Event):
+    client = await Client.connect("http://localhost:7233")
+    async with Worker(
+        client,
```
macrogreg, line 40: Is `client` the right name here. Technically it is, but semantically you are actually developing a "host" (aka an app that serves workflow worker functionality).
Can this be named "connection" or similar?

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+from datetime import timedelta
+from temporalio import activity, workflow
+from temporalio.client import Client
+from temporalio.worker import Worker
+
+@activity.defn
+async def create_greeting_activity(salutation: str, name: str) -> str:
+    return f"{salutation}, {name}!"
+
+@workflow.defn
```
macrogreg, line 67: What is the value of having the `workflow.defn` tag? IF it does something - what? If not, why have it?

python/phase-2.md
```diff
@@ -0,0 +1,830 @@
...
+    async def run(self, name: str) -> str:
+        while True:
+            # Store greeting (this is typed)
+            self._current_greeting = await workflow.execute_local_activity(
+                create_greeting_activity,
+                self._salutation,
+                name,
+                start_to_close_timeout=timedelta(seconds=5),
+            )
+            workflow.logger.debug("Greeting set to %s", self._current_greeting)
```
macrogreg, line 84: How / where did you set the log destination?

python/phase-2.md
bergundy: `execute` matches the API, not `start`, the client `execute_workflow` method copies this signature.

python/phase-2.md
bergundy: Or if we add more overloads

python/phase-2.md
bergundy: 👍

python/phase-2.md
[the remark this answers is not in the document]
cretz: Yes, but it makes cross-SDK typing a bit weird (e.g. if you have a Go function that takes two parameters, what does that look like from a Python client that always wraps). After in-meeting discussion, I think we've settled on multiple top-level positional parameters being acceptable and serialized as such in the payload like in other SDKs.

python/phase-2.md
bergundy: I haven't written any serious Python in a while so that would be an empty statement :-P

python/phase-2.md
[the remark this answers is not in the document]
bergundy: I think there's value in allowing reuse of well known Python APIs, it puts less of a barrier for adoption by reducing the amount of new APIs a developer needs to learn.

python/phase-2.md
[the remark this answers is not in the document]
cretz: > FTR, I'd be fine with the proposed solution

:+1:

> From the type system perspective?  Are you saying this isn't doable?

It's doable but ugly. It's why in Java you have `WorkflowClient.start(myWorkflow::myMethod)` for start and `myWorkflow.myMethod()` for execute. Two completely different ways do mostly the same thing whereas here it's `client.start_workflow(MyWorkflow.my_method)` or `client.execute_workflow(MyWorkflow.my_method)` to do either. (same for activities and child workflows and such).

> I haven't tried this is in Python but with TS you could infer the possible signals given a class definition and still preserve the handle.signal('name') signature (in TS name would be keyof T['signals']).

Python has the concept of literals, but not really that advanced. Technically we could make the `@workflow.signal` wrap the method and return something like `WorkflowSignal[ParamType]` that we can then limit the client side to, but I am trying at least at first to not have the decorators make wrappers (yet).

> In TS this wasn't the case, I haven't looked deeply enough into the Python type system to see if that'd be possible.

The untyped stub may not be _that_ ugly, just really untyped, heh. It's one thing to `handle.signal("my_signal_name")` which is clearly dynamic, it's quite another to `stub.my_signal_name()` which makes one think of a defined method. This gets even crazier if the signal name does not match Python function naming standards.

python/phase-2.md
cretz: >  If users want to use a "collection" the API should support it, I never said we should provide a collection primitive.

I think it's ok to ask them to convert to a list. The user can do this:

```python
my_activities = {"act-1": my_func1, "act-2": my_func2}
to_register = [activity.defn(fn, name=name) for name, fn in my_activities]
```

python/phase-2.md
cretz: I don't mind changing to `@workflow.execute` if others agree too.

python/phase-2.md
cretz: It marks a class as a workflow and all the validation/constraints that come with that and it is a place where the name is set (or derived from the thing below it). We have no other options at this time, but it is important to use as a way to mark this class as a workflow with those inherent constraints.

python/phase-2.md
cretz: This is a Python thing. See https://docs.python.org/3/howto/logging.html, https://docs.python.org/3/howto/logging-cookbook.html, https://docs.python.org/3/library/logging.html, etc. Think of it like slf4j+logback or log4j in Java. Temporal SDKs provide a workflow logger that is smart enough to not log during replay.

python/phase-2.md
bergundy: After taking a stab at functional workflows with Python, I'm now on "team classes".

Below is what the TS API would look like translated to Python.
The main issue I see with this approach is that Python doesn't have anonymous functions (it does have lambdas but those are quite limited).
This limitation requires extra typing which we can avoid with the class based approach.

```python
class State(dataclass):
  blocked: bool

unblock_signal = workflow.define_signal('unblock')

async def my_workflow(start_blocked: bool):
  state = State(blocked=start_blocked)

  @workflow.signal(unblock_signal)
  def unblock():  # Would have been better as an anonymous function
     state.blocked = False

  await workflow.wait_condition(lambda: not state.blocked)
  ...
```

To summarize, while there are advantages to the functional approach and a lot of the concerns of associating signals and queries with a workflow and verifying that those are handled can be done with the functional approach, it would be less convenient in Python, I'm dropping this issue.

python/phase-2.md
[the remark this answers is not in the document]
bergundy: After our zoom meeting, I realized that what I was suggesting is very close to what is stated in this proposal given that the signal and query references passed to "typed handle" methods must be members of the workflow class that was used to create the handle.

python/phase-2.md
cretz: Technically you could have:

```python
@workflow.signal_defn
def unblock_signal(some_arg: str) -> None:
    ...

@workflow.defn
async def my_workflow() -> None:
    @unblock_signal
    def unblock(some_arg: str) -> None:
      doSomething()
```

And I think I could type it, but that's a lot of magic.

python/phase-2.md
cretz: I support another verb here if we want, but I think we still need it to start with a verb.

python/phase-2.md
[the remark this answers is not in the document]
cretz: The question is whether learning a new API is worse than the existing API not working exactly like you expect and Python documents (e.g. sub-second sleeps/timeouts).

Sushisource (approved):

python/phase-2.md
cretz: I have updated this to allow base classes to have the decorator but it is not inherited

python/phase-2.md
cretz: I have added a new section concerning the separation of contract and impl

python/phase-2.md
cretz: Done

cretz:
I have added new "Activities as Methods" and "Workflow Contracts" sections to the bottom.

@bergundy 
> There are some missing Workflow APIs in this proposal but it's okay, it doesn't need to be complete.

Which ones are missing?

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+        pass
+    
+    @activity.defn
+    def some_activity(self, arg1: str) -> List[str]:
+        return ["Yay"]
+```
+
+You cannot just `workflow.execute_activity(ActivityClass.some_activity, "foo")` because technically the first parameter
+is `self` there. But sometimes they may want to call
+`workflow.execute_activity(ActivityClass.some_activity, ActivityClass("init-arg1"), "foo")` and who are we to say they
```
Sushisource, line 852: This specifically we really do want to prevent I believe. The workflow calling the activity cannot make assertions about the state of the activity class since it may well be run on another worker. So, allowing them to pass in instance in the workflow code is pretty misleading.

python/phase-2.md
```diff
@@ -828,4 +829,168 @@ Notes:
...
+        arg2: LocalParamType2,
+        /,
+        *,
+    ```
+  This means if you had a real two-param activity, you'd now be able to have an optional first parameter. Also, I am not
+  yet confident how this will look when we remove all of these overloads once Mypy gets their `ParamSpec` support.
+* We could have a completely separate call like so:
+
+    ```python
+    workflow.execute_activity_method(ActivityClass.some_activity, "foo")
```
bergundy, line 900: I don't have a strong preference here, some other options you've mentioned are completely valid too.

This is the cleanest IMHO with the current limitations of python typings.

The proxy approach introduces new concepts that I think we should avoid, it also make goto definition harder (lessons learned from TS).

The other valid option is have the `method=True` flag but I feel like the signature would be more confusing with additional overloads.

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+    activities = ActivityClass.__new__(ActivityClass)
+    workflow.execute_activity(activities.some_activity, "foo")
+    ```
+  This is nice because it doesn't require the Python SDK to care (and people have been doing this in class methods for
+  factories for a while now). We can even provide a helper that does basically this like so:
+
+    ```python
+    activities = workflow.proxy_activity_class(ActivityClass)
+    workflow.execute_activity(activities.some_activity, "foo")
+    ```
```
Sushisource, line 876: This is probably my favorite option in this list but I still don't completely love it.

Pretty sure you already said this but there's really no way to have the decorator detect and strip the self param for usage specifically in this situation when being called from a workflow? It's a bit like what you suggest with wrapping the function below, except without needing to have `defn` and `method_defn` be different things. If we did that, I'm worried people would forget constantly and use the wrong one (although, ideally we could just blow up at init time if they did)

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+    ```python
+    @overload
+    async def execute_activity(
+        activity: Callable[[Any, LocalParamType], LocalReturnType],
+        arg: LocalParamType,
+        /,
+        *,
+        method: Literal[True],
+    ```
+  This seems like a fairly sane way, but requires callers pass a special parameter to call
```
Sushisource, line 923: Sane, but, 100% forgettable and if I understand correctly, won't blow up until they actually try to execute the activity if they do

python/phase-2.md
```diff
@@ -828,4 +829,168 @@ Notes:
...
+
+```python
+@workflow.defn
+class MyWorkflow(ABC):
+  @workflow.run
+  @abstractmethod
+  def run(self, arg: str) -> List[str]:
+    raise NotImplementedError
+
+@workflow.defn(name=MyWorkflow.__name__)
```
bergundy, line 964: Is this required with a subclass?

python/phase-2.md
```diff
@@ -828,4 +829,168 @@ Notes:
...
+    return Worker(workflows=[MyWorkflowImpl])
+```
+
+Notes:
+
+* This has no special logic to handle this approach. The same approach can be used to separate activity class method
+  contracts from their impls.
+* It is unfortunate that we require a `@workflow.defn` on both the base and impl classes.
+  * We want to require this decorator at client-side _and_ worker-side to ensure name sharing
+  * With `Protocol` it is obvious why, there is technically no code shared
```
bergundy, line 983: Maybe it'd work if we supported `@workflow.defn(protocol=MyWorkflow)` where the decorator would validate (at runtime) that the class implements the protocol and would "copy" the method decorators over..

python/phase-2.md
```diff
@@ -828,4 +829,168 @@ Notes:
...
+  * With abstract base classes, we maybe could support `@workflow.defn(extends=MyWorkflow)` but that doesn't give you
+    much
+  * I am hesitant to try to implicitly derive the base name because there are other use cases for inheritance besides
```
bergundy, line 986: 👍

python/phase-2.md
```diff
@@ -828,4 +829,168 @@ Notes:
...
+    just contract/impl separation
+* It is unfortunate that we require a `@workflow.run` on both the base and impl classes.
```
bergundy, line 988: All of these complexities with abstract classes make me wonder if it's worth supporting them at all.
Maybe we shouldn't support inheritance if we can support protocols.

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+
+Notes:
+
+* This has no special logic to handle this approach. The same approach can be used to separate activity class method
+  contracts from their impls.
+* It is unfortunate that we require a `@workflow.defn` on both the base and impl classes.
+  * We want to require this decorator at client-side _and_ worker-side to ensure name sharing
+  * With `Protocol` it is obvious why, there is technically no code shared
+  * With abstract base classes, we maybe could support `@workflow.defn(extends=MyWorkflow)` but that doesn't give you
+    much
```
Sushisource, line 985: Why does it not give you much? It should be able to set the name correctly, yes? If so, you could do 

`@workflow.defn(implements=MyWorkflow)` for protocols and I think that's quite nice.

I have no problem with both places declaring the same decorators. I think it's a feature rather than a downside. It's double checking you're consistent, and it makes it visually obvious without referring to the protocol/base what the class is doing. Super minimal amount of extra typing for some additional clarity.

python/phase-2.md
bergundy: Or at least not support abstract classes but I'm not convincing myself.

bergundy:
> I have added new "Activities as Methods" and "Workflow Contracts" sections to the bottom.
> 
> @bergundy
> 
> > There are some missing Workflow APIs in this proposal but it's okay, it doesn't need to be complete.
> 
> Which ones are missing?

AFAICT these ones:

- `patch`
- `deprecate_patch`
- `random`
- `uuid`

Also if logging is built-in you'll need to support metrics too.

bergundy:
Sorry, accidentally closed this.

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+    async def complete_with_greeting(self) -> None:
+        self._complete.set()
+
+    @workflow.query
+    async def current_greeting(self) -> str:
+        return self._current_greeting
+
+def test_workflow(client: Client) -> None:
+    # Start the workflow
+    handle = await client.start_workflow(GreetingWorkflow.run, "John")
```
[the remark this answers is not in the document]
bergundy: https://github.com/temporalio/proposals/pull/55/files#diff-2f71769a191b3cb7cd9cf0dafd3d0ce210744f634a028078edad2fe46628c417R927

python/phase-2.md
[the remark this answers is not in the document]
bergundy: If the type is defined (optional in Python) we will use it as a hint

python/phase-2.md
```diff
@@ -0,0 +1,996 @@
...
+        self._salutation = "Hello"
+        self._salutation_update = asyncio.Event()
+        self._complete = asyncio.Event()
+
+    @workflow.run
+    async def run(self, name: str) -> str:
+        while True:
+            # Store greeting (this is typed)
+            self._current_greeting = await workflow.execute_local_activity(
+                create_greeting_activity,
```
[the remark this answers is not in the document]
bergundy: https://github.com/temporalio/proposals/pull/55/files#diff-2f71769a191b3cb7cd9cf0dafd3d0ce210744f634a028078edad2fe46628c417R836

python/phase-2.md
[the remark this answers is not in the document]
cretz: I don't think, at least currently, there is enough value to have payload metadata unique to the Python SDK just for named parameters. If there does become enough value, we can add it.

python/phase-2.md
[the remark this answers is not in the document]
cretz: Yes, we use generics in `execute_workflow` to keep the type (of course many Python developers don't develop with types, but even type inference will work in most IDEs).

python/phase-2.md
cretz: In this case it's just passing as a parameter, just as if it was a top-level function and `ActivityClass` was the first parameter instead of "self". At least at compile time, we can't tell them they can't pass the first legitimate parameter of a two-parameter function.

python/phase-2.md
cretz: > This is probably my favorite option in this list but I still don't completely love it.

I too don't like it, but there is no perfect option at the moment to separate top-level functions from methods

> there's really no way to have the decorator detect and strip the self param for usage specifically in this situation when being called from a workflow?

Not at compile time because the decorator at compile time can't differentiate between method w/ "self" and top-level function with just a param. They just look like params at compile time (at runtime I can inspect and determine it's a method instead of a top-level function).

> If we did that, I'm worried people would forget constantly and use the wrong one

Concur

python/phase-2.md
cretz: It will blow up with the type checker (because it'd pick the other overload which would require two parameters, one for self and one for the real parameter). But we have to remember that many Python developers don't include type checking. There is no common Python compilation step to catch this stuff anyways. Of course at runtime I can do all sorts of things to fix or error in bad situations.

python/phase-2.md
cretz: See points below this about annoying inheritance concerns.

python/phase-2.md
cretz: > Why does it not give you much? It should be able to set the name correctly, yes? If so, you could do

It doesn't give you much because you're already extending `MyWorkflow` with the Python syntax. I don't need a custom extension approach too. I can derive it at runtime. The only benefit would be getting the name right, but if we move to proper inherited decorators that benefit wouldn't even exist either.

> If so, you could do `@workflow.defn(implements=MyWorkflow)` for protocols and I think that's quite nice.

This bullet was discussing abstract base classes. Protocols are meant to be separate from the implementation. In Python if you want explicit "implements" you'd use an abstract base class and properly extend. There is no benefit to using a Protocol over an abstract base class if you're already sharing code.

> I have no problem with both places declaring the same decorators

My only problem is getting the name right. The two classes need to have the same name.

python/phase-2.md
cretz: > Maybe we shouldn't support inheritance if we can support protocols.

There is no reason not to support inheritance. The only problems you are seeing here are brought on by the fact that you want to separate contract from the implementation . Normal inheritance poses no problems and we need to have no special support for it.

cretz:
@bergundy - Lets consider versioning and random helpers as phase 3

python/phase-2.md
```diff
@@ -923,6 +924,47 @@ What do we do? Here are the notes/options (activity options left off of calls fo
...
+
+        def some_activity1(self, arg: str) -> List[str]:
+            ...
+
+        def some_activity2(self, arg: int) -> List[int]:
+            ...
+
+    some_activity1 = workflow.activity_method_reference(ActivityClass.some_activity1)
+    some_activity2 = workflow.activity_method_reference(ActivityClass.some_activity2, name="some_other_name")
+
```
Sushisource, line 943: Thanks for writing this. Personally, I don't mind it, but I can see how people would get annoyed by it for sure. Given that the other approach does type-level prevent you from using the wrong call for the type of activity, I'm happy with that.

python/phase-2.md
```diff
@@ -993,4 +1035,192 @@ Notes:
...
+* You have to opt-in to this
+
+Instead, we can offer this in the `temporalio.workflow` package:
+
+```python
+def set_signal_handler(name: str, handler: Callable) -> None:
+    ...
+
+def set_query_handler(name: str, handler: Callable) -> None:
+    ...
```
Sushisource, line 1160: What I'm wondering is if we can do something like this (solving the req-resp problem as an example). Signals are still statically defined, but, reusable.

```python
class ReqRespSignalUser:
    def __init__(self):
        self.res = asyncio.Future()

    async def request_handler(self, arg):
        raise NotImplementedError

    async def make_request(self, external_wf_handle, arg):        
        await external_wf_handle.signal(ReqRespSignalUser.req_resp_sig, "hi") 
        return await self.res

    @workflow.signal
    async def req_resp_sig(self, req_or_resp: str) -> None:
        # Made up but you get the idea
        if req_or_resp.is_req:
            self.request_callback(req_or_resp)
        else:
            self.res.set_result(req_or_resp)

@workflow.defn
class MyWorkflow(ReqRespSignalUser):
    # Now this workflow can call self.make_request(other_workflow, "whatever")
    
    def request_handler(self, arg):
        return "request handled bruh"
```

mfateev: What about a class that is not part of the workflow? Think about a library class.

Sushisource: I think this works for that - since the library author would have to know to `set_signal_handler` as the traditional way, instead they make a class like this one which explicitly defines the signal handler, and then that class can be inherited by the workflow, or by another class which is eventually inherited by the workflow, etc.

cretz: The problem described in this section is for dynamically _named_ signals/queries. The req/resp for the _same named_ signal is next section. The solution you're showing doesn't relate to dynamically named signals.

We need `set_signal_handler` and `set_query_handler` for dynamically named signals/queries. We could use user-land code and the catch-all signal/query handler but I listed the downsides. After some thought, I figured the ability to set these handlers from inside the workflow was acceptable (though we'd encourage the other approach of course).

Sushisource: What I'm getting at is that dynamic names are not really required. You can accomplish all the same things you'd want to do with them by switching on some message type, as you know and we talked about before.

So the question becomes, what do dynamic names solve for you, and being able to have self-contained logic that depends on signals is one reason and my example shows how you can still have that self contained behavior w/o a dynamically named signal.

So I'm wondering if we can come up with a motivating example that shows how dynamically named signals are actually needed to solve some problem that can't be solved this way, I can't quite think of one and that'd be a good thing to have so we can know "Oh ok we do need them for such and such reason"

cretz: > So I'm wondering if we can come up with a motivating example that shows how dynamically named signals are actually needed to solve some problem that can't be solved this way, I can't quite think of one and that'd be a good thing to have so we can know "Oh ok we do need them for such and such reason"

I've been told "DSL"s, e.g.

```python
@dataclass
class DslInput:
  signals: Dict[str, Action]
  queries: Dict[str, TypeOfData]

@workflow.defn
class MyDslWorkflow:
  @workflow.rum
  async def run(input: DslInput):
    for name, action in input.signals
      workflow.set_signal_handler(name, action_handler_for(action))
    for name, type_of_data in input.queries
      workflow.set_query_handler(name, query_handler_for(type_of_data))
```

The reason we would support this natively instead of the user-land `dynamic=True` approach of my mixin above are in the bullets I laid out.

cretz:
I think this proposal has about run its course. Based on discussions herein and off-PR, I think we have arrived at a reasonable set of features. Of course some of this may change at implementation time as we learn more.

Last call for anything else that seems off here before I merge.

python/phase-2.md
Sushisource: Seems to me this can still easily work the same way as my example - the DSL input can be switched on, either directly in the workflow or in some kind of mix-in like my example.

We know we get some nice benefits with static only signals, and it still seems to me nothing necessitates dynamic ones, or is even really substantially easier with them.

If we want to add it to stay consistent, I suppose that's fine, but seems like a good opportunity to simplify in the absence of a super compelling reason to keep them

cretz: > Seems to me this can still easily work the same way as my example - the DSL input can be switched on, either directly in the workflow or in some kind of mix-in like my example.

But you don't get custom names driven by the input, or if you do, you don't get the benefit of Temporal logic for not-found signal/query handlers/names.

> We know we get some nice benefits with static only signals, and it still seems to me nothing necessitates dynamic ones, or is even really substantially easier with them.

The example I gave of custom, dynamic naming while still getting all the benefits of the default Temporal logic handling of unknown signals/queries necessitates it. I think to support your argument you have to either:

* Convince everyone that you don't need dynamically named signals/queries (e.g. from workflow input)
* Show how you could do this at runtime and still get the same Temporal logic/errors of unhandled signals/queries

> If we want to add it to stay consistent, I suppose that's fine, but seems like a good opportunity to simplify in the absence of a super compelling reason to keep them

I don't want to take that opportunity here personally to make that decision. But I definitely agree that we should discourage this approach, probably going as far as not documenting its use when documenting normal signal/query documentation.

cretz: What I can do if it's worth it is mark `set_signal_handler` and `set_query_handler` as experimental (instead of just discouraged) and if you can convince the team of removal, I can remove.

python/phase-2.md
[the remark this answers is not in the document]
mfateev: I like this consistency especially as do plan to support signaling and querying activities in the future.

cretz:
Merging!

PR #56, opened 2022-03-29
TS connections proposal

bergundy:
New public API for configuring starter and worker connections.


typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+- Verbose
+- Confusing when to use each connection type
+
+### Implicit connection reuse with similar options
+
+```ts
+// worker.ts
+import { Worker } from '@temporalio/worker';
+
+// worker1 and worker2 share the same underlying connection based on given options
```
cretz, line 54: Can you give details on how this is achieved? Are you using some kind of shared global state in TypeScript?

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+```
+
+#### Pros
+
+- One way to create clients
+- Explicit sharing model
+
+#### Cons
+
+- Connnection isn't really reused (not possible)
```
cretz, line 124: Why is it not possible to reuse the connection? Is it because the client is pure TS gRPC and the worker is core? If we're not going to back the client by core, then that limitation exists for every client vs worker wrt connection reuse. 

But it doesn't mean connections can't be used _across workers_. You just store an internal, lazily created Core connection no the client when it's first needed. Then every worker created from that client can use the same connection.

The only "Con" I see with this approach if pure TS gRPC client remains is that for worker-only situations, you are creating/connecting via pure TS gRPC unnecessarily.

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+
+- Short
+- Easy path from sample with local server to real world use case
+
+#### Cons
+
+- Implicit connection sharing is non-obvious
+- No way to opt out of connection sharing (can be addressed if needed)
+
+### Mix of options and explicit clients (recommended)
```
cretz, line 74: Why recommended/ And why not list the cons of having multiple ways to create a connection and having the user have to care about connection reuse to get its benefits?

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+// starter.ts
+import { Connection, WorkflowClient } from '@temporalio/client';
+
+const client = await WorkflowClient.create({ connectionOptions: { address }, namespace });
+
+const connection = await Connection.create({ address });
+const client = new WorkflowClient({ connection, namespace });
+```
+
+### Python SDK approach
```
cretz, line 100: FWIW, this is also the Go and Java SDK approach. This is more accurately called the "Every SDK we have except TypeScript approach".

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+
+There's no benefit in having 2 client classes except for shorter method names on
+each of them (e.g. `WorkflowClient.execute` would need to be named `Client.executeWorkflow`).
+
+Some of the proposed alternatives above would make more sense if the SDK exposed a single `Client` class.
+
+### Why do we need both native and JS clients?
+
+- Rust Core does not support pluggable clients and must use a native client.
+- We do not want native extensions as a (strict) dependency for "starters" because those might be deployed on environments where
```
cretz, line 139: This is unfortunate but understandable. I do wonder if we could flip the script a bit and give core a gRPC client backed by TS. That is gross too of course. But it should be noted with this limitation, we flat out can't share connections regardless of approach shown.

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+import { Worker } from '@temporalio/worker';
+
+// worker1 and worker2 share the same underlying connection based on given options
+const worker1 = await Worker.create({ connectionOptions: { address }, namespace, taskQueue: 'tq1' });
+const worker2 = await Worker.create({ connectionOptions: { address }, namespace, taskQueue: 'tq2' });
+
+// starter.ts
+import { WorkflowClient } from '@temporalio/client';
+
+const client = await WorkflowClient.create({ connectionOptions: { address }, namespace });
```
lorensr, line 61:
```suggestion
const client = await WorkflowClient.create({ connection: { address }, namespace });
```

I like this shorter option name (for client and worker). I think it would work in this scenario, in which passing in a Connection instance isn't an option.

typescript/connections.md
bergundy: Yes, we'd have to use some shared global state

typescript/connections.md
bergundy: I regret recommending it. I want to think about this some more

typescript/connections.md
bergundy: It's not the Java approach because Java has `WorkflowClient` and the service stub.

typescript/connections.md
bergundy: If we default to using the Core client we could reuse connections, I'm debating whether we want to go that route.
The JS client has more features and flexibility so it'd be hard to give it up.

typescript/connections.md
bergundy: Flipping it might be the best direction for this SDK but we're not quite there yet.

typescript/connections.md
cretz: Definitely don't need to give up the JS client. I just want to make sure that connection reuse across workers is clear.

typescript/connections.md
bergundy: Yes, but that'd be more like the explicit connection proposal not passing around the high level client.

typescript/connections.md
bergundy: Yes, for sure possible across workers but not between a worker and a client.

typescript/connections.md
```diff
@@ -0,0 +1,151 @@
...
+#### Pros
+
+- Explicit connection reuse
+
+#### Cons
+
+- Verbose
+- Confusing when to use each connection type
+
+### Implicit connection reuse with similar options
```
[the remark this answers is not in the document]
Sushisource: This is also my favorite. The somewhat obfuscated nature of the perf/resource usage isn't so much a concern in Node where things like that are already happening left and right.

It solves the problem of unifying the API without requiring dramatic work inside Core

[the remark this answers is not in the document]
cretz: This is my least favorite. You gonna hash the connection options and then store a connection based on that hash in some global map? This is very unclear to me from a dev POV.

typescript/connections.md
```diff
@@ -0,0 +1,209 @@
...
+```ts
+// worker.ts
+await Core.install({ serverOptions: { address, namespace } });
+// Worker uses a connection created by the Core singleton
+const worker = await Worker.create({ taskQueue });
+```
+
+## Proposals
+
+### Explicit `Connection` objects
```
cretz, line 18: After discussion and re-review, I am afraid I begrudgingly support this option as best.

I think it is very important to have these goals:

* Connection sharing should be default, not opt-in
* That you are sharing a connection should be clear
  * Worker creation therefore must be a two-step process since they need to be created from a common base of shared something
* Prefer to avoid global state

I think asking them to create a connection, just as if gRPC required a channel to be created explicitly (which I know it does not in all gRPC impls, but it does in many) is the best way forward. It does mean an extra step. I'd also add:

* I would like to just call it `WorkerConnection` or something similar I dunno. `NativeConnection` reads a bit rough to me.
* I would prefer `WorkflowClient` constructor just accept the connection instead of `connection.service`. It just feels more right to create a client from a connection than from a connection service (you can type the param as `connection: { service: WorkflowService }` if you must but I prefer accepting exactly what's returned from `Connection.create` since that's easier to understand where it came from for API browsing).

lorensr: > That you are sharing a connection should be clear

Why is this important for majority of devs to know?

bergundy: I tend to agree with @cretz, and suggest that we go with "Explicit Connection objects".
That approach is explicit, future proof for when we want to support different connection types, and flexible (for example we could extend this API to create implicit connections when given options).

bergundy: I'd like to call the connection `NativeConnection` in case we want to use it in the client in the future or we support running workers with non-native connections.

I would change the `WorkflowClient` constructor but I don't want to tie that change to this proposal because I might take it a step further and unify the 2 clients into a single `Client` in the future.

cretz: > Why is this important for majority of devs to know?

Clarity of when connections are being made is helpful like any other form of clarity. We need a really good reason to argue against clarity. It's a tradeoff. The question is whether it's more important for them not to know.

eplarson: Another reason this approach is good: dependency injection. In frameworks like Nest.js, the most natural thing is to create a connection within the DI context, then inject it into workers (or wherever else) upon creation. 

Even without explicit connection creation, you'd still want to create a `ConnectionOptions` object within DI context and inject _that_ into the workers. So a 2-step creation is a given when using DI, and it's much clearer to just create the connection directly.

Regarding the confusion about when to use each connection type, it seems like that could be solved with more specific types for `Worker.create()` and `new WorkflowClient()`.

Sushisource (approved):
Explicit types gewd

typescript/connections.md
```diff
@@ -0,0 +1,221 @@
...
+const connection = await NativeConnection.create({ address });
+const worker = await Worker.create({ connection, namespace, taskQueue });
+
+// starter.ts
+
+// same as today but Connection becomes async to make the `getSystemInfo` call
+import { Connection, WorkflowClient } from '@temporalio/client';
+
+const connection = await Connection.create({ address });
+const client = new WorkflowClient(connection.service, { namespace });
```
lorensr, line 36:
```suggestion
const client = new WorkflowClient(connection, { namespace });
```

+1 to Chad's suggestion

typescript/connections.md
lorensr: > Why is it important to hide it from all devs?

Simplicity / lower mental load for the majority for whom connection constructing doesn't matter. And it wouldn't have to be completely hidden—`Worker.create` API doc can say "Creates a gRPC connection to the server, or reuses an existing connection if another Worker has previously been created with the same `connectionOptions`".

> That approach is explicit, future proof for when we want to support different connection types, and flexible (for example we could extend this API to create implicit connections when given options).

I'd be happy with this extension, and using it for most of samples. In which case the above API doc could be in `WorkerOptions.connectionOptions` rather than `Worker.create`.

mjameswh: Seems I'm a little bit to the show, but for what it it's worth, here's my two cents...

First of all, I think that being explicit on connection instantiation will be wonderful from a DI point of view.

I'm however concerned that, in its current form, this proposal exposes the user to some lower-level aspects of how Temporal TS works. This might not be an issue for users that are already comfortable with Temporal, but may be a turn off for new users.

Here are some of the irritants I see:
1) as a new user, I need to know that there are two types of connection object, and I have to know which one to provide when creating a client or setting up a worker... And even worst, this knowledge is likely to change at some point in time which means that code examples I see on the internet may not work or may be non optimal in the future.
2) as a user, I need to understand the cost associated with the connection object and therefore, that I am in charge of reusing the connection object. Simply having a connection object doesn't make it explicit that I am meant to reuse it and that doing so is desirable.
3) Knowing that I can reuse the connection now makes me wonder about the lifecycle of objects I create in association with that connection. Am I expected to have some kind of global connection that I expose to my whole application, and then create a client attached to it whenever I need one, and then forget about it? Is there a cost associated ok to have multiple clients attached to one connection? Do I need to explicitly dispose of client/workers? How do I dispose of a connection, and what will be the consequence on worker/clients I created from it? All of this can be clarified with good documentation, but that means that I, as a user, have to take the time to read that doc...

bergundy: I've been putting off this reply for a while, sorry for that.

You raise some valid points as usual @mjameswh.

> as a new user, I need to know that there are two types of connection object, and I have to know which one to provide when creating a client or setting up a worker... And even worst, this knowledge is likely to change at some point in time which means that code examples I see on the internet may not work or may be non optimal in the future.

I see us at some point soon adding a method to create workers and clients from connection options instead of passing around connection objects.

 > as a user, I need to understand the cost associated with the connection object and therefore, that I am in charge of reusing the connection object. Simply having a connection object doesn't make it explicit that I am meant to reuse it and that doing so is desirable.

I agree, this would be solved by accepting connection options and defaulting to implicitly sharing connections.
That being said, I'm not convinced this is the right approach. In general I'd like to avoid implicit global state and not sure how to solve this differently.

> Knowing that I can reuse the connection now makes me wonder about the lifecycle of objects I create in association with that connection. Am I expected to have some kind of global connection that I expose to my whole application, and then create a client attached to it whenever I need one, and then forget about it? Is there a cost associated ok to have multiple clients attached to one connection? Do I need to explicitly dispose of client/workers? How do I dispose of a connection, and what will be the consequence on worker/clients I created from it? All of this can be clarified with good documentation, but that means that I, as a user, have to take the time to read that doc...

Good point, the answer is yes, you're supposed to keep a single connection (per role - starter / worker) and reuse that throughout the app. Client objects are relatively cheap so those don't have to be reused but typically we suggest reusing them.
As a rule of thumb, I'm trying to keep "expensive" constructors in async factory functions (e.g. `create`), that should be a guiding factor for users to decide what to reuse.
While this answer might not be completely satisfactory, we'll make up for this complexity with good samples and docs.


I'm open to other suggestions but I feel like we took the most forwards compatible and flexible path without introducing implicitness and global state.

PR #59, opened 2022-06-30
Add proposal for VSCode debugger plugin

bergundy:


debugger.md
```diff
@@ -0,0 +1,160 @@
...
+    Note over VSCode,Lang: supply workflow history path and replay server port
+    Lang->>Core: start replayer and listen on port
+    Note over Lang,Core: set breakpoint on first workflow task
+    Core->>Lang: activate workflow
+    Lang->>Debugger: break
+    Debugger->>VSCode: break
+    VSCode->>-User: debugger paused on breakpoint
+```
+
+#### Set breakpoint on workflow task
```
lorensr, line 137: Can you also set a breakpoint on a line of code?

bergundy: That's just standard debugging

lorensr: Right—my question is whether it still works if I start execution via the main panel. Also wondering if the history list will update, like highlighting the current task.

bergundy: Yes

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+The current workflow task will be highlighed when being processed.
+
+#### Debug adapter
+
+Use to step over SDK code and customized per language
+
+#### Core replay debug server
+
+A webserver opened by Core, listens on port allocated by the VSCode plugin.
+Has a single endpoint for setting breakpoints on workflow task started events (by event ID).
```
bergundy, line 80: Totally forgot the other API to push the currently processing workflow task to vscode

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+- Core replay debug server
+- Lang middleware
+
+#### Main panel
+
+Has input to enter workflow ID and optional run ID with submit button to start a debug session.
+Alternatively can have a file (path) input to use a history JSON or binary file (as exported by web and tctl or
+downloaded with a client).
+
+The credentials for connecting to a server should be stored in the system keychain (can use
```
cretz, line 53: Just researched, there's a vscode secrets API for this: https://code.visualstudio.com/api/references/vscode-api#ExtensionContext.secrets (commit: https://github.com/microsoft/vscode/commit/d3965a2b4dd54f1a415fa370fea90b84bc7d9b7c, associated PR specifically mentions so extension authors don't need keytar)

Also may need a bit info on where/how the server address and creds are inputted by the user. Also would suggest supporting in-vscode-config creds if they so choose. No detail needed, just want to know how I point to my server. Is there a vscode "servers" concept that can be leveraged?

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+
+- VSCode plugin
+  - Main panel / Workflow history panel
+  - Debug adapter
+- Core replay debug server
+- Lang middleware
+
+#### Main panel
+
+Has input to enter workflow ID and optional run ID with submit button to start a debug session.
```
cretz, line 49: > submit button to start a debug session.

Can you clarify whether this starts a vscode debug session just like it would without the Temporal extension? I just want to confirm that this:

* Starts a replayer w/ deadlock detection off to do normal code stepping
* This main panel is just a easy interface to a launch.json they could otherwise setup
* The history panel is just a panel like call stack and locals that is just giving more info (and allows some stepping at an event/task level)

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+
+Use to step over SDK code and customized per language
+
+#### Core replay debug server
+
+A webserver opened by Core, listens on port allocated by the VSCode plugin.
+Has a single endpoint for setting breakpoints on workflow task started events (by event ID).
+
+In order to indicate to Lang that an activation has a breakpoint associated with it, we'll add a new field to
+[`WorkflowActivation`][workflow-activation]
```
cretz, line 83: I am a bit confused about this point. Why is lang involved? Isn't the current workflow task info delivered on a side channel? And isn't whether something has hit a breakpoint or not between core and vscode? I don't think lang has anything do to here. I do understand if the vscode side has to step out until they get to user code.

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+
+> NOTE: This assumes initial support of Core based languages, could be extended to other languages eventually
+
+#### Lang middleware
+
+Intercepts activations and sets breakpoints
```
cretz, line 89: Not completely sure lang has a role to play at all. Setting breakpoints should be done in the debugger not lang.

The only thing I could see the lang side having is the ability to start a replayer in debug mode with some information to core about where the RPC should be hosted.

debugger.md
cretz, line 49: > enter workflow ID and optional run ID

We also need a way to set where the actual workflow code even is so the replayer can even be setup. Just a workflow ID and run ID is not enough to start a replayer.

debugger.md
```diff
@@ -0,0 +1,160 @@
...
+
+The IDE plugin shouldn't require writing the boilerplate for setting up the replay worker and should be able to "just
+work" with zero setup. That being said, while a typical replay worker setup is more or less standard, the only
+meaningful customization options are interceptors, data converters and "sinks" (loggers and such although those
+typically do not run during replay).
+
+The IDE plugin should support this customization either by accepting a file for overriding the worker options or by
+automatically detecting it given a convention (e.g. `/path/to/project/.temporal-replayer-options.ts` - name TBD).
+
+### History visualization
```
cretz, line 101: How is the history visualization different than the history panel?

debugger.md
bergundy: Thanks for the tip, will change to use builtin secrets.

bergundy: I was thinking we'd have a form in the extension panel to configure the server settings.
I'll add to the proposal.

debugger.md
bergundy: Confirmed.
I'll add clarifications to the proposal.

debugger.md
bergundy: Communication between core and vscode is async.
After thinking about this some more, we can do all of this with minimal lang interference and no changes to core.
We could intercept activations and notify vscode when SDK starts to process each workflow task, vscode could use that information to set a breakpoint if required.

debugger.md
bergundy: After discussing on slack I think we have a better approach now, I'll modify this proposal accordingly.

debugger.md
bergundy: This is just some implementation recommendation, I can restructure to clarify.

debugger.md
bergundy: Hmm.. could be in the input there or plugin config file (where you'd set your interceptors and data converters).

debugger.md
Sushisource: I think you still need some small change to core. It doesn't need to be specific to this, though.

Right now lang doesn't know which workflow task it's on. More than one activation can exist for one workflow task. This problem can be solved by attaching some kind of "workflow task ID" to each activation, though, and then it's easy to know when you've moved from one WFT to the next.

Unless I'm just not following what this was supposed to be for in the first place.

bergundy: I was thinking we can use this: https://github.com/temporalio/sdk-core/blob/15ea5d3ee580e2a226eefdcf0d8444b2ae30a997/protos/local/temporal/sdk/core/workflow_activation/workflow_activation.proto#L30

Sushisource: That is probably stable, but, a bit of semantic abuse. It's not hard to add something more appropriate to core.

debugger.md
bergundy: It's currently not possible for Core based SDKs to correlate history events with activation jobs.
@Sushisource we should probably add that ability.

PR #61, opened 2022-07-27
Propose enhanced stack trace queries

bergundy:
[Rendered](https://github.com/temporalio/proposals/blob/ehnanced-stack-trace-query/enhanced-stack-trace-queries.md)

### DIFF
--- a/enhanced-stack-trace-queries.md
+++ b/enhanced-stack-trace-queries.md
@@ -1,0 +1,84 @@
+# Enhanced stack trace queries
+
+## Backround
+
+The SDKs have a built-in `__stack_trace` query that can be used display where the workflow is blocked at the time of the
+query (the last processed workflow task).
+
+The response to a `__stack_trace` query is a single string representing a list of stack individual stack traces
+separated by double newlines.
+
+Example from the TS SDK:
+
+```
+    at someInternalFunction (/path/to/workflows/my-workflow.ts:18:10)
+    at someWorkflowFunction (/path/to/workflows/my-workflow.ts:60:20)
+
+    at someInternalFunction2 (/path/to/workflows/my-workflow.ts:100:30)
+    at someWorkflowFunction2 (/path/to/workflows/my-workflow.ts:123:5)
+```
+
+While this is valuable for developers to be able to understand how their code correlates to event histories, it is quite
+limited since it requires looking up the actual code to get the full picture.
+
+## Proposal
+
+SDKs can optionally implement an `__enhanced_stack_trace` query (name TBD) where instead of returning the stack traces
+as a single string, they return a pre-defined structure, that structure will also include file sources.
+
+The web UI can use this response to show the workflow code and highlight locations where it is blocked.
+
+### Proposed structure
+
+```ts
+/**
+ * A pointer to a location in a file
+ */
+export interface FileLocation {
+  filename: string;
+  line: number;
+  column: number;
+}
+
+/**
+ * Used as the result for the enhanced stack trace query
+ */
+export interface EnhancedStackTrace {
+  /**
+   * Mapping of file name to file contents
+   */
+  sources: Record<string, string>;
+  locations: FileLocation[];
+}
+```
+
+## Details
+
+### TS SDK
+
+In the TS SDK, the worker already has access to a single source map where it can look up the sources and fulfill this
+query.
+
+### Other SDKs
+
+Other SDKs may not have access to the source in production environments but in certain environments like local
+development those sources should be present, the exact solution should be researched per SDK.
+
+## Alternatives considered
+
+### SDK provides source control URLs for the UI to look up the code in
+
+The problem with this approach is that the workflow code is composed from several sources, internal and external.
+It would be very difficult to map all of these sources.
+
+Another problem with this approach is that, when developing locally, workflow code would not be checked into the
+upstream source control system and it would be difficult to produce a local URL, at that point we'd be better off using
+file paths.
+
+## Future work
+
+When we have support for (not yet-planned) worker queries, we could build a more advanced UI that will let you follow
+the execution of the workflow over time. This proposal could serve as the basis for that.
+
+The reason why worker queries are required for this is because it would require fetching history, replaying a workflow
+from the beginning, and recording the stack traces at every "activation" point (typically the workflow task boundary).
--- a/enhanced-stack-trace-queries.md
+++ b/enhanced-stack-trace-queries.md
@@ -31,24 +31,51 @@ The web UI can use this response to show the workflow code and highlight locatio
 ### Proposed structure
 
 ```ts
+export interface SDKInfo {
+  name: string;
+  version: string;
+}
+
+/**
+ * Represents a slice of a file starting at lineOffset
+ */
+export interface FileSlice {
+  content: string;
+  lineOffset: number;
+}
+
 /**
  * A pointer to a location in a file
  */
 export interface FileLocation {
-  filename: string;
+  /**
+   * Path to source file (absolute or relative).
+   * When using a relative path, make sure all paths are relative to the same root.
+   */
+  filepath: string;
   line: number;
   column: number;
+  /**
+   * function/goroutine/thread/greenlet/whatever name, if it exists.
+   * (Not sure if this will be used but doesn't hurt to have some more information)
+   */
+  context?: string;
+}
+
+export interface StackTrace {
+  locations: FileLocation[];
 }
 
 /**
  * Used as the result for the enhanced stack trace query
  */
 export interface EnhancedStackTrace {
+  sdk: SDKInfo;
   /**
    * Mapping of file name to file contents
    */
-  sources: Record<string, string>;
-  locations: FileLocation[];
+  sources: Record<string, FileSlice[]>;
+  stacks: StackTrace[];
 }
 ```
 
--- a/enhanced-stack-trace-queries.md
+++ b/enhanced-stack-trace-queries.md
@@ -40,7 +40,13 @@ export interface SDKInfo {
  * Represents a slice of a file starting at lineOffset
  */
 export interface FileSlice {
+  /**
+   * slice of a file with `\n` (newline) line terminator.
+   */
   content: string;
+  /**
+   * Only used possible to trim the file without breaking syntax highlighting.
+   */
   lineOffset: number;
 }
 
@@ -52,14 +58,20 @@ export interface FileLocation {
    * Path to source file (absolute or relative).
    * When using a relative path, make sure all paths are relative to the same root.
    */
-  filepath: string;
-  line: number;
-  column: number;
+  filePath: string;
   /**
-   * function/goroutine/thread/greenlet/whatever name, if it exists.
-   * (Not sure if this will be used but doesn't hurt to have some more information)
+   * If possible, SDK should send this, required for displaying the code location.
    */
-  context?: string;
+  line?: number;
+  /**
+   * If possible, SDK should send this.
+   */
+  column?: number;
+  /**
+   * Function name this line belongs to (if applicable).
+   * Used for falling back to stack trace view.
+   */
+  functionName?: string;
 }
 
 export interface StackTrace {
@@ -72,7 +84,9 @@ export interface StackTrace {
 export interface EnhancedStackTrace {
   sdk: SDKInfo;
   /**
-   * Mapping of file name to file contents
+   * Mapping of file path to file contents.
+   * SDK may choose to send no, some or all sources.
+   * Sources might be trimmed, and some time only the file(s) of the top element of the trace will be sent.
    */
   sources: Record<string, FileSlice[]>;
   stacks: StackTrace[];
@@ -80,6 +94,11 @@ export interface EnhancedStackTrace {
 ```
 
 ## Details
+
+### Opt-in to sending sources
+
+Even though the user's payload codec applies to the response, they may not want to send code to the UI for security
+reasons. User should opt-in to this feature.
 
 ### TS SDK
 


enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Used as the result for the enhanced stack trace query
+ */
+export interface EnhancedStackTrace {
+  /**
+   * Mapping of file name to file contents
+   */
+  sources: Record<string, string>;
+  locations: FileLocation[];
+}
```
lorensr, line 52:
```suggestion
  sdk: string;  
}
```

`'typescript' | 'go'` etc

Might be helpful

lorensr (approved):

enhanced-stack-trace-queries.md
bergundy: Good idea, but I'd do `sdk: { name, version }`

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,84 @@
...
+
+/**
+ * Used as the result for the enhanced stack trace query
+ */
+export interface EnhancedStackTrace {
+  /**
+   * Mapping of file name to file contents
+   */
+  sources: Record<string, string>;
```
mjameswh, line 50: A few things to consider here:
- This would only contain files that are mentioned in actual stack traces, right?
- Would there be any value including sources for the SDK itself, or does it make sense to exclude them?
- Should we worry about some limit on maximum query response length? AFAIK query responses are subject to grpc's max packet size of 4 mb, but is there some other, lower limit that could be an issue?

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,84 @@
...
+  locations: FileLocation[];
```
mjameswh, line 51: Shouldn't it be `FileLocation[][]`, ie. the enhanced stack trace for each blocked promise? Or do you intend to only return the top most location for each blocked promise?

bergundy: I was intending to return the top-most location for each blocked promise but that's a good point it'd be valuable to know the stack, just not sure how to visualize that, @stevekinney

enhanced-stack-trace-queries.md
bergundy: > A few things to consider here:
> 
> * This would only contain files that are mentioned in actual stack traces, right?

Yes

> * Would there be any value including sources for the SDK itself, or does it make sense to exclude them?

We typically cut the stack trace at the SDK level in the stack trace query so those should only make it to the UI in rare situations, we'll probably 

> * Should we worry about some limit on maximum query response length? AFAIK query responses are subject to grpc's max packet size of 4 mb, but is there some other, lower limit that could be an issue?

That's a concern for sure, one option for dealing with this is to limit the source code sent to X wrapping lines for each location but I'm not sure it's worth the optimization here, worth considering.

mjameswh: Forget this. If we agree on trimming down the list of files being transferred to only _user_ files that are indeed mentioned in the stack traces, then I don't think this is an immediate issue. We might have to deal with some huge source files at some point in the future, but that can be dealt later.

lorensr: Even if just the first user source file, could be large. More concerned with unnecessary data/load time than avoiding reaching max possible. Would do whatever we wanted to display, like 15 lines on either side, to reduce load time.

mjameswh: >  like 15 lines on either side

Not that easy given the proposed structure... We might have several references to a same file, at different locations. We would have to capture 15 lines on either side starting from each active position in that file... And then, we end up with a sparse file... How can the UI match line numbers between the stack trace and the sparse source file?

lorensr: Could do:

```
sources: Record<Filename, Lines>;

Lines = Record<number, string>;
```

enhanced-stack-trace-queries.md
bergundy: cc @canvanooo

enhanced-stack-trace-queries.md
bergundy: Or better yet:
```
sources: Record<string, ContentWithOffset>

interface ContentWithOffset {
  content: string;
  lineOffset: number;
}
```

lorensr: What's an example?

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,84 @@
...
+The web UI can use this response to show the workflow code and highlight locations where it is blocked.
+
+### Proposed structure
+
+```ts
+/**
+ * A pointer to a location in a file
+ */
+export interface FileLocation {
+  filename: string;
```
Sushisource, line 38: There could be name collisions - this should probably be `filePath`, which maybe you meant, but we'll want to make it clear.

The comment will want to explain how this path should be relative to the "project root" however that is traditionally defined for the language.

enhanced-stack-trace-queries.md
bergundy: What do you mean?

enhanced-stack-trace-queries.md
bergundy: Yes, this should be the full path, I'll clarify this.
Could be absolute path or relative as long as all paths are relative to the same root.

enhanced-stack-trace-queries.md
Sushisource: Agreed here. It's not a stack trace without... the stack :smile: 

I think you might want to ship slightly more data so there's at least something valuable even if sources aren't provided.

Ex:

```ts
export interface StackFrame {
    location: FileLocation
    // This can be the "someInternalFunction" part of "at someInternalFunction (/path/to/workflows/my-workflow.ts:18:10)"
    context: String 
    // We can add stuff here in a backwards compat manner
}

export interface StackTrace {
    stack: StackFrame[]
}
```

Then this line becomes
`stacks: StackTrace[]`

bergundy: Not sure what would go in context but it's a good idea to make `StackFrame` an object.

I'll definitely update the proposal to something like this.

We could a "step out" button in the UI to jump to the caller's location.

We don't have to support this immediately in the UI but it should be trivial to add and having the full trace definitely adds value.

Thanks both for pointing this out.

enhanced-stack-trace-queries.md
lorensr: I meant like JSON values. If I have 3 stack traces and we display 15 lines before/after:

```
line 115 src/foo.js 
line 120 src/foo.js 
line 2000 src/foo.js 
```

would your `sources` contain this?

```
sources: {
  src/foo.js: { content: '<contents of lines 100 through 2016>', lineOffset: 100 }
}
```

mjameswh: I think it should be

`sources: Record<string, ContentWithOffset[]>`

Note the array.

enhanced-stack-trace-queries.md
mjameswh: Another nice thing to add would be some details would allow linking the promise to a corresponding workflow history event, when appropriate…

enhanced-stack-trace-queries.md
lorensr: That would enable less content. And we could deduplicate overlapping content.

Overlapping:

```
sources: {
  src/foo.js: [
    { content: '<contents of lines 100 through 131>', lineOffset: 100 }
    { content: '<contents of lines 105 through 136>', lineOffset: 105 }
    { content: '<contents of lines 1985 through 2016>', lineOffset: 1985 }        
  ]
}
```

Dedup:

```
sources: {
  src/foo.js: [
    { content: '<contents of lines 100 through 136>', lineOffset: 100 }
    { content: '<contents of lines 1985 through 2016>', lineOffset: 1985 }        
  ]
}
```

FE could count newlines in the content.

bergundy: You are both correct

enhanced-stack-trace-queries.md
bergundy: Yes, that is something I want to add as well but those aren't directly connected to workflow stack traces which show where the user code is blocked.
I did want to include for each event, the stack trace that lead to creating that event but the plan was to do it in the "advanced" proposal, the one that requires worker queries so it can be done in all SDKs.

bergundy: It should be easy enough to do in TS since we already capture all of these stack traces but we don't in other SDKs at the moment.
I'm debating whether I should add it in this proposal or not.

bergundy: I'll keep this thread alive for reference.

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+ */
+export interface FileSlice {
+  content: string;
+  lineOffset: number;
+}
+
+/**
+ * A pointer to a location in a file
+ */
+export interface FileLocation {
```
bergundy, line 50: I'm considering renaming this to `StackFrame`

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * Path to source file (absolute or relative).
+   * When using a relative path, make sure all paths are relative to the same root.
+   */
+  filepath: string;
```
bergundy, line 55: `filePath`?

bergundy: `path`?

Sushisource: `filePath`

enhanced-stack-trace-queries.md
Sushisource: > Not sure what would go in context 

Like in my comment - you often get something like the function name when you dump the stack. That can be useful to see, especially if full source isn't available.

bergundy: Yeah, I figured that out after commenting, for some reason I missed the intent there. It's now a part of FileLocation.

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+export interface SDKInfo {
+  name: string;
+  version: string;
+}
+
+/**
+ * Represents a slice of a file starting at lineOffset
+ */
+export interface FileSlice {
+  content: string;
```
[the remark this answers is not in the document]
mjameswh: Given that the worker will anyway have to do something like a split in order to extract the range of lines to be sent, then either returning an array of lines or normalizing the EOL sequence to a single \n seems like a trivial task.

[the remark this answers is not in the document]
mjameswh: I also note that once serialized to JSON, an array of line has essentially the same length as a normalized multiline string (EOL become `\\n` after JSON serialization).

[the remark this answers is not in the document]
cretz: > seems like a trivial task.

Definitely trivial, but without a comment saying that's what's expected, nobody will know to do it. I don't mind whichever way is chosen so long as it's made known.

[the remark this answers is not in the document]
bergundy: I'll mention that content lines must end with `\n` and that it's the SDK's responsibility

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+ * A pointer to a location in a file
+ */
+export interface FileLocation {
+  /**
+   * Path to source file (absolute or relative).
+   * When using a relative path, make sure all paths are relative to the same root.
+   */
+  filepath: string;
+  line: number;
+  column: number;
```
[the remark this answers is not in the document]
bergundy: In that case the UI would need to degrade to something like the current stack trace view. Maybe the SDK should detect this and return a NotImplemented error back to the UI which would fallback to string based stack trace.

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+  /**
+   * function/goroutine/thread/greenlet/whatever name, if it exists.
```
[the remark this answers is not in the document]
bergundy: It's not necessarily a coroutine, each location of the trace might include this.
See example from JS:

https://v8.dev/docs/stack-trace-api

```
ReferenceError: FAIL is not defined
   at Constraint.execute (deltablue.js:525:2)
   at Constraint.recalculate (deltablue.js:424:21)
   at Planner.addPropagate (deltablue.js:701:6)
   at Constraint.satisfy (deltablue.js:184:15)
   at Planner.incrementalAdd (deltablue.js:591:21)
   at Constraint.addConstraint (deltablue.js:162:10)
   at Constraint.BinaryConstraint (deltablue.js:346:7)
   at Constraint.EqualityConstraint (deltablue.js:515:38)
   at chainTest (deltablue.js:807:6)
   at deltaBlue (deltablue.js:879:2)
```

Other languages support this too.

enhanced-stack-trace-queries.md
```diff
@@ -0,0 +1,111 @@
...
+
+/**
+ * Used as the result for the enhanced stack trace query
+ */
+export interface EnhancedStackTrace {
+  sdk: SDKInfo;
+  /**
+   * Mapping of file name to file contents
+   */
+  sources: Record<string, FileSlice[]>;
```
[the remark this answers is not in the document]
bergundy: Yes,

I'm now on the fence whether the SDK / UI should fallback to the old mechanism or omit some of the data here.
Maybe we should add a capabilities object to the response so the UI can adjust the view accordingly.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: My concern is the general term "context". I wonder if there is some other term. Maybe `contextName` I guess like you had in a comment above. I don't have a good solution here, it's just that when I think of context, I think of arbitrary values.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: :+1: It should be opt-in from the worker side I think. Shouldn't be able to issue a query to get source code by default.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: Oh wait, you mean the function name as the `context`? I think that may be a separate value too. Go and Python have coroutine/task name _and_ function name.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: I think just making line/column optional is fine. In Java and others when compiling without debug options, you'll get method and what not with no line numbers (I forget whether file is absent).

Sushisource (approved):
This lgtm

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: I don't think the coroutine or task name provides value for the UI though, do you think otherwise?

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: Hmmm.. would you opt-in to returning the code? Why? Is this because of the security concern with remote codecs?

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: I'm adding:
If worker cannot get the required information for enhancing the stack trace, it will still respond with a structured stack trace but sources and some location fields might be missing.

I'll also make it the worker's responsibility to specify which fields are going to be included in the response.
Something like:
```ts
interface Capabilities {
  lineNumbers?: boolean;
  sources?: boolean;
  ...
}
```

It's more info but makes it easier for the UI to pick the right visualization.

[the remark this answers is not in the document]
Sushisource: Why do that instead of simply using the presence or absence of those things to determine whether or not to render them? All the information is there already, this is just an opportunity to accidentally incorrectly construct the types.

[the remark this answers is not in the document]
mjameswh: > Why do that instead of simply using the presence or absence of those things to determine whether or not to render them? All the information is there already, this is just an opportunity to accidentally incorrectly construct the types.

I’d also add that some information might be available on some files/frame and not on other. Flags at top level might not help much in such cases…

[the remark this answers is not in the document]
bergundy: SGTM, you both make a good point

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: Nah, that's fine. Is there a better word we can use besides "context" for what is essentially "function name" in all of our SDKs?

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: Yes, because of security concern in general (unrelated to remote codecs). IMO we should not have a built-in query that can return code. But I understand that can be annoying to ask developers to opt in to. I would welcome others' opinions here.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: Thinking about this some more, blindly sending chunks to the UI will hurt the ability to have proper syntax highlighting.
I think we should keep this structure for future optimization but I'd start with sending entire files for now and make the UI deal with presenting snippets instead of entire files.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: I'm just curious why the code is different from data in that regard.
I'd add a worker option just in case, just not sure if the default should be on or off.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: A large Java trace for example may be sending over dozens of large files here. This is all starting to make me rethink sending sources is that important. I definitely won't use it as a user, but then again I haven't seen all of the complaints from users that can't see their code in the UI (if they even exist).

How do other "APM" or error tracking systems do this? So when Sentry shows code for an error, what is it doing? Does it show code for every element in the trace or only the top element of the trace? How do they obtain the source code to map to the lines? Is it gathered like we'd be doing here when you invoke Sentry "report" or whatever? I'll try to dig into https://github.com/getsentry/sentry-go a bit.

Maybe we should only send the file of the top element of the trace?

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
cretz: Because it is clear to a user they are sending data. It is not clear there is a secret enabled by default backdoor to their source code. Remove the "enabled by default" from that statement and it becomes more reasonable.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: Renamed to function name

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: Clarified this is opt-in.

enhanced-stack-trace-queries.md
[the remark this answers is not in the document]
bergundy: Updated comment

bergundy:
Address all the comments, merging since we already have 2 approvals.
If anyone has other concerns, comment and I can open a followup PR.

I thought more about adding information to correlate stacks with history events, to show where a command is generated from code but that can added in the future.

PR #57, opened 2022-04-28
Ruby SDK proposal Phase 1

antstorm:
# Summary

This is the first phase of the API proposal for the new Ruby SDK

# Motivation

We want to provide a Temporal SDK for Ruby users, allowing them to leverage 
Temporal in their applications.

# Detailed design

This is an interface proposal and all the details are included in the proposal itself

# Drawbacks

- implementation cost, both in term of code size and complexity
- another SDK for Temporal to support

# Alternatives

Alternatives to parts of the proposal are mentioned in the proposal itself

# Adoption strategy

There are no current developers to consider here. However there are a few companies
using a Coinbase authored Ruby SDK. There is no current plan to support that migration.
But this shouldn't be too hard and a migration path might get provided by the community.

# How we teach this

The full documentation on the SDK will follow once we start getting closer to a stable release.

# Unresolved questions

- RBS vs Sorbet for static typing
- FFI vs Rutie for Core SDK integration

PR #63, opened 2022-08-08
Ruby SDK lightweight proposal

antstorm:
# Summary

After receiving lots of feedback on the [original Ruby SDK proposal](https://github.com/temporalio/proposals/pull/57), we've decided to take a slightly different approach and proceed with the Ruby SDK by submitting this lightweight proposal. The idea here is to try and get a high-level alignment on the general look and feel of the whole SDK without focusing too much on the details unlocking the development process.

The details will of course be addressed and ironed out in the subsequent more details proposals.

### DIFF
--- a/ruby/sdk-lightweight.md
+++ b/ruby/sdk-lightweight.md
@@ -23,7 +23,7 @@ method that will be called to execute an activity.
 # The activity class name will be used as activity name unless overridden
 class SumActivity < Temporal::Activity
   # Optionally override the name
-  activity 'sum-activity'
+  activity_name 'sum-activity'
 
   # return value of the #execute method will be used as an activity result
   def execute(a, b)
@@ -55,7 +55,7 @@ require 'sum_activity'
 # Similar to activities the class name will be used as a workflow name unless overridden
 class SumWorkflow < Temporal::Workflow
   # Optionally override the name
-  workflow 'sum-workflow'
+  workflow_name 'sum-workflow'
 
   def execute(a, b)
     # - Positional arguments passed to activity (inline with other SDKs)
@@ -85,7 +85,8 @@ end
 require 'temporal/client'
 require 'sum_workflow'
 
-connection = Temporal::Connection.new('localhost:7233')
+# Connections are eager and a lazy option will be introduced later
+connection = Temporal::Connection.connect('localhost:7233')
 client = Temporal::Client.new(connection, namespace: 'demo')
 
 # workflow is referenced by the class
@@ -106,7 +107,7 @@ client.start_workflow('sum-workflow', 1, 2, id: 'id-1', task_queue: 'demo')
 ```ruby
 class MyWorkflow < Temporal::Workflow
   # Optionally override the name of the workflow
-  workflow 'my.workflow'
+  workflow_name 'my.workflow'
 end
 
 client.start_workflow(MyWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
@@ -122,7 +123,7 @@ require 'some_other_workflow'
 require 'sum_activity'
 require 'some_other_activity'
 
-connection = Temporal::Connection.new('localhost:7233')
+connection = Temporal::Connection.connect('localhost:7233')
 worker = Temporal::Worker.new(
   connection,
   namespace: 'demo',
@@ -165,7 +166,7 @@ end
 
 class ParentWorkflow < Temporal::Workflow
   def execute
-    workflow.execute_workflow(ChildWorkflow, 'parent', id: 'id-2')
+    workflow.execute_child_workflow(ChildWorkflow, 'parent', id: 'id-2')
   end
 end
 ```
@@ -192,13 +193,20 @@ require 'sum_activity'
 class AsyncWorkflow < Temporal::Workflow
   def execute
     activity_promise = async { workflow.execute_activity(SumActivity, 1, 2) }
-    workflow_promise = async { workflow.execute_workflow(ChildWorkflow, 'parent') }
+    workflow_promise = async { workflow.execute_child_workflow(ChildWorkflow, 'parent') }
 
     # Block execution until both promises are resolved
-    workflow.wait_for(activity_promise, workflow_promise)
+    activity_result, workflow_result = async.all(activity_promise, workflow_promise).await
 
-    activity_promise.result # => 3
-    workflow_promise.result # => 'Hello, parent!'
+    activity_result # => 3
+    workflow_result # => 'Hello, parent!'
+    
+    promise = async { workflow.execute_activity(SumActivity, 3, 4) }
+    
+    # individual promises can also be waited on
+    result = promise.await
+    
+    result # => 7
   end
 end
 ```
@@ -208,19 +216,20 @@ A bit more involved example:
 ```ruby
 class AnotherAsyncWorkflow < Temporal::Workflow
   def execute
-    promise_1 = async { wait_and_sum(1, 2, delay: 10) }
-    promise_2 = async { wait_and_sum(3, 4, delay: 15) }
+    promise_1 = async { wait_and_sum(1, 2, delay: 15) }
+    promise_2 = async { wait_and_sum(3, 4, delay: 10) }
 
     # The same exact method can be called in a blocking fashion
     wait_and_sum(5, 6, delay: 5) # => 11
 
     # Block execution until any promise is resolved
-    async.wait_for(async.any(promise_1, promise_2))
+    async.any(promise_1, promise_2).await
 
-    promise_1.result # => 3
-
-    # Because of the longest delay this would block until resolved
+    # Returns immediately because it will already be resolved (shorter delay)
     promise_2.result # => 7
+
+    # This would block until the promise is resolved (longer delay)
+    promise_1.result # => 3
   end
 
   private
@@ -232,6 +241,11 @@ class AnotherAsyncWorkflow < Temporal::Workflow
 end
 ```
 
+Methods `async.all` and `async.any` return promises themselves, so they can be used just like the
+promises returned by an `async {}` call.
+
+Also worth noting that `async {}` calls can be nested to model more complex business logic.
+
 
 ## Activity heartbeats & cancellations
 
@@ -241,9 +255,9 @@ block (that will not raise until the block has executed) or opt-out of this defa
 altogether and use `activity.cancelled?` flag to determine when an activity was cancelled.
 
 ```ruby
-class MyActivity < Temporal::Activity
+class LongRunningActivity < Temporal::Activity
   def execute
-    loop do
+    10.times do
       sleep 1
       activity.heartbeat('ping')
     end
@@ -256,21 +270,24 @@ end
 Here is what it will look like from a workflow perspective:
 
 ```ruby
-class MyWorkflow < Temporal::Activity
+class InnerCancelWorkflow < Temporal::Workflow
   def execute
-    # Async blocks double as cancellation scopes
-    promise = async { workflow.execute_activity(LongRunningActivity) }
+    # the block supplies an optional argument
+    async do |outer_promise|
+      async do
+        workflow.sleep(30)
+        outer_promise.cancel
+      end
 
-    workflow.sleep(10)
-
-    # Blocking call
-    promise.cancel
+      # One of these blocking activities will end up getting cancelled 
+      #   unless they all finish within 30 seconds
+      workflow.execute_activity(LongRunningActivity)
+      workflow.execute_activity(LongRunningActivity)
+      workflow.execute_activity(LongRunningActivity)
+    end.await
   end
 end
 ```
-
-*NOTE: While `promise.cancel` is blocking, it can be wrapped in an `async` block for non-blocking
-cancellations.*
 
 #### Alternative approaches
 
@@ -289,14 +306,17 @@ can be used. In which case cancellation will raise after the block is executed:
 ```ruby
 class MyWorkflow < Temporal::Activity
   def execute
-    workflow.execute_activity(SomeActivity)
+    # Can get cancelled
+    workflow.execute_activity(NonCriticalActivity)
 
-    workflow.shield do
-      promise_1 = async { workflow.execute_activity(AnotherActivity) }
-      promise_2 = async { workflow.execute_activity(SomeOtherActivity) }
-
-      workflow.wait_for(promise_1, promise_2)
-    end
+    # This block will delay any cancellations until it is finished
+    async(shield: true) do
+      workflow.execute_activity(CriticalActivity)
+      workflow.execute_activity(OtherCriticalActivity)
+    end.await
+    
+    # Can get cancelled
+    workflow.execute_activity(AnotherNonCriticalActivity)
   rescue Temporal::Error::WorkflowCancelled
     ...
   end
@@ -311,32 +331,40 @@ end
 ```ruby
 class CalculatorWorkflow < Temporal::Workflow
   # Both signal and query handlers can be defined as Symbols (when the name matches the
-  # method) or a Hash when name is different (or cannot be used as a method name)
-  signals :add, 'my.signal.subtract' => :subtract
-  queries :get_value, 'my.query.value' => :get_value
+  # method) or overriden when the name is different (or cannot be used as a method name)
+  signal :add
+  signal :subtract, name: 'my.signal.subtract'
+  
+  query :value # can use a getter here
+  query :get_previous_value, name: 'my.query.previous_value'
 
   # It is preferable in this case to initialize the values before the execution to
   # support starting a workflow with a signal
   def initialize
     @value = 0
+    @previous_value = 0
   end
 
   def execute
-    workflow.wait_for { @value >= 42 }
+    async.await_for { @value >= 42 }
   end
 
   private
+  
+  attr_reader :value
 
   def add(val)
+    @previous_value = @value
     @value += val
   end
 
   def subtract(val)
+    @previous_value = @value
     @value -= val
   end
 
-  def get_value
-    @value
+  def get_previous_value
+    @previous_value
   end
 end
 ```
@@ -351,20 +379,20 @@ class CalculatorWorkflow < Temporal::Workflow
   def execute
     value = 0
 
-    workflow.on_signal('add') do |val|
+    workflow.on_signal(:add) do |val|
       value += val
     end
 
-    workflow.on_signal('subtract') do |val|
+    workflow.on_signal('my.signal.subtract') do |val|
       value -= val
     end
 
-    workflow.on_query('get_value') do
+    workflow.on_query(:value) do
       value
     end
 
     # A condition block will pause until the enclosed statement returns true
-    workflow.wait_for { value >= 42 }
+    async.await_for { value >= 42 }
   end
 end
 ```
@@ -375,7 +403,7 @@ end
 ```ruby
 require 'calculator_workflow'
 
-connection = Temporal::Connection.new('localhost:7233')
+connection = Temporal::Connection.connect('localhost:7233')
 client = Temporal::Client.new(connection, namespace: 'demo')
 
 handle = client.start_workflow(SumWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
@@ -398,11 +426,8 @@ class ParentWorkflow < Temporal::Workflow
 
     # All these calls are blocking, but can be executed in an async block
     handle.signal(:add, 34)
-    handle.query(:get_value) # => 34
     handle.signal(:subtract, 4) # This will send a 'my.signal.subtract' signal as defined
-    handle.query(:get_value) # => 30
     handle.signal(:add, 12)
-    handle.query(:get_value) # => 42
 
     # Block until workflow finishes
     handle.result


ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+
+#### Alternative approaches
+
+- Using plain methods as activities. This is not feasible in Ruby since non-namespaced methods
+  are global without a clear way to reference them (would have to be `method(:my_activity)`)
+
+- It is possible to allow multiple activities per class, for example:
+
+```ruby
+class CalculatorActivity < Temporal::Activity
```
Sushisource, line 38:
```suggestion
class CalculatorActivities < Temporal::Activity
```

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+  end
+end
+```
+
+#### Alternative approaches
+
+- For the same exact reasons it is not feasible to use non-namespaced methods for workflow
+
+- Similar to activities it should be possible to define multiple workflows per class, however I
+  don't think this would be used often
```
Sushisource, line 79: I don't think it's advisable either. It becomes too easy to think you can change state in one workflow and read it in another which would cause all sorts of determinism issues.

So I think we should actually explicitly prevent this if we can

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+
+```ruby
+require 'sum_activity'
+
+class SumWorkflow < Temporal::Workflow
+  def execute(a, b)
+    # - .execute! is a shorthand for .execute().await
+    # - only positional arguments are supported (inline with other languages), keyword
+    #   arguments are reserved for options
+    SumActivity.execute(a, b, start_to_close_timeout: 10)
```
Sushisource, line 69: Is this example supposed to be `execute!` - since there's no await?

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+```ruby
+class CalculatorWorkflow < Temporal::Workflow
+  def execute
+    value = 0
+
+    workflow.on_signal('add') do |val|
+      value += val
+    end
+
+    workflow.on_query('get_value') do
```
Sushisource, line 171: Did we already determine we were going with this way of registering handlers rather than as class methods? There are a lot of discoverability / client usability benefits to them being class methods, so I'd advocate for that if it's doable.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+  # Generate custom stubs to be used in workflows:
+  # - CalculatorActivity.add
+  # - CalculatorActivity.subtract
+  activities :add, :subtract
+
+  def add(a, b)
+    a + b
+  end
+
+  def subtract(a, b)
```
cretz, line 48: IF you have the time, can you show what heartbeating and cancellation look like?

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,194 @@
...
+
+```ruby
+require 'temporal/client'
+require 'sum_workflow'
+
+connection = Temporal::Connection.new('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+# workflow is referenced with a stub, not just class
+handle = client.start_workflow(SumWorkflow.execute(1, 2), id: 'id-1', task_queue: 'demo')
```
cretz, line 92: Can you provide some background on how Ruby intercepts this call?

ruby/sdk-lightweight.md
cretz, line 69: Can you provide some background on how Ruby intercepts this call? I assume you do some runtime check to see whether it's an activity or not? So if I were calling a regular method, it'd still happen as normal correct? Are there concerns about users reading this and not knowing whether they are invoking an activity or just a normal method?

cretz, line 69: Can I see an example of starting multiple activities and waiting on the first to finish? Just curious how this looks in Ruby.

cretz (commented):
Not commenting on specific approaches at this time. I am just requesting more information in some comments ahead of discussion meeting.

ruby/sdk-lightweight.md
antstorm: Yes, my bad 👍

ruby/sdk-lightweight.md
antstorm: There are downsides to using class methods as well, e.g. you won't get the benefit of using enclosed variables. But I'll see if that's feasible

antstorm: Added an alternative approach where these can be defined as methods instead

ruby/sdk-lightweight.md
```diff
@@ -13,7 +13,7 @@ will be revisited when Ruby's typing system matures to support non-basic use-cas
 
 ## Defining an activity
 
-Activities are defined as classes ([a very common practise](https://guides.rubyonrails.org/active_job_basics.html)
+Activities are defined as classes ([a very common practice](https://guides.rubyonrails.org/active_job_basics.html)
```
cretz, line 16: 🇺🇸! :-)

ruby/sdk-lightweight.md
antstorm: As discussed on the call earlier — Ruby has class (static) methods and instance method that don't overlap:

```ruby
class Test
  def self.method_1
    puts "class method 1"
  end

  def self.method_2
    puts "class method 2"
  end

  def method_1
    puts "instance method 1"
  end

  def method_3
    puts "instance method 3"
  end
end

Test.method_1
# => class method 1

Test.method_2
# => class method 2

Test.method_3
# => undefined method `method_3' for Test:Class (NoMethodError)

t = Test.new

t.method_1
# => instance method 1

t.method_3
# => instance method 3

t.method_2
# => undefined method `method_2' for #<Test:0x000000012c025c38> (NoMethodError)
```

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,314 @@
...
+- It is possible to allow multiple activities per class, for example:
+
+`calculator_activities.rb`
+
+```ruby
+class CalculatorActivities < Temporal::Activity
+  # Generate custom stubs on the class to be used in workflows:
+  # - CalculatorActivities.add
+  # - CalculatorActivities.subtract
+  activities :add, :subtract
```
cretz, line 44: For others knowledge, this can also have `, 'custom-activity-name' => :divide`

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,314 @@
...
+require 'sum_workflow'
+require 'sum_activity'
+require 'calculator_activities'
+
+connection = Temporal::Connection.new('localhost:7233')
+worker = Temporal::Worker.new(
+  connection,
+  namespace: 'demo',
+  task_queue: 'demo',
+  workflows: [SumWorkflow], # registers all workflows in the class
```
cretz, line 130:
```suggestion
  workflows: [SumWorkflow], # registers each class as a workflow
```

Because, after discussion, a workflow class can only have one workflow. And, after our discussion, we decided to always require the entrypoint to be `execute` instead of allowing that to be customized since it's inline w/ normal Ruby nomenclature and there's no benefit in customizing it.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,314 @@
...
+
+```ruby
+require 'temporal/client'
+require 'sum_workflow'
+
+connection = Temporal::Connection.new('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+# workflow is referenced via a stub, not just a class
+handle = client.start_workflow(SumWorkflow.execute(1, 2), id: 'id-1', task_queue: 'demo')
```
cretz, line 94: Can you confirm that if I called this activity method `add` or something, that this would be `SumWorkflow.add`? Meaning the activity invocation name will match the activity implementation name? (granted you may have to do a reverse map lookup to get the _actual_ string name if they overrode it)

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,314 @@
...
+  end
+end
+```
+
+This approach uses `Thread#raise` that will eject the execution out of the current context. In cases
+where this is not acceptable we will allow to either wrap a critical code in an `activity.shield`
+block (that will not raise until the block has executed) or opt-out of this default behaviour
+altogether and use `activity.cancelled?` flag to determine when an activity was cancelled.
+
+Here is what it will look like from a workflow perspective:
```
cretz, line 175: We will have to have a (maybe separate) discussion about workflow cancellation and its cascading effects. We can't use `Thread#raise` in workflows so you very well may want to turn to something like https://docs.temporal.io/typescript/cancellation-scopes. But for activity cancellation, this looks perfect.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,314 @@
...
+## Signal and query a workflow
+
+```ruby
+require 'calculator_workflow'
+
+connection = Temporal::Connection.new('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+handle = client.start_workflow(SumWorkflow.execute(1, 2), id: 'id-1', task_queue: 'demo')
+handle.signal('add', 34)
```
cretz, line 308: You support `client.start_workflow(SumWorkflow.execute(1, 2), ...` so why not support `handle.signal(SumWorkflow.add(34))` here? (of course still supporting the string option for when you don't have the method handy)

cretz (commented):
Added some comments since I may be absent for most of the in-person discussion.

Most of this LGTM!

ruby/sdk-lightweight.md
antstorm: Yes, that's right, but this is a wrong line :)

Basically when invoking the workflow you would always have `SomeWorkflow.execute` method, but when calling an activity from a workflow the name would correspond to the name on your activity class:

```ruby
class MyActivities < Temporal::Activity
  activities :add, :subtract, 'custom-activity' => :divide
  ...
end

class MyWorkflow < Temporal::Workflow
  def execute
    MyActivities.add!(1, 2)

    # non-! version returns a future
    future = MyActivities.subtract(42, 10)
    
    # Schedules an activity with a type "custom-activity"
    MyActivities.divide!(4, 2)
  end
end
```

ruby/sdk-lightweight.md
cretz: I definitely want to avoid things like underscore prefixes and pieces like that. Also remember we want these things to be typed some day in the future.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+given that the language is object-oriented). Activities are expected to implement an `#execute`
+method that will be called to execute an activity.
+
+`sum_activity.rb`
+
+```ruby
+# The activity class name will be used as activity name unless overridden
+class SumActivity < Temporal::Activity
+  # Optionally override the name
+  activity 'sum-activity'
```
bergundy, line 26:
```suggestion
  activity_name 'sum-activity'
```

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+
+`sum_workflow.rb`
+
+```ruby
+require 'sum_activity'
+
+# Similar to activities the class name will be used as a workflow name unless overridden
+class SumWorkflow < Temporal::Workflow
+  # Optionally override the name
+  workflow 'sum-workflow'
```
bergundy, line 58:
```suggestion
  workflow_name 'sum-workflow'
```

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+```ruby
+require 'sum_activity'
+
+class AsyncWorkflow < Temporal::Workflow
+  def execute
+    activity_promise = async { workflow.execute_activity(SumActivity, 1, 2) }
+    workflow_promise = async { workflow.execute_workflow(ChildWorkflow, 'parent') }
+
+    # Block execution until both promises are resolved
+    workflow.wait_for(activity_promise, workflow_promise)
```
bergundy, line 198: Is this still going to be available, or do we only want `async.all` and `async.any`?

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+class AnotherAsyncWorkflow < Temporal::Workflow
+  def execute
+    promise_1 = async { wait_and_sum(1, 2, delay: 10) }
+    promise_2 = async { wait_and_sum(3, 4, delay: 15) }
+
+    # The same exact method can be called in a blocking fashion
+    wait_and_sum(5, 6, delay: 5) # => 11
+
+    # Block execution until any promise is resolved
+    async.wait_for(async.any(promise_1, promise_2))
```
bergundy, line 218: In one of our conversations, these combinators returned promises. Doesn't look like this is reflected in the proposal.

ruby/sdk-lightweight.md
bergundy: We also discussed `all` returning a promise that resolves with an array of results.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+```ruby
+class MyWorkflow < Temporal::Activity
+  def execute
+    # Async blocks double as cancellation scopes
+    promise = async { workflow.execute_activity(LongRunningActivity) }
+
+    workflow.sleep(10)
+
+    # Blocking call
+    promise.cancel
```
bergundy, line 267: You'd want to "race" between the activity and the timer to avoid unhandled rejection.
I'd replace this example to avoid suggesting this pattern.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+```ruby
+class MyWorkflow < Temporal::Activity
+  def execute
+    workflow.execute_activity(SomeActivity)
+
+    workflow.shield do
+      promise_1 = async { workflow.execute_activity(AnotherActivity) }
+      promise_2 = async { workflow.execute_activity(SomeOtherActivity) }
+
+      workflow.wait_for(promise_1, promise_2)
```
bergundy, line 298: Also looks outdated

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,437 @@
...
+## Register signal and query handlers
+
+`calculator_workflow.rb`
+
+```ruby
+class CalculatorWorkflow < Temporal::Workflow
+  # Both signal and query handlers can be defined as Symbols (when the name matches the
+  # method) or a Hash when name is different (or cannot be used as a method name)
+  signals :add, 'my.signal.subtract' => :subtract
+  queries :get_value, 'my.query.value' => :get_value
```
bergundy, line 316: Needs to be updated with recent conclusions.

ruby/sdk-lightweight.md
antstorm: Sorry, this is outdated. Pushed new version

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+```ruby
+redis = Redis.new('localhost:6379')
+
+Temporal::Worker.new(
+  ...,
+  activities: [CalculatorActivities.new(redis)]
+)
+```
+
+*NOTE: This however will come with a heavy warning — the activity must be strictly thread-safe as
```
cretz, line 152: This warning should probably exist for _all_ activities, not just ones they instantiate. Because we should be instantiating the class-activities at registration time too. This is so the user can use (thread-safe) fields of the class across activities if they so choose.

Basically, the lifetime of all activity classes should be at least the worker level, whether we instantiate them or they do. It'd be confusing if class state was per-activity for one way of registering and per-worker for another.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+  private
+
+  def wait_and_sum(a, b, delay:)
+    workflow.sleep(delay)
+    workflow.execute_activity(SumActivity, a, b)
+  end
+end
+```
+
+Methods `async.all` and `async.any` return promises themselves, so they can be used just like the
```
cretz, line 244: Note that Python combines these to just [wait](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait) and has an option for whether you want first completed or all. This is important because there's actually multiple wait styles:

* first success - first one finishes successfully or _all_ finish unsuccessfully (JS equiv to `Promise.any`)
* first completed - first one finishes or fails (JS equiv to `Promise.race`)
* all - all complete

You can keep the "any" and "all" separate, but you need to define how you handle a promise completed with a failure in the "any" case and you may need another "any" for the alternative approach

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+end
+```
+
+Here is what it will look like from a workflow perspective:
+
+```ruby
+class InnerCancelWorkflow < Temporal::Workflow
+  def execute
+    # the block supplies an optional argument
+    async do |outer_promise|
```
cretz, line 276: I know this is just a param name, but I wonder if we want to call this a "cancel scope" instead of a "promise". They can't do anything but cancel this, right?

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+```ruby
+class CalculatorWorkflow < Temporal::Workflow
+  # Both signal and query handlers can be defined as Symbols (when the name matches the
+  # method) or overriden when the name is different (or cannot be used as a method name)
+  signal :add
+  signal :subtract, name: 'my.signal.subtract'
+  
+  query :value # can use a getter here
+  query :get_previous_value, name: 'my.query.previous_value'
```
cretz, line 339: We need to consider how to provide "deserialization types" to these too, though it can be later on.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+require 'some_other_activity'
+
+connection = Temporal::Connection.connect('localhost:7233')
+worker = Temporal::Worker.new(
+  connection,
+  namespace: 'demo',
+  task_queue: 'demo',
+  workflows: [SumWorkflow, SomeOtherWorkflow],
+  activities: [SumActivity, SomeOtherActivity]
+)
```
cretz, line 133: I think it's gonna be important to provide an `activity_runner` here which defaults to `ThreadPoolRunner` or something. I think it's really important to be able to let people make async activities in their favorite async frameworks. Arguably it could be a per-activity runner instead of a per-worker activity runner.

Being able to run thousands of workflows and thousands of activities on just a couple of threads is very nice for scale.

cretz (approved):
Added comments here because I wasn't available during review.

Looks great! The async stuff looks very intuitive. My comments are minor and should not hold up development.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+```ruby
+require 'temporal/worker'
+require 'sum_workflow'
+require 'some_other_workflow'
+require 'sum_activity'
+require 'some_other_activity'
+
+connection = Temporal::Connection.connect('localhost:7233')
+worker = Temporal::Worker.new(
+  connection,
```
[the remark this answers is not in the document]
bergundy: > worker is created from a client (which it essentially is in all of our SDKs except unfortunately this one)

Only Go and Python.
I'm really on the fence here TBH, I like the simplicity of only having a `Client` and a `Worker` but OTOH like having the flexibility and "correctness" of this separate layer.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+require 'temporal/client'
+require 'sum_workflow'
+
+# Connections are eager and a lazy option will be introduced later
+connection = Temporal::Connection.connect('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+# workflow is referenced by the class
+handle = client.start_workflow(SumWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
+handle.result # => 3
```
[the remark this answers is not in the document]
bergundy: I'd punt on this for now, can be added later

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+given that the language is object-oriented). Activities are expected to implement an `#execute`
+method that will be called to execute an activity.
+
+`sum_activity.rb`
+
+```ruby
+# The activity class name will be used as activity name unless overridden
+class SumActivity < Temporal::Activity
+  # Optionally override the name
+  activity_name 'sum-activity'
```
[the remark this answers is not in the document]
bergundy: Had the same thoughts, can be added later

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: And Java whose `WorkerFactory` accepts a client. TS of course can't share the client at this time.

ruby/sdk-lightweight.md
bergundy: `all` equivalent in JS is `Promise.all`, it also has `Promise.allSettled` that doesn't throw when a promise is rejected.
I find the JS approach simpler but don't have a strong preference, up to @antstorm to decide.

ruby/sdk-lightweight.md
bergundy: 👍  I'd rather this only be a cancellation "handle" an not give access to the promise which could lead to deadlock if used incorrectly.

ruby/sdk-lightweight.md
bergundy: +1

ruby/sdk-lightweight.md
bergundy: I see where you're coming from but I would punt on this for now.

ruby/sdk-lightweight.md
bergundy: Wondering we should require the user to instantiate their own classes, it eliminates questions of "when / where is this instantiated" and makes it so there's a single way to register activities.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
bergundy: True, but the `WorkflowClient` accepts a `WorkflowServiceStubs` instance so this layer not completely hidden.

ruby/sdk-lightweight.md
Sushisource: +1 to different top level names. Easier to scan and read at a glance.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+```ruby
+require 'temporal/client'
+require 'sum_workflow'
+
+# Connections are eager and a lazy option will be introduced later
+connection = Temporal::Connection.connect('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+# workflow is referenced by the class
+handle = client.start_workflow(SumWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
```
drewhoskins-stripe, line 93: task_queue: not sure if your intent here is to show the most common pattern.
With Nexus, temporal is moving in the direction of hiding the task queue from the caller wherever possible.
Even without Nexus though, it seems best if the workflow provides its own namespace and task queue whenever possible.  Can we do that?  I think only when you're calling indirectly via the name (without the class) would you strictly need to provide a task queue.  And I expect calling via class to be the 95% case.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+Methods `async.all` and `async.any` return promises themselves, so they can be used just like the
+promises returned by an `async {}` call.
+
+Also worth noting that `async {}` calls can be nested to model more complex business logic.
+
+
+## Activity heartbeats & cancellations
+
+This approach uses `Thread#raise` that will eject the execution out of the current context. In cases
```
drewhoskins-stripe, line 252: 👍

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+and eliminates the risk of forgetting to actually start an async block. However this means that
+we'll need to find a solution for unhandled promise rejections, which we'll cover separately in a
+more detailed workflow proposal.
+
+We will introduce a helper method `async` to execute a block of code asynchronously:
+
+```ruby
+require 'sum_activity'
+
+class AsyncWorkflow < Temporal::Workflow
```
drewhoskins-stripe, line 193: 👍

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+```ruby
+class ChildWorkflow < Temporal::Workflow
+  def execute(name)
+    "Hello, #{name}!"
+  end
+end
+
+class ParentWorkflow < Temporal::Workflow
+  def execute
+    workflow.execute_child_workflow(ChildWorkflow, 'parent', id: 'id-2')
```
drewhoskins-stripe, line 169: FWIW, at Stripe, execute takes a context variable explicitly.  This makes it easier for folks to discover the stuff they're supposed to interact with.  Putting those methods on the workflow gets confusing because there are a bunch of other methods you're not supposed to use.
I'm not sure it's a good idea here without static type declarations but figured I'd mention it as an option.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+end
+```
+
+#### Alternative approaches
+
+- Overall using `Thread#raise` is considered bad practice because it might raise in the
+  middle of some process. We might want to consider opting in to this behaviour instead
+
+
+## Workflow cancellations
```
drewhoskins-stripe, line 298: Not sure where to bring this up, but kinda related, the default for temporal-ruby when the workflow implementation throws is to fail the workflow, whereas in Java, the workflow keeps retrying unless you specifically throw a fatal error.  
I'm guessing we'll switch since we're moving to rust-core but wanted to double check.  I would be in favor of defaulting to retrying since that unlocks the ability to deploy code to fix workflows and seems more inline with temporal's ethos.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+# Similar to activities the class name will be used as a workflow name unless overridden
+class SumWorkflow < Temporal::Workflow
+  # Optionally override the name
+  workflow_name 'sum-workflow'
+
+  def execute(a, b)
+    # - Positional arguments passed to activity (inline with other SDKs)
+    # - Keyword arguments are reserved for options
+    # - This call is blocking
+    workflow.execute_activity(SumActivity, a, b, start_to_close_timeout: 10)
```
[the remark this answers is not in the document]
drewhoskins-stripe: Also, because you're going with one-activity-per-class (yay!), it's great if the activity author specifies the timeouts and retry policies rather than the caller, because the activity author is the expert (often they are two different people). In that world, the caller only overrides them in in rare cases, which may make it easier for you to stomach an explicit options class rather than the pattern of keyword-args-as-options.

FYI, if you're not aware, Temporal is working on adding ExecuteAtMostOnce vs ExecuteAtLeastOnce annotations to the SDKs.
(see https://temporalio.notion.site/Replay-roadmap-highlights-print-version-632692bcd46549719f91bb9b7ac859bd "shorthand retry policies")

This is another aspect that pushes us toward wanting more stuff declared on the activity.

drewhoskins-stripe (commented):
Nice proposal and I like a couple of the improvements.
Made a few comments--I've missed some of the previous back-and-forth but hopefully some of it is valuable.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: This is correct. We encourage single struct arguments everywhere. So maybe `activity_param` as a shortcut would help. The break can be (mostly) clean on activity definition, but worker invocation needs to be able to call activities from other languages and we often allow people to make stubs of those other-language activities in order to have typing help during invocation (so maybe doesn't apply here _yet_).

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: This is a poll and a "wait for" or timeout or whatever makes sense. Unsure about naming. In other dynamic languages that do have async support we've just called it `result`, but if it's important to have a verb in front, no problem.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: Due to how Temporal API works and that activities can be called across languages, retry policies and timeouts must be caller-time values. You can't know a Ruby activity's retry policy when called from Go.

However, we encourage people to use normal code sharing practices so _they_ can centralize their retry policies. So if you're a Ruby-only shop you can put them on this class and reference them, but if you're not maybe you'll have them in a proto contract or just known on the caller side.

Most users agree that definition-time activity options make sense, but the core primitives of Temporal don't support that yet.

ruby/sdk-lightweight.md
cretz: I don't think we can reasonably do this. Many users of the client don't even have the workflow definition present. All of our SDKs require task queue at workflow start time and I don't think Ruby should be unique here. Also, it is quite common for the same workflow to be on different task queues (e.g. task queue versioning).

As with other parts, users are encouraged to build on top of these primitives to wrap/centralize this stuff though. So maybe you have a static start method on your workflow class for Ruby users that knows the task queue to use.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
Sushisource: The stubbing is probably the best reason.

We should definitely make sure to drive people towards one arg in every other situation, though.

Perhaps we could have some methodology for declaring stubs which is specific about them being external, and only allow more than one arg there, but I can see how that starts to make for perhaps too broad of an API surface for middling benefit.

ruby/sdk-lightweight.md
cretz: In theory, at least with all of our other SDKs, all methods on `workflow` package/module/whatever are for use in the workflow. So if they are unsure what to interact with, "just look at the workflow module". If there are methods you are not supposed to use in workflows, they shouldn't be in that module.

There is definitely a lot of discussion about the pros/cons of implicit vs explicit contexts.

ruby/sdk-lightweight.md
drewhoskins-stripe: What I would recommend would be a way for a workflow/activity author to provide a default configuration that's usable by its in-language clients.  If clients can't use those defaults for whatever reason, they have to provide.  This could be two separate caller methods.  Many shops will be able to exclusively use the nice one.

Just because everybody can't have nice things doesn't mean most people can't.  And it'd be better for us to provide it than for everybody to have to roll their own--Maxim was saying as much at Replay.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: We are trying to keep from having two different ways to provide the options to an activity at this low level. But this is really easy with just a constant on the activity they can reference. We are also trying to keep Ruby from being special here compared to all other SDKs.

ruby/sdk-lightweight.md
cretz: I think that may be a general discussion for all SDKs. I don't think just Ruby should deviate at this time.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+    # This block will delay any cancellations until it is finished
+    async(shield: true) do
+      workflow.execute_activity(CriticalActivity)
+      workflow.execute_activity(OtherCriticalActivity)
+    end.await
+    
+    # Can get cancelled
+    workflow.execute_activity(AnotherNonCriticalActivity)
+  rescue Temporal::Error::WorkflowCancelled
+    ...
```
jeffschoner, line 321: Do you expect that more activities can be run here as well? I do like that creating a canceled context does not seem to be necessary. I've seen that abstraction be confusing to workflow authors.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+      async do
+        workflow.sleep(30)
+        outer_promise.cancel
+      end
+
+      # One of these blocking activities will end up getting cancelled 
+      #   unless they all finish within 30 seconds
+      workflow.execute_activity(LongRunningActivity)
+      workflow.execute_activity(LongRunningActivity)
+      workflow.execute_activity(LongRunningActivity)
```
jeffschoner, line 286: I assume these must raise when canceled because there will not be an available return result. Should that error type be called out like in other parts of the proposal?

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+  def get_previous_value
+    @previous_value
+  end
+end
+```
+
+#### Alternative approach
+
+To provide extra flexibility in handling signals and queries (especially from external classes and
```
jeffschoner, line 374: I'm in favor of the the pattern above instead of the one below if possible. I've been able to compare workflow authors' use of the above approach in Java with the below in temporal-ruby since we support both equally at Stripe. I've seen more confusion leading to code defects with the callback-based approach. Mainly authors don't know when they should register their signal handler, and by not understanding that the code is not truly multithreaded. The above approach solves the first problem well, by taking way this unnecessary control. I've yet to find a case where registering a callback later is necessary. The second problem remains but seems to be less confusing because receipt of signals is less intertwined with their "main" code.

This got bad enough that we ended up eventually banning the below approach, abstracting over it in favor of an approach similar to that of the Go SDK. That is another option to consider here potentially as well if you do want to offer an inline approach. It ends up looking something like:

```
  queue = workflow.get_signal_queue(:foo)
  signal = queue.receive # blocks indefinitely
  signal = queue.receive(timeout: 60) # blocking with timeout, raises on timeout
  maybe_signal = queue.receive_now # returns nil if there is no signal available
  queue.empty? # true if there are no signals
```

However, there's not an obvious way to inline queries like this.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+## Signal and query a workflow
+
+```ruby
+require 'calculator_workflow'
+
+connection = Temporal::Connection.connect('localhost:7233')
+client = Temporal::Client.new(connection, namespace: 'demo')
+
+handle = client.start_workflow(SumWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
```
jeffschoner, line 409: What about signal-to-start/start-with-signal functionality? I recall there being some design discussions of this in temporal-ruby since the optimal API was non-obvious at first

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+## Extensibility
+
+To allow extending existing classes with the workflow code without passing workflow context
+explicitly we will expose a module with all the necessary mixins:
+
+```ruby
+require 'temporal/workflow/concern'
+
+class MyClass
+  include Temporal::Workflow::Concern
```
jeffschoner, line 448: Have you instead considered providing something like `DynamicWorkflow` and `DynamicActivity` from other SDKs for extensibility instead? I've found this to be very powerful in Java

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+Also worth noting that `async {}` calls can be nested to model more complex business logic.
+
+
+## Activity heartbeats & cancellations
+
+This approach uses `Thread#raise` that will eject the execution out of the current context. In cases
+where this is not acceptable we will allow to either wrap a critical code in an `activity.shield`
+block (that will not raise until the block has executed) or opt-out of this default behaviour
+altogether and use `activity.cancelled?` flag to determine when an activity was cancelled.
```
jeffschoner, line 255: I worry about the use of `.canceled?` because when the bit gets flipped happens at unexpected times (only when heartbeating is done). I've seen a lot of confusion around this already in all SDKs, but I think the confusion would be heightened if the workflow author was able to check this bit because it would give them a false sense that it may be updated more frequently.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+
+We will introduce a helper method `async` to execute a block of code asynchronously:
+
+```ruby
+require 'sum_activity'
+
+class AsyncWorkflow < Temporal::Workflow
+  def execute
+    activity_promise = async { workflow.execute_activity(SumActivity, 1, 2) }
+    workflow_promise = async { workflow.execute_child_workflow(ChildWorkflow, 'parent') }
```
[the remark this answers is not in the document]
cretz: Yes, inside the block itself is synchronous (unless you nest `async`). So `execute_child_workflow` twice in the same async block just means that the whole block is async/background, but the second won't execute until after the first is complete.

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+## Activity heartbeats & cancellations
+
+This approach uses `Thread#raise` that will eject the execution out of the current context. In cases
+where this is not acceptable we will allow to either wrap a critical code in an `activity.shield`
+block (that will not raise until the block has executed) or opt-out of this default behaviour
+altogether and use `activity.cancelled?` flag to determine when an activity was cancelled.
+
+```ruby
+class LongRunningActivity < Temporal::Activity
+  def execute
```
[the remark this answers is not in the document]
cretz: > the cursoring

Can you clarify what you mean here? Of course an activity is just any user code they want, we only provide helpers for cancellation, heartbeating, and obtaining some activity info. That means basically everything you can do in Ruby is supported.

ruby/sdk-lightweight.md
cretz: > Do you expect that more activities can be run here as well?

Yes, this is a feature in all Temporal SDKs. But note, cancellation is a reported state not an event from the Temporal server. So you can't get a second cancellation. And if this error is not bubbled out, the workflow won't be considered cancelled.

ruby/sdk-lightweight.md
cretz: I believe the full team agrees with you here. While our examples and suggestions will be the explicit methods over callbacks, we still often need the callbacks to exist for dynamic registration based on input. So I think both being present with strong emphasis on the method-based approach is ideal (and what Python does).

ruby/sdk-lightweight.md
cretz: :+1: Should specify here in proposal. What we do in most SDKs is have a separate client method, but in Python we realized this was cleaner just making the start signal and start signal args as params to the regular start call.

ruby/sdk-lightweight.md
cretz: Those are a different type of extensibility than this.

This provides workflow primitives in non-workflow code so you can build workflow libraries/helpers/mixins.

What you are talking about wrt `DynamicWorkflow`/`DynamicActivity` from Java is the ability to catch unregistered workflows/activities in catch-all implementations. While this is often useful, we haven't built it in many places besides Java, but it'd be fairly easy to build.

ruby/sdk-lightweight.md
cretz: While `Thread.raise` is a good way to signal cancellation, people also need to be able to know their activity is in a cancelled state. The presence of this boolean is not the same as suggesting it be the primary way to determine cancellation.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
antstorm: @cretz great suggestion on specifying serializers/deserializers, I had exactly the same approach in mind.

On the subject of single vs multiple arguments — while it makes perfect sense I think that it opens up more questions. The main one being — what should that single argument type be? Ruby doesn't have a good default solution. A Hash (dictionary) is easily serializable/deserializable to JSON, however it needs to support have variable value types (which RBS has, but Sorbet doesn't). Ruby's `Struct` doesn't have any native serialization support (unless you convert it to Hash first) and doesn't have any typing support (unless you use Sorbet's proprietary `T::Struct`, but that's a different data type altogether). Right now it feels like the extra cost for developers will be outweighing the benefits.

A good example here is Protobuf where they had to develop a whole new data type to support serialization and type checks. I wonder if we want to have this built in the SDK or simply be flexible enough that someone can built the support for Protobuf (or JSON, `T::Struct` or whatever else) on top of our SDKs.

I also feel like this is something that needs to be agreed on across all SDKs and added as a separate proposal. Using a single SDK to drive this change has the risk of excluding some people who might have an opinion on it from the conversation.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
antstorm: This seems less important at the high level. The main points are:

1. There should be a way to check workflow status from a handle without waiting for it to finish
2. We might need to rename the `#result` method to indicate that it's polling/blocking

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
antstorm: We had a few discussions on this, I propose we revisit this part at a later point when we have more details (such as the configuration, converters and interceptors) crystallised

ruby/sdk-lightweight.md
antstorm: If we expect the Worker to instantiate client provided activity classes at the registration time then there is no need for accepting classes and we can just accept instances. My biggest worry is that we're exposing every single Ruby SDK user to the thread-safety issues that they might not be very familiar with (Ruby is known for not being 100% thread safe, which is the main reason for the Global Interpreter Lock).

By accepting a class rather than an instance I was hoping that we can create a new Activity instance only on a dedicated thread when processing an activity task therefore making it much easier and safer for the majority. If we don't care about this then I'm happy with only accepting instances (I don't like the discrepancy with workflows, but I can probably live with it)

ruby/sdk-lightweight.md
antstorm: Right now it's important to agree that `async.all` means "all finished" (completed or failed) and `async.any` means "first finished" (completed or failed).

I agree that top-level names are the best for readability and can be added later for more granular control. But overall this doesn't feel very important since we have `async.await_for` that can be used instead.

ruby/sdk-lightweight.md
antstorm: I remember that Maxim felt strong about this one, but I think we can punt on it for now, especially since it doesn't really affect the API much

ruby/sdk-lightweight.md
antstorm: Very good point, yes 👍

ruby/sdk-lightweight.md
antstorm: I'll make sure there's a clear separation of the runner, which we can later expose as a configurable argument. But it's probably something that we don't need in the lightweight proposal :)

ruby/sdk-lightweight.md
cretz: I think accepting classes makes plenty of sense and reads better, with the option of accepting instances too for advanced use.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
antstorm: @drewhoskins-stripe I agree with you on this and in fact this is how the current SDK behaves allowing you to specify it on the Activity class and then optionally override when calling an Activity. This is a step away from the other SDKs, but I think that the key is extensibility of the SDK allowing you to easily add support for this behaviour (for example via an interceptor) and maybe publish it as a gem.

Using a single class to combine timeouts and retry policy options does make it easier (although admittedly forces more typing):

```ruby
workflow.execute_activity(SumActivity, a, b, timeouts: Temporal::Timeouts.new(start_to_close: 10), retry_policy: Temoral::RetryPolicy.new(...))
```

@cretz @bergundy @Sushisource what do you think? 

@cretz I think the cross-language argument doesn't really apply here because these calls similarly can't know remote argument/result serialization. But that doesn't mean we don't want to support it for same-language use-cases and via a stub for remote.

ruby/sdk-lightweight.md
antstorm: Agree with both points — it's a very useful property to have (and definitely made the current SDK much easier to use for many engineers) and also that this should come as a cross-SDK proposal.

@cretz are you aware of any ongoing efforts to propose this?

ruby/sdk-lightweight.md
antstorm: @drewhoskins-stripe not sure where the miscommunication happened, but all the public methods (without exceptions) on the `workflow` context will be available for general use. This is true in the current SDK as well.

Passing the context explicitly suggests that you would define these methods on the `Temporal::Workflow` which then has the potential of colliding with user-defined methods and also creates inconsistency in the API (calling a class vs a string name vs utility method such as `now` and `sleep`)

ruby/sdk-lightweight.md
```diff
@@ -0,0 +1,462 @@
...
+handle = client.start_workflow(SumWorkflow, 1, 2, id: 'id-1', task_queue: 'demo')
+handle.result # => 3
+```
+
+#### Alternative approaches
+
+- Starting a foreign workflow from a string name
+
+```ruby
+client.start_workflow('sum-workflow', 1, 2, id: 'id-1', task_queue: 'demo')
```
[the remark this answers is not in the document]
antstorm: We had this discussion with Temporal and decided that the SDK should be less opinionated here, but instead have the flexibility for someone to build the "nice" version on top of it. This of course has the downside of the "nice" version becoming more popular than the default version, but should that happen Temporal can choice to maintain both.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
antstorm: Yep, what Chad said ^. Basically:


```async
async do
  workflow.execute_activity(Activity1)
  
  promise_1 = async do
    workflow.execute_activity(Activity2)
    workflow.execute_activity(Activity3)
  end

  promise_2 = async { workflow.execute_activity(Activity4) }
  
  async.all(promise_1, promise_2).await
end
```

This will not start `Activity2` and `Activity4` until `Activity1` has completed. Then start `Activity2` and `Activity4` at the same time, but not `Activity3`. After `Activity2` is completed, `Activity3` will get executed. And the whole outer `async` block will complete after all activities have completed. Hope that makes sense

ruby/sdk-lightweight.md
antstorm: Unless these activities implement some custom cancellation handling they will raise `Temporal::Error::ActivityCancelled`. This will bubble up to the outer `async` block and raise in the workflow. Based on the error super class we can treat this as workflow cancellation (@bergundy please keep me honest here)

ruby/sdk-lightweight.md
antstorm: The main purpose for the alternative approach here is to provide the flexibility of extending the behaviour:

```ruby
class MyWorkflow < Temporal::Workflow
  def execute
    ExecutionTimeQueryHandler.track(workflow) do
       ...
    end
  end
end

class ExecutionTimeQueryHandler
  def self.track(workflow, &block)
    @start_time = workflow.now
    workflow.on_signal(:get_time) { workflow.now - @time }
    block.call
  end
end
```

But this is more of an "advanced usage" rather than the default approach we'd recommend

ruby/sdk-lightweight.md
antstorm: I think params to the `#start_workflow` call make the most sense, especially since the signature of the method is identical otherwise. I think this is also getting into more details than the lightweight proposal was meant for :)

ruby/sdk-lightweight.md
antstorm: Agree with Chad that this is a different type of extensibility.

I think that `DynamicWorkflow` and `DynamicActivity` are best implemented using interceptors (aka middleware) — you'd catch a `WorkflowNotFound` / `ActivityNotFound` error and handle it via a dedicated class. This way we don't need a special concept for this use-case and allow for many others.

ruby/sdk-lightweight.md
antstorm: Also worth noting that with the Rust Core the cancellation mechanism is abstracted away from heartbeating (still coupled internally, but this is not exposed to the SDK itself). So in the future cancellations might get propagated to a non-heartbeating code, so we do need to have some mechanism in place to check the state of an activity from within.

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: > what do you think?

Does Ruby offer a splat/spread operator? I don't think there is much value in a "Timeouts" grouping. But there may be value in an "ActivityOptions" grouping that you can splat, clone/override, etc.

> I think the cross-language argument doesn't really apply here because these calls similarly can't know remote argument/result serialization.

To clarify, we are talking about definition-site activity options. Invocation from non-Ruby won't have access to those options. You might as well just make them a constant on the class and reference them at call time instead of having it built in. I don't think there is enough benefit in having an official Temporal way for declaration-site options that only half-work and have no server support and make the Ruby SDK special.

I think tackling declaration-site options is a reasonable thing to do holistically, not shoehorned into a not-yet-built SDK. We know we need call-site options anyways, lets get that built first for parity purposes.

ruby/sdk-lightweight.md
cretz: I am not aware of any, but I'd of course support it. It'd need server side assistance and the upcoming Nexus framework expects to be able to support that as well (where the Nexus handler decides these options, not the caller).

ruby/sdk-lightweight.md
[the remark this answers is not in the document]
cretz: I don't think we should group timeouts. They are often unrelated to each other. It's not like a "timeout policy".

bergundy:
Looks like we have enough approvers and no new comments in a while.
I'm merging this.
As I'm sure we're all aware this is bound to change when we get to the implementation.

PR #64, opened 2022-08-21
Initial proposal for Nexus & ALO

rylandg:
This is a proposal for Nexus & ALO, two new concepts which can be leveraged by Temporal. 

[Rendered version](https://github.com/temporalio/proposals/blob/nexus-alo/nexus-and-alo/README.md)


nexus-and-alo/problem-statements.md
```diff
@@ -0,0 +1,81 @@
...
+
+When looking at Workflows through this lens, the ideology generally holds true. There is one case where things are not as clear and thats when it comes to invoking stateful operations of an arbitrary length. To be completely and totally accurate, I’m specifically talking about operations (RPC calls that result in compute) that are both:
+
+- Arbitrary length - the potential duration is completely unbounded (aka any call over a network)
+- Identifiable - if you can’t get back to it, why are we even talking about it?
+
+I will note that for the sake of practicality I’m over simplifying things. Technically speaking there is potential entropy even within local calls on your OS, but the chances are so low and the behavior is so well defined that its not worth addressing.
```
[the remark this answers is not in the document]
Sushisource: Heh, I was gonna make the same comment here. As a general note I would go for reduced verbosity, say more with less.

This last paragraph even self-admits its superfluousness

nexus-and-alo/usage-scenarios.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Solution:** Temporal users can now define and create a Nexus which can be used to surface any Temporal functionality to other Temporal consumers. When calling into another Nexus from a Temporal context (eg: Workflow), every call will surfaced as an ALO, even if the Nexus implementation did not define them as such. This is because Temporal can automatically retry calls to any external endpoint and therefore seamlessly provide the ALO experience even when the endpoint was not implemented as an ALO originally.
```
Sushisource, line 20: This probably can't be entirely true. The automatic retrying will likely only meet user expectations if the endpoint is structured properly by providing idempotency.

It also raises the question, why would Nexus expose something that isn't an ALO? It's not clear to me at this point, reading serially, how an ALO and not-ALO are distinguished.

nexus-and-alo/usage-scenarios.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Problem statement:** Calling into external APIs is an inconsistent experience as each API developer makes different choices, adds unique bugs and provides different levels of guarantees. As a consumer looking for consistency and similarity of guarantees between APIS , this reality provides a far from ideal experience.
+
+**Solution:** While most API developers will likely not adopt Temporal or the ALO concept immediately, this is not required to provide a consistent experience for consuming miscellaneous external APIs. API consumers or producers looking to provide a more reliable experience for those calling their API can simply wrap the existing API with a Temporal Nexus. To be clear, the developer would not be required to use any other Temporal primitives (Workflows & Activities) in this scenario at all, just Nexus. Wrapping an API (even one that you don’t own) with Nexus would immediately provide significant improvements in resiliency, availability and developer experience. All calls to the API would come with automatic retries out of the box, and ALOs could be manually surfaced by connecting various endpoints of the backing API to the expected ALO methods.
```
Sushisource, line 32: I get that this doc can't really get into implementation details, but, this bit in particular feels a bit handwave-y.

If it was reduced to "Nexus should be able to wrap existing APIs to provide a Temporal-workflow-like experience" I would've learned the same amount.

In particular
>ALOs could be manually surfaced by connecting various endpoints of the backing API to the expected ALO methods

Is hard to understand.

What do I get here besides auto-retries? (Even just that is pretty nice!) Maybe a brief example would be clarifying?

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+Because of this, ALOs will also be required to have a unique Id. Furthermore, because server side generation of unique Ids is inherently lossy, ALO **will be required to be created with a client-side generated unique Id**.
+
+### A non-Temporal example of ALOs
+
+To concretely understand the value of ALOs, it's easiest to showcase how they could be leveraged as part of an HTTP API. The example below will not necessarily be supported initially, and is not using best practices. For instance, the ALOs are all being stored exclusively in memory on a single machine, this means there is no durability and potential race conditions if running multiple servers to handle traffic. 
+
+// TODO (Chad): you may have thoughts on how to improve this
+
+```go
+package main
```
Sushisource, line 60: IMO this code sample is less useful than fleshing out the above gRPC sample with request/response details would be.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+2. User registers provided endpoint to Nexus Worker, similarly to how you would register Workflows and Activities with a Worker. 
+3. User calls the external Nexus Service using a Workflow-specific Nexus API which is provided by Temporal.
+    ```go
+    // calling a Nexus from a Temporal Workflow
+    func MyWorkflow(ctx workflow.Context, name string) (string, error) {
+        client := temporalnexus.NewWorkflowClient(temporalnexus.WorkflowClientOptions{Endpoint: "temporal:72aA64A0jhOL99B2"})
+        // Start some alo
+        alo, err := client.StartALO(ctx, temporalnexus.WorkflowStartALOOptions{
+            Name: "foo",
+            ID:   "my-id",
```
Sushisource, line 80: Feel free to resolve if too detailed: It may make a lot of sense to only allow users to specify a "human name" (which need not be unique) and have the client always use a UUID for the real ID, and return that.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+	err = builder.AddWorkflowALOHandler("foo", temporalnexus.WorkflowALOHandlerOptions{
+		Workflow: SayHelloWorkflow,
+		Options:  temporalclient.StartWorkflowOptions{TaskQueue: "my-task-queue"},
+	})
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	// Add queries and signals
+	err = builder.AddQuerySyncHandler("foo/get", temporalnexus.QuerySyncHandlerOptions{
```
Sushisource, line 246: Are these queries and signals supposed to interact with the previously registered worlkflow? That's what it seems like but it's unclear how they are tied together.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+	// Start Temporal worker
+	temporalWorker := temporalworker.New(temporalClient, "my-task-queue", temporalworker.Options{})
+	temporalWorker.RegisterWorkflow(SayHelloWorkflow)
+	if err := temporalWorker.Start(); err != nil {
+		log.Fatal(err)
+	}
+	defer temporalWorker.Stop()
+
+	// Start Nexus worker
+	worker, err := worker.New(worker.Options{Client: client, Service: "my-service", Handler: builder.BuildHandler()})
```
Sushisource, line 274: One thing that leads to complications in our current design is how workers can be crammed into the middle of any application.

It's worth considering a design where, rather than users instantiating a worker, we use a plugin type model where we provide the "worker" binary, and users provide the handlers to it at start time. This is more of a convention-over-configuration approach, but can be a lot better at ensuring a consistent and correct experience.

Food for thought.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+	// Wait for completion
+	log.Print("Worker started, ctrl+c to stop")
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, os.Interrupt)
+	<-signalCh
+}
+```
+
+### Calling Nexus from Temporal
```
Sushisource, line 290: It seems a bit counter intuitive but maybe move this sample in front of the other, as this is something we all understand already and it helps explain what's going on with the Nexus handler example

Sushisource (commented):
Interesting stuff, curious to see how this evolves.

It might help to structure feedback if, after starting with the conceptual basics, you show a motivating example of "here's something that would be hard to do today, made easier and accomplished with nexus".

EX: Take the cross-namespace call example and solve it with the new interfaces

nexus-and-alo/usage-scenarios.md
rylandg: There is actually an example of this already. The manual HTTP ALO implementation provided in the ALO strawman is meant to demonstrate this:

https://github.com/temporalio/proposals/blob/nexus-alo/nexus-and-alo/alo-strawman.md#a-non-temporal-example-of-alos

rylandg: Its more than just retries, basically any API which surfaces the building blocks of an ALO (resource based APIs) could be surfaced as one. 

That being said, regardless of my view on the actual solution this section is not clear enough and I will improve it. Thanks for calling that out!

rylandg (commented):
Thank you so much @lorensr & @Sushisource for taking the time to leave such meaningful and helpful feedback.

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+
+### Proposal for initial interface of the ALO
+
+```protobuf
+// An ALO is started via the Start method
+message AloInfo {
+  string id = 1;
+  Status status = 2;
+  map<string, string> metadata = 3;
+  repeated string http_result_urls = 4; // used for providing webhook/callback URLs for async completion of ALO
```
bergundy, line 24: This seems like an implementation detail that's leaked to the transport agnostic info struct.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+## Terminology
+
+**Nexus Service Instance:** Formally registered Nexus Service with a Nexus backend (eg: Temporal).  Nexus Instances must always have a unique identifier.
+
+**Nexus Task:** Unit of work which is run as a result of a request reaching the Nexus endpoint. Nexus Tasks are developed and operated similar to WorkflowTasks and ActivityTasks (eg: run and hosted on Nexus Workers). Within a Nexus Task Handler, developers can transform incoming payloads, make routing decisions and more.
+
+**Nexus Task Handler:** Source code which is designed to handle incoming Nexus Tasks. Nexus Task Handler is similar to a Workflow or Activity definition and is hosted by a Nexus Worker. 
+
+**Nexus Worker:** A special purpose Worker that pulls Nexus Tasks for its registered Nexus Instances from the Server and executes them using associated Nexus Task Handlers.
```
bergundy, line 22: Users might want to consume nexus requests using standard a HTTP / gRPC stack.
I would note that that conversion of requests to tasks just one implementation option.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Consuming a Nexus Service from Temporal
+
+1. User asks the Nexus Service author for the endpoint of the Nexus Service Instance they are trying to consume. For Temporal based Nexus Service Instances, the endpoint will not be HTTP or gRPC but rather a custom uri formulation.
+2. User registers provided endpoint with their SDK to avoid hardcoding the endpoint into all of their application logic. 
+3. User calls the external Nexus Service using a Workflow-specific Nexus API which is provided by Temporal.
+    ```go
+    // calling a Nexus from a Temporal Workflow
+    func MyWorkflow(ctx workflow.Context, name string) (string, error) {
+        client := temporalnexus.NewAloClient(temporalnexus.AloClientOptions{Endpoint: "temporal:72aA64A0jhOL99B2"})
```
bergundy, line 76: The mapping should be done outside of the workflow.
Temporal will have a map of `nexus service type:service instance` -> `nexus endpoint`.

This simplifies writing workflows since you don't need to inject configuration into them and hides the transport detail from the workflow.

Some other ideas that came up during our earlier discussions were to assign an "environment" (e.g. prod / dev) to each worker so workflows don't have to specify the instance name, it'll default to the current environment.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+        // Worker definition...
+    }
+    ```
+1. Nexus is now able to process requests to its frontend endpoint.
+
+### Consuming a Nexus Service from Temporal
+
+1. User asks the Nexus Service author for the endpoint of the Nexus Service Instance they are trying to consume. For Temporal based Nexus Service Instances, the endpoint will not be HTTP or gRPC but rather a custom uri formulation.
+2. User registers provided endpoint with their SDK to avoid hardcoding the endpoint into all of their application logic.
```
bergundy, line 71: Based on our recent discussions, this should no longer be required since Temporal server will hold the mapping.

nexus-and-alo/nexus-strawman.md
bergundy: The whole idea of registration on the server side should be expanded in this document IMHO.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+1. User asks the Nexus Service author for the endpoint of the Nexus Service Instance they are trying to consume. For Temporal based Nexus Service Instances, the endpoint will not be HTTP or gRPC but rather a custom uri formulation.
+2. User registers provided endpoint with their SDK to avoid hardcoding the endpoint into all of their application logic. 
+3. User calls the external Nexus Service using a Workflow-specific Nexus API which is provided by Temporal.
+    ```go
+    // calling a Nexus from a Temporal Workflow
+    func MyWorkflow(ctx workflow.Context, name string) (string, error) {
+        client := temporalnexus.NewAloClient(temporalnexus.AloClientOptions{Endpoint: "temporal:72aA64A0jhOL99B2"})
+        // Start some ALO
+        alo, err := client.StartAlo(ctx, temporalnexus.StartAloOptions{
```
bergundy, line 78: nit: start options don't seem like they should be Temporal specific.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+        }
+        // Workflow implementation...
+    }
+    ```
+
+## Concepts
+
+### Nexus Instantiation
+
+Instantiating a Nexus Service makes it possible for the Nexus Backend to receive traffic meant for that specific Instance. Nexus Instantiation does make your Nexus Service Instance endpoint(s) available for traffic, but until you register and run corresponding Nexus Task Workers, no forward progress will be made on incoming requests. In this way, sending requests at this stage is similar to starting Workflows without any Workflow Workers running.
```
bergundy, line 94: I'd specify that this only applies in the temporal specific protocol and even then I'm not sure this is desired, it implies that Temporal is persisting request auth info.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Registration:**
+
+For the sake of this proposal, we will only consider the case where Nexus is registered via `nexus` CLI. In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance. Registering a Nexus Service Instance would be done with the following command structure:
+
+`nexus service register --name my-service --description "My service" --backend nexus-backend.example.com`
```
bergundy, line 100: I thought a service has a type and an instance name, is this no longer the case?

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+Nexus registered successfully with instance name "my-service"!
+Available at:
+	Temporal: temporal:72aA64A0jhOL99B2
+```
+
+**What is the behavior after running this command?**
+
+![Nexus behavior](./images/behavior.png)
+
+The Nexus Instance is now registered and available via Nexus backend and should be reachable via the provided endpoint. Sending requests to the endpoint is possible even at this stage, but there should be no expectation of a response or of forward progress of the request. This is because while the Nexus Instance is registered and available, we have not run any corresponding Nexus Workers. In this way, sending requests at this stage is similar to starting Workflows without any Workflow Workers running.
```
bergundy, line 118: Requires more information about the "temporal" protocol, is it not a sync request / response protocol?
If so, how is this possible?

nexus-and-alo/nexus-strawman.md
bergundy: I think we want to have the flexibility of a library at this stage, what your'e suggesting @Sushisource can be added later.

bergundy (commented):
Did we intentionally leave out how you'd call Nexus from non-workflow code?
I'd be curious to see this and I think it might affect some decisions made in this proposal.

nexus-and-alo/nexus-strawman.md
Sushisource: Yes, completely. It just avoids a footgun where the user doesn't actually use unique IDs. The call returns the UUID, which they then persist. The passed-in ID may be non unique (though typically, will also be unique) and serves as a human readable alternative.

nexus-and-alo/nexus-strawman.md
lorensr: Nexus service instance is a URI? So mapping is like `processOrder:mynexus.myco.com` -> `temporal:72aA64A0jhOL99B2`? And then you have no client and do this?

```go
alo, err := StartAlo(ctx, nexus.StartAloOptions{
  Type: "processOrder",
  Nexus: "mynexus.myco.com",
  ID: "my-business-id"
}
```

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,348 @@
...
+
+- Making external RPC calls or other I/O operations.
+- Using a full-fledged Temporal client for kicking off Workflows or interacting with other Temporal resources.
+- Interacting with the originating Request through a formal Response.
+
+### Nexus Task Worker
+
+In order to make forward progress on incoming Nexus Tasks, a valid Nexus Worker for the Nexus Instance must be registered.
+
+// TODO (Chad): improve if needed
```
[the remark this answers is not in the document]
lorensr: Like the idea of this snippet building on the last one by referencing `HandleMyServiceRequest`

cretz:
I apologize I haven't been able to review this due to other outstanding work. I will try to get to it today.

nexus-and-alo/nexus-strawman.md
bergundy: The ID should be unique globally (only in the target Nexus?). It's used as an idempotency key, for workflow based ALOs we'll use this ID as the workflow ID.
@mfateev mentioned at the conference that in a workflow we can default to a deterministic UUID, I agree, we do the same for child workflows now.

nexus-and-alo/alo-strawman.md
bergundy: To add to this, the result URLs might not be `http`.
There should also be a way to set the result URL later on, for example if my workflow code has something like:
```
alo, err = StartALO()
// ... run arbitrary logic
result = alo.Result()
```

I might want to set the callback URL in case the result is not available yet to avoid having to poll.

nexus-and-alo/nexus-strawman.md
bergundy: +1 to removing the client layer.

nexus-and-alo/nexus-strawman.md
cretz, line 100: > --backend nexus-backend.example.com`

To confirm here for others that may be getting lost in the indirections like me - This is telling a "Nexus Backend Server" (i.e. probably implemented natively by Temporal) that this service is a think this backend can receive Nexus requests/tasks for.

This not to be confused with registering/mapping the _outbound_ services to the other Temporal server that would be made on your Temporal server.

So basically if you make a Temporal-workflow-backed Nexus ALO request from a Temporal workflow, it goes:

    Temporal worker ->
    outbound Temporal server ->
    inbound Nexus server ->
    Nexus worker ->
    Temporal server (probably same process as inbound Nexus server) ->
    Temporal worker

We probably need a sequence diagram showing this

nexus-and-alo/nexus-strawman.md
cretz: I don't have the details, but I believe @mfateev wants requests to be able to survive when all workers are down. Of course Nexus is not Temporal specific, so it's more like "backend does whatever it needs to with requests to ensure delivery".

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+### Nexus Task hosting
+
+Now that the Nexus Service is registered with a Nexus Backend, we have to define, register and run a Nexus Task Handler in order to handle the incoming requests to our endpoint.
+
+The Nexus payloads will be passed through the system before eventually being picked up by an appropriate Nexus Task Worker. Just as with Workflow and Activity Tasks, Nexus Tasks are not parsed or deeply introspected before being delivered. This also means that DataConverter functionality is unhindered and can be fully used with Nexus. 
+
+Here is a proposed API of the Nexus Task Handler:
+
+// TODO (Chad): show Nexus Handler without HTTP (Temporal specific).
```
cretz, line 137: I am not sure I can without performing actual implementation. I suppose it'd be something like:

```go
type NexusHandler interface {
  ServeNexus(nexus.ResponseWriter, *nexus.Request)
}
```

Which matches https://pkg.go.dev/net/http#Handler. In other languages it may be different.

So that code below is just POC toy code and it probably won't look like that. I hesitate doing implementation in markdown.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
cretz: I don't think the HTTP code should be in the markdown. Ignore that HTTP code exists.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+- Making external RPC calls or other I/O operations.
+- Using a full-fledged Temporal client for kicking off Workflows or interacting with other Temporal resources.
+- Interacting with the originating Request through a formal Response.
+
+### Nexus Task Worker
+
+In order to make forward progress on incoming Nexus Tasks, a valid Nexus Worker for the Nexus Instance must be registered.
+
+// TODO (Chad): improve if needed
```
cretz, line 196: I think this gets a bit too much into implementation. I don't think it's fair to design what this Go API looks exactly like now. I suppose it'd be very similar to HTTP serving w/ HTTP handlers in Go and different in other languages.

nexus-and-alo/nexus-strawman.md
cretz: > Are these queries and signals supposed to interact with the previously registered worlkflow?

Yes, the ALO identifier is the workflow identifier in this use case. We may have to rethink this a bit since workflow identifiers only have to be unique by implementing namespace whereas ALO identifiers need to be unique by caller _and_ implementing namespace.

nexus-and-alo/nexus-strawman.md
cretz: My gut reaction is also gross. We need to be flexible and non-global in everything we do. We don't need to care about how users use this library to build their applications. It will be totally normal to have cheap, ephemeral workers IMO. But the key point is that we don't care. Sure we can suggest things, but to bake in anything about process boundaries is way too inflexible.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+	log.Print("Worker started, ctrl+c to stop")
+	signalCh := make(chan os.Signal, 1)
+	signal.Notify(signalCh, os.Interrupt)
+	<-signalCh
+}
+```
+
+### Calling Nexus from Temporal
+
+```go
```
cretz, line 292: As with the other code samples, this is not what the code will end up looking like. I can already see a few differences:

* No separate client, service names are don on calls
* No being forced to reference backend protos directly
* No required identifier for ALO
* No differentiation in-workflow between ALO and sync call (but we do want that differentiation out-of-workflow)
  * This was subject to post-conference discussion. If we can make sync calls implement everything an ALO implements from a caller POV in default ways (e.g. identifier is request ID, status is always "done", and cancel never works), the API can be shared.

nexus-and-alo/usage-scenarios.md
```diff
@@ -0,0 +1,32 @@
...
+- [Exposing non-Temporal application via Nexus](#exposing-non-temporal-application-via-nexus)
+
+## Exposing Temporal application to a non-Temporal consumer (Nexus + ALO)
+
+**Problem statement:** One of the most common patterns we see from Temporal users is the need to expose a generic API for internal consumers who are not familiar with Temporal. The most common solution for this problem today is to stand up a separate infrastructure dependency in front of Temporal (such as a load balancer, HTTP server, GRPC Gateway) which defines the formal API & contract for consumers.
+
+**Solution:** Temporal users can now define and create a Nexus which represents the desired shape of their GRPC/HTTP API. Because Nexus gives users full control of how to parse/handle incoming requests, they can choose to expose their Temporal logic without anyone knowing that they used Temporal behind the scenes. If the consumer of their API has adopted the ALO standard, they can also “automagically” expose their Nexus APIs as ALOs. The capabilities of Nexus completely and totally obviate the need for a standalone API in these scenarios.
```
cretz, line 14: To clarify somewhere in this doc (because it can be hard to see end goals):

Also, in later versions assuming you don't need advanced runtime request customization, high-level abstractions will make this so transparent to users that a Nexus service definition is all that a client will require to make calls and all that an implementer will require to implement.

The end goal would support someone writing a Nexus service definition. That service definition is discoverable (marketplace and/or just an URL on the endpoint). Client can code-gen from that service definition in any language to have type-safe invocations. Implementer can code-gen from that service definition in non-Temporal-specific way to have simple, typed Nexus interface, or in Temporal-specific way to have a workflow (or top-level activity) interface ready to implement. Of course we'd leverage as much existing industry things as we could.

nexus-and-alo/usage-scenarios.md
cretz: > why would Nexus expose something that isn't an ALO? It's not clear to me at this point, reading serially, how an ALO and not-ALO are distinguished.

This tripped me up at first too, since technically "arbitrary" in ALO means it could be short. However, it does make sense that a user know whether the call is a request/response or a request/interactable-and-come-back-later. Basically all HTTP/gRPC calls today are the former, so the whole point of ALOs is to define the latter while letting Nexus support both.

cretz (approved):
Looks good enough to me to get to the implementation. I hesitated to update code in here because I really need to sit down and ponder API decisions.

cretz:
After this high-level proposal, I think we need a (mostly-non-markdown) technical proposal for the MVP that contains:

* Protos for what a Nexus backend must implement to serve:
  * Nexus service registration
  * Nexus service requests
  * Nexus workers
* Protos for the commands and events that Temporal server must implement to support Nexus invocations from a Temporal worker
* CLI documentation for what a `nexus` CLI may implement at first
* Nexus Go-and-one-other-language API (i.e. unimplemented interfaces/structs/methods) for implementing Nexus workers including:
  * Nexus request handling
  * ALO request handling
* Nexus Go-and-one-other-language API for a simple Nexus client
  * Very simple since we aren't focusing on non-Temporal invocations
* Temporal Go-and-1-other-language SDK updates for:
  * Invoking Nexus calls from a workflow
  * Implementing an ALO backed by a Temporal workflow (may be just simple mapping)
  * Invoking Nexus calls from an SDK client (maybe this is just getting a Nexus client w/ the same conn params)

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+The current belief is that ALO should never become a generic Durable Object and therefore it will never be possible to surface arbitrary methods as part of the ALO itself. Instead you would surface those endpoints as either stateless short lived API methods, or as top level ALOs themselves. That being said, there is significant room for extension in the ALO surface area such a Listing & Filtering, and the ability to Pause and Resume.
+
+### Proposal for initial interface of the ALO
+
+```protobuf
+// An ALO is started via the Start method
+message AloInfo {
+  string id = 1;
+  Status status = 2;
+  map<string, string> metadata = 3;
```
rylandg, line 23:
```suggestion
```

nexus-and-alo/alo-strawman.md
rylandg: Will make that change

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+  repeated string http_result_urls = 4; // used for providing webhook/callback URLs for async completion of ALO
+
+  enum Status {
+    STATUS_UNSPECIFIED = 0;
+    RUNNING = 1;
+    COMPLETED = 2;
+  }
+}
+
+service ALOHandler {
```
rylandg, line 33:
```suggestion
service AloHandler {
```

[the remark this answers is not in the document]
rylandg: Fixed and ty!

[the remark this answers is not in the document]
rylandg: Indeed.

This is meant to be the highest level representation of ALO capabilities. The fact its defined with proto language is not super relevant. Is it confusing the way I have it now?

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+  rpc Start(StartRequest) returns (StartResponse);
+  rpc GetInfo(GetInfoRequest) returns (AloInfo);
```
rylandg, line 35:
```suggestion
  rpc GetInfo(GetInfoRequest) returns (GetInfoResponse);
```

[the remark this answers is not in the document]
rylandg: Adjusted

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,193 @@
...
+
+![Full client Id flow](./images/full-client-id-flow.png)
+
+Because of this, ALOs will also be required to have a unique Id. Furthermore, because server side generation of unique Ids is inherently lossy, an ALO `StartRequest` **must include a client-side generated unique Id**.
+
+### A non-Temporal example of ALOs
+
+To concretely understand the value of ALOs, it's easiest to showcase how they could be leveraged as part of an HTTP API. The example below will not necessarily be supported initially, and is not using best practices. For instance, the ALOs are all being stored exclusively in memory on a single machine, this means there is no durability and potential race conditions if running multiple servers to handle traffic. 
+
+// TODO (Chad): you may have thoughts on how to improve this
```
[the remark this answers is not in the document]
rylandg: I completely get what you are saying. I just think its really critical for this proposal to have one concrete example of that "very low-level handler". Its the most common feedback so far when people look at this proposal.

nexus-and-alo/milestones.md
```diff
@@ -0,0 +1,50 @@
+# Milestones
+
+## MVP - Phase 0 (Namespace to Namespace)
+
+**Rationale:** Keeps requirements as minimal as possible and makes the primary focus unblocking X-Namespace calls.
+
+**Requirements:**
+
+* Existing Temporal users
+* Only target scenarios where Temporal Namespace is calling to another Temporal Namespace
```
rylandg, line 10:
```suggestion
* Only target scenarios where Temporal Namespace is calling to another Temporal Namespace. To be explicit, we will not initially support client-to-Temporal-ALO calls. 
```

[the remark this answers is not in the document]
rylandg: Clarified

nexus-and-alo/milestones.md
```diff
@@ -0,0 +1,50 @@
...
+**Rationale:** Assuming we have learned significant lessons from internal APIs during Phase 1, we are now ready to target services which are leveraged by external consumers and not just internal ones. This will bring in significantly higher security requirements as exposing something publicly is far more risky than exposing something internally.
+
+**Requirements:**
+
+```diff
+- Existing Temporal users
+- Minimum additional authentication and authorization requirements
+- Target teams looking to expose Temporal functionality internally, both in Temporal to Temporal cases and non-Temporal to Temporal cases
+
++ Target teams looking to expose Temporal functionality internally and externally, both in Temporal to Temporal cases and non-Temporal to Temporal cases
```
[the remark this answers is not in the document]
rylandg: That's a fair assumption.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+- [Concepts](#concepts)
+  - [Nexus Instantiation](#nexus-instantiation)
+  - [Nexus CRUD](#nexus-crud)
+  - [Nexus Task hosting](#nexus-task-hosting)
+  - [Nexus Task Worker](#nexus-task-worker)
+  - [Calling Nexus from Temporal](#calling-nexus-from-temporal)
+
+## Terminology
+
+**Nexus Service Instance:** Formally registered Nexus Service with a Nexus backend (eg: Temporal).  Nexus Instances must always have a unique identifier.
```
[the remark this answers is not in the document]
rylandg: I'm going to keep this for now. I've already removed every other concept per your feedback (service blueprint, service prototype etc). I think calling it Service creates a potential problem for the future.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Nexus Task:** Unit of work which is run as a result of a request reaching the Nexus endpoint. Nexus Tasks are developed and operated similar to WorkflowTasks and ActivityTasks (eg: run and hosted on Nexus Workers). Within a Nexus Task Handler, developers can transform incoming payloads, make routing decisions and more.
```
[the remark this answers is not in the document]
rylandg: Love that change.

nexus-and-alo/nexus-strawman.md
rylandg: Mostly agree with chad on this one.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Nexus Task Handler:** Source code which is designed to handle incoming Nexus Tasks. Nexus Task Handler is similar to a Workflow or Activity definition and is hosted by a Nexus Worker. 
+
+**Nexus Worker:** A special purpose Worker that pulls Nexus Tasks for its registered Nexus Instances from the Server and executes them using associated Nexus Task Handlers. 
+
+## User flows
+
+### Hosting a Nexus Service
+
+1. User instantiates Nexus Service Instance using provided `nexus` CLI commands (also possible via SDK or API) and receives an Instance ID. Nexus is now able to receive requests to its frontend endpoint (though they don't get processed until step 4).
```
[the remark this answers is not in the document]
rylandg: This is inescapable because we cannot guarantee global uniqueness in many realistic contexts. The Cloud is a great example of this. Is your Nexus name unique to the Namespace? Account? Cluster?

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+1. User defines a Nexus Task Handler that will handle the processing of Nexus Tasks for the instantiated Nexus Service Instance.
+1. User hosts their Nexus Task Handler using a Nexus Worker. They must specify the Nexus Service Instance ID when starting the Nexus Worker.
+
+	// TODO (Chad): show how registration would work for generic handler
```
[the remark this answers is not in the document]
rylandg: I think the code is one of the most important aspects of this proposal and I absolutely disagree that we should defer to hand wave explanations. We absolutely reserve the right to change implementation details and interfaces after this proposal is finalized. Seeing a practical manifestation is a key blocker for many learners.

nexus-and-alo/nexus-strawman.md
rylandg: Will update

nexus-and-alo/nexus-strawman.md
rylandg: I tend to side with Roey and Chad on this one. What I proposed to Max is that we always require client side ID, but in the case of calling an ALO from a Workflow, we could automatically handle the ID generation state-fully and therefore not explicitly require it from the caller.

nexus-and-alo/nexus-strawman.md
rylandg: Will adjust this section

nexus-and-alo/nexus-strawman.md
rylandg: I've removed the concept of type entirely from the proposal. I do believe it will come back but its a two way door.

nexus-and-alo/nexus-strawman.md
```diff
@@ -0,0 +1,361 @@
...
+
+**Registration:**
+
+For the sake of this proposal, we will only consider the case where Nexus is registered via `nexus` CLI. In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance. Registering a Nexus Service Instance would be done with the following command structure:
+
+`nexus service register --name my-service --description "My service" --backend nexus-backend.example.com` 
+
+This would result in the following output being displayed
+
+// TODO (Chad): if this spits out a Temporal uri the cli should probably have an option to denote that
```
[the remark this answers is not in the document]
rylandg: It doesn't need to be for a non-Temporal user initially.

nexus-and-alo/nexus-strawman.md
rylandg: Agree with Chad but can elaborate.

nexus-and-alo/nexus-strawman.md
rylandg: Yea but I just want to see a theoretical version of the only handler we will actually support with MVP0.

nexus-and-alo/nexus-strawman.md
rylandg: Look thats completely ok but I need to show something. Can you maybe just redo all the samples with pseudocode? I really don't care about the specific interfaces, I just want to demonstrate the capabilities.

nexus-and-alo/usage-scenarios.md
rylandg: Chad said it well.

rylandg (commented):
Really appreciate all of this additional feedback. I've made some immediate adjustments but have a backlog of fixes I will make based on these discussions.

nexus-and-alo/alo-strawman.md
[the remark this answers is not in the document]
cretz: Your chosen definition language is fine, it's just that when people read "service", they thing gRPC service and not language interface.

Also note that this interface will probably be quite different per language. Each language has their own way of handling request/response.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
cretz: > Is your Nexus name unique to the Namespace? Account? Cluster?

Endpoint. When I call service "foo" on "bar.temporal.cloud", that's all I need. An ID has nothing to do with it.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
cretz: For raw Nexus handling, probably just like raw HTTP handling in Go. Maybe something like:

```go

type myNexusHandler struct{}

func (myNexusHandler) ServeNexus(nexus.ResponseWriter, *nexus.Request) error {
  panic("Not implemented")
}

func main() {
  // Create Nexus client
  client, err := nexus.DialClient(ctx, nexus.ClientOptions{Target: "nexus-backend.example.com"})
  if err != nil {
    log.Fatal(err)
  }
  defer client.Close()

  // Start worker
  worker, err := nexus.NewWorker(nexus.WorkerOptions{
    Client: client,
    Handler: myNexusHandler{},
    Service: "foo",
  }
  if err != nil {
    log.Fatal(err)
  } else if err = worker.Start(ctx); err != nil {
    log.Fatal(err)
  }
  defer worker.Stop()

  // Wait for signal
  signalCh := make(chan os.Signal, 1)
  signal.Notify(signalCh, os.Interrupt)
  <-signalCh
}
```

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
cretz: Then we don't need to have this "TODO" because it doesn't apply initially since there is no externally-addressable URI the CLI spits out.

nexus-and-alo/nexus-strawman.md
cretz: I put one at https://github.com/temporalio/proposals/pull/64/files/33399efe89392c6fb71658b4dfbc077143b956bd#r963677093 for general Nexus handling

nexus-and-alo/nexus-strawman.md
cretz: Let's just go with the example that exists below then. No improvement needed for the proposal.

nexus-and-alo/alo-strawman.md
cretz: Please don't remove the metadata, only the `http_result_urls`

nexus-and-alo/milestones.md
cretz: I was hoping we would support this in the initial version

nexus-and-alo/alo-strawman.md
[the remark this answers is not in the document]
rylandg: Ok I can clarify that with a disclaimer

nexus-and-alo/milestones.md
rylandg: I will say its not a requirement but if its possible within scope of mvp I'm not opposed.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
rylandg: I see, for me endpoint is one valid form of id. In other words we are more aligned than not.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
rylandg: What about the URI that you give out to other Temporal users to leverage your Nexus?

nexus-and-alo/alo-strawman.md
rylandg, line 24:
```suggestion
```

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
cretz: That's your general Temporal server since we don't have a Nexus frontend yet.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
bergundy: +1

nexus-and-alo/nexus-strawman.md
bergundy: @cretz then why would you suggest calling "Nexus Task" a "Nexus Request"?

Part of the objectives of this proposal is to make a non-Temporal specific standard and to allow Nexus requests to be sent over open protocols like HTTP.

cretz: It is a non-Temporal specific standard, and Nexus requests (one day soon) will be able to be invoked over HTTP. But what you're referring to is _request handling_, not _request sending_. I think we all agree request sending needs to be HTTP, but I question whether the language that Nexus workers speak to the backend to handle requests needs to be abstracted over HTTP.

bergundy: Nexus requests can be sent over temporal-specific, or other protocols (like HTTP and gRPC).

The request handler can be implemented using standard HTTP and gRPC stack or when Temporal is the gateway using a Temporal Nexus worker.
I see value in designing a generic, modular Nexus SDK API to enable reuse of most of the handler code in both worker and HTTP / gRPC scenarios.

There are a few layers here on the request handling (backend) side:
1. gateway
2. request -> task dispatcher
3. task -> handler dispatcher
4. handler

Layer 1 can be any HTTP / gRPC gateway or Temporal.
Layer 2 is only required when using a non-standard protocol and will be implemented by Temporal server. Alternative implementations are allowed of course (we'll probably end up with a test server dispatcher).
Layer 3 is only relevant if using layer 2.
Layer 4 can be used on top of layer 3, or layer 1 with an additional thin adapter layer.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
rylandg: Not true. The URI is actually something above the level of GRPC URI etc. In the Temporal to Temporal case its assumed there is discovery and visibility between any relevant clusters.

rylandg:
Hey everyone! I have meticulously combed through the feedback and believe i've either directly addressed it or persisted it outside of this PR. I will now close this PR and reopen to make the review experience better.

nexus-and-alo/nexus-strawman.md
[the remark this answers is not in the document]
lorensr: I'm also inclined toward a single term. What's the potential problem?

nexus-and-alo/nexus-strawman.md
lorensr: I like the idea of being able to set up a Nexus directly backed by serverless functions (1 -> 4, Requests are sent from gateway to function over HTTP)

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,192 @@
...
+### Proposal for initial interface of the ALO
+
+```protobuf
+// An ALO is started via the Start method
+message AloInfo {
+  string id = 1;
+  Status status = 2;
+  map<string, string> metadata = 3;
+
+  enum Status {
```
drewhoskins-stripe, line 25: These values don't match what you listed above.
In the offsite, we talked about an `ENQUEUED`-ish status.  That would be the state before deduplication, rate limiting, or other async validations.
I notice you mentioned a `REJECTED` status above.  `ENQUEUED` would be the status that might still result in a `REJECTED`, whereas `RUNNING` would no longer be able to transition to that state.

nexus-and-alo/alo-strawman.md
```diff
@@ -0,0 +1,192 @@
...
+}
+```
+
+### A bit about Ids
+
+![Lost the client Id](./images/lost-client-id.png)
+
+One of the most common debates around ALOs is whether unique Ids should be a fundamental part of the specification and design. 
+
+Considering ALOs represent the eventual completion of an operation, its impossible to imagine a design where they aren’t identifiable. Note that I did not say “addressable” because technically in the case of asynchronous result delivery it may not be required for the ALO itself to be addressable. That being said, even in the async result delivery case, the result must be delivered with a unique correlation Id.
```
drewhoskins-stripe, line 46: What is the value of the abstraction if I can't address it?   Wouldn't `AloHandler` not work if we couldn't address it?

rylandg:
For posterity! This new is the new PR.

https://github.com/temporalio/proposals/pull/66

PR #66, opened 2022-09-08
Nexus alo

rylandg:
This is a proposal for Nexus & ALO, two new concepts which can be leveraged by Temporal. This is the second PR for the proposal as the first one was growing too large to work with.

[Rendered version](https://github.com/temporalio/proposals/blob/nexus-alo/nexus/README.md)


nexus/README.md
```diff
@@ -0,0 +1,45 @@
+# Nexus & ALOs
```
rylandg, line 1:
```suggestion
# Nexus
```

rylandg, line 1:
```suggestion
# Nexus
```

rsnodgrass:
@rylandg Just a thought, reading texts that mention Arbitrary Length Operation (ALO) doesn't quickly make sense without the additional explanation that the "length" is actually length of time. Calling these Arbitrary Duration Operations (ADO) as terminology might make this much clearer conceptually. Especially for those new to thinking in terms of Durable Execution.

nexus/alo-strawman.md
```diff
@@ -0,0 +1,61 @@
...
+**Note:** Proto is merely a medium for conveying a contract and is not a requirement or key aspect of this sample.
+
+```protobuf
+// An ALO is started via the Start method
+message AloInfo {
+  string id = 1;
+  Status status = 2;
+  map<string, string> metadata = 3;
+
+  enum Status {
```
drewhoskins-stripe, line 27: These values don't match what you listed above.
In the offsite, we talked about an ENQUEUED-ish status. That would be the state before deduplication, rate limiting, or other async validations.
I notice you mentioned a REJECTED status above. ENQUEUED would be the status that might still result in a REJECTED, whereas RUNNING would no longer be able to transition to that state.

nexus/alo-strawman.md
```diff
@@ -0,0 +1,61 @@
...
+}
+```
+
+### A bit about Ids
+
+![Lost the client Id](./images/lost-client-id.png)
+
+One of the most common debates around ALOs is whether unique Ids should be a fundamental part of the specification and design. 
+
+Considering ALOs represent the eventual completion of an operation, its impossible to imagine a design where they aren’t identifiable. Note that I did not say “addressable” because technically in the case of asynchronous result delivery it may not be required for the ALO itself to be addressable. That being said, even in the async result delivery case, the result must be delivered with a unique correlation Id.
```
drewhoskins-stripe, line 48: A bit confused here.  What is the value of the abstraction if I can't address it? Wouldn't AloHandler not work if we couldn't address it?

drewhoskins-stripe:
> Especially for those new to thinking in terms of Durable Execution

Or just "Durable operations"  I tried "Durable Async Operations" at Stripe but people complained that there's already a DAO concept.

rsnodgrass:
To be honest, any of those suggested above so are clearer to the intent of the term than "Arbitrary Length Operation (ALO)":

*** Arbitrary Duration Operation (ADO)** or Arbitrary Duration Procedure (ADP) or Arbitrary Duration Activity (ADA)
* Durable Async Operations (DAO)
* **Durable Asynchronous Procedure (DAP)**
* Asynchronous Durable Operation (ADO) or Asynchronous Durable Execution (ADE)
* **Durable Operation (DO)**
* Durable Eventual Operation (DEO)

I really liked the original use of "Arbitrary" because it immediately is distinct from pre-conceived notions of what people think when they hear "asynchronous". The definition of arbitrary: *based on **random choice** or personal whim, rather than any reason or system* also clearly indicates that execution will likely not occur in order within completely independent execution plans (or workflows).

I still like Arbitrary Duration Operation (ADO) the best about of the above, with Durable Asynchronous Procedure (DAP) a close second, and Durable Operation (DO) third.

nexus/alo-strawman.md
cretz: Wrt ENQUEUED I think, with the removal of the requirement that Nexus requests are persisted, there is no persisted status before it is running. So while it technically could be in that state, I don't think we'd ever be able to return this info structure while it's waiting to be started.

cretz:
> I still like Arbitrary Duration Operation (ADO) the best about of the above

This gives me flashbacks to my old COM/MFC days ([ADO](https://en.wikipedia.org/wiki/ADO)). While I can understand the concern that "length" is more ambiguous than "duration" (which is unrelated to "durable" or durability), I personally like "ALO" from a pure acronym/nomenclature POV and am not concerned that the three words that make it up are expected to be descriptive enough.

Basically I see it as "long running operations" (i.e. [AIP-151](https://google.aip.dev/151)) but "any length" of time instead of "long running".

joebowbeer:
> > I still like Arbitrary Duration Operation (ADO) the best about of the above
> 
> This gives me flashbacks to my old COM/MFC days ([ADO](https://en.wikipedia.org/wiki/ActiveX_Data_Objects)).

I still think that leading with ADO or any other acronym or initialism is creating unnecessary hurdles for initiates, adding conceptual complexity for no gain.

You want to create a world in which this kind of Operation is the only thing anyone cares about, right? So just call it Operation in your context. (Or if you want to make a land grab, call it Cloud Operation.)

If you ever want to talk about the other kinds of operations that programmers used to worry about, create an acronym for that: NADO = Non-Arbitrary-Duration-Operation.

nexus/nexus-strawman.md
```diff
@@ -0,0 +1,358 @@
...
+
+Instantiating a Nexus Service makes it possible for the Nexus Backend to receive traffic meant for that specific Instance. Nexus Instantiation does make your Nexus Service Instance endpoint(s) available for traffic, but until you register and run corresponding Nexus Workers, no forward progress will be made on incoming requests. In this way, sending requests at this stage is similar to starting Workflows without any Workflow Workers running.
+
+**Note:** Meaningfully persisting the in flight Nexus Requests is not a product level requirement. Some users may not want Nexus Request payloads to be persisted and we should leave an opening for that in the design.
+
+**Registration:**
+
+For the sake of this proposal, we will only consider the case where Nexus is registered via `nexus` CLI. In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance. Registering a Nexus Service Instance would be done with the following command structure:
+
+`tctl nexus register --name my-service --description "My service" --backend nexus-backend.example.com`
```
cretz, line 102:
```suggestion
`nexus register --name my-service --description "My service" --backend nexus-backend.example.com`
```

This is not Temporal specific

nexus/nexus-strawman.md
```diff
@@ -0,0 +1,358 @@
...
+
+### Nexus Instantiation
+
+Instantiating a Nexus Service makes it possible for the Nexus Backend to receive traffic meant for that specific Instance. Nexus Instantiation does make your Nexus Service Instance endpoint(s) available for traffic, but until you register and run corresponding Nexus Workers, no forward progress will be made on incoming requests. In this way, sending requests at this stage is similar to starting Workflows without any Workflow Workers running.
+
+**Note:** Meaningfully persisting the in flight Nexus Requests is not a product level requirement. Some users may not want Nexus Request payloads to be persisted and we should leave an opening for that in the design.
+
+**Registration:**
+
+For the sake of this proposal, we will only consider the case where Nexus is registered via `nexus` CLI. In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance. Registering a Nexus Service Instance would be done with the following command structure:
```
cretz, line 100: > In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance.

To clarify, of course anything you can do via a CLI you could do via low level API (that's what the CLI uses)

nexus/nexus-strawman.md
```diff
@@ -0,0 +1,358 @@
...
+
+## User flows
+
+### Hosting a Nexus Service
+
+1. User instantiates Nexus Service Instance using provided `nexus` CLI commands (also possible via SDK or API) and receives an Instance URI. Nexus is now able to receive requests to its frontend endpoint (though they don't get processed until step 4).
+1. User defines a Nexus Request Handler that will handle the processing of Nexus Requests for the instantiated Nexus Service Instance.
+1. User hosts their Nexus Request Handler using a Nexus Worker. They must specify the Nexus Service Instance URI when starting the Nexus Worker.
+
+    // TODO (Chad): show how registration would work for generic handler
```
cretz, line 32: You have that below in the "Nexus Request Handler hosting"

cretz (approved):
This all looks good to me. I think we can get started on technical design (basically the protocol/language APIs implemented)

cretz:
> You want to create a world in which this kind of Operation is the only thing anyone cares about, right?

The purpose of denoting an "ALO" is that, yes, there is a concept of a fast "sync operation" from a client POV (but not necessarily different API from inside a Temporal workflow, they all look async). When we get to a code generation stage this will be even more stark where some calls will return a value and others will return a promise of a value.

nexus/nexus-strawman.md
rylandg: I discussed with you that for the sake of this proposal I will not have Nexus as a standalone concept.

nexus/nexus-strawman.md
rylandg: Agreed.

rylandg, line 100:
```suggestion
For the sake of this proposal, we will only consider the case where Nexus is registered via `nexus` CLI (which will be possible via an API that is exposed). In the future we may evaluate and consider alternative methods of registering a Nexus Service Instance. Registering a Nexus Service Instance would be done with the following command structure:
```

nexus/nexus-strawman.md
rylandg:
```suggestion
```

nexus/alo-strawman.md
rylandg: The general thought process is that addressable is not equivalent to identifiable. For example, should it be possible to have a ALO operation completed via a Kafka backend? If the answer is yes, its not possible to really address an item in a queue, especially if its expected to be delivered back to the caller and not requested synchronously. In this case you still need a unique & shared Id in order to correlate the result delivery with the original invocation, but that may not be delivered through an explicit call to GetResult.

rylandg (commented):
Really appreciate the discussion @rsnodgrass & @joebowbeer. At minimum I will make sure we have some internal discussions around the ALO acronym and whether it will cause us problems down the road. I think there were some great alternative suggestions made in this proposal so worst case we have something to fall back on :)

nexus/alo-strawman.md
rylandg: Chad is correct. We've removed the explicit persistence requirement and therefore this isn't a relevant scenario anymore.

rylandg:
We are officially closing this proposal and are in the technical design phase.

PR #67, opened 2022-09-29
Propose new search attribute APIs

bergundy:
Rendered here: https://github.com/bergundy/proposals/blob/ts-search-attributes/typescript/search-attributes.md

### DIFF
--- a/typescript/search-attributes.md
+++ b/typescript/search-attributes.md
@@ -1,0 +1,86 @@
+## Search attributes V2
+
+### Why do we need this?
+
+In the current version of the SDK, search attributes are always arrays. In order for the server to support visibility
+on top of standard persistence, we need to redo the search attributes API to enforce single values for all search
+attributes except for keywords. The server will eventually drop support for multiple search attribute values for
+non-keywords.
+
+### Minimal proposal
+
+Make the E2E experience consistent, including query construction for listing workflows and search attribute
+registration.
+
+`workflows.ts`
+
+```ts
+import { searchAttributes } from '@temporalio/common';
+
+export const customerAttribute = searchAttributes.defineAttribute(searchAttributes.Text, 'customer');
+export const dateOfPurchaseAttribute = searchAttributes.defineAttribute(Date, 'dateOfPurchase');
+
+export async function myWorkflow() {
+  // Get the value of a search attribute - returns string | undefined here
+  const customer = customerAttribute.value();
+  // ... random workflow logic ...
+
+  // Set the value of a search attribute, all set and unset calls are buffered and flushed out as a single command at
+  // the end of an activation
+  dateOfPurchaseAttribute.set(new Date());
+  // Explicit unset API
+  dateOfPurchaseAttribute.unset();
+  // ...
+}
+```
+
+`client.ts`
+
+```ts
+import { searchAttributes } from '@temporalio/common';
+import { Client, Q } from '@temporalio/client';
+import { customerAttribute, dateOfPurchaseAttribute, myWorkflow } from './workflows';
+
+const client = new Client(/* options */);
+
+// Only available in OSS for now
+await client.operator.registerSearchAttributes(customerAttribute, dateOfPurchaseAttribute);
+// Other methods will be exposed too
+
+const iterator = client.workflow.list({
+  query: Q.filter(
+    // WorkflowType is a pre-defined search attribute
+    Q.and(searchAttributes.WorkflowType.eq(myWorkflow.name), customerAttribute.eq('customer-id-1234')).orderBy(
+      dateOfPurchaseAttribute,
+      'ASC' // Optionally specify DESC | ASC
+    )
+  ),
+  pageSize: 100, // optional
+});
+
+// Returns something like the TS wrapper returned by Describe - the API returns
+// temporal.api.workflow.v1.WorkflowExecutionInfo
+for await (const { workflowId, ...rest } of iterator) {
+  const handle = client.workflow.getHandle(workflowId);
+  await handle.cancel();
+}
+
+const handle = client.getHandle(someWorkflowId);
+const info = await handle.describe();
+const customer = info.getAttribute(customerAttribute); // includes runtime validation - returns string | undefined
+// -- OR --
+const customer = info.getAttribute<string>(customer); // no runtime validation - returns string | undefined
+```
+
+### What about the current APIs?
+
+Search attributes are exposed in the SDK today in 3 places:
+
+1. In the client - `WorkflowClient.describe()` return value (`WorkflowExecutionDescription.searchAttributes`)
+2. In the workflow - `workflowInfo()` return value (`WorkflowInfo.searchAttributes`)
+3. `upsertSearchAttributes` call from the workflow
+
+We'll deprecate all of these APIs and point to the alternatives presented above.
+
+If `upsertSearchAttributes` is called in the same activation after as the new `set|unset` APIs are used, it will trigger
+flushing the buffered changes.
--- a/typescript/search-attributes.md
+++ b/typescript/search-attributes.md
@@ -15,14 +15,17 @@ registration.
 `workflows.ts`
 
 ```ts
-import { searchAttributes } from '@temporalio/common';
+import { defineAttribute, Keyword, Text } from '@temporalio/common/search-attributes';
 
-export const customerAttribute = searchAttributes.defineAttribute(searchAttributes.Text, 'customer');
-export const dateOfPurchaseAttribute = searchAttributes.defineAttribute(Date, 'dateOfPurchase');
+export const tagsAttribute = defineAttribute(Keyword, 'tags');
+export const customerAttribute = defineAttribute(Text, 'customer');
+export const dateOfPurchaseAttribute = defineAttribute(Date, 'dateOfPurchase');
 
 export async function myWorkflow() {
   // Get the value of a search attribute - returns string | undefined here
   const customer = customerAttribute.value();
+  // Get the value of a search attribute - returns string[] (which will be empty if unset)
+  const tags = tagsAttribute.value();
   // ... random workflow logic ...
 
   // Set the value of a search attribute, all set and unset calls are buffered and flushed out as a single command at
@@ -37,9 +40,9 @@ export async function myWorkflow() {
 `client.ts`
 
 ```ts
-import { searchAttributes } from '@temporalio/common';
+import { WorkflowType } from '@temporalio/common/search-attributes';
 import { Client, Q } from '@temporalio/client';
-import { customerAttribute, dateOfPurchaseAttribute, myWorkflow } from './workflows';
+import { customerAttribute, dateOfPurchaseAttribute, myWorkflow, tagsAttribute } from './workflows';
 
 const client = new Client(/* options */);
 
@@ -50,7 +53,7 @@ await client.operator.registerSearchAttributes(customerAttribute, dateOfPurchase
 const iterator = client.workflow.list({
   query: Q.filter(
     // WorkflowType is a pre-defined search attribute
-    Q.and(searchAttributes.WorkflowType.eq(myWorkflow.name), customerAttribute.eq('customer-id-1234')).orderBy(
+    Q.and(WorkflowType.eq(myWorkflow.name), customerAttribute.eq('customer-id-1234')).orderBy(
       dateOfPurchaseAttribute,
       'ASC' // Optionally specify DESC | ASC
     )
@@ -67,9 +70,19 @@ for await (const { workflowId, ...rest } of iterator) {
 
 const handle = client.getHandle(someWorkflowId);
 const info = await handle.describe();
-const customer = info.getAttribute(customerAttribute); // includes runtime validation - returns string | undefined
+const customer = info.getSearchAttribute(customerAttribute); // includes runtime validation - returns string | undefined
+const tags = info.getSearchAttribute(tagsAttribute);
+// ^ includes runtime validation - returns string[] (empty if not set)
+
+// TODO: should we have a string based API? Will it return single or multiple values?
+// It could look like this:
+
+const customer = info.getSearchAttribute<string>('customer'); // no runtime validation - may return undefined
+const tag = info.getSearchAttribute<string>('tags'); // no runtime validation - returns first tag or undefined
+
 // -- OR --
-const customer = info.getAttribute<string>(customer); // no runtime validation - returns string | undefined
+const [customer] = info.getSearchAttribute<string>('customer'); // returns an array that may be empty
+const tags = info.getSearchAttribute<string>('tags');
 ```
 
 ### What about the current APIs?


typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
+## Search attributes V2
+
+### Why do we need this?
+
+In the current version of the SDK, search attributes are always arrays. In order for the server to support visibility
+on top of standard persistence, we need to redo the search attributes API to enforce single values for all search
+attributes except for keywords. The server will eventually drop support for multiple search attribute values for
+non-keywords.
```
lorensr, line 8: We could maintain the array & support for multiple values when backed by ES, while supporting single-value arrays for standard persistence. There are use cases for it. Was it decided that there are not enough use cases to warrant the complexity?

bergundy: We don't want different experience based on the backend.
Research shows that multiple values are almost never used.

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+Make the E2E experience consistent, including query construction for listing workflows and search attribute
+registration.
+
+`workflows.ts`
+
+```ts
+import { searchAttributes } from '@temporalio/common';
+
+export const customerAttribute = searchAttributes.defineAttribute(searchAttributes.Text, 'customer');
+export const dateOfPurchaseAttribute = searchAttributes.defineAttribute(Date, 'dateOfPurchase');
```
Sushisource, line 21: I might drop `define` here. You're not really defining it in the verb sense, especially if it already exists, rather you are referring to some existing definition. Perhaps `attributeDefinition` then?

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+await client.operator.registerSearchAttributes(customerAttribute, dateOfPurchaseAttribute);
+// Other methods will be exposed too
+
+const iterator = client.workflow.list({
+  query: Q.filter(
+    // WorkflowType is a pre-defined search attribute
+    Q.and(searchAttributes.WorkflowType.eq(myWorkflow.name), customerAttribute.eq('customer-id-1234')).orderBy(
+      dateOfPurchaseAttribute,
+      'ASC' // Optionally specify DESC | ASC
+    )
```
Sushisource, line 56: This looks nice but I wouldn't underestimate the complexity of a query builder. This isn't a small amount of work to implement properly, especially if it's type-safe.

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+for await (const { workflowId, ...rest } of iterator) {
+  const handle = client.workflow.getHandle(workflowId);
+  await handle.cancel();
+}
+
+const handle = client.getHandle(someWorkflowId);
+const info = await handle.describe();
+const customer = info.getAttribute(customerAttribute); // includes runtime validation - returns string | undefined
+// -- OR --
+const customer = info.getAttribute<string>(customer); // no runtime validation - returns string | undefined
```
Sushisource, line 72: Is this argument supposed to be `'customer'` ?

Sushisource (commented):
Overall I like this. The query builder is perhaps a bit much, but it's independent of the rest of this. I would just call that out as maybe an additional improvement for later.

typescript/search-attributes.md
bergundy: I'd be fine with just `attribute` TBH.

typescript/search-attributes.md
bergundy: Yes, thanks for catching it

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+export const customerAttribute = searchAttributes.defineAttribute(searchAttributes.Text, 'customer');
+export const dateOfPurchaseAttribute = searchAttributes.defineAttribute(Date, 'dateOfPurchase');
+
+export async function myWorkflow() {
+  // Get the value of a search attribute - returns string | undefined here
+  const customer = customerAttribute.value();
+  // ... random workflow logic ...
+
+  // Set the value of a search attribute, all set and unset calls are buffered and flushed out as a single command at
+  // the end of an activation
```
cretz, line 29: > and unset calls are buffered and flushed out as a single command at the end of an activation

I assume last write wins when altering the same attribute before next command?

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Minimal proposal
+
+Make the E2E experience consistent, including query construction for listing workflows and search attribute
+registration.
+
+`workflows.ts`
+
+```ts
+import { searchAttributes } from '@temporalio/common';
```
cretz, line 18: What does `searchAttributes` look like? Can I iterate all of them inside the workflow? Can I do things like unset all of the ones with a certain prefix? (i.e. a replace-all)

typescript/search-attributes.md
cretz: For me it's not about difficulty of building, it's about maintenance and compatibility. How well defined are list filters? If you can give me a grammar that you guarantee won't change in an incompatible way, then I can support a programmatic builder. Even then though, though there's some type safety, you'll end up where SQL query builders always end up where people will escape back into string literals for extreme flexibility.

Also, this API needs to be fledged out some more. Why `Q` when it's a "list filter" in our docs? Why not a builder pattern? How can I embed a custom query string? Why is "orderBy" on a "query" part of something called a "filter"? (granted that's a Temporal naming problem).

typescript/search-attributes.md
cretz, line 72: We should standardize on the word "search attribute" or "attribute".

cretz, line 72: I assume you'll do the same for memo here and for the in-workflow part? (though memo can be any type and is not part of your filter work)

cretz, line 72: Can we mark this new API experimental until implemented in another SDK?

cretz (approved):

typescript/search-attributes.md
bergundy: I checked with the server team and they've confirmed that the entire list of capabilities is documented here: https://docs.temporal.io/concepts/what-is-a-list-filter.

I'm not proposing this as the only way, The `list` API will accept a structured query and a plain string, we can of course make the query builder accept strings too but I don't think we need that at this point.

The doc page mentions list filters but it actually talks about queries, while the projection part of the query is omitted in Temporal, the rest is a query, it's composed of filters and ordering operators.
filter is not a what you get when you use `Q.filter`, it's just a factory for creating a new query with a single filter.
Alternatively, we could have a `Q()` method or constructor instead, I'm open to discussing that.

Essentially Q is a builder.

You can compose multiple operations:
```
Q.filter(...).filter(...).orderBy(...).orderBy(...)
```

NOTE that queries are immutable and `filter` and `orderBy` return new instances.

typescript/search-attributes.md
bergundy: Yeah, we should probably just stick to search attribute.

bergundy: There's no need to do this for memo, but I do think that the "upsert memo" API in the workflow module will do batching as well.

bergundy: Yes, we should mark any new APIs experimental for a certain period IMHO

typescript/search-attributes.md
cretz: Is this builder solving a need people have?

typescript/search-attributes.md
cretz: Getting a single memo value by key for a specific type makes sense though, right? `info.getMemo(myStringMemo)` and `info.getMemo<string>("my-string-memo")`

Sushisource: Works for me

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+import { customerAttribute, dateOfPurchaseAttribute, myWorkflow } from './workflows';
+
+const client = new Client(/* options */);
+
+// Only available in OSS for now
+await client.operator.registerSearchAttributes(customerAttribute, dateOfPurchaseAttribute);
+// Other methods will be exposed too
+
+const iterator = client.workflow.list({
+  query: Q.filter(
```
Quinn-With-Two-Ns, line 51: does `Q.filter` produce a filter query as a string or an object?

typescript/search-attributes.md
Quinn-With-Two-Ns, line 56: Can constructing a filter always succeed? or should we have a users call a `.build()` method at the end to return any errors in the construction?

bergundy: Discoverability of possible operators, learning our query syntax, escaping values.

bergundy: Avoiding typos

bergundy: They always succeed or throw on construction.

typescript/search-attributes.md
bergundy: Hmm.. right now memos can be any user object so we can't do runtime validation like we do for search attributes.
ATM users can access their memo as: `info.memo['my-string-memo'] as string`.

Adding the getter doesn't provide that much value IMHO but I'll consider adding it for API consistency with search attributes.

typescript/search-attributes.md
bergundy: object

typescript/search-attributes.md
bergundy: Yes, I thought that was obvious but I can add a comment.

typescript/search-attributes.md
bergundy: I kind of like the verb `define` for consistency with `defineQuery` and `defineSignal`.

typescript/search-attributes.md
bergundy: This is just a set of functions and constant, it includes things like:

```
defineAttribute
WorkflowType // built-in attribute
```

typescript/search-attributes.md
cretz: Sorry, to clarify I mean as in a struggle people have today. They can typo lots of strings, e.g. task queues, I am just wondering if anyone is clamoring for this.

Also, can we get an exact query grammar? (if not in this issue, just in docs somewhere)

typescript/search-attributes.md
cretz: Was mostly obvious, no comment needed

typescript/search-attributes.md
cretz: Can I iterate all of them inside the workflow? Can I do things like unset all of the ones with a certain prefix? (i.e. a replace-all)

typescript/search-attributes.md
cretz: We added [workflow.memo](https://python.temporal.io/temporalio.workflow.html#memo) and [workflow.memo_value](https://python.temporal.io/temporalio.workflow.html#memo_value) (and the client handle describe response equivalents) after discussion because there is no simple way to deserialize a memo value without manually invoking a converter.

No need to add it now, but I think making memo and search attributes APIs match as much as possible makes sense (though obviously the latter is much more limited on types).

typescript/search-attributes.md
bergundy: The doc page I linked above.

I don't think people are clamoring for this, I just this it's a general improvement and is pretty straightforward to do.

typescript/search-attributes.md
bergundy: You'd have to define your own set of search attributes, it's up to you how to access those.
We can expose an API to list the set of search attributes, do you think it's valuable?
Since search attributes have to be registered, I don't know if there's a real need for it.

typescript/search-attributes.md
bergundy: If we had type hints in TS I'd do it, but we currently don't support that.

typescript/search-attributes.md
dnr: Maybe not "clamoring", but I would have appreciated a query builder for the one spot in the server where I had to construct a query for internal use. Mostly just so I don't have to think about quoting or escaping.

typescript/search-attributes.md
mjameswh: TypeScript doesn't allow object destructuring on imports statements. That means that one would always have to write `searchAttributes.defineAttribute(...)`, instead of simply `defineAttribute(...)`.

I think `defineAttribute` should be outside of this object, and `searchAttributes` should contain only build-in search attributes...

bergundy: There's always:
```ts
const { defineAttribute, Keyword } = searchAttributes
```

Alternative (which I don't love):
```ts
import { defineAttribute, Keyword } from '@temporalio/common/lib/search-attributes'
```

Alternative 2 (possible) with some extra files:
```ts
import { defineAttribute, Keyword } from '@temporalio/common/search-attributes'
```

Alternative 3 (really don't love):
```
import { defineAttribute, Keyword } from '@temporalio/search-attributes'
```

bergundy: Maybe let's do (2)?

typescript/search-attributes.md
lorensr: i'll clamor to avoid thinking about quotes in quotes 😊

typescript/search-attributes.md
```diff
@@ -0,0 +1,86 @@
...
+    // WorkflowType is a pre-defined search attribute
+    Q.and(searchAttributes.WorkflowType.eq(myWorkflow.name), customerAttribute.eq('customer-id-1234')).orderBy(
```
mjameswh, line 53: I think it should be possible to call `orderBy` only on the output of `filter`... Not on a "boolean operator fragment" (call it whatever you want, I mean the object returned by `and` or `or`).

typescript/search-attributes.md
lorensr, line 72:
```suggestion
const customer = info.getAttribute<string>(customerAttribute); // no runtime validation - returns string | undefined
const customer = info.getAttribute('customer'); // no runtime validation - returns unknown
```

lorensr (approved):

typescript/search-attributes.md
mjameswh: Was thinking more about this:

```
import { defineAttribute, SearchAttributes } from '@temporalio/common';

const customerAttribute = defineAttribute('string', 'customer');

Q.filter(customerAttribute, 'johnny')
 .filter(SearchAttributes.workflowType, 'CustomerWorkflow');
```

typescript/search-attributes.md
bergundy: You are correct, `and` and `or` are different, they construct `Filter` objects not `Query` objects.

typescript/search-attributes.md
lorensr: I like 2

mjameswh: `SearchAttributes` is essentially an enum, from user's POV

typescript/search-attributes.md
lorensr: Agreed. My guess is that's what Roey intended. `Q.filter().orderBy()`

typescript/search-attributes.md
bergundy: I like 2 as well.
Not sure why `SearchAttributes` are an enum..

typescript/search-attributes.md
mjameswh: Yeah... There's some example around where parenthesis are misplaced, but as long as we agree on the principle, I'm fine.

typescript/search-attributes.md
mjameswh: I think "built in search attributes" should be grouped together. Not formally an enum, but an object on which every field is a builtin search attribute.

Having an export named simply `WorkflowType` gives the impression that it is a type. Writing it with a lower case makes it only marginally better.

bergundy: Sure, maybe have a `builtins` constant with all of the builtin search attributes.

mjameswh (approved):
LGTM

typescript/search-attributes.md
```diff
@@ -0,0 +1,99 @@
...
+import { Client, Q } from '@temporalio/client';
+import { customerAttribute, dateOfPurchaseAttribute, myWorkflow, tagsAttribute } from './workflows';
+
+const client = new Client(/* options */);
+
+// Only available in OSS for now
+await client.operator.registerSearchAttributes(customerAttribute, dateOfPurchaseAttribute);
+// Other methods will be exposed too
+
+const iterator = client.workflow.list({
```
cretz, line 53: Quick question as I start to work on this in Python. So this solves https://github.com/temporalio/sdk-features/issues/75 too? I assume this is interceptable? Also, can you show a quick example of status filtering? Is `ExecutionStatus` a search attribute?

bergundy: This will solve the high level SDK list API, assuming you want to add the query builder (you should IMHO).
I don't see much value in making it intercept-able if there are no headers to pass to the server.
`ExecutionStatus` along with the rest of the [default search attributes](https://docs.temporal.io/visibility#default-search-attributes) will be defined in the SDK.

For search attributes that we know are enums I'd definitely restrict the type (I think it's only ExecutionStatus).

Something like:
```ts
Q.filter(builtinSearchAttributes.ExecutionStatus.eq('RUNNING'))
```

cretz: >  I don't see much value in making it intercept-able if there are no headers to pass to the server.

While I thought similar, I was overridden on workflow handle "describe". I think we are now forced to be consistent here.

bergundy:
Closed in favor of #75

PR #68, opened 2022-10-04
[Ruby SDK] Type hinting, Converters, Codecs and Errors

antstorm:
# Summary

This proposal outlines how data conversion will work in Ruby SDK along with the Error hierarchy. This is required to proceed with the implementation of the Ruby SDK.

### DIFF
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -141,9 +141,11 @@ And a payload codec is expected to implement this interface:
 ```ruby
 interface _PayloadCodec
   # Takes an array of Payload objects and returns an array of encoded Payload object
+  # The resulting Array is expected to hold at least one Payload
   def encode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
 
   # Takes an array of Payloads object and returns an array of decoded Payload object
+  # The resulting Array is expected to hold at least one Payload
   def decode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
 end
 ```
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -1,0 +1,273 @@
+# Ruby SDK - Type hinting, Converters, Codecs and Errors
+
+This proposal goes into details regarding the implementation of data conversion pipeline (including
+type hints, converters and codecs) as well as outlines the hierarchy of exception classes.
+
+## Type hinting
+
+The Ruby typing system is not something we can currently rely on for specifying the types of all the
+inputs and outputs. Instead we will add custom "annotations" (technically these are just method
+calls) to the activity/workflow definitions in order to provide type hinting. Here's the basic
+example:
+
+```ruby
+class SumActivity < Temporal::Activity
+  input Integer, Integer
+  output Integer
+
+  def execute(a, b)
+    a + b
+  end
+end
+```
+
+We will support most basic Ruby types (e.g. `Integer`, `Float`, `String`, `Symbol`, `Boolean`,
+`Array`, `Hash`, etc). This will allow the SDK to perform a runtime check of the arguments and
+return values to ensure they match the specified types.
+
+We will also leverage this system to allow SDK users to specify their own types:
+
+```ruby
+class MyInput < Struct.new(:a, :b)
+  def self.load(hash)
+    new(hash['a'], hash['b'])
+  end
+
+  def dump
+    { 'a' => a, 'b' => b } # explicit for illustration purposes, can use #to_h instead
+  end
+end
+
+class MyOutput < Struct.new(:sum)
+  def self.load(sum)
+    new(sum.to_i)
+  end
+
+  def dump
+    sum.to_s # string used for illustration purposes, can use Integer directly instead
+  end
+end
+
+class MyWorkflow < Temporal::Workflow
+  input MyInput
+  output MyOutput
+
+  def execute(input)
+    MyOutput.new(input.a + input.b)
+  end
+end
+```
+
+Custom types are expected to implement the following interface:
+
+```ruby
+interface _CustomType[U]
+  # Take a primitive type and return an instance of the implementing class
+  def self.load: (U) -> instance
+
+  # Returns a trivially serializable type representing the custom object
+  def dump: -> U
+end
+```
+
+The same exact concept applies to signals and queries:
+
+```ruby
+class MyWorkflow < Temporal::Workflow
+  signal :add, input: Integer
+  query :get_sum, output: MyOutput
+
+  def execute
+    ...
+  end
+end
+```
+
+Notes:
+
+- This feature is 100% optional. Without these definitions the output from converters will be used
+  directly as arguments and return values
+- The full potential of this solution is obviously limited to a single language SDKs
+- When sending a custom input type to another SDK, workflow/activity definitions (without the
+  implementation) can still be leveraged
+- A custom output type sent to another SDK can still be `dump`ed into a matching JSON or other
+  serializable structure
+- The same approaches apply to inputs and outputs received from other SDKs
+
+
+## Payload handling
+
+Ruby SDK will use the model proposed [here](https://docs.temporal.io/docs/concepts/what-is-a-data-converter/)
+to distinguish between payload converters and codecs.
+
+The main difference is that a Converter translates between any Ruby object and a Payload, while
+a Codec is a lower level construct that encodes/decodes from a Payloads protobuf to a Payloads
+protobuf. The former is intended to support specific serialisation formats, while the latter is for
+modifying a binary that is transmitted over the wire.
+
+
+### Converters
+
+A payload converter is expected to implement the following interface:
+
+```ruby
+interface _PayloadConverter[U]
+  # Returns whether or not it can decode a given Payload
+  def can_decode?: (Temporal::Payload) -> Bool
+
+  # Takes any Ruby object and returns a Payload object or nil if unable to convert
+  def to_payload: (U) -> Temporal::Payload
+
+  # Takes a Payload object and returns a Ruby object. Only called on a matching encoding
+  def from_payload: (Temporal::Payload) -> U
+end
+```
+
+The SDK will provide an implementation of basic converters — Null (for `nil`s), Bytes (any stream of
+bytes) and JSON (using the `oj` gem and potentially allowing the SDK user to specify the
+[mode](https://github.com/ohler55/oj/blob/develop/pages/Modes.md)). We will also provide a
+`Composite` converter to combine multiple converters behind the same interface.
+
+
+### Codecs
+
+And a payload codec is expected to implement this interface:
+
+```ruby
+interface _PayloadCodec
+  # Takes an array of Payload objects and returns an array of encoded Payload object
+  def encode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
+
+  # Takes an array of Payloads object and returns an array of decoded Payload object
+  def decode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
+end
+```
+
+
+### Configuring client/worker
+
+In order to use the defined converters and codecs, these need to be supplied to a client and worker.
+
+```ruby
+converter = Temporal::Converters::Composite.new(
+  Temporal::Converters::Null,
+  Temporal::Converters::Bytes,
+  MyProtobufConverter.new
+)
+codecs = [EncryptionCodec.new(KEY_ID), CompressionCodec]
+
+connection = Temporal::Connection.new('http://localhost:7233')
+client = Temporal::Client.new(
+  connection,
+  'my-namespace',
+  payload_converter: converter,
+  payload_codecs: codecs
+)
+
+worker = Temporal::Worker.new(
+  connection,
+  namespace: 'demo',
+  task_queue: 'demo',
+  workflows: [...],
+  activities: [...],
+  payload_converter: converter,
+  payload_codecs: codecs
+)
+```
+
+*This does indeed involve some duplication and we might simplify this initialization later
+on. But it's outside the scope for this proposal.*
+
+Notes:
+
+- Converters and codecs both can be classes and class instances as long as the given object
+  implements the required interface
+- This API is still being figured out to accommodate future features and configuration params
+
+
+## Data conversion
+
+All the previously outlined modules are required to facilitate the process of data conversion.
+Here's the diagram representing it:
+
+![Data conversion](./images/data-conversion.png)
+
+This describes the process that every piece of data (workflow/activity args and return values,
+signal inputs, query results, headers, memos, etc) goes through getting transformed from the
+user-land object to something that can be transmitted over the wire.
+
+*NOTE: As mentioned previously, the "Type hints" are optional and will be omitted altogether unless
+defined.*
+
+The reverse process is applied when converting the data from a protobuf received over the wire back
+to a use-land object.
+
+
+## Errors
+
+Here's the Error hierarchy that is planned for the Ruby SDK:
+
+```ruby
+module Temporal
+  # top level error superclass
+  class Error < StandardError
+
+  # superclass for all errors within the SDK itself
+  class InternalError < Error
+
+  class ClientError < InternalError # errors specific to Client
+  class MissingNamespace < ClientError
+  ...
+
+  class WorkerError < InternalError # errors specific to Worker
+  class TooManyStickyWorkflows < WorkerError
+  ...
+
+  # superclass for connection and network errors
+  class NetworkError < Error
+
+  class TimeoutError < NetworkError
+  class HostUnreachable < NetworkError
+  ...
+
+  # superclass for errors related to the server and it's responses
+  class ServerError < Error
+
+  class MissingWorkflowHistory < ServerError
+  class ExpectedSingleCloseEvent < ServerError
+  ...
+
+  # superclass for all API failure responses
+  class APIError < ServerError
+
+  class WorkflowExecutionAlreadyStarted < APIError
+  class NamespaceNotActive < APIError
+  ...
+
+  # superclass for type hinting errors
+  class TypeError < Error
+
+  class WrongType < TypeError
+  class UnsupportedType < TypeError
+  ...
+
+  # superclass for workflow errors
+  class WorkflowError < Error
+
+  class Cancelled < WorkflowError
+  class UnsupportedQuery < WorkflowError
+  class DeadLockDetected < WorkflowError
+  ...
+
+  # superclass for activity errors
+  class ActivityError < Error
+
+  class MissingActivity < ActivityError
+  class MissingHeartbeat < ActivityError
+  ...
+```
+
+Notes:
+
+- This is just an example, but obviously more error types are expected
+- Some concrete error class placement might be confusing and will be clarified later in the process
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -147,6 +147,12 @@ error otherwise), an encoding payload converter (the one used with the `Composit
 only expected to handle values and Payload matching it's specified encoding.
 
 
+#### JSON Converter
+
+The default JSON converter will respect [the interface](https://ruby-doc.org/stdlib-3.1.2/libdoc/json/rdoc/JSON.html#module-JSON-label-Custom+JSON+Additions)
+provided by the Ruby's JSON module to serialize/deserialize an object.
+
+
 ### Codecs
 
 And a payload codec is expected to implement this interface:
@@ -162,66 +168,6 @@ interface _PayloadCodec
   def decode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
 end
 ```
-
-
-### Configuring client/worker
-
-In order to use the defined converters and codecs, these need to be supplied to a client and worker.
-
-```ruby
-converter = Temporal::Converters::Composite.new(
-  Temporal::Converters::Null,
-  Temporal::Converters::Bytes,
-  MyProtobufConverter.new
-)
-codecs = [EncryptionCodec.new(KEY_ID), CompressionCodec]
-
-connection = Temporal::Connection.new('http://localhost:7233')
-client = Temporal::Client.new(
-  connection,
-  'my-namespace',
-  payload_converter: converter,
-  payload_codecs: codecs
-)
-
-worker = Temporal::Worker.new(
-  connection,
-  namespace: 'demo',
-  task_queue: 'demo',
-  workflows: [...],
-  activities: [...],
-  payload_converter: converter,
-  payload_codecs: codecs
-)
-```
-
-*This does indeed involve some duplication and we might simplify this initialization later
-on. But it's outside the scope for this proposal.*
-
-Notes:
-
-- Converters and codecs both can be classes and class instances as long as the given object
-  implements the required interface
-- The order of both the `Composite` encoding converters and payload codecs matters
-- This API is still being figured out to accommodate future features and configuration params
-
-
-## Data conversion
-
-All the previously outlined modules are required to facilitate the process of data conversion.
-Here's the diagram representing it:
-
-![Data conversion](./images/data-conversion.png)
-
-This describes the process that every piece of data (workflow/activity args and return values,
-signal inputs, query results, headers, memos, etc) goes through getting transformed from the
-user-land object to something that can be transmitted over the wire.
-
-*NOTE: As mentioned previously, the "Type hints" are optional and will be omitted altogether unless
-defined.*
-
-The reverse process is applied when converting the data from a protobuf received over the wire back
-to a use-land object.
 
 
 ## Errors
@@ -402,3 +348,65 @@ interface _FailureConverter
   def from_failure: (Temporal::Failure) -> Temporal::Error
 end
 ```
+
+
+### Configuring client/worker
+
+In order to use the defined converters and codecs, these need to be supplied to a client and worker.
+
+```ruby
+converter = Temporal::Converters::Composite.new(
+  Temporal::Converters::Null,
+  Temporal::Converters::Bytes,
+  MyProtobufConverter.new
+)
+codecs = [EncryptionCodec.new(KEY_ID), CompressionCodec]
+
+connection = Temporal::Connection.new('http://localhost:7233')
+client = Temporal::Client.new(
+  connection,
+  'my-namespace',
+  payload_converter: converter,
+  payload_codecs: codecs,
+  failure_converter: MyFailureConverter.new,
+)
+
+worker = Temporal::Worker.new(
+  connection,
+  namespace: 'demo',
+  task_queue: 'demo',
+  workflows: [...],
+  activities: [...],
+  payload_converter: converter,
+  payload_codecs: codecs
+)
+```
+
+*This does indeed involve some duplication and we might simplify this initialization later
+on. But it's outside the scope for this proposal.*
+
+Notes:
+
+- All converters and codecs are optional and defaults will be used in case none are provided
+- Converters and codecs both can be classes and class instances as long as the given object
+  implements the required interface
+- The order of both the `Composite` encoding converters and payload codecs matters
+- This API is still being figured out to accommodate future features and configuration params
+
+
+## Data conversion
+
+All the previously outlined modules are required to facilitate the process of data conversion.
+Here's the diagram representing it:
+
+![Data conversion](./images/data-conversion.png)
+
+This describes the process that every piece of data (workflow/activity args and return values,
+signal inputs, query results, headers, memos, etc) goes through getting transformed from the
+user-land object to something that can be transmitted over the wire.
+
+*NOTE: As mentioned previously, the "Type hints" are optional and will be omitted altogether unless
+defined.*
+
+The reverse process is applied when converting the data from a protobuf received over the wire back
+to a use-land object.
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -222,53 +222,33 @@ module Temporal
   # superclass for all errors within the SDK itself
   class InternalError < Error
 
-  class ClientError < Error # errors specific to Client
-  ...
+  # errors specific to Client
+  class ClientError < Error
 
-  class WorkerError < InternalError # errors specific to Worker
-  ...
+  # errors specific to Worker
+  class WorkerError < InternalError
 
   # superclass for connection and network errors
-  class NetworkError < Error
-
-  class TimeoutError < NetworkError
-  class HostUnreachable < NetworkError
-  ...
+  class RPCError < Error
 
   # superclass for errors related to the server and it's responses
   class ServerError < Error
-
   class UnexpectedError < ServerError
-  ...
 
   # superclass for all API failure responses
   class APIError < ServerError
 
-  class WorkflowExecutionAlreadyStarted < APIError
-  class NamespaceNotFound < APIError
-  class NamespaceNotActive < APIError
-  ...
-
   # type hinting errors
   class TypeError < Error
-  ...
 
   # superclass for workflow errors
   class WorkflowError < Error
 
-  class Cancelled < WorkflowError
-  class UnsupportedQuery < WorkflowError
-  class DeadLockDetected < WorkflowError
-  ...
-
   # superclass for activity errors
   class ActivityError < Error
-
-  class MissingActivity < ActivityError
-  ...
 ```
 
 Notes:
 
-- This is just an example, but obviously more error types are expected
+- This list is not exhaustive, more errors will be added as parts of the SDK are getting built
 - Some concrete error class placement might be confusing and will be clarified later in the process
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -115,10 +115,11 @@ interface _PayloadConverter[U]
   # Returns whether or not it can decode a given Payload
   def can_decode?: (Temporal::Payload) -> Bool
 
-  # Takes any Ruby object and returns a Payload object or nil if unable to convert
+  # Takes any Ruby object and returns a Payload object
   def to_payload: (U) -> Temporal::Payload
 
-  # Takes a Payload object and returns a Ruby object. Only called on a matching encoding
+  # Takes a Payload object and returns a Ruby object
+  # Only called after a positive response from the #can_decode? method
   def from_payload: (Temporal::Payload) -> U
 end
 ```
@@ -127,6 +128,10 @@ The SDK will provide an implementation of basic converters — Null (for `nil`s)
 bytes) and JSON (using the `oj` gem and potentially allowing the SDK user to specify the
 [mode](https://github.com/ohler55/oj/blob/develop/pages/Modes.md)). We will also provide a
 `Composite` converter to combine multiple converters behind the same interface.
+
+To signal the `Composite` converter that a converter is unable to encode provided data it should
+throw a `Temporal::IncompatibleData` error. In this case the `Composite` converter will try the
+other converters it was initialized with.
 
 
 ### Codecs
@@ -215,12 +220,10 @@ module Temporal
   # superclass for all errors within the SDK itself
   class InternalError < Error
 
-  class ClientError < InternalError # errors specific to Client
-  class MissingNamespace < ClientError
+  class ClientError < Error # errors specific to Client
   ...
 
   class WorkerError < InternalError # errors specific to Worker
-  class TooManyStickyWorkflows < WorkerError
   ...
 
   # superclass for connection and network errors
@@ -233,22 +236,19 @@ module Temporal
   # superclass for errors related to the server and it's responses
   class ServerError < Error
 
-  class MissingWorkflowHistory < ServerError
-  class ExpectedSingleCloseEvent < ServerError
+  class UnexpectedError < ServerError
   ...
 
   # superclass for all API failure responses
   class APIError < ServerError
 
   class WorkflowExecutionAlreadyStarted < APIError
+  class NamespaceNotFound < APIError
   class NamespaceNotActive < APIError
   ...
 
-  # superclass for type hinting errors
+  # type hinting errors
   class TypeError < Error
-
-  class WrongType < TypeError
-  class UnsupportedType < TypeError
   ...
 
   # superclass for workflow errors
@@ -263,7 +263,6 @@ module Temporal
   class ActivityError < Error
 
   class MissingActivity < ActivityError
-  class MissingHeartbeat < ActivityError
   ...
 ```
 
--- a/ruby/type-hinting-converters-codecs-errors.md
+++ b/ruby/type-hinting-converters-codecs-errors.md
@@ -97,7 +97,7 @@ Notes:
 
 ## Payload handling
 
-Ruby SDK will use the model proposed [here](https://docs.temporal.io/docs/concepts/what-is-a-data-converter/)
+Ruby SDK will use the model proposed [here](https://docs.temporal.io/concepts/what-is-a-data-converter)
 to distinguish between payload converters and codecs.
 
 The main difference is that a Converter translates between any Ruby object and a Payload, while
@@ -193,7 +193,7 @@ module Temporal
   class UnexpectedError < RPCError
 
   # superclass for all API failure responses
-  class APIError < ServerError
+  class FailureError < Error
 
   # type hinting errors
   class TypeError < Error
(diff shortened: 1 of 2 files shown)


ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+
+### Converters
+
+A payload converter is expected to implement the following interface:
+
+```ruby
+interface _PayloadConverter[U]
+  # Returns whether or not it can decode a given Payload
+  def can_decode?: (Temporal::Payload) -> Bool
```
bergundy, line 116: I'm fine with this approach replacing the "EncodingDataConveter" concept in other SDKs.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+```ruby
+module Temporal
+  # top level error superclass
+  class Error < StandardError
+
+  # superclass for all errors within the SDK itself
+  class InternalError < Error
+
+  class ClientError < InternalError # errors specific to Client
```
bergundy, line 218: Why is `ClientError` and `Internal` error?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+  class MissingNamespace < ClientError
+  ...
+
+  class WorkerError < InternalError # errors specific to Worker
+  class TooManyStickyWorkflows < WorkerError
```
bergundy, line 223: You don't need this error

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+```ruby
+module Temporal
+  # top level error superclass
+  class Error < StandardError
+
+  # superclass for all errors within the SDK itself
+  class InternalError < Error
+
+  class ClientError < InternalError # errors specific to Client
+  class MissingNamespace < ClientError
```
bergundy, line 219: I think this is called `NamespaceNotFoundError` in other SDKs.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  class TimeoutError < NetworkError
+  class HostUnreachable < NetworkError
+  ...
+
+  # superclass for errors related to the server and it's responses
+  class ServerError < Error
+
+  class MissingWorkflowHistory < ServerError
+  class ExpectedSingleCloseEvent < ServerError
```
bergundy, line 237: These could just be a generic "UpexpectedError" IMHO.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+  class Cancelled < WorkflowError
+  class UnsupportedQuery < WorkflowError
+  class DeadLockDetected < WorkflowError
+  ...
+
+  # superclass for activity errors
+  class ActivityError < Error
+
+  class MissingActivity < ActivityError
+  class MissingHeartbeat < ActivityError
```
bergundy, line 266: What are these?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  class MissingWorkflowHistory < ServerError
+  class ExpectedSingleCloseEvent < ServerError
+  ...
+
+  # superclass for all API failure responses
+  class APIError < ServerError
+
+  class WorkflowExecutionAlreadyStarted < APIError
+  class NamespaceNotActive < APIError
```
bergundy, line 244: I think NamespaceNotFound goes in the APIError category.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+  ...
+
+  # superclass for type hinting errors
+  class TypeError < Error
```
bergundy, line 248: Is there not a built-in TypeError in ruby?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  class WrongType < TypeError
+  class UnsupportedType < TypeError
```
bergundy, line 251: nit: This is too detailed IMHO.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+The reverse process is applied when converting the data from a protobuf received over the wire back
+to a use-land object.
+
+
+## Errors
+
+Here's the Error hierarchy that is planned for the Ruby SDK:
+
+```ruby
+module Temporal
```
bergundy, line 211: You should express all of the `Failure` classes here too.

bergundy (commented):
Overall LGTM.

I'd define the failure converter while tackling this.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  def execute(a, b)
+    a + b
+  end
+end
+```
+
+We will support most basic Ruby types (e.g. `Integer`, `Float`, `String`, `Symbol`, `Boolean`,
+`Array`, `Hash`, etc). This will allow the SDK to perform a runtime check of the arguments and
+return values to ensure they match the specified types.
```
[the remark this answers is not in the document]
antstorm: Agree, but that's at least something. We'll aim to adopt a proper typing system as it matures

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+  def dump: -> U
+end
+```
+
+The same exact concept applies to signals and queries:
+
+```ruby
+class MyWorkflow < Temporal::Workflow
+  signal :add, input: Integer
+  query :get_sum, output: MyOutput
```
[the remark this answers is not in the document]
antstorm: It does, yes, you just list all of them like this:

```ruby
class MyWorkflow < Temporal::Workflow
  signal :add, input: Integer
  signal :divide, input: Float
  query :get_sum, output: MyOutput
  query :get_something_else, output: OtherOutput
end
```

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+A payload converter is expected to implement the following interface:
+
+```ruby
+interface _PayloadConverter[U]
+  # Returns whether or not it can decode a given Payload
+  def can_decode?: (Temporal::Payload) -> Bool
+
+  # Takes any Ruby object and returns a Payload object or nil if unable to convert
+  def to_payload: (U) -> Temporal::Payload
```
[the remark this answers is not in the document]
antstorm: As long as we're not abusing the `nil` here, I'm fine with a special error thrown to make it similar to other SDKs. I'll update the proposal

ruby/type-hinting-converters-codecs-errors.md
antstorm: Yes, good point. The logic here was — it's an internal error specific to the client, but these probably won't be internal errors, but network/usage errors

ruby/type-hinting-converters-codecs-errors.md
antstorm: That works 👍

ruby/type-hinting-converters-codecs-errors.md
antstorm: `MissingActivity` for when an activity is not registered with a worker and the other one I thought might be used for when an activity fails because of a missed heartbeat (but I guess it manifests itself as a timeout error, so probably not needed)

ruby/type-hinting-converters-codecs-errors.md
antstorm: Good point 👍

ruby/type-hinting-converters-codecs-errors.md
antstorm: There's `ArgumentError`, but that normally related to the argument count. There's also `TypeError`, which is normally thrown when the supplied type can not be implicitly converted to the expected type. Do you recon we should use the default error here?

ruby/type-hinting-converters-codecs-errors.md
antstorm: I'll trim these down

ruby/type-hinting-converters-codecs-errors.md
antstorm: I'll add these, sure

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+Custom types are expected to implement the following interface:
+
+```ruby
+interface _CustomType[U]
+  # Take a primitive type and return an instance of the implementing class
+  def self.load: (U) -> instance
+
+  # Returns a trivially serializable type representing the custom object
+  def dump: -> U
+end
```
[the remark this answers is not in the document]
antstorm: There's `load`/`dump` methods used by some serialization classes such as `JSON` and `Marshall` (but not user-defined classes). However I don't think there's anything this conflicts with

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+  output Integer
+
+  def execute(a, b)
+    a + b
+  end
+end
+```
+
+We will support most basic Ruby types (e.g. `Integer`, `Float`, `String`, `Symbol`, `Boolean`,
+`Array`, `Hash`, etc). This will allow the SDK to perform a runtime check of the arguments and
```
cretz, line 25: We should support everything in Ruby that can be serialized/deserialized via standard Ruby JSON (maybe this means we should set `object_class` during `JSON.parse` to this value?). Whatever the Ruby "way" is to JSON deserialize into known class structures we should support.

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
cretz: We should use Ruby's and not ours if we can. Ideally there should be nothing Temporal specific about converting classes to/from JSON.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+The Ruby typing system is not something we can currently rely on for specifying the types of all the
+inputs and outputs. Instead we will add custom "annotations" (technically these are just method
+calls) to the activity/workflow definitions in order to provide type hinting. Here's the basic
+example:
+
+```ruby
+class SumActivity < Temporal::Activity
+  input Integer, Integer
+  output Integer
```
cretz, line 16: How strict is this gonna be? For example, if the args come in as two strings, are we gonna fail the workflow task? More importantly, if a workflow returns a bool, are we not gonna complete the workflow?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+    ...
+  end
+end
+```
+
+Notes:
+
+- This feature is 100% optional. Without these definitions the output from converters will be used
+  directly as arguments and return values
+- The full potential of this solution is obviously limited to a single language SDKs
```
cretz, line 90: > The full potential of this solution is obviously limited to a single language SDKs

I think I disagree with this statement and if that's the case, we should discuss in which cases this is true. JSON is the default and is supported by all SDKs and no SDK has anything language specific about JSON by default. Can you clarify why this solution is limited to this SDK? Every SDK I know of can shape a class/interface/struct to work with other languages' equivalents.

ruby/type-hinting-converters-codecs-errors.md
cretz: I don't understand this. Where do I set the encoding string to be used in a dictionary? We shouldn't call `can_decode` in order for trying to decode when we should be using the encoding to determine that instead, right

Also, why does a high-level payload converter need to know whether it can decode. IMO a data converter has a single payload converter and that's it. If the payload converter wants to delegate, it can, but the high-level interfaces is just a simple to/from.

Arguably this would be called `can_convert` or similar since decoding is not the common word we use on converters.

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
cretz: There may be some confusion on the Python side (which I think is the best approach, though I am biased) considers a throwing of a payload converter to be a failure. "encoding payload converters" (i.e. ones that a higher level converter can delegate to based on that metadata field) return `None` (same as `nil`) to signal they can't convert something and to fall through. 

So Python definitely doesn't use exceptions as control flow, it uses nil _kinda_ like here but only for encoding-specific converters. Very simply, a converter should either convert a payload or throw, no nil. A encoding-specific converter should either convert a payload or return nil if it's not expected to handle it (or still throw if it is expected to handle it but something else is wrong). A caller of a converter should never have to handle nil.

Combining the concept of encoding-specific payload converters with the general concept of payload conversion is folly IMO.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  # Takes a Payload object and returns a Ruby object. Only called on a matching encoding
+  def from_payload: (Temporal::Payload) -> U
+end
+```
+
+The SDK will provide an implementation of basic converters — Null (for `nil`s), Bytes (any stream of
+bytes) and JSON (using the `oj` gem and potentially allowing the SDK user to specify the
+[mode](https://github.com/ohler55/oj/blob/develop/pages/Modes.md)). We will also provide a
+`Composite` converter to combine multiple converters behind the same interface.
```
cretz, line 129: IMO the "Composite" converter should require its children to implement a separate interface telling you the encoding so you can keep a dictionary. Iterating converters is for encoding not decoding.

ruby/type-hinting-converters-codecs-errors.md
cretz, line 119: Arguably I should be able to convert multiple values into a single payload or multiple payloads into multiple values or whatever. Python does this, but 1:1 is probably fine too.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+
+### Codecs
+
+And a payload codec is expected to implement this interface:
+
+```ruby
+interface _PayloadCodec
+  # Takes an array of Payload objects and returns an array of encoded Payload object
+  def encode: (Array[Temporal::Payload]) -> Array[Temporal::Payload]
```
cretz, line 139: Need to clarify that while the resulting array doesn't have to be the same length as the parameter array, it does have to have at least 1 value. Same for `decode`.

ruby/type-hinting-converters-codecs-errors.md
cretz, line 219: Can you clarify whether you expect to catch-and-convert these at the service level or the client level? If I call describe namespace on the client using the lower-level service, will this be thrown?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  class WorkerError < InternalError # errors specific to Worker
+  class TooManyStickyWorkflows < WorkerError
+  ...
+
+  # superclass for connection and network errors
+  class NetworkError < Error
+
+  class TimeoutError < NetworkError
+  class HostUnreachable < NetworkError
```
cretz, line 230: I am not sure you can reliably capture these. I think it'd be more reasonable, instead of trying to guess at all the ones you can catch, to instead just have an `RPCError` from Core that contains the gRPC status code and message. Then we can specialize as needed. I think you are overthinking this and adding way too many errors.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,273 @@
...
+
+  class WrongType < TypeError
+  class UnsupportedType < TypeError
+  ...
+
+  # superclass for workflow errors
+  class WorkflowError < Error
+
+  class Cancelled < WorkflowError
+  class UnsupportedQuery < WorkflowError
```
cretz, line 258: Where is this thrown/caught? Usually this is just an error returned to core from internal code. I can't think of the user-facing value of this. I strongly recommend not assuming/guessing which of these types of errors are needed, but instead letting them come about organically during development.

ruby/type-hinting-converters-codecs-errors.md
antstorm: There unfortunately isn't a standard way of doing this.  There's `oj` gem (a popular JSON lib) which has an optional [Object mode](https://github.com/ohler55/oj/blob/develop/pages/Modes.md#object-mode) that I have used in the former SDK, but that obviously won't be compatible with other SDKs

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
antstorm: But this isn't a converter. We can make this fit into the existing interface (used in `JSON`, `YAML` and `Marshall`), but it won't act as type hinting, more like a custom converter injected an a workflow/activity level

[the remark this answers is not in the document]
cretz: To clarify, does Ruby have a way to customize an object for JSON conversion? We should tell people to use that then. Sure we still say _which_ objects, but the "how" is via Ruby conversion.

ruby/type-hinting-converters-codecs-errors.md
antstorm: The emphasis is on the "**full** potential". Meaning that if your Ruby workflow is expecting a non-JSON compatible type as an input, then the calling SDK (assuming non-Ruby) will have to adjust the payload to match the expectations. It is possible, but requires extra work.

Imagine you are expecting a custom class (`MyInput`) that consists of two `DateTime` values — `from` and `to`. In a Ruby-only setup the object is `dump`ed and `load`ed using the same exact class, so the experience is seamless. However in when a caller is a non-Ruby SDK — it'll have to produce a JSON that decodes into something that `MyInput.load` expects (and JSON obviously doesn't have a `DateTime` type).

So while it can work, it is not using the full potential, since either encoding or decoding will have to be tuned to match the expectations of a type hint.

ruby/type-hinting-converters-codecs-errors.md
antstorm: I see, that makes sense

ruby/type-hinting-converters-codecs-errors.md
cretz: > Meaning that if your Ruby workflow is expecting a non-JSON compatible type as an input

This should not be possible unless you did a custom conversion. By default all serializable types should only support cross-SDK compatible types.

> In a Ruby-only setup the object is dumped and loaded using the same exact class, so the experience is seamless

Ah, so you did custom conversion support. w/ dump and load. Well of course custom conversion code is per SDK.

I think we're on the same page. I was just concerned that we were doing something by default that wasn't cross-SDK compatible, but I see you mean not by default.

ruby/type-hinting-converters-codecs-errors.md
antstorm: I think it's a bit clearer when you are restricted to 1:1 conversion and other combinations can be achieved by putting values in an array (or whatever collection type)

ruby/type-hinting-converters-codecs-errors.md
antstorm: Added the clarification

ruby/type-hinting-converters-codecs-errors.md
antstorm: It should be the client-level in my opinion. So any lower-level calls should return the original protobuf response

ruby/type-hinting-converters-codecs-errors.md
antstorm: Good point, let me simplify this

ruby/type-hinting-converters-codecs-errors.md
antstorm: Agree, I'll clear the specific errors out and make a note that this is not an exhaustive list

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
antstorm: Rails has one (via the `as_json` method that is expected to return an intermediary object that is then easily converted into JSON, like a Hash, Array or String). However we can't depend on Rails, since not everyone is using it.

Ruby allows objects to implement a `to_json` method, but it is expected to generate a JSON String directly (e.g. returning a Hash instead will throw an error).

In both cases there isn't an established way of doing the reverse conversion.

Also, I don't see how JSON is relevant here. This should work with any format, be it YAML, XML or a custom binary.

ruby/type-hinting-converters-codecs-errors.md
antstorm: Yes, exactly, `dump`/`load` is for converting from the object to a converter input and vice versa. It doesn't dictate the format of the payload. A payload converter does.

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
cretz: :+1: On the rest of the points.

> Also, I don't see how JSON is relevant here. This should work with any format, be it YAML, XML or a custom binary.

We don't hide the fact that by default we use the language's JSON features, not some "generic" serialization features. In languages like Go and Java, you can customize the object's JSON formatting/parsing via JSON-specific tags/annotations. It is important to those users to know we use stdlib JSON and they can customize in that way.

ruby/type-hinting-converters-codecs-errors.md
antstorm: > Is there a general "object-from-hash" in Ruby? Sounds like not.

Yep, there isn't.

> "How do I say something is a list of my class w/ the custom load method?"

Potentially we can support the `[MyClass]` notation for the `input`/`output` params. Do you think we should add it straight away?

> So if there are one or two really popular libraries for this in Ruby, we could consider supporting/encouraging as optional dependencies.

Agree, but I don't think there is a consensus on what these options are. I suggest we delay this until we start gathering feedback from the actual users rather than speculate.

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
antstorm: Right, I see, that makes sense 👍 Ruby doesn't have it, so type hints should return something that one of the Payload Converters then can pick up (including their custom converters that they might add).

ruby/type-hinting-converters-codecs-errors.md
cretz: > Do you think we should add it straight away?

The concern from the Python side was that there are more than arrays, e.g. lists, sets, dictionaries, etc. If this would take care of most people's concerns, sure. Though really we should strongly encourage one single object for every input and output and to make everything else attributes/fields/members/whatever-you-call-them on that object.

antstorm: In that case maybe let's not add it just yet and steer people away from it by not making it easier (they can already implement whatever wrapper they need with the provided interface)

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
cretz: Can you confirm what happens in the JSON payload converter when it sees a class that doesn't have a `dump` (or `load` on decode)?

cretz (approved):
LGTM.

Just one nit confirming what happens in a payload converter for a class that doesn't have `dump`/`load`. Do we give a "JSON serialization not supported for object" or is there some fallback we can support? Maybe `to_json` + `as_json` like the Rails people? Again, if there's a popular Ruby model library we can support by simple reflection we should. Or if not, we need to make sure the JSON payload converter is really easy to convert/extend to support their favorite modelling library in a centralized way so they don't have to have `dump`/`load` everywhere. (a oj-based converter for make a great first sample in `samples-ruby`)

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
antstorm: Sure, it then attempts to serialize it using the JSON library. This works well for built-in data types, but obviously fails for custom classes.

One thing to correct — it won't be the JSON payload converter attempting to call `#dump` on a class, this needs to happen before its result is passed to a high-level converter (`Composite` by default). This is done because the result of `dump` might be better converted by using another converter (like `Bytes` or your custom converter)

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
antstorm: Had a discussion in a separate thread and agreed on what's stated above ^. The proposal has been updated to include this part

ruby/type-hinting-converters-codecs-errors.md
antstorm: Updated

antstorm:
@cretz thank you!

I agree that we can figure out a fallback mechanism for the most common use-cases. Supporting Rails' `to_json`/`as_json` makes sense, however we'll need to amend it since it wasn't meant as a 2-way conversion — it is mostly used to represent objects in JSON responses to API calls, not deserialize it afterwards. But point taken and I'll dig a bit more into what might be a good popular alternative to tap into.

ruby/type-hinting-converters-codecs-errors.md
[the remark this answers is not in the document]
cretz: > obviously fails for custom classes.

Not true though, because of https://ruby-doc.org/stdlib-3.0.0/libdoc/json/rdoc/JSON.html#module-JSON-label-Custom+JSON+Additions right?

> One thing to correct — it won't be the JSON payload converter attempting to call #dump on a class, this needs to happen before its result is passed to a high-level converter (Composite by default).

I disagree with this. I think it should only be called at JSON conversion time if present. I don't think the other two converters (bytes and null) should be enough to default to calling dump/load regardless of the child encoding converters. It is up to each encoding converter to handle objects its own way, composite should do nothing special about that.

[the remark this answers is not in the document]
cretz: To clarify here, maybe we don't want generic `dump` and `load`, but instead `dump_json` and `load_json` since we shouldn't have generic customization per object, only specific converter customization.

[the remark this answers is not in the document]
antstorm: @cretz ah, interesting, I wasn't aware of that interface in the JSON class. I haven't ever seen this being used in real world applications, though. Therefore I'm not 100% comfortable depending on this interface.

I also don't want this to only work with JSON. What if I want to use a protobuf here instead without forcing my inputs/outputs to be protobuf generated classes? In this case I would imagine my type hint to return a protobuf object when called `dump` and a protobuf converter would pick it up and encode into a string.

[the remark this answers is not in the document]
antstorm: Resolved separately

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+
+  # errors specific to Worker
+  class WorkerError < InternalError
+
+  # superclass for connection, network and response errors
+  class RPCError < Error
+  class UnexpectedError < RPCError
+
+  # superclass for all API failure responses
+  class APIError < ServerError
```
cretz, line 196: I think this should be called "Failure" or similar since it is basically 1:1 with "Temporal Failures" and it's not always "API" related (at least not obviously to the user which doesn't use our API). Also, it should carry the raw proto with it probably.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+end
+```
+
+At this point we are not going to be converting user errors back to their original classes. These
+will be decoded into `ApplicationError`s while keeping the original class name, message, stack trace
+and other details. Here's an example of handling an activity error:
+
+```ruby
+class ProcessPaymentActivity < Temporal::Activity
+  class InsufficientFunds < Temporal::NonRetryableError; end
```
cretz, line 237: I didn't see `Temporal::NonRetryableError` in the hierarchy before. Can you clarify its superclass?

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+
+class RenewSubscriptionWorkflow < Temporal::Workflow
+  def execute
+    ...
+
+    begin
+      workflow.execute_activity(ProcessPaymentActivity, account, amount)
+    rescue Temporal::ActivityError => error
+      if error.cause.is_a?(Temporal::ApplicationError)
+        case error.cause.original_class
```
cretz, line 269: "original_class" isn't something available on our failure proto. Where is this field populated from? If it's "type" we should probably keep that name because other languages use it too and "class" is a Ruby concept.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+        when 'ProcessPaymentActivity::InsufficientFunds'
```
cretz, line 270: How do you know how many levels deep to qualify this? If this was `class MyError` in `MyCompany::MyTemporalWorkflows` module, what would this value be? In other SDKs, we have left this unqualified when storing on the `type` in `ApplicationError`.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+
+    ...
+  end
+end
+```
+
+Application errors will be encoded/decoded using the same Payload converters as the other payloads.
+This also means that if an error class implements `#to_json` / `.json_create` (as suggested by the
+[Ruby's JSON module](https://ruby-doc.org/stdlib-3.1.2/libdoc/json/rdoc/JSON.html#module-JSON-label-Custom+JSON+Additions))
+then we will be able to recreate the original error class:
```
cretz, line 290: Where are you deserializing this from? The first error detail payload and ignore the rest?

It can be confusing for a user to know that half the time you deserialize into a known type and half the time you don't based on whether the class supports it. To clarify, you are going to always try to instantiate a class of the app error's "type" and then always try to deserialize JSON and on any error you'll leave it as app error? As a user, am I expected to handle both cases just in case the potentially-separate implementer of the error failed JSON serialization?

We don't support this in any SDK today, can we not support this in Ruby either until we have a holistic solution here? All SDKs suffer this issue, I don't think we should solve now.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+        # handle payment network error
+        # e.status is available as expected
+      end
+
+      ...
+    end
+  end
+```
+
+To allow customisation of this behaviour we will expose an interface for failure conversion:
```
cretz, line 340: I think this should be the only place we allow customization

ruby/type-hinting-converters-codecs-errors.md
antstorm: It is the `type`, however since this is `Temporal::ApplicationError` (and not a protobuf Failure) I think we can rename it to provide a bit more context. `Type` doesn't make much sense within Ruby.

I can keep the `type` available as well in case you feel strongly about it

ruby/type-hinting-converters-codecs-errors.md
antstorm: This will be the result of a `.class.name` call on the error itself. In Ruby it's quite clear, because you can't reference a class without all of the modules it's nested in:

```ruby
module MyCompany
  module MyTemporalWorkflows
    class MyError < StandardError; end

    puts MyError.name # prints "MyCompany::MyTemporalWorkflows::MyError"
  end
end
```

ruby/type-hinting-converters-codecs-errors.md
antstorm: It is the only place. Unless you're referring to JSON additions, but you can't possibly expect users to manually convert every possible ApplicationFailure `type` in their custom failure converter

ruby/type-hinting-converters-codecs-errors.md
cretz: It does if you are talking about consistency across APIs. We don't embed the word API hardly anywhere. They don't know that API is even involved.

ruby/type-hinting-converters-codecs-errors.md
cretz: I support other names if `type` is unclear. But `original_class` is clearly a Ruby-ism and incorrect when an activity failed somewhere else that just set the type string.

ruby/type-hinting-converters-codecs-errors.md
cretz: Java also does this which is qualified, but others may not. This has never been a problem before because we have never used "type" to reference a class. This would be the first time "type" is ever used that way that I know of which is why I think we need a more general discussion. "type" is just a general user string, not meant for machine reading or having any expectations around it. Often "type" is just used in places like retry policy's non-retryable-types.

cretz: > This will be the result of a .class.name call on the error itself.

:+1:

> you can't reference a class

We have never used "type" to reference a class and if we're gonna start doing that, we need a more general discussion. "type" is a user string that has no expected format and is not meant for machine reading. It just happens to default to the class name, but it's really meant to be an "error type", often used for matching with user-defined strings in retry policy's non-retryable types.

ruby/type-hinting-converters-codecs-errors.md
cretz: > Unless you're referring to JSON additions, but you can't possibly expect users to manually convert every possible ApplicationFailure type in their custom failure converter

I am referring to that. We should not auto-deserialize to user-defined errors in Ruby until we have a general discussion about how that works in SDKs. Reusing the "type" to mean "class" and "error details" to mean "class info" would be a new thing to Temporal.

ruby/type-hinting-converters-codecs-errors.md
antstorm: Well, this is not true, since all the protobufs clearly define a `Temporal::Api`. So both `Failure` and `Api` have the same source and are never shown to the user.

But I'm not too fussy about it, happy to call it `Failure` if you think it's clearer (I don't)

cretz: One of the primary reasons for these errors is to hide/convert the protobufs. I don't mind something besides "Failure", I just think "API" is not best. I choose "Failure" because it aligns with other SDKs and docs and there is no value in uniqueness here.

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+- When sending a custom input type to another SDK, workflow/activity definitions (without the
+  implementation) can still be leveraged
+- A custom output type sent to another SDK can still be `dump`ed into a matching JSON or other
+  serializable structure
+- The same approaches apply to inputs and outputs received from other SDKs
+
+
+## Payload handling
+
+Ruby SDK will use the model proposed [here](https://docs.temporal.io/docs/concepts/what-is-a-data-converter/)
```
bergundy, line 100:
```suggestion
Ruby SDK will use the model proposed [here](https://docs.temporal.io/concepts/what-is-a-data-converter/)
```

ruby/type-hinting-converters-codecs-errors.md
bergundy: I find it weird that ApplicationError subclasses ServerError.
I would change the hierarchy and possibly rename APIError to something like FailureError.
The alternative (which @cretz will disagree with) is to call the failures base class TemporalFailure and suffix its subclasses with Failure like Java and TS do.
I think it's nice to cleanly mark and group all of these failure types this way.

ruby/type-hinting-converters-codecs-errors.md
bergundy: I would keep the other SDK's convention of `ApplicationError.nonRetryable`, it maps directly to the protocol and is easier to spot when inspecting a workflow's history in the UI.

ruby/type-hinting-converters-codecs-errors.md
bergundy: I'd stay close to the protos and the other SDKs here, the protos define the field as "type".

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+    end
+  end
+```
+
+To allow customisation of this behaviour we will expose an interface for failure conversion:
+
+```ruby
+interface _FailureConverter
+  # Takes any Ruby error and returns a Failure object
+  def to_failure: (Temporal::Error) -> Temporal::Failure
```
bergundy, line 345: Arguably this should take an ruby `Error`

ruby/type-hinting-converters-codecs-errors.md
```diff
@@ -0,0 +1,412 @@
...
+
+  # Takes a Failure object and returns a Ruby error
+  def from_failure: (Temporal::Failure) -> Temporal::Error
```
bergundy, line 348: The return type here should be APIError (or whatever the base class of the failure classes are)

ruby/type-hinting-converters-codecs-errors.md
bergundy: I'd punt on this but I agree with the general direction.

bergundy (commented):
I saw in the image that `MyType` gets serialized as binary instead of JSON string, is this intentional?

I don't see the "default" failure converter mentioned anywhere in this doc, might as well add that here.

ruby/type-hinting-converters-codecs-errors.md
cretz: > The alternative (which @cretz will disagree with) is to call the failures base class TemporalFailure and suffix its subclasses with Failure like Java and TS do.

I don't disagree with this here. I only disagree in Python where errors are suffixed with `Error`. I'm all for a common suffix on these and If unless it's expected to always have the `Error` suffix, I think `Failure` works.

ruby/type-hinting-converters-codecs-errors.md
antstorm: Sounds good, I can see how this is a bigger discussion across all the SDKs. I'll leave this part off for now 👍

ruby/type-hinting-converters-codecs-errors.md
antstorm: @bergundy are you suggesting nesting it further (e.g. `Temporal::ApplicationError::NonRetryable`)? Or something else?

ruby/type-hinting-converters-codecs-errors.md
antstorm: I've renamed the error to `FailureError` and made it a direct subclass of `Temporal::Error` 👍

ruby/type-hinting-converters-codecs-errors.md
antstorm: What would be the expected way for the user to set the `type` field if it's a user string? Are they expected to always `raise ApplicationError.new(type: 'my-error-type')`?

ruby/type-hinting-converters-codecs-errors.md
antstorm: Agreed to remove this from this proposal until there's a cross-SDK solution

ruby/type-hinting-converters-codecs-errors.md
antstorm: Yes, good point, I'll update it 👍

ruby/type-hinting-converters-codecs-errors.md
bergundy: No, I'm suggesting that ApplicationError would have a retryable flag.

ruby/type-hinting-converters-codecs-errors.md
antstorm: Actually, I'm not sure that's the case. This should be the same as the input to `#to_failure` since we're allowing the users to provide their own implementation, right?

ruby/type-hinting-converters-codecs-errors.md
antstorm: So in order to raise a non-retryable error you'd need to either:

```ruby
class MyError < Temporal::ApplicationError; end
...
raise MyError.new('error message', non_retryable: true)
```

or:

```ruby
class MyError < Temporal::ApplicationError
  def initialize(message)
    super(message, non_retryable: true)
  end
end
...
raise MyError.new('error message')
```

am I interpreting this correctly? As in — there's no simple way of defining a non-retryable error type without additional code?

I still think a superclass approach is better. Maybe we can make it more explicit `Temporal::NonRetryableApplicationError`?

antstorm:
@bergundy 

> I saw in the image that MyType gets serialized as binary instead of JSON string, is this intentional?

That is intentional. It shows that "Type Hinting" is completely decoupled from "Payload Converters". In case a type hint returns a binary string, there's no need to treat it differently from a value (without a type hint) that itself is a binary string, so therefore it should be serialized using the "Bytes" converter, not the "JSON".

> I don't see the "default" failure converter mentioned anywhere in this doc, might as well add that here.

Added (right after the `_FailureConverter` interface definition) 👍

ruby/type-hinting-converters-codecs-errors.md
bergundy: This is fine IMHO:

```ruby
class MyError < Temporal::ApplicationError
  def initialize(message)
    super(message, non_retryable: true)
  end
end
```

We also provide factory methods to instantiate retryable and non-retryable application failures in other SDKs.

ruby/type-hinting-converters-codecs-errors.md
cretz: Yes. But our default failure converter defaults it to the error class name for them if they don't create an application error but we are still wrapping it as one (e.g. for activity failures).

bergundy (approved):

PR #69, opened 2022-10-11
Mass replay verification public API

Sushisource:
Rendered: https://github.com/Sushisource/proposals/blob/replay-verify-sdk-api/typescript/mass-replay.md

### DIFF
--- a/typescript/mass-replay.md
+++ b/typescript/mass-replay.md
@@ -1,0 +1,89 @@
+# Mass Replay Verification API
+
+Users often manually verify that they have not broken compat with existing
+workflow histories when making changes by downloading histories and replaying
+them with a replay worker.
+
+We aim to make that simpler and faster with the below changes, which should
+be similarly applicable to all SDKs but are laid out here in TS first.
+
+For flexibility and compose-ability reasons, it is desirable to separate the
+construction of a set of histories the user might care about from the 
+downloading of those histories and the replay of those histories. Hence
+the base level APIs proposed are (in order):
+
+## Fetching
+See https://github.com/temporalio/proposals/pull/67
+
+Which includes functions for listing workflows according to some query params.
+Those functions return something like `Iterable<WorkflowExecutionInformation>`
+
+## Downloading
+This is simply a matter of (lazily) mapping the iterable:
+`Iterable<WorkflowExecutionInformation> -> AsyncIterable<WorkflowIDAndHistory>`
+
+Preserving the workflow id is important, since history does not include it
+by itself.
+
+By itself, this would look something like:
+`const historyIterator = client.workflow.downloadLazily(workflowIds)`
+
+Which would be sugar for creating a returning an async iterator which calls
+`getExecutionHistory` on each id. This might include doing so concurrently
+(not required for initial implementation).
+
+
+## Replaying
+Pass the lazy iterable of histories into a function which instantiates the
+worker and feeds the histories to it:
+
+```typescript
+export interface ReplayResults {
+  // True if any workflow failed replay
+  readonly hadAnyFailure: boolean;
+  // Maps run id to information about the replay failure
+  readonly failureDetails: Map<string, WorkflowError>;
+}
+
+// Critically, ReplayWorkerOptions will get a `failFast` option, which will
+// default to true. If set, we throw on the first nondeterminism error. If not
+// set, failure information is in the returned results object.
+
+export class Worker {
+  // etc etc...
+  public static async runReplayHistories(
+    options: ReplayWorkerOptions,
+    histories: AsyncIterable<History>
+  ): Promise<ReplayResults> {}
+  // etc etc...
+}
+```
+
+Under the hood this creates a Rust stream, which is fed the histories
+downloaded from the lang side as they become available, and is passed into 
+`init_replay_worker`.
+
+## All together now
+A user doing this using these base APIs would write something like:
+```typescript
+const iterator = client.workflow.list({
+  // Made up nonsense syntax. See other proposal.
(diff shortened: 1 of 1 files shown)


typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+For flexibility and compose-ability reasons, it is desirable to separate the
+construction of a set of histories the user might care about from the 
+downloading of those histories and the replay of those histories. Hence
+the base level APIs proposed are (in order):
+
+## Fetching
+See https://github.com/temporalio/proposals/pull/67
+
+Which includes functions for listing workflows according to some query params.
+Those functions return something like `Iterable<WorkflowExecutionInformation>`
```
bergundy, line 19: This should be an async iterable as well to support pagination.

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+`Iterable<WorkflowExecutionInformation> -> AsyncIterable<WorkflowIDAndHistory>`
+
+Preserving the workflow id is important, since history does not include it
+by itself.
+
+By itself, this would look something like:
+`const historyIterator = client.workflow.downloadLazily(workflowIds)`
+
+Which would be sugar for creating a returning an async iterator which calls
+`getExecutionHistory` on each id. This might include doing so concurrently
```
bergundy, line 32: Pretty easy to limit the concurrency with rxjs, it's built into `flatMap`.

typescript/mass-replay.md
Sushisource: Heh, yeah I originally had it that way, but the proposal doesn't have an `await` on list.

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+
+## All together now
+A user doing this using these base APIs would write something like:
+```typescript
+const iterator = client.workflow.list({
+  // Made up nonsense syntax. See other proposal.
+  query: 'task_queue=foo AND status=closed ORDER BY end_time DESC',
+  pageSize: 100,
+});
+const histories = client.workflow.downloadLazily(iterator);
```
bergundy, line 74: Let's add an options object to pass in `concurrency`.

typescript/mass-replay.md
Sushisource: It's not just a matter of that, it's also a matter of buffering that and sending it down the pipe into the rust side appropriately. Shouldn't be all that hard though.

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+const result = worker.runReplayHistories({ whatever: 'options' }, histories).await;
```
bergundy, line 75: You write too much rust :)

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+## But wait there's more
+Arguably, the above is still a bit too manual. We could, in addition to the
+base APIs, provide a more "all-in-one" approach:
+
+```typescript
+const result = worker.validateCompatabilityWithQueue(
+    client, 'task_queue_foo', { sample_strategy: mostRecent(500) }
+).await;
```
bergundy, line 85:
```suggestion
const result = await worker.validateCompatabilityWithQueue(
    client, 'task_queue_foo', { sampleStrategy: mostRecent(500) }
);
```

bergundy: I'd consider making this a static method like the runReplayHistories method.
Or make them both instance methods but I prefer static to save the extra call.

bergundy (approved):
Nice proposal.

All of the functionality is here.
I mostly had nits and minor suggestions.

typescript/mass-replay.md
bergundy: https://github.com/temporalio/proposals/pull/67/files#diff-7d87e4d9af7e84955486e56bc4d577483f5fa34479dbe84b2ceeb50f4d64a81cR66

bergundy: It's a little hard to see with all of the comments.

Sushisource: ooooo

typescript/mass-replay.md
Sushisource: Oh I meant it to be, I just forgot the capital letter here (and in the other spot too actually)

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+const histories = client.workflow.downloadLazily(iterator);
+const result = await Worker.runReplayHistories({ whatever: 'options' }, histories);
+```
+
+## But wait there's more
+Arguably, the above is still a bit too manual. We could, in addition to the
+base APIs, provide a more "all-in-one" approach:
+
+```typescript
+const result = await Worker.validateCompatabilityWithQueue(
```
cretz, line 83: Still needs to have the workflow code that applies to the task queue, but at this point, I think this is such a tiny bit of sugar that it can be a "second phase".

cretz (approved):
LGTM

typescript/mass-replay.md
```diff
@@ -0,0 +1,89 @@
...
+  readonly failureDetails: Map<string, WorkflowError>;
+}
+
+// Critically, ReplayWorkerOptions will get a `failFast` option, which will
+// default to true. If set, we throw on the first nondeterminism error. If not
+// set, failure information is in the returned results object.
+
+export class Worker {
+  // etc etc...
+  public static async runReplayHistories(
```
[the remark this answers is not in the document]
Sushisource: Yeah, that's totally fine

PR #70, opened 2022-12-07
.NET SDK - Phase 1

cretz:
# .NET SDK - Phase 1 - Proposal Review

This is the proposal for the first phase of the .NET SDK.

* See it [rendered](https://github.com/cretz/temporal-proposals/blob/dotnet-phase-1/dotnet/sdk-phase-1.md) (or can use the "rich diff" link in upper right; rendered link will disappear on PR merge)
* The playground/POC repo I was toying with concepts and a Rust bridge is [here](https://github.com/cretz/temporal-sdk-dotnet-poc). It was just for testing concepts. It is not accurate and may be deleted later.

---

dotnet

### DIFF
--- a/dotnet/sdk-phase-1.md
+++ b/dotnet/sdk-phase-1.md
@@ -19,42 +19,28 @@ Notes for this proposal:
 The .NET SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-core), utilizing "P/Invoke" and a
 custom-built bridge.
 
-The high-level goals for the .Net SDK are:
+The high-level goals for the .NET SDK are:
 
 * Be as simple as possible from a caller POV
 * Be as lightweight as reasonable with regards to dependencies
 * Be reasonably flexible with reasonable defaults
-* Fit well into the .Net ecosystem
+* Fit well into the .NET ecosystem
 * Minimum versions supported: .NET Core 3.1, .NET Standard 2.0, and .NET Framework 4.6.2
 * All common 64-bit platforms supported, 32-bit maybe later
 
 ## Repository Management
 
 The existing repository at https://github.com/temporalio/sdk-dotnet has hundreds of commits, dozens of issues, open PRs,
-GH pages, etc that don't align with this proposal necessarily. After some thought, it has been decided to:
-
-* Create a new https://github.com/temporalio/sdk-dotnet-incubating repo for the very first commits
-* After that repo has some code and API doc gen capability:
-  * Rename https://github.com/temporalio/sdk-dotnet to https://github.com/temporalio/legacy-dotnet (note the
-    intentionally absent "sdk" term for SEO reasons)
-    * Add README note on https://github.com/temporalio/legacy-dotnet pointing to
-      https://github.com/temporalio/sdk-dotnet
-    * Disable GH pages
-    * Archive it
-    * Consider deletion at some point in the future
-  * Rename https://github.com/temporalio/sdk-dotnet-incubating to https://github.com/temporalio/sdk-dotnet
-  * Update DNS for https://dotnet.temporal.io/ to point to vercel-based API docs page
-
-Doing the repo rename was deemed easier than going through manually closing issues, closing PRs, creating empty `main`
-branch, and keeping `master` branch around for those reading older code.
+GH pages, etc that don't align with this proposal necessarily. After some thought, it has been decided to move that
+repository and create https://github.com/temporalio/sdk-dotnet again fresh.
 
 ## Namespace/Package
 
 * Will publish as `Temporalio`
   * Need to reserve prefix `Temporalio` on NuGet, see
     https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation
-* Namespace will be `Temporal`
-  * See discussion topics later about this being `Temporal.Sdk` instead
+* Namespace will be `Temporalio`
+  * See discussion topics later about this being `Temporalio.Sdk` instead
 
 ## Workflow Definition
 
@@ -64,7 +50,7 @@ Here's a definition for an other-language-defined workflow (from Python README):
 namespace MyNamespace;
 
 using System.Text.Json.Serialization;
-using Temporal.Workflow;
+using Temporalio.Workflow;
 
 public record GreetingInfo
 {
@@ -97,7 +83,7 @@ public interface IGreetingWorkflow
 Notes:
 
 * We use regular .NET JSON serialization (see data conversion later)
-* Workflow attributes, and in the future workflow runtime features, are in the `Temporal.Workflow` namespace
+* Workflow attributes, and in the future workflow runtime features, are in the `Temporalio.Workflow` namespace
 * We only support workflow interfaces right now because we don't have a workflow implementation yet
   * For now, everything non-static must have some form of `[Workflow` attribute, and not have a default implementation
 * Every workflow interface must have the non-inheritable `Workflow` attribute
@@ -120,7 +106,7 @@ Notes:
 * Users are encouraged to set a `static` `readonly` `Ref` property for use by clients
   * `Refs<T>.Instance` is how it is obtained
     * Also will do validation which will be shared in the future by the workflow worker
-    * `Refs` is in `Temporal` not `Temporal.Workflow`
+    * `Refs` is in `Temporalio` not `Temporalio.Workflow`
   * This is backed by a dynamic proxy
   * Really old versions of .NET (i.e. would be rarely used by users, < C# 8), don't allow a `static readonly` property on the
     interface and therefore they have to make their `Ref` elsewhere
@@ -132,12 +118,12 @@ Common static definition:
 ```csharp
 namespace MyNamespace;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public static class GreetingActivities
 {
     [Activity(Name = "create_greeting_activity")]
-    public static async Task<string> CreateGreeting(GreetingInfo info)
+    public static async Task<string> CreateGreetingAsync(GreetingInfo info)
     {
         return $"{info.Salutation}, {info.Name}!";
     }
@@ -149,7 +135,7 @@ Common instance definition:
 ```csharp
 namespace MyNamespace;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public class GreetingActivities
 {
@@ -190,7 +176,7 @@ Notes:
 The activity context looks like this:
 
 ```csharp
-namespace Temporal.Activity;
+namespace Temporalio.Activity;
 
 public class ActivityContext
 {
@@ -229,16 +215,16 @@ Notes:
 
 ### Async Completion
 
-Simply `throw new Temporal.Activity.CompleteAsyncException` from the activity.
+Simply `throw new Temporalio.Activity.CompleteAsyncException` from the activity.
 
 ## Client
 
-Inside the `Temporal.Client` namespace, we have the following types:
+Inside the `Temporalio.Client` namespace, we have the following types:
 
 * `TemporalClient` - namespace-specific, stateless high-level client to Temporal
   * See later for most of its API design
   * Accepts `ITemporalConnection`
-  * Implements `ITemporalClient` for DI use (which itself implements `Temporal.Worker.IWorkerClient` for worker use)
+  * Implements `ITemporalClient` for DI use (which itself implements `Temporalio.Worker.IWorkerClient` for worker use)
 * `TemporalConnection` - namespace-agnostic, stateful low-level client to Temporal
   * Has `WorkflowService`, `OperatorService`, etc for making raw gRPC calls
   * Supports lazy connectivity
@@ -249,7 +235,7 @@ Inside the `Temporal.Client` namespace, we have the following types:
 Example of starting the workflow defined earlier in document, signalling it, then waiting on its results:
 
 ```csharp
-using Temporal.Client;
+using Temporalio.Client;
 using MyNamespace;
 
 // Connect
@@ -269,7 +255,7 @@ Console.WriteLine(await handle.GetResultAsync());
 Example of doing the same via already started workflow:
 
 ```csharp
-using Temporal.Client;
+using Temporalio.Client;
 using MyNamespace;
 
 // Connect
@@ -293,7 +279,7 @@ user would rather not provide the type until fetching the result. See the defini
 Looks like:
 
 ```csharp
-namespace Temporal.Client;
+namespace Temporalio.Client;
 
 public class TemporalClient : ITemporalClient
 {
@@ -318,7 +304,13 @@ public class TemporalClient : ITemporalClient
     public async Task<WorkflowHandle<TResult>> StartWorkflowAsync<T, TResult>(
         Func<T, Task<TResult>> workflow, T arg, StartWorkflowOptions options) { /*...*/ }
 
-    public async Task<WorkflowHandle<Nothing>> StartWorkflowAsync(
+    public async Task<WorkflowHandle> StartWorkflowAsync(
+        Func<Task> workflow, StartWorkflowOptions options) { /*...*/ }
+
+    public async Task<WorkflowHandle> StartWorkflowAsync<T>(
+        Func<T, Task> workflow, T arg, StartWorkflowOptions options) { /*...*/ }
+
+    public async Task<WorkflowHandle> StartWorkflowAsync(
         string workflow, object[] args, StartWorkflowOptions options) { /*...*/ }
 
     public async Task<WorkflowHandle<TResult>> StartWorkflowAsync<TResult>(
@@ -326,7 +318,7 @@ public class TemporalClient : ITemporalClient
 
     // GetWorkflowHandle
 
-    public WorkflowHandle<Nothing> GetWorkflowHandle(
+    public WorkflowHandle GetWorkflowHandle(
         string workflowID, string? runID = null, string? firstExecutionRunID = null) { /*...*/ }
 
     public WorkflowHandle<TResult> GetWorkflowHandle<TResult>(
@@ -342,7 +334,7 @@ public class TemporalClient : ITemporalClient
 
     // ListWorkflowsAsync
 
-    public WorkflowExecutionEnumerator ListWorkflows(
+    public IAsyncEnumerator<WorkflowExecution> ListWorkflows(
         string query,
         ListWorkflowsOptions? options = null,
     ) { /*...*/ }
@@ -353,10 +345,12 @@ public class TemporalClient : ITemporalClient
 
 Notes:
 
+* There is no lazy connection/client like there is in other languages
+  * Originally the code had a `Lazy` connection option, but the team decided they wanted it removed
 * There will be `ExecuteWorkflowAsync` helper methods for each `StartWorkflowAsync` overload there
 * It was decided forcing an options object is easier than lots of parameters for start workflow
-* Notice that there is a string-based workflow `StartWorkflowAsync` call that accepts a return type and one that doesn't
-  and returns a handle for `Nothing`. The user can specify a return type on `WorkflowHandle.GetResultAsync` if they
+* Notice that there is a string-based workflow `StartWorkflowAsync` call that accepts a return type and one that
+  doesn't. The user can specify a return type on `WorkflowHandle.GetResultAsync` if they
   want instead of at start time (or just never altogether if the return doesn't matter to the user).
 * For string-based calls to `StartWorkflowAsync`, we are accepting all args at once via `object[]`. But since array
   types are covariant in .NET, this means you could have a workflow that accepts `string[]` and forget that needs to
@@ -374,7 +368,7 @@ Notes:
   * We could have taken a proxy approach, but that would have made the code look like
     `client.GetWorkflowHandle<GreetingWorkflow>("my wf id", wf => wf.RunAsync)` which is not much better and is hard to
     get right with all of the overloads so you can have the lambda as the last param. This is what Azure Durable
-    Entities does.
+    Entities does. There is discussion on this later.
 * Cancellation tokens (and RPC metadata and RPC timeout and such) will be in the call options, not as the last parameter
   in a bunch of overloads like many other .NET methods
 * Older versions of .NET don't have `IAsyncEnumerator` and we don't want to go out of our way to provide a special
@@ -382,27 +376,27 @@ Notes:
   * Older versions can just use raw gRPC or not have this feature. These older .NET versions are really old and we don't
     expect hardly any users on them.
   * Note, `ListWorkflows` is not `ListWorkflowsAsync` because the first page is lazily fetched on first iteration
-  * `WorkflowExecutionEnumerator` has a `MapHistories` on it that returns `IAsyncEnumerator<WorkflowHistory>` (or is an
-    extension method)
+  * We need an easy way to map this to histories. Maybe a `GetHandle` extension method on the execution (that accepts a
+    client), and a `SelectHistories` extension method on `IAsyncEnumerator<WorkflowExecution>` that accepts a client.
 
 ### WorkflowHandle
 
 Looks like:
 
 ```csharp
-namespace Temporal.Client;
+namespace Temporalio.Client;
 
-public class WorkflowHandle<TResult>
+public class WorkflowHandle
 {
     public string ID { get; }
     public string? RunID { get; }
     public string? ResultRunID { get; }
     public string? FirstExecutionRunID { get; }
 
-    public async Task<TResult> GetResultAsync(
+    public async Task GetResultAsync(
         bool followRuns = true, RPCOptions? options = null) { /*...*/ }
 
-    public async Task<TResultLocal> GetResultAsync<TResultLocal>(
+    public async Task<TResult> GetResultAsync<TResult>(
         bool followRuns = true, RPCOptions? options = null) { /*...*/ }
 
     public async Task SignalAsync(string signal, object[] args, RPCOptions? options = null) { /*...*/ }
@@ -442,6 +436,12 @@ public class WorkflowHandle<TResult>
 
 #endif
 }
+
+public class WorkflowHandle<TResult> : WorkflowHandle
+{
+    public new async Task<TResult> GetResultAsync(
+        bool followRuns = true, RPCOptions? options = null) { /*...*/ }
+}
 ```
 
 Notes:
@@ -462,8 +462,8 @@ Notes:
 Usage:
 
 ```csharp
-using Temporal.Client;
-using Temporal.Worker;
+using Temporalio.Client;
+using Temporalio.Worker;
 using MyNamespace;
 
 // Connect
@@ -486,9 +486,9 @@ await worker.ExecuteAsync(cts.Token);
 Looks like:
 
 ```csharp
-namespace Temporal.Worker;
+namespace Temporalio.Worker;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public class TemporalWorkerOptions
 {
@@ -551,7 +551,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Worker;
+namespace Temporalio.Worker;
 
 class WorkflowReplayer
 {
@@ -580,7 +580,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Converters;
+namespace Temporalio.Converters;
 
 public record DataConverter(
     IPayloadConverter PayloadConverter,
@@ -597,18 +597,18 @@ public record DataConverter(
 
 public interface IPayloadConverter
 {
-    Temporal.Api.Common.V1.Payload ToPayload(object value);
+    Temporalio.Api.Common.V1.Payload ToPayload(object value);
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload);
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload);
 }
 
 public interface IEncodingPayloadConverter
 {
     string Encoding { get; }
 
-    bool TryToPayload(object value, out Temporal.Api.Common.V1.Payload payload);
+    bool TryToPayload(object value, out Temporalio.Api.Common.V1.Payload payload);
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload);
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload);
 }
 
 public record PayloadConverter(IEnumerable<IEncodingConverter> EncodingConverters) : IPayloadConverter
@@ -620,9 +620,9 @@ public record PayloadConverter(IEnumerable<IEncodingConverter> EncodingConverter
         new BinaryProtoConverter(),
         new JsonPlainConverter() }) { /*...*/ }
 
-    public Temporal.Api.Common.V1.Payload ToPayload(object value) { /*...*/ }
+    public Temporalio.Api.Common.V1.Payload ToPayload(object value) { /*...*/ }
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload) { /*...*/ }
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload) { /*...*/ }
 }
 
 public class BinaryNullConverter : IEncodingPayloadConverter { /*...*/ }
@@ -633,41 +633,41 @@ public class JsonPlainConverter : IEncodingPayloadConverter { /*...*/ }
 
 public interface IFailureConverter
 {
-    Temporal.Api.Failure.V1.Failure ToFailure(
+    Temporalio.Api.Failure.V1.Failure ToFailure(
         Exception exception, IPayloadConverter payloadConverter);
 
     Exception ToException(
-        Temporal.Api.Failure.V1.Failure failure, IPayloadConverter payloadConverter);
+        Temporalio.Api.Failure.V1.Failure failure, IPayloadConverter payloadConverter);
 }
 
 public record FailureConverter(bool EncodeCommonAttributes = false) : IFailureConverter
 {
-    public Exception ToException(Temporal.Api.Failure.V1.Failure failure, IPayloadConverter payloadConverter) { /*...*/ }
+    public Exception ToException(Temporalio.Api.Failure.V1.Failure failure, IPayloadConverter payloadConverter) { /*...*/ }
 
-    public Temporal.Api.Failure.V1.Failure ToFailure(Exception exception, IPayloadConverter payloadConverter) { /*...*/ }
+    public Temporalio.Api.Failure.V1.Failure ToFailure(Exception exception, IPayloadConverter payloadConverter) { /*...*/ }
 }
 
 public interface IPayloadCodec
 {
-    Task<IEnumerable<Temporal.Api.Common.V1.Payload>> Encode(
-        IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads);
+    Task<IEnumerable<Temporalio.Api.Common.V1.Payload>> Encode(
+        IReadOnlyCollection<Temporalio.Api.Common.V1.Payload> payloads);
 
-    Task<IEnumerable<Temporal.Api.Common.V1.Payload>> Decode(
-        IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads);
+    Task<IEnumerable<Temporalio.Api.Common.V1.Payload>> Decode(
+        IReadOnlyCollection<Temporalio.Api.Common.V1.Payload> payloads);
 }
 
 public record CompositePayloadCodec(IEnumerable<IPayloadCodec> Codecs) : IPayloadCodec
 {
-    public Task<IEnumerable<Temporal.Api.Common.V1.Payload>> Decode(
-        IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads) { /*...*/ }
+    public Task<IEnumerable<Temporalio.Api.Common.V1.Payload>> Decode(
+        IReadOnlyCollection<Temporalio.Api.Common.V1.Payload> payloads) { /*...*/ }
     
-    public Task<IEnumerable<Temporal.Api.Common.V1.Payload>> Encode(
-        IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads) { /*...*/ }
+    public Task<IEnumerable<Temporalio.Api.Common.V1.Payload>> Encode(
+        IReadOnlyCollection<Temporalio.Api.Common.V1.Payload> payloads) { /*...*/ }
 }
 
 public static class ConverterExtensions
 {
-    public static Task<IEnumerable<Temporal.Api.Common.V1.Payload>> ToPayloads(
+    public static Task<IEnumerable<Temporalio.Api.Common.V1.Payload>> ToPayloads(
         this IDataConverter converter,
         IReadOnlyCollection<object> values) { /*...*/ }
 
@@ -676,22 +676,22 @@ public static class ConverterExtensions
     // set of payloads than it was given.
     public static Task<IEnumerable<object>> ToValues(
         this IDataConverter converter,
-        IReadOnlyCollection<IEnumerable<Temporal.Api.Common.V1.Payload>> payloads,
+        IReadOnlyCollection<IEnumerable<Temporalio.Api.Common.V1.Payload>> payloads,
         IReadOnlyCollection<Type> types) { /*...*/ }
 
     public static Task<Exception> ToException(
         this IDataConverter converter,
-        Temporal.Api.Failure.V1.Failure failure) { /*...*/ }
+        Temporalio.Api.Failure.V1.Failure failure) { /*...*/ }
 
-    public static Task<Temporal.Api.Failure.V1.Failure> ToFailure(
+    public static Task<Temporalio.Api.Failure.V1.Failure> ToFailure(
         this IDataConverter converter,
         Exception exception) { /*...*/ }
 
-    public static T? ToValue<T>(this IPayloadConverter converter, Temporal.Api.Common.V1.Payload payload) { /*...*/ }
+    public static T? ToValue<T>(this IPayloadConverter converter, Temporalio.Api.Common.V1.Payload payload) { /*...*/ }
 
-    public static Task EncodeFailure(this IPayloadCodec codec, Temporal.Api.Failure.V1.Failure failure) { /*...*/ }
+    public static Task EncodeFailure(this IPayloadCodec codec, Temporalio.Api.Failure.V1.Failure failure) { /*...*/ }
 
-    public static Task DecodeFailure(this IPayloadCodec codec, Temporal.Api.Failure.V1.Failure failure) { /*...*/ }
+    public static Task DecodeFailure(this IPayloadCodec codec, Temporalio.Api.Failure.V1.Failure failure) { /*...*/ }
 }
 ```
 
@@ -705,7 +705,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Exceptions;
+namespace Temporalio.Exceptions;
 
 public abstract class TemporalException : Exception
 {
@@ -717,9 +717,9 @@ public class FailureException : TemporalException
     internal FailureException(
         string? message = null,
         Exception? inner = null,
-        Temporal.Api.Failure.V1.Failure? failure = null) : base(message, inner) { /*...*/ }
+        Temporalio.Api.Failure.V1.Failure? failure = null) : base(message, inner) { /*...*/ }
 
-    public Temporal.Api.Failure.V1.Failure? Failure { get; private init; }
+    public Temporalio.Api.Failure.V1.Failure? Failure { get; private init; }
 }
 
 // Intentionally extensible
@@ -743,7 +743,7 @@ public class AppException : FailureException
 
     public IEnumerable<object>? OutboundDetails { get; protected init; }
 
-    public IReadOnlyCollection<Temporal.Api.Common.V1.Payload>? InboundDetails { get; protected init; }
+    public IReadOnlyCollection<Temporalio.Api.Common.V1.Payload>? InboundDetails { get; protected init; }
 }
 
 // Many more not listed
@@ -766,7 +766,7 @@ Notes:
 For client, looks like:
 
 ```csharp
-namespace Temporal.Client.Interceptors;
+namespace Temporalio.Client.Interceptors;
 
 public interface IClientInterceptor
 {
@@ -812,7 +812,7 @@ public record StartWorkflowInput(
 
 Notes:
 
-* Worker interceptors look very similar, though they are in `Temporal.Worker.Interceptors` and have
+* Worker interceptors look very similar, though they are in `Temporalio.Worker.Interceptors` and have
   `IWorkerInterceptor.InterceptActivity(ActivityInboundInterceptor)` and
   `IWorkerInterceptor.InterceptWorkflow(WorkflowInboundInterceptor)`
 * Like Go and Python, if an interceptor provided to a client implements both `IClientInterceptor` _and_
@@ -827,7 +827,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Testing
+namespace Temporalio.Testing
 
 public class ActivityEnvironment
 {
@@ -904,7 +904,7 @@ Notes:
 .NET provides several
 [extensions to the runtime library](https://learn.microsoft.com/en-us/dotnet/standard/runtime-libraries-overview#extensions-to-the-runtime-libraries)
 which many users use when making apps. We discuss here some of those patterns and how we can support those, either by
-our existing design, or with helpers in a `Temporal.Extensions` namespace.
+our existing design, or with helpers in a `Temporalio.Extensions` namespace.
 
 With DI and configuration and such, this is very analogous to Spring Boot for Java.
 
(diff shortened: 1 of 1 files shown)


dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+using MyNamespace;
+
+// Connect
+var client = await TemporalClient.ConnectAsync(new("localhost:7233"));
+
+// Get workflow handle
+var handle = await client.GetWorkflowHandle<string>("my-workflow-id");
+
+// Send signal
+await handle.Signal(IGreetingWorkflow.Ref.CompleteWithGreetingAsync);
```
Sushisource, line 282: Maybe uprade this to the one that sends an arg just so we see what that looks like

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+  * Cannot think of a good way to make this better. We could:
+    * Use varargs (i.e. `params`), but not having options as the last parameter is ugly DX
+    * Use some kind of forced wrapper, e.g. `IArgumentCollection`, but it's ugly too
+    * Accept args in options instead of as param, but that options set is reused for other calls
+    * Accept both single-arg and multi-arg w/ "unset" defaults, but that is confusing and most .NET devs use their
+      positional parameters unnamed
+    * Have separate overloads for no-arg, 1-arg, and multi-arg but that's not any clearer
+* For `GetWorkflowHandle`, we cannot accept a workflow reference because in .NET if you don't type all type arguments
+  (i.e. including that parameter type), you have to help the type inferencer out. So since they'd have to provide a type
+  argument anyways, we make it where they either can here or wait until `GetResultAsync` to do so.
```
Sushisource, line 373: This is the saddest bit for me. Having to refer to the workflow _result_ type rather than the _workflow_ type feels like a nasty disconnect.

It seems like we could do what is done for signalling/querying and have it not be too ugly? Something like:

```csharp
var handle = await client.GetWorkflowHandle(IGreetingWorkflow.Ref.RunAsync, "my-workflow-id");
```

And now this can do the inference since it's getting the result type from the `RunAsync` function. Right?

The downside here is of course that saying `RunAsync` here feels a bit odd, but I know I'd prefer to do it this way instead of possibly getting the return type wrong.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+    public string ID { get; }
+    public string? RunID { get; }
+    public string? ResultRunID { get; }
+    public string? FirstExecutionRunID { get; }
+
+    public async Task<TResult> GetResultAsync(
+        bool followRuns = true, RPCOptions? options = null) { /*...*/ }
+
+    public async Task<TResultLocal> GetResultAsync<TResultLocal>(
+        bool followRuns = true, RPCOptions? options = null) { /*...*/ }
```
Sushisource, line 406: Possibly consider using a marker type like `Untyped`, and only adding this method to `WorkflowHandle<Untyped>` via an extension method. (And add an explicit `client.GetUntypedWorkflowHandle` which returns that)

That way there's an explicit way of creating an untyped workflow handle, and you avoid being able to specify a different result type when getting the result than was specified when you created the handle, which ideally is a state that can't be represented.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+var client = await TemporalClient.ConnectAsync(new("localhost:7233"));
+
+// Build worker options
+var options = new TemporalWorkerOptions("task queue");
+options.AddActivities(new GreetingActivities());
+
+// Run until Ctrl+C
+var worker = new TemporalWorker(client, options);
+var cts = new CancellationTokenSource();
+Console.CancelKeyPress += (sender, eventArgs) => {
```
Sushisource, line 479: Ok I have to say this is the nicest way of adding an additional signal handler I've seen in a language for a minute

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+    eventArgs.Cancel = true;
+    cts.Cancel();
+};
+await worker.ExecuteAsync(ctx.Token);
```
Sushisource, line 483: Should be `cts`

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+    public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options) { /*...*/ }
+
+    public Task ExecuteAsync(CancellationToken stoppingToken) { /*...*/ }
+}
+```
+
+Notes:
+
+* Like other options, technically all worker options are optional at compile time but we fail at runtime if task queue
+  is not set
```
Sushisource, line 520: If this is the case why have the no-arg constructor at all?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+    public void AddActivityDefinition(ActivityDefinition definition) { /*...*/ }
+
+    // Many more options not listed
+}
+
+public class TemporalWorker
+{
+    public TemporalWorker(IWorkerClient client, TemporalWorkerOptions options) { /*...*/ }
+
+    public Task ExecuteAsync(CancellationToken stoppingToken) { /*...*/ }
```
Sushisource, line 513: Why `Execute` rather than `Run` here?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+
+#endif
+}
+```
+
+Notes:
+
+* We have intentionally called this `WorkflowReplayer` in .NET since this kind of wordiness is clearer
+* Will need to accept a core runtime in the options here, not sure if we want that as part of worker options or if
+  replayer really needs its own options set that's 99% the same with worker. Maybe just a param.
```
Sushisource, line 576: I think a different options type is worth it. There are quite a lot of options that do not apply to replay worker, definitely nowhere near 99% the same.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+    { }
+}
+
+public interface IPayloadConverter
+{
+    IEnumerable<Temporal.Api.Common.V1.Payload> ToPayloads(
+        IReadOnlyCollection<object> values);
+
+    IEnumerable ToValues(
+        IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads, IReadOnlyCollection<Type>? types = null);
```
Sushisource, line 604: Is the way this works that the two collections must be the same size and the type for each payload is specified by the corresponding index in the types list? Or are there type tags inside the payloads and the types can be in any order?

If the orders must match, this seems easily error prone

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+}
+
+public interface IPayloadCodec
+{
+    IEnumerable<Temporal.Api.Common.V1.Payload> Encode(IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads);
+
+    IEnumerable<Temporal.Api.Common.V1.Payload> Decode(IReadOnlyCollection<Temporal.Api.Common.V1.Payload> payloads);
+}
+
+public record CompositePayloadCodec(IEnumerable<PayloadCodec> Codecs) : IPayloadCodec
```
Sushisource, line 661: Should be `IEnumerable<IPayloadCodec>`?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+Notes:
+* Several more not listed. We will also include non-failure exceptions (e.g. `WorkflowAlreadyStartedException`) here
+  instead of in the client area. This namespace is the .NET equivalent of
+  https://pkg.go.dev/go.temporal.io/api/serviceerror + https://pkg.go.dev/go.temporal.io/sdk/temporal wrt errors.
+* `AppException` is not `ApplicationException` to avoid clash with `System.ApplicationException`
+* `AppException` is the only inheritable exception
+* We have to differentiate outbound (converted) details and inbound (not converted) details since, like some other SDKs,
+  we cannot convert this stuff until the type is provided to do so
+  * Should we store the converter on the exception for easy conversion? Then we could have
+  `public T GetDetail<T>(int index = 0)`
```
Sushisource, line 732: I think that makes sense and is safer

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+  * We already need to apply for https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation
+  * Maybe we can persuade https://github.com/ritterim/temporal to deprecate and
+    https://learn.microsoft.com/en-us/nuget/nuget-org/deprecate-packages#transfer-popularity-to-a-newer-package but it
+    may require some ownership work
+
+#### Why `Temporalio`/`Temporal` NuGet/namespace and not `Temporalio.Sdk`/`Temporal.Sdk` NuGet/namespace?
+
+* Seemed simpler at first to just have `Temporal`
+* After more thought, maybe `Temporal.Sdk` would be better? It'd conform better with
+  https://learn.microsoft.com/en-us/dotnet/standard/design-guidelines/names-of-namespaces
```
Sushisource, line 954: I'd definitely do this one

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+* Many other supporting types
+
+Example of starting the workflow defined earlier in document, signalling it, then waiting on its results:
+
+```csharp
+using Temporal.Client;
+using MyNamespace;
+
+// Connect
+var client = await TemporalClient.ConnectAsync(new("localhost:7233"));
```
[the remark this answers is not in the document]
cretz: Not by default. I don't believe it is our standard. We encourage people to connect eagerly to find connection issues before their first call (or before their worker just hangs). However, `Lazy` is an option here, but I call it `ConnectAsync` (like I call it `connect` in Python) because it connects by default. If I had my way we wouldn't even support lazy :-)

dotnet/sdk-phase-1.md
cretz: From the signature later in `WorkflowHandle`, it would be `await handle.Signal(IGreetingWorkflow.Ref.UpdateInfoAsync, new GreetingInfo() { Name = "John" });`. If it is unclear enough I can alter the sample.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+* There will be `ExecuteWorkflowAsync` helper methods for each `StartWorkflowAsync` overload there
+* It was decided forcing an options object is easier than lots of parameters for start workflow
+* Notice that there is a string-based workflow `StartWorkflowAsync` call that accepts a return type and one that doesn't
+  and returns a handle for `Nothing`. The user can specify a return type on `WorkflowHandle.GetResultAsync` if they
+  want instead of at start time (or just never altogether if the return doesn't matter to the user).
+* For string-based calls to `StartWorkflowAsync`, we are accepting all args at once via `object[]`. But since array
+  types are covariant in .NET, this means you could have a workflow that accepts `string[]` and forget that needs to
+  just be the first value of the object array.
+  * Cannot think of a good way to make this better. We could:
+    * Use varargs (i.e. `params`), but not having options as the last parameter is ugly DX
```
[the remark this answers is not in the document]
cretz: > I think I would actually prefer options to come before args

It is very normal in programming languages to have thing your calling followed immediately by its arguments. This is also the case in all of our SDKs to date (though Go strangely has the options before everything). Also, having options last means they can be left off in cases where they are not needed (e.g. signals have no options _yet_) and they look better when they become multiline which is common. And it is also common in .NET.

Note this bullet only applies to people using the untyped form of starting a workflow. I hope it is not commonly used anyways.

dotnet/sdk-phase-1.md
cretz: > It seems like we could do what is done for signalling/querying and have it not be too ugly? Something like:
> ```csharp
> var handle = await client.GetWorkflowHandle(IGreetingWorkflow.Ref.RunAsync, "my-workflow-id");
> ```

This is the conundrum I spoke about in Slack. I so wish we could do that. Say you have this signature for `GetWorkflowHandle` that looks like `StartWorkflowAsync` to support your snippet there:

```csharp
public WorkflowHandle<TResult> GetWorkflowHandle<T, TResult>(
    Func<T, Task<TResult>> workflow,
    string workflowID,
    string? runID = null,
    string? firstExecutionRunID = null
)
```

You get this error:

> error CS0411: The type arguments for method 'TemporalClient.GetWorkflowHandle<T, TResult>(Func<T, Task<TResult>>, string, string?, string?)' cannot be inferred from the usage. Try specifying the type arguments explicitly.

This is because, I assume due to type reification, .NET requires all type arguments to be inferencable. So your example changes to:

```csharp
var handle = await client.GetWorkflowHandle<GreetingInfo, string>(IGreetingWorkflow.Ref.RunAsync, "my-workflow-id");
```

And no, you can't specify just one type arg, you have to specify them all. And no, there is no `_` or `?` or `unknown` or similar placeholder for an that "arg" type param there. And no you can't give it some covariant like `object` (e.g. accept `Func<object, Task<TResult>>`) and expect it to work (you get errors like "Argument 1: cannot convert from 'method group' to 'Func<object, Task<string>>'" or at the least, force a type arg anyways).

This is just a known limitation in the .NET world. Their generic type inferencer just doesn't allow you to ignore types you don't provide. As one SO commenter states:

> As a rule, you're not going to have any luck with type inference when your method signature has fewer types of parameters than it has generic type parameters.

Meaning if you don't accept the types from the user in your signature, the inferencer won't be able to figure them out. I have spent a long time on this problem, enough to make me even consider switching to the proxy+lambda approach of Azure durable entities. But I figured:

* This only occurs on `GetWorkflowHandle` and, in the future, `GetExternalWorkflowHandle` on workflows
* People need a way to specify the return type anyways when using untyped handles
* I can provide an no-type-arg-required form for people that don't care about the return
* I can let people defer their type designation until they actually need the result (untyped when getting handle, specify result at result time)
* Other than the result type, there is no other benefit to accepting the workflow run method reference here. At least in Python I can force all signals/queries on the workflow handle to be of the same workflow class the run call is on, but not .NET.

After pouring over this forever, I believe what I have here is the best solution.

dotnet/sdk-phase-1.md
cretz: I can make this an extension method, Will think on it. I can also provide something like https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.cast. I like the simplicity of having it here for discoverability and for anyone to force-ably override the result type, but will ponder.

dotnet/sdk-phase-1.md
cretz: :+1: Will fix next commit (may wait for other reviews).

dotnet/sdk-phase-1.md
cretz: From below:

> `ExecuteAsync(CancellationToken)` was specifically done this way to match `Microsoft.Extensions.Hosting.BackgroundService.ExecuteAsync`

dotnet/sdk-phase-1.md
cretz: Mentioned way later in document, but this follows https://learn.microsoft.com/en-us/dotnet/core/extensions/options which states:

>  When using the options pattern, an options class:
> * Must be non-abstract with a public parameterless constructor

I did bring this up in the `#dotnet-sdk` channel on public Slack on whether to choose this pattern or immutable records (my personal favorite), but this is the safest bet for all of our options.

dotnet/sdk-phase-1.md
cretz: Core reuses the same config type :-)

But yeah, makes sense. Maybe I'll make this and the worker options share a common base class and/or maybe some conversion between the two or overload that accepts worker config or something. My goal is to allow people to share their worker options during replaying to ensure they get everything right. But there are some things that are defined at the client level (e.g. data converter) anyways they have to provide to a replayer.

dotnet/sdk-phase-1.md
cretz: > If the orders must match, this seems easily error prone

They must match (same size and order) and it is error prone. In Python where we do similar, we disregard if the count isn't the same. I put a lot of thought into this.

* It is an optional argument
* This is a low-level converter implementer concern. The caller is usually me inside the SDK which ensures proper values.
* For those converting manually, they'll be using some higher level helper that will probably only do one value at a time anyways (e.g. `MyConverter.ToValue("foo")`) that then calls this with the generic type there
* We choose at the lowest level to have a collection form of this instead of single payload conversion, but maybe we should change that to just be single? But I'd still provide a `Type` not a generic due to how I have to do it at runtime.
* Technically I could have `ToValues(IReadOnlyCollection<KeyValuePair<Payload, Type>>)` but that assumes I always have a type...maybe I should assume that? In cases where the type is known from the payload, e.g. a proto, the type can be determined without this parameter...but I guess technically they have to provide _some_ type at the call site even if just an object

After some thought, I am considering changing this to:

```
object? ToValue(Payload payload, Type type)
```

With an extension for:

```
T ToValue<T>(Payload payload)
```

This would match https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializer.deserialize. I am trying to think through the ramifications of singular value payload-to-value conversion instead of multiple. I can't think of any (like I can with codec).

dotnet/sdk-phase-1.md
cretz: Yes, will fix next commit

dotnet/sdk-phase-1.md
cretz: The concern there is, for inbound details (i.e. from proto) I have to carry around a converter as a private field inside the exception. And for outbound details (i.e. not yet to proto) I have to cast this type arg at runtime. So it starts to look something like:

```csharp
public T GetDetail<T>(int index = 0)
{
  // If there are inbound details use the converter, otherwise cast from outbound
  if (InboundDetails != null)
  {
    return converter!.ToValue<T>(InboundDetails[index]);
  }
  return (T) OutboundDetails[index];
}
```

I _guess_ that's ok so long as the user doesn't get the type wrong.

dotnet/sdk-phase-1.md
cretz: I'm starting to lean this way too. I think `Temporal.Sdk` _does_ sound better. But then I would want `Temporalio.Sdk` NuGet package? Technically the NuGet package and the actual in-use namespace can be different, and it happens plenty, but most like it the same.

Maybe I should just go ahead and use `Temporalio.Sdk` for the namespace and NuGet package? `Temporalio` isn't too ugly as a root namespace right?

dotnet/sdk-phase-1.md
Sushisource: OK, this is the part I was missing:
> Meaning if you don't accept the types from the user in your signature, the inferencer won't be able to figure them out. 

So signal works because one of the args is one of the generic types. Frustrating.

I think it'd still be nice if there was some static function, ex: `MyWorkflow.makeHandle(client)` which would do the write thing with types, but since I guess we don't control the whole interface/class, only certain functions, that doesn't quite work.

dotnet/sdk-phase-1.md
Sushisource: `ToValues(IReadOnlyCollection<KeyValuePair<Payload, Type>>)` was what I was going to suggest, but is maybe a tiny bit annoying for people to zip together. It is safer, though.

I think only having `ToValue` is also very reasonable. That's how most deserialization API's I've ever seen work. Just more obvious.

dotnet/sdk-phase-1.md
cretz: > Frustrating

Exactly.

> that doesn't quite work.

Yeah, no matter how much I twist, I can't find a way given a workflow run method reference only, how to make a return-type-inferred handle from it if that run method has an argument.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1022 @@
...
+Notes:
+
+* We use regular .NET JSON serialization (see data conversion later)
+* Workflow attributes, and in the future workflow runtime features, are in the `Temporal.Workflow` namespace
+* We only support workflow interfaces right now because we don't have a workflow implementation yet
+  * For now, everything non-static must have some form of `[Workflow` attribute, and not have a default implementation
+* Every workflow interface must have the non-inheritable `Workflow` attribute
+  * Optional `Name` positional property can be given. Default is the unqualified type name with the "I" prefix removed
+    if it's an interface (which it has to be currently) and the next letter is also capital
+    * So what does `IDWorkflow` become? Yes, `DWorkflow` :-( Weighed as an acceptably rare cost for obvious benefit
```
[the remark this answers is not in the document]
cretz: If I was forced to do this, I would just never infer a name ever. Workflow names not having little language-specific things like these prefixes be default I think is important. This would only make more typing for 99% of users who won't care for just one user that may be surprised we removed the `I` .NET expects to be the first char in some case where they want to reference it by string.

dotnet/sdk-phase-1.md
cretz: :+1: I'll change it. I'll also change payload codec to properly be async-capable.

cretz: I have changed payload converter to only deal with single values, changed payload codec to support async, and added extension methods to these converters.

(leaving this discussion unresolved for posterity)

dotnet/sdk-phase-1.md
cretz: Fixed.

dotnet/sdk-phase-1.md
cretz: Added this

Sushisource (approved):
Looking good!

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+### Async Completion
+
+Simply `throw new Temporal.Activity.CompleteAsyncException` from the activity.
+
+## Client
+
+Inside the `Temporal.Client` namespace, we have the following types:
+
+* `TemporalClient` - namespace-specific, stateless high-level client to Temporal
+  * See later for most of its API design
```
Quinn-With-Two-Ns, line 239: Would this client include other server APIs like `Schedules`?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+The .NET SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-core), utilizing "P/Invoke" and a
+custom-built bridge.
+
+The high-level goals for the .Net SDK are:
+
+* Be as simple as possible from a caller POV
+* Be as lightweight as reasonable with regards to dependencies
+* Be reasonably flexible with reasonable defaults
+* Fit well into the .Net ecosystem
+* Minimum versions supported: .NET Core 3.1, .NET Standard 2.0, and .NET Framework 4.6.2
```
Quinn-With-Two-Ns, line 28: Just for my own understanding how did we choose these versions?

dotnet/sdk-phase-1.md
cretz: Yes

bergundy (commented):
Very nice proposal, well though out.
I got up to the "Client" section, I plan to continue on Monday.

dotnet/sdk-phase-1.md
marcusbooyah: It looks like .NET core 3.1 will be unsupported by MSFT in just a couple of days. [Spurce](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core)

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: > but IXXx would be rejected but that's probably taking it too far.

This would force people to violate the .NET standard. So `IIDWorkflow` or `IQRWorkflow` or whatever, valid names, would fail.

I think we can count on people always using the universally accepted `I` prefix for interfaces.

dotnet/sdk-phase-1.md
cretz: :+1: .NET core 3.1 support doesn't significantly harm us though.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
+# .NET SDK - Phase 1
+
+The .NET SDK will be implemented in three loosely defined phases:
+
+* Phase 1 - Initial project, activity worker, and client
+* Phase 2 - Workflow worker
+* Phase 3 - Deterministic workflow sandbox and/or static analyzer
```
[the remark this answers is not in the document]
cretz: Replayer part of phase 2 and test framework I'm actually mostly including as part of phase 1. It's actually the very first thing I am completing fully (something I wish I did in Python) so I can write my tests. I have basically already completed it in my POC.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+    public string Salutation { get; init; } = "Hello";
+
+    [JsonPropertyName("name")]
+    public string Name { get; init; } = "<unknown>";
+}
+
+[Workflow]
+public interface IGreetingWorkflow
+{
+    static readonly IGreetingWorkflow Ref = Refs<IGreetingWorkflow>.Instance;
```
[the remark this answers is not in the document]
cretz: I didn't really tbh. The ref pattern is a result of research into referencing methods and the only way implemented after weighing it against the other two options (proxy lambda and source generation).

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+    intentionally absent "sdk" term for SEO reasons)
+    * Add README note on https://github.com/temporalio/legacy-dotnet pointing to
+      https://github.com/temporalio/sdk-dotnet
+    * Disable GH pages
+    * Archive it
+    * Consider deletion at some point in the future
+  * Rename https://github.com/temporalio/sdk-dotnet-incubating to https://github.com/temporalio/sdk-dotnet
+  * Update DNS for https://dotnet.temporal.io/ to point to vercel-based API docs page
+
+Doing the repo rename was deemed easier than going through manually closing issues, closing PRs, creating empty `main`
```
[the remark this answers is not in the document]
cretz: Concur hence "Consider deletion at some point in the future" above this. But doesn't have to be at first.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+  * Optional `Name` positional property can be given. Default is the unqualified type name with the "I" prefix removed
+    if it's an interface (which it has to be currently) and the next letter is also capital
+    * So what does `IDWorkflow` become? Yes, `DWorkflow` :-( Weighed as an acceptably rare cost for obvious benefit
+* Every workflow must explicitly have one and only one method with a non-inheritable `WorkflowRun` attribute
+  * Can be in base class, but must also be overridden in this workflow class w/ the attribute even if it just delegates
+    to super
+  * Method must return a `Task`
+* Workflow signals are methods with the non-inheritable `WorkflowSignal` attribute
+  * Can be in base class, but if overridden, must also have the attribute
+  * Method must return a `Task` and can accept arguments
```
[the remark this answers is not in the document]
cretz: > Is there a way to support both sync and async handlers?

Technically yes, but note that the presence of `Task` return does not require that the function be async. In fact you'll get a warning if you mark a `Task`-returning function as `async` if you don't `await` anywhere in it. But I understand it can be annoying to return a `Task` if you have no need for asynchrony in the signal.

After confirming as I have with activities that this won't cause type inference ambiguity I can support non-async signals I believe. I kinda liked the always-one-way myself.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+  * Optional `Name` positional property can be given. Default is the unqualified method name with "Async" trimmed off if
+    present.
+* Workflow queries are methods with the non-inheritable `WorkflowQuery` attribute
+  * Can be in base class, but if overridden, must also have the attribute
+  * Method cannot return a `Task`, but can accept arguments
```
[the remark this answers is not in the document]
cretz: Yes, I believe so. I have to confirm as I did with activities that method references are properly disambiguated to use the more specific overload when inferred, but otherwise I think we can.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+  * Optional `Name` positional property can be given. Default is the unqualified method name.
+* Users are encouraged to set a `static` `readonly` `Ref` property for use by clients
```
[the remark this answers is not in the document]
cretz: Oh, are you maybe asking for like a "Rejected Alternatives" section?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+The activity context looks like this:
+
+```csharp
+namespace Temporal.Activity;
+
+public class ActivityContext
+{
+    public static ActivityContext Current { get; }
+
+    public static bool HasCurrent { get; }
```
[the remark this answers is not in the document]
cretz: Hrmm, yeah both have benefits. In this case, `HasX` lets you know you can access property `X`.

I now wonder if, instead of a `Current` (so you have to `ActivityContext.Current.Heartbeat()`) I should just make these static (so you `ActivityContext.Heartbeat()`).

I also considered changing the name of this class to just `Activity` but it clashes with `Temporal.Activity`

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
bergundy: The test framework makes total sense as first thing to be implemented.
The replayer (+mass replayer + history downloading + list API) could be a separate item or at least noted here in phase 2.
Still a nit so don't feel obligated to change anything.

dotnet/sdk-phase-1.md
mjameswh: Not specifically pushing for more recent versions, but maybe we should set a soft goal and a hard goal... I mean, I don't see any reason to go out of our way to support .Net core 3.1 or Framework 4.6.2, but as long as they don't make things more complicated than they need to, then keeping them would be nice.

mjameswh: Question: This proposal is about phase 1 (client + activity worker), which should be not too hard to make compatible across .Net editions. However, I expect that workflow workers, and particularly sandboxing, might require going more in depth on language features, making it much more difficult to implement on some of these. Do we have some insights on this at this point?

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
bergundy: Why accept a `Lazy` option instead of a static method with that name and avoid the runtime error altogether?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+    public string Namespace { get; }
+    public DataConverter DataConverter { get; }
+    public IBridgeClientProvider BridgeClientProvider { get; }
+    public ITemporalConnection Connection { get; }
+    public TemporalClientOptions Options { get; }
+
+    // StartWorkflowAsync
+
+    public async Task<WorkflowHandle<TResult>> StartWorkflowAsync<TResult>(
+        string workflow, object[] args, StartWorkflowOptions options) { /*...*/ }
```
bergundy, line 313: With the proxy lambda approach, would we be able to support type safe multi argument methods?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+using Temporal.Activity;
+
+public class TemporalWorkerOptions
+{
+    public TemporalWorkerOptions() { /*...*/ }
+    public TemporalWorkerOptions(string taskQueue) { /*...*/ }
+
+    public string? TaskQueue { get; set; }
+
+    public List<ActivityDefinition> ActivityDefinitions { get; set; } = new();
```
bergundy, line 500: Can you provide more info on what's an ActivityDefinition and when a user might want to add a definition directly?

dotnet/sdk-phase-1.md
cretz: > Not specifically pushing for more recent versions, but maybe we should set a soft goal and a hard goal... I mean, I don't see any reason to go out of our way to support .Net core 3.1 or Framework 4.6.2, but as long as they don't make things more complicated than they need to, then keeping them would be nice.

Yeah, not wanting to go out of our way. This is what https://www.nuget.org/packages/Microsoft.Azure.WebJobs.Extensions.DurableTask, a similar library, supports. But you can't have "soft" and "hard", basically you need to have CI for the lowest and most recent at least.

> Do we have some insights on this at this point?

I do not. I have not even put any worthwhile research into what .NET sandboxing approaches may exist for us.

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: > history downloading + list API

I have included that in this proposal, but not the replayer. I consider it an extension of workflows (the Python SDK separated that work unnecessarily, this one won't).

>  Still a nit so don't feel obligated to change anything.

Will clarify on next commit. But I'm not trying to be detailed in what will be supported in future versions. The way I want people to think about it is "this phase has what is here at high level and what is needed to support that, and every future phase has future stuff". If the high-level part is not here or mentioned/referenced, it's deferred.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+  * Inheritance rules apply from the earlier activity definition section
+  * Users can use the lower-level `AddActivityDefinition` but we left this name unwieldy on purpose, we prefer
+    `AddActivities`
+* Since workflows are not currently implemented, at least one activity is required
+* Activities can be sync or async
+  * For future `Workflow.ExecuteActivity` uses, we have confirmed .NET can disambiguate in an overload between `Task`
+    return type and generic return type with no constraints
+  * Async activities are executed using the normal scheduler
+    * No need yet for a custom scheduler
+  * Sync activities are executed with `ThreadPool`
```
bergundy, line 538: Is this a fixed size thread pool? How does this work with the worker limitations of `maxConcurrent(Local)Activities`?
Can user get themselves into a situation where the worker polls on an activity and there's no free slot on the pool to execute it?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+
+```csharp
+namespace Temporal.Worker;
+
+class WorkflowReplayer
+{
+    public WorkflowReplayer(TemporalWorkerOptions options) { /*...*/ }
+
+    public async Task<WorkflowReplayResult> ReplayWorkflow(
+        WorkflowHistory history, bool throwOnReplayFailure = true) { /*...*/ }
```
bergundy, line 561: I think we always want to throw here and return void.

dotnet/sdk-phase-1.md
mjameswh: >But you can't have "soft" and "hard", basically you need to have CI for the lowest and most recent at least.

Sure... We can start by setting the CI to number X, but know that if you face any difficulty that can be resolved by updating to X+1, then you just do so, without remorse and without loosing a week researching/demonstrating/arguing about how it would be possible to get around that in version X...

bergundy (approved):
Overall LGTM.
As I mentioned a few times in my comments inline, I would like to see some justification to why the lambda proxy pattern was dismissed but I'm okay with the current direction too.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+
+* We use regular .NET JSON serialization (see data conversion later)
+* Workflow attributes, and in the future workflow runtime features, are in the `Temporal.Workflow` namespace
+* We only support workflow interfaces right now because we don't have a workflow implementation yet
+  * For now, everything non-static must have some form of `[Workflow` attribute, and not have a default implementation
+* Every workflow interface must have the non-inheritable `Workflow` attribute
+  * Optional `Name` positional property can be given. Default is the unqualified type name with the "I" prefix removed
+    if it's an interface (which it has to be currently) and the next letter is also capital
+    * So what does `IDWorkflow` become? Yes, `DWorkflow` :-( Weighed as an acceptably rare cost for obvious benefit
+* Every workflow must explicitly have one and only one method with a non-inheritable `WorkflowRun` attribute
```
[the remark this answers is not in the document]
cretz: In these languages requiring a certain method name w/ no compile-time limitation is not usually how you see these implemented. If you want to mark a method for a purpose, you use the language's preferred way of marking it. Otherwise, you use inheritance/interfaces to define method name requirements.

The benefits for inheritance are not necessarily more than you could get requiring people use a certain method name. It's more about code legibility for inheritance at that point. But yes, you can say "every workflow must have a `RunAsync` method on that class" just as easily as you could say "every workflow must have a method marked `WorkflowRun` on that class".

Also, this is consistent with Python and Java and PHP, the other SDKs that use language metadata to mark workflow entry points.

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: For one you'd have to make every method virtual, you can't proxy normal methods. I will add a some notes discussing proxying.

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: Concur and I am on the same fence. One thing that doesn't come into play is the verbosity - .NET developers definitely don't mind that.

One reason many of these things are instances instead of static is for DI purposes. I considered making a `IActivityContext` interface, but I couldn't have a static `Current` property on it in _all_ versions of .NET we are targeting (because you can't have any non-interface-methods on interfaces in older .NET versions). So that's why I have it like this.

dotnet/sdk-phase-1.md
cretz: Yes, but we have avoided those in Python too where we could have done a similar thing. In addition to requiring the non-interface methods be virtual, it also introduces some difficulties (e.g. I can't stop you from making other calls in the lambda either).

dotnet/sdk-phase-1.md
cretz: In the notes I have:

> Activity definitions are the underlying type of the activity collection (they have just a name and method to call)

It's harmless to expose this in the same way as you could call `temporalio.activity.defn(name="foo")(myFunction)` in Python. I can't think of how it might be used, but with these options classes, I usually see in .NET the raw underlying type with helpers to add. You don't usually make things private in options classes.

dotnet/sdk-phase-1.md
cretz: > Is this a fixed size thread pool?

This is a global .NET thing: https://learn.microsoft.com/en-us/dotnet/api/system.threading.threadpool. There is not even a way to make a custom thread pool in .NET. It's kinda weird but very common for .NET developers to use this.

> How does this work with the worker limitations of maxConcurrent(Local)Activities?
Can user get themselves into a situation where the worker polls on an activity and there's no free slot on the pool to execute it?

It'll just pool up work until a thread is open for use. So they could hit heartbeat timeouts and, if I don't have a thread set aside for workflows (I probably won't) task timeouts there. I can't really think of how to improve this. They should set `SetMinThreads` globally probably, but if everyone properly uses async, we shouldn't make them (those are non-blocking but still use the same thread pool).

You might think you could warn a user with ".NET thread pool has fewer threads than concurrent activity limit", but again with async that's unfair. And people may be totally ok with having a received activity wait on an available thread (workflows less so, will discuss when we get there). And .NET'ters are used to this limitation.

I am open to alternatives, but short of building custom thread pools the user can pass around, there isn't much we can do besides docs that I can think of.

dotnet/sdk-phase-1.md
cretz: I could have `ReplayWorkflow` that always throws and `TryReplayWorkflow` that returns this value, but I think this pattern is simpler. It's helpful for people if they wanted to write their own mapper and collect the results. Same as Python.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+{
+    public WorkflowReplayer(TemporalWorkerOptions options) { /*...*/ }
+
+    public async Task<WorkflowReplayResult> ReplayWorkflow(
+        WorkflowHistory history, bool throwOnReplayFailure = true) { /*...*/ }
+
+#if NETCOREAPP3_0_OR_GREATER
+
+    public IAsyncEnumerator<WorkflowReplayResult> ReplayWorkflows(
+      IAsyncEnumerator<WorkflowHistory> histories, bool throwOnReplayFailure = true) { /*...*/ }
```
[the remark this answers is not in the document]
cretz: Yes, technically the caller could write this whole method. It was decided on Python and others to provide an option to eagerly throw on first failure instead of return. No reason for .NET to deviate.

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: We need a common entry point for lazy and non-lazy here as we did in Python IMO. This is even more important for .NET where people will use configuration separately from where they make the call (https://learn.microsoft.com/en-us/dotnet/core/extensions/configuration will work with our options by intention).

Also, lazy should be rare/discouraged IMO. And we have to carry it around as an option anyways. Having one entry point to client connection is clearer than overloads.

It's only a runtime error if they try to use the connection constructor in a non-lazy way. I think it's worth always having the constructor available for those wanting to manually construct a connection though (unless there's a good reason, constructors are often made visible to users in .NET).

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: > but the proxy lambda approach seems like a valid alternative.

Not sure I agree, but I'll explain in proposal. In Java we regret it because the call is literally on the proxy. In .NET, I guess I can ask people to please only use simple single calls in the lambdas and not do advanced things with it (we can't prevent all method calls on proxies).

If we forced people to make interfaces for everything they do, like Durable Entities and Temporal Java do, then it'd make more sense. I was hoping to avoid that. Proxies require virtual methods.

Worst case, we can add it on top.

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
mjameswh: >technically there's no reason why query handlers can not call external non-workflow APIs to respond using both workflow state and external data

Do we support async query handlers in any lang already? I'm probably missing some context or knowledge here, but it seems to me like there are quite a few things that makes this undesirable (ie. doesn't a query has to complete in the same WFT?), and even impossible in sandboxed langs...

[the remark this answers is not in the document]
cretz: > Do we support async query handlers in any lang already?

Yes, Python. When originally designed, the concern was that there were async things in the language that a query handler might want to do. In retrospect though after using for a while, I didn't really find any `async` constructs that one would use in a query handler. So I may want to remove it, we'll see.

> seems to me like there are quite a few things that makes this undesirable (ie. doesn't a query has to complete in the same WFT?), and even impossible in sandboxed langs

Just because a query has to complete in the same WFT doesn't mean it can't use something async in the language. Some utilities in languages that are effectively immediate are marked async and in some languages you can't even invoke them outside of an async context. Having said that, in most languages invoking one of these even outside of an async context is often doable.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+
+public class TemporalClient : ITemporalClient
+{
+    public static async Task<TemporalClient> ConnectAsync(TemporalClientConnectOptions options) { /*...*/ }
+
+    public TemporalClient(ITemporalConnection connection, TemporalClientOptions options) { /*...*/ }
+
+    public string Namespace { get; }
+    public DataConverter DataConverter { get; }
+    public IBridgeClientProvider BridgeClientProvider { get; }
```
mjameswh, line 306: Do you really intend to expose this? Why?

dotnet/sdk-phase-1.md
mjameswh: I don't really want to push in that direction, but of curiosity, would it works if workflows had to extend from a base interface, with the `RunAsync` method's return type being a generic parameter on the interface?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+
+public class WorkflowHandle<TResult>
+{
+    public string ID { get; }
+    public string? RunID { get; }
+    public string? ResultRunID { get; }
+    public string? FirstExecutionRunID { get; }
+
+    public async Task<TResult> GetResultAsync(
+        bool followRuns = true, RPCOptions? options = null) { /*...*/ }
```
mjameswh, line 403: Why do we have `RpcOptions` on all `WorkflowHandle` methods, but nowhere else? If we want to allow changing `RpcOptions` only for some specific calls/series of code, can't we have some way to wrap these calls in some "withGrpcOptions" context (ie. using thread local or async local) rather than adding this extra argument everywhere?

dotnet/sdk-phase-1.md
cretz: Yes, if we force everyone to implement an interface that specifies the return type generically, then we can have functions that return a handle for that. So if you have:

```csharp
class MyWorkflow : IWorkflow<MyReturnType>
{
  public async Task<MyReturnType> RunAsync() { /*...*/ }
}
```

Then you can have:

```
public WorkflowHandle<TResult> GetWorkflowHandle<TResult>(IWorkflow<T> workflow...
```

But, you'd have to have a different interface for one that accepts an arg vs no arg (and to do it properly, a different one for no return too because there is no generic `void`/"unit" type in .NET)

Changing up the entire SDK to avoid some user specifying return type when they happen to manually get a handle is a bit much.

dotnet/sdk-phase-1.md
cretz: I do. I have a similar thing in Python.

My reasoning here is that people want to make their own clients or wrap their own clients, but we still need this usable from a worker. We messed this up in Go and now people can't wrap clients the way they want because we cast back to the concrete type when using in a worker. So we have this interface that `ITemporalClient` implements (left out for brevity):

```csharp
namespace Temporal.Worker;

public interface IWorkerClient
{
    public string Namespace { get; }
    public Temporal.Converters.DataConverter DataConverter { get; }
    public Temporal.Client.IBridgeClientProvider BridgeClientProvider { get; }
}
```

So if you wrap your client, so long as it still implements `IWorkerClient` (which is implemented by `ITemporalClient` which is probably what you'd conform to), you're ok. So you'd just pass through the bridge client provider from the underlying client.

I would definitely entertain better options here. The goal is "Allow easy wrapping/impl of the Temporal client interface while still allowing that to be used to create workers".

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+
+    public async Task TerminateAsync(
+        string reason? = null, object[]? args = null, RPCOptions? options = null) { /*...*/ }
+
+    public async Task<WorkflowExecutionDescription> DescribeAsync(RPCOptions? options = null) { /*...*/ }
+
+#if NETCOREAPP3_0_OR_GREATER
+
+    public async Task<WorkflowHistory> FetchHistoryAsync(
+        WorkflowHistoryEventFilterType eventFilterType = WorkflowHistoryEventFilterType.AllEvents,
```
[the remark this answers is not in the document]
cretz: I default them. But I think they should be exposed. If we're allowed to wait on events and filter them so should a user.

dotnet/sdk-phase-1.md
cretz: > Why do we have RpcOptions on all WorkflowHandle methods, but nowhere else?

I didn't make it clear, but but `RpcOptions` are children of `StartWorkflowOptions` and `ListWorkflowsOptions`. And they are on all async activity handle calls (left off for brevity). Every gRPC call we make has the ability to set RPC options.

>  If we want to allow changing RpcOptions only for some specific calls/series of code, can't we have some way to wrap these calls in some "withGrpcOptions" context (ie. using thread local or async local) rather than adding this extra argument everywhere?

We allow RPC metadata (but not timeout) at the client level. Timeout doesn't make sense at a client level. You only add the option if you need for a call when you make that call. This matches Python and others (but advanced client options left off for brevity).

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: Added discussion point on lambda vs method reference

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: Added discussion point on lambda vs method reference

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1052 @@
...
+    }
+}
+```
+
+Notes:
+
+* Activities can be on static or non-static methods
+  * Even though our proxying library requires virtual methods for interception, we're never intercepting, only
+    _referencing_.
+    * This means if someone _does_ invoke an instance method on a `Ref`, it'd call it normally. We can't help this.
```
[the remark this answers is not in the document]
cretz: Added discussion point on lambda vs method reference

dotnet/sdk-phase-1.md
cretz: Added discussion point on lambda vs method reference

dotnet/sdk-phase-1.md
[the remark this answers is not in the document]
mjameswh: What bothers me is that they get in the way of `RpcOptions`, which means that a user who wants to set `RpcOptions` is forced to explicitly set all of these other (very rarely used) options. But maybe named arguments would help there, not sure how they interact with optional args. Otherwise, I'd suggest an overload on the signature with `(RpcOptions)` only.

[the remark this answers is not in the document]
cretz: > But maybe named arguments would help there, not sure how they interact with optional args.

Right. You can ignore optional args and use named args. So you can have:

```csharp
await myHandle.FetchHistoryAsync(options: new(timeout=TimeSpan.FromMinutes(3)));
```

That reminds me I do need to change that arg name to `rpcOptions` instead of just `options`.

dotnet/sdk-phase-1.md
mjameswh: >My reasoning here is that people want to make their own clients or wrap their own clients, but we still need this usable from a worker. We messed this up in Go and now people can't wrap clients the way they want because we cast back to the concrete type when using in a worker.

I get it.

Could we have a single `asWorkerClient()` method, that returns an opaque `WorkerClient` instead? The key difference here would be that this object would be guaranteed to be immutable (we don't have that certainty on a user implementation of `IWorkerClient`) and that `BridgeClientProvider` would not be publicly exposed.

I can see why devs might want to extends `TemporalClient`, but I'd prefer not provide them a footgun if we can avoid that...

cretz:
Made team-suggested changes. Closing this PR in favor of a new PR for final comments before merge.

PR #71, opened 2022-12-14
.NET SDK - Phase 1 - Final Comment Period

cretz:
# .NET SDK - Phase 1 - Proposal Review - Final Comment Period

This is the proposal for the first phase of the .NET SDK.

**Please see the initial comment PR at #70 where most discussion took place**.

* See it [rendered](https://github.com/cretz/temporal-proposals/blob/dotnet-phase-1/dotnet/sdk-phase-1.md) (or can use the "rich diff" link in upper right; rendered link will disappear on PR merge)
* The playground/POC repo I was toying with concepts and a Rust bridge is [here](https://github.com/cretz/temporal-sdk-dotnet-poc). It was just for testing concepts. It is not accurate and may be deleted later.

---

image

### DIFF
--- a/dotnet/sdk-phase-1.md
+++ b/dotnet/sdk-phase-1.md
@@ -19,42 +19,28 @@ Notes for this proposal:
 The .NET SDK will be backed by [sdk-core](https://github.com/temporalio/sdk-core), utilizing "P/Invoke" and a
 custom-built bridge.
 
-The high-level goals for the .Net SDK are:
+The high-level goals for the .NET SDK are:
 
 * Be as simple as possible from a caller POV
 * Be as lightweight as reasonable with regards to dependencies
 * Be reasonably flexible with reasonable defaults
-* Fit well into the .Net ecosystem
+* Fit well into the .NET ecosystem
 * Minimum versions supported: .NET Core 3.1, .NET Standard 2.0, and .NET Framework 4.6.2
 * All common 64-bit platforms supported, 32-bit maybe later
 
 ## Repository Management
 
 The existing repository at https://github.com/temporalio/sdk-dotnet has hundreds of commits, dozens of issues, open PRs,
-GH pages, etc that don't align with this proposal necessarily. After some thought, it has been decided to:
-
-* Create a new https://github.com/temporalio/sdk-dotnet-incubating repo for the very first commits
-* After that repo has some code and API doc gen capability:
-  * Rename https://github.com/temporalio/sdk-dotnet to https://github.com/temporalio/legacy-dotnet (note the
-    intentionally absent "sdk" term for SEO reasons)
-    * Add README note on https://github.com/temporalio/legacy-dotnet pointing to
-      https://github.com/temporalio/sdk-dotnet
-    * Disable GH pages
-    * Archive it
-    * Consider deletion at some point in the future
-  * Rename https://github.com/temporalio/sdk-dotnet-incubating to https://github.com/temporalio/sdk-dotnet
-  * Update DNS for https://dotnet.temporal.io/ to point to vercel-based API docs page
-
-Doing the repo rename was deemed easier than going through manually closing issues, closing PRs, creating empty `main`
-branch, and keeping `master` branch around for those reading older code.
+GH pages, etc that don't align with this proposal necessarily. After some thought, it has been decided to move that
+repository and create https://github.com/temporalio/sdk-dotnet again fresh.
 
 ## Namespace/Package
 
 * Will publish as `Temporalio`
   * Need to reserve prefix `Temporalio` on NuGet, see
     https://learn.microsoft.com/en-us/nuget/nuget-org/id-prefix-reservation
-* Namespace will be `Temporal`
-  * See discussion topics later about this being `Temporal.Sdk` instead
+* Namespace will be `Temporalio`
+  * See discussion topics later about this being `Temporalio.Sdk` instead
 
 ## Workflow Definition
 
@@ -64,7 +50,7 @@ Here's a definition for an other-language-defined workflow (from Python README):
 namespace MyNamespace;
 
 using System.Text.Json.Serialization;
-using Temporal.Workflow;
+using Temporalio.Workflow;
 
 public record GreetingInfo
 {
@@ -97,7 +83,7 @@ public interface IGreetingWorkflow
 Notes:
 
 * We use regular .NET JSON serialization (see data conversion later)
-* Workflow attributes, and in the future workflow runtime features, are in the `Temporal.Workflow` namespace
+* Workflow attributes, and in the future workflow runtime features, are in the `Temporalio.Workflow` namespace
 * We only support workflow interfaces right now because we don't have a workflow implementation yet
   * For now, everything non-static must have some form of `[Workflow` attribute, and not have a default implementation
 * Every workflow interface must have the non-inheritable `Workflow` attribute
@@ -120,7 +106,7 @@ Notes:
 * Users are encouraged to set a `static` `readonly` `Ref` property for use by clients
   * `Refs<T>.Instance` is how it is obtained
     * Also will do validation which will be shared in the future by the workflow worker
-    * `Refs` is in `Temporal` not `Temporal.Workflow`
+    * `Refs` is in `Temporalio` not `Temporalio.Workflow`
   * This is backed by a dynamic proxy
   * Really old versions of .NET (i.e. would be rarely used by users, < C# 8), don't allow a `static readonly` property on the
     interface and therefore they have to make their `Ref` elsewhere
@@ -132,12 +118,12 @@ Common static definition:
 ```csharp
 namespace MyNamespace;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public static class GreetingActivities
 {
     [Activity(Name = "create_greeting_activity")]
-    public static async Task<string> CreateGreeting(GreetingInfo info)
+    public static async Task<string> CreateGreetingAsync(GreetingInfo info)
     {
         return $"{info.Salutation}, {info.Name}!";
     }
@@ -149,7 +135,7 @@ Common instance definition:
 ```csharp
 namespace MyNamespace;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public class GreetingActivities
 {
@@ -190,7 +176,7 @@ Notes:
 The activity context looks like this:
 
 ```csharp
-namespace Temporal.Activity;
+namespace Temporalio.Activity;
 
 public class ActivityContext
 {
@@ -229,16 +215,16 @@ Notes:
 
 ### Async Completion
 
-Simply `throw new Temporal.Activity.CompleteAsyncException` from the activity.
+Simply `throw new Temporalio.Activity.CompleteAsyncException` from the activity.
 
 ## Client
 
-Inside the `Temporal.Client` namespace, we have the following types:
+Inside the `Temporalio.Client` namespace, we have the following types:
 
 * `TemporalClient` - namespace-specific, stateless high-level client to Temporal
   * See later for most of its API design
   * Accepts `ITemporalConnection`
-  * Implements `ITemporalClient` for DI use (which itself implements `Temporal.Worker.IWorkerClient` for worker use)
+  * Implements `ITemporalClient` for DI use (which itself implements `Temporalio.Worker.IWorkerClient` for worker use)
 * `TemporalConnection` - namespace-agnostic, stateful low-level client to Temporal
   * Has `WorkflowService`, `OperatorService`, etc for making raw gRPC calls
   * Supports lazy connectivity
@@ -249,7 +235,7 @@ Inside the `Temporal.Client` namespace, we have the following types:
 Example of starting the workflow defined earlier in document, signalling it, then waiting on its results:
 
 ```csharp
-using Temporal.Client;
+using Temporalio.Client;
 using MyNamespace;
 
 // Connect
@@ -269,7 +255,7 @@ Console.WriteLine(await handle.GetResultAsync());
 Example of doing the same via already started workflow:
 
 ```csharp
-using Temporal.Client;
+using Temporalio.Client;
 using MyNamespace;
 
 // Connect
@@ -293,7 +279,7 @@ user would rather not provide the type until fetching the result. See the defini
 Looks like:
 
 ```csharp
-namespace Temporal.Client;
+namespace Temporalio.Client;
 
 public class TemporalClient : ITemporalClient
 {
@@ -318,7 +304,13 @@ public class TemporalClient : ITemporalClient
     public async Task<WorkflowHandle<TResult>> StartWorkflowAsync<T, TResult>(
         Func<T, Task<TResult>> workflow, T arg, StartWorkflowOptions options) { /*...*/ }
 
-    public async Task<WorkflowHandle<Nothing>> StartWorkflowAsync(
+    public async Task<WorkflowHandle> StartWorkflowAsync(
+        Func<Task> workflow, StartWorkflowOptions options) { /*...*/ }
+
+    public async Task<WorkflowHandle> StartWorkflowAsync<T>(
+        Func<T, Task> workflow, T arg, StartWorkflowOptions options) { /*...*/ }
+
+    public async Task<WorkflowHandle> StartWorkflowAsync(
         string workflow, object[] args, StartWorkflowOptions options) { /*...*/ }
 
     public async Task<WorkflowHandle<TResult>> StartWorkflowAsync<TResult>(
@@ -326,7 +318,7 @@ public class TemporalClient : ITemporalClient
 
     // GetWorkflowHandle
 
-    public WorkflowHandle<Nothing> GetWorkflowHandle(
+    public WorkflowHandle GetWorkflowHandle(
         string workflowID, string? runID = null, string? firstExecutionRunID = null) { /*...*/ }
 
     public WorkflowHandle<TResult> GetWorkflowHandle<TResult>(
@@ -342,7 +334,7 @@ public class TemporalClient : ITemporalClient
 
     // ListWorkflowsAsync
 
-    public WorkflowExecutionEnumerator ListWorkflows(
+    public IAsyncEnumerator<WorkflowExecution> ListWorkflows(
         string query,
         ListWorkflowsOptions? options = null,
     ) { /*...*/ }
@@ -353,10 +345,12 @@ public class TemporalClient : ITemporalClient
 
 Notes:
 
+* There is no lazy connection/client like there is in other languages
+  * Originally the code had a `Lazy` connection option, but the team decided they wanted it removed
 * There will be `ExecuteWorkflowAsync` helper methods for each `StartWorkflowAsync` overload there
 * It was decided forcing an options object is easier than lots of parameters for start workflow
-* Notice that there is a string-based workflow `StartWorkflowAsync` call that accepts a return type and one that doesn't
-  and returns a handle for `Nothing`. The user can specify a return type on `WorkflowHandle.GetResultAsync` if they
+* Notice that there is a string-based workflow `StartWorkflowAsync` call that accepts a return type and one that
+  doesn't. The user can specify a return type on `WorkflowHandle.GetResultAsync` if they
   want instead of at start time (or just never altogether if the return doesn't matter to the user).
 * For string-based calls to `StartWorkflowAsync`, we are accepting all args at once via `object[]`. But since array
   types are covariant in .NET, this means you could have a workflow that accepts `string[]` and forget that needs to
@@ -374,7 +368,7 @@ Notes:
   * We could have taken a proxy approach, but that would have made the code look like
     `client.GetWorkflowHandle<GreetingWorkflow>("my wf id", wf => wf.RunAsync)` which is not much better and is hard to
     get right with all of the overloads so you can have the lambda as the last param. This is what Azure Durable
-    Entities does.
+    Entities does. There is discussion on this later.
 * Cancellation tokens (and RPC metadata and RPC timeout and such) will be in the call options, not as the last parameter
   in a bunch of overloads like many other .NET methods
 * Older versions of .NET don't have `IAsyncEnumerator` and we don't want to go out of our way to provide a special
@@ -382,27 +376,27 @@ Notes:
   * Older versions can just use raw gRPC or not have this feature. These older .NET versions are really old and we don't
     expect hardly any users on them.
   * Note, `ListWorkflows` is not `ListWorkflowsAsync` because the first page is lazily fetched on first iteration
-  * `WorkflowExecutionEnumerator` has a `MapHistories` on it that returns `IAsyncEnumerator<WorkflowHistory>` (or is an
-    extension method)
+  * We need an easy way to map this to histories. Maybe a `GetHandle` extension method on the execution (that accepts a
+    client), and a `SelectHistories` extension method on `IAsyncEnumerator<WorkflowExecution>` that accepts a client.
 
 ### WorkflowHandle
 
 Looks like:
 
 ```csharp
-namespace Temporal.Client;
+namespace Temporalio.Client;
 
-public class WorkflowHandle<TResult>
+public class WorkflowHandle
 {
     public string ID { get; }
     public string? RunID { get; }
     public string? ResultRunID { get; }
     public string? FirstExecutionRunID { get; }
 
-    public async Task<TResult> GetResultAsync(
+    public async Task GetResultAsync(
         bool followRuns = true, RPCOptions? options = null) { /*...*/ }
 
-    public async Task<TResultLocal> GetResultAsync<TResultLocal>(
+    public async Task<TResult> GetResultAsync<TResult>(
         bool followRuns = true, RPCOptions? options = null) { /*...*/ }
 
     public async Task SignalAsync(string signal, object[] args, RPCOptions? options = null) { /*...*/ }
@@ -442,6 +436,12 @@ public class WorkflowHandle<TResult>
 
 #endif
 }
+
+public class WorkflowHandle<TResult> : WorkflowHandle
+{
+    public new async Task<TResult> GetResultAsync(
+        bool followRuns = true, RPCOptions? options = null) { /*...*/ }
+}
 ```
 
 Notes:
@@ -462,8 +462,8 @@ Notes:
 Usage:
 
 ```csharp
-using Temporal.Client;
-using Temporal.Worker;
+using Temporalio.Client;
+using Temporalio.Worker;
 using MyNamespace;
 
 // Connect
@@ -486,9 +486,9 @@ await worker.ExecuteAsync(cts.Token);
 Looks like:
 
 ```csharp
-namespace Temporal.Worker;
+namespace Temporalio.Worker;
 
-using Temporal.Activity;
+using Temporalio.Activity;
 
 public class TemporalWorkerOptions
 {
@@ -551,7 +551,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Worker;
+namespace Temporalio.Worker;
 
 class WorkflowReplayer
 {
@@ -580,7 +580,7 @@ Notes:
 Looks like:
 
 ```csharp
-namespace Temporal.Converters;
+namespace Temporalio.Converters;
 
 public record DataConverter(
     IPayloadConverter PayloadConverter,
@@ -597,18 +597,18 @@ public record DataConverter(
 
 public interface IPayloadConverter
 {
-    Temporal.Api.Common.V1.Payload ToPayload(object value);
+    Temporalio.Api.Common.V1.Payload ToPayload(object value);
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload);
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload);
 }
 
 public interface IEncodingPayloadConverter
 {
     string Encoding { get; }
 
-    bool TryToPayload(object value, out Temporal.Api.Common.V1.Payload payload);
+    bool TryToPayload(object value, out Temporalio.Api.Common.V1.Payload payload);
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload);
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload);
 }
 
 public record PayloadConverter(IEnumerable<IEncodingConverter> EncodingConverters) : IPayloadConverter
@@ -620,9 +620,9 @@ public record PayloadConverter(IEnumerable<IEncodingConverter> EncodingConverter
         new BinaryProtoConverter(),
         new JsonPlainConverter() }) { /*...*/ }
 
-    public Temporal.Api.Common.V1.Payload ToPayload(object value) { /*...*/ }
+    public Temporalio.Api.Common.V1.Payload ToPayload(object value) { /*...*/ }
 
-    object? ToValue(Type type, Temporal.Api.Common.V1.Payload payload) { /*...*/ }
+    object? ToValue(Type type, Temporalio.Api.Common.V1.Payload payload) { /*...*/ }
 }
 
 public class BinaryNullConverter : IEncodingPayloadConverter { /*...*/ }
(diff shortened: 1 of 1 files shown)


dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1101 @@
...
+```
+
+Cons:
+
+* Every method must be virtual or on an interface
+  * Cannot proxy regular methods which is a bit off-putting. This works for Azure because they make you opt-in to the
+    typed approach
+    [by forcing interfaces](https://learn.microsoft.com/en-us/azure/azure-functions/durable/durable-functions-dotnet-entities#accessing-entities-through-interfaces) which we don't want to do
+  * Cannot proxy static methods
+    * We should support static activities, so you'd have method references anyways probably
```
nickwilliams-codynamic, line 1024: Forgive my ignorance, but what does proxy mean in this context? I thought all Temporal might need to do with these functions is run something before and/or after innovation, is there something more than that?

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1101 @@
...
+    * Accept both single-arg and multi-arg w/ "unset" defaults, but that is confusing and most .NET devs use their
+      positional parameters unnamed
+    * Have separate overloads for no-arg, 1-arg, and multi-arg but that's not any clearer
+* For `GetWorkflowHandle`, we cannot accept a workflow reference because in .NET if you don't type all type arguments
+  (i.e. including that parameter type), you have to help the type inferencer out. So since they'd have to provide a type
+  argument anyways, we make it where they either can here or wait until `GetResultAsync` to do so.
+  * We could have taken a proxy approach, but that would have made the code look like
+    `client.GetWorkflowHandle<GreetingWorkflow>("my wf id", wf => wf.RunAsync)` which is not much better and is hard to
+    get right with all of the overloads so you can have the lambda as the last param. This is what Azure Durable
+    Entities does. There is discussion on this later.
```
nickwilliams-codynamic, line 371: I know there is a desire to not force interfaces but I think this all becomes much cleaner if there is an interface something like

```
interface IWorkflow<TResult>
{
   Task RunAsync();
}
```

Would this then allow you to just have the below?
```
WorkflowHandle<MyResultType> handle = client.GetWorkflowHandle<MyWorkflow>("workflow-id-here");
```

Personally I don't see the problem in having to implement an interface for my workflow. It feels quite dotnet like to me, and seems to produce a much cleaner public API IMO.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1101 @@
...
+* We support the https://learn.microsoft.com/en-us/dotnet/core/extensions/options pattern with all of our `Options`
+  classes
+  * Some of the options things like activities method references can't by loaded from a file, so we should take a look
+    at helpers for converting activity strings to activity classes reflectively
+  * We need a sample showing configuring workers via config file akin to Spring Boot
+* See the host section for how it all may come together inside a worker service
+
+### Logging
+
+https://learn.microsoft.com/en-us/dotnet/core/extensions/logging
```
nickwilliams-codynamic, line 942: Speaking purely from an consumer point of view I think it would be a poor decision not to include this as the default logging implementation. Microsoft.Extensions.Logging is the de-facto logging mechanism in dotnet and has been for a while now and its interface seems to be very stable.

I think the vast majority of consumers would end up creating their own `DotnetTemporalLogger` implementation which all just bridge this gap. Without understanding the how ITemporalLogger would work you'd lose some of the freebies you get from using the standard `ILogger<>` implementation (like automatic class names, parameterized logs, etc.)

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1101 @@
...
+
+### HostBuilder and Worker Services
+
+https://learn.microsoft.com/en-us/dotnet/core/extensions/generic-host and
+https://learn.microsoft.com/en-us/dotnet/core/extensions/workers
+
+* Same discussion around dependencies as logging, this would need `Microsoft.Extensions.Hosting`
+* Our `Temporalio.Worker.TemporalWorker` is not a subclass of `BackgroundService` to avoid the dependency
+  * `TemporalWorker.ExecuteAsync` matches `BackgroundService.ExecuteAsync` very intentionally, so this is incredibly
+    easy for a user to create and we will of course have a sample
```
nickwilliams-codynamic, line 966: I think this one is more understandable than the logging one. I would wonder if it would be worth including in some sort of hosting extensions NuGet package though?

I've seen even reasonably experienced dotnet developers really struggle to understand how to:
- Properly handle graceful shutdowns
- Actively make decisions about how to handle exceptions in the service itself
- Sub-optimal blocking with `while(true)` type loops

If we we're to include a sane BackgroundService implementation somewhere I think it would go a long way to supporting the `Be as simple as possible from a caller POV` and `Be reasonably flexible with reasonable defaults` goals

nickwilliams-codynamic:
This looks really good @cretz. I've added a couple of comments, a couple duplicate the conversations from #70 so feel free to close them off if these decisions have been made already.

The only thing I'm struggling with is the `Ref` pattern, it feels unnatural to me... that being said I don't really understand the proxying requirements which the SDK has. Do you know if there is anywhere I can go and read up about that? I had a look at the Typescript SDK and it seems that everything is just a reference to a function. Which doesn't seem too much difference to having references to a dotnet `Func<>` on the face of it.

dotnet/sdk-phase-1.md
```diff
@@ -0,0 +1,1101 @@
...
+    at helpers for converting activity strings to activity classes reflectively
+  * We need a sample showing configuring workers via config file akin to Spring Boot
+* See the host section for how it all may come together inside a worker service
+
+### Logging
+
+https://learn.microsoft.com/en-us/dotnet/core/extensions/logging
+
+* Unfortunately there is no logging in the standard library for .NET, therefore in order to support this we'd have to
+  add a dependency on https://www.nuget.org/packages/Microsoft.Extensions.Logging
```
ReubenBond, line 945: Note that your libraries should add a dependency to the `.Abstractions` package, `Microsoft.Extensions.Logging.Abstractions`.
I recommend that you do go this route, rather than creating your own abstraction, however thin.

dotnet/sdk-phase-1.md
cretz: > Forgive my ignorance, but what does proxy mean in this context?

http://www.castleproject.org/projects/dynamicproxy/

> I thought all Temporal might need to do with these functions is run something before and/or after innovation, is there something more than that?

You need to be able to reference the function. When you have a client and need to call a workflow/signal/query, or in a workflow and need to call an activity/child, you need a typed reference to them for type safety. C# doesn't have method references, so we need to reference them somehow.

> The lambda sample here looks far more idiomatic to c#

But did you know that lambda is not invoking the thing? Instead it's just using a lambda to be able to provide a method reference for use by what does schedule the work.

Here's the problem: I have a workflow/activity instance method, and I need to reference it (not call it) via a start workflow or start activity call. What is the best way to do that in C#?

dotnet/sdk-phase-1.md
cretz: > I know there is a desire to not force interfaces but I think this all becomes much cleaner if there is an interface

There is definitely this desire. None of our other SDKs require implementing an interface. Workflow methods are allowed to return value (or empty task), allowed to accept any number of arguments technically, etc.

Your interface example doesn't return `TResult`. Even if it did, how does that interface help in the definition of type safe signals and queries? We choose the attribute approach here as it's a better way to color methods than an interface ,which limits flexibility (and we'd still have to have attributes anyways for signals and queries).

Here's an exercise. Show workflows all permutations of workflow entrypoints that return no values, and accept 0 or more args. Then show how a client can invoke them in a type safe manner. Then do the same for signal and query methods.

dotnet/sdk-phase-1.md
cretz: This may have convinced me, I may go ahead and add a dependency on this package (as much as I absolutely abhor required dependencies for optional things like logging in a library like ours).

dotnet/sdk-phase-1.md
cretz: > I would wonder if it would be worth including in some sort of hosting extensions NuGet package though?

I considered that, but it'd be like a 3 line project. From a different section:

> Could have a separate `Temporalio.Extensions` project ...

cretz:
@nickwilliams-codynamic - Thanks for the review!

> The only thing I'm struggling with is the Ref pattern, it feels unnatural to me... that being said I don't really understand the proxying requirements which the SDK has. Do you know if there is anywhere I can go and read up about that? 

I gave exercises at the bottom of two of your comments above that should help here.

> I had a look at the Typescript SDK and it seems that everything is just a reference to a function. Which doesn't seem too much difference to having references to a dotnet Func<> on the face of it.

But you can't have a reference to a instance method as a `Func<>` without instance of the object in .NET unlike Java and many other languages. The only purpose of the `Ref` pattern is to make a fake instance so you can reference instance methods. You are more than welcome to instantiate a dummy version to do it. Could also use a lambda where I instantiate a fake instance and you pretend to make a single call so I know which method you want to invoke, but that is ugly IMO (ntm it only works with interface or virtual methods).

If you could show me a good way to reference a (non-virtual) instance method on a class in .NET without instantiating the class, I'd definitely welcome it. I need the method reference for proper typing. The fake instance via DynamicProxy using this `Ref` pattern is the best I could come up with.

dotnet/sdk-phase-1.md
cretz: :+1: That seems to be the consensus, so I may do this. The problem we have had in the past with dependencies like these, no matter how prevalent and stable, is people eventually complain. "Why are you making me take a dependency for logging that I don't want?"

Here's a question though: When `7.0.0` of that library drops, which version should I depend on? And when a user comes to me and says "this other library I'm using in the same app depends on the other incompatible version" what do I tell them?

ReubenBond: Now that .NET ships versions alternating between Long Term Service and Standard Term Service, you might find that certain companies have rules about always using an LTS version (even though .NET 6's patch lifecycle extends only 6 months past .NET 7's patch lifecycle). Anecdotally, this is especially true for banks. Given that, always adopting the latest LTS once the previous LTS is nearing EoL is probably a safe strategy. You can also multi-target and use conditionals in your project so that .NET 6 users pull in 6.0 and .NET 7 users pull in 7.0, but it's a hassle and the gain is dubious in the case of abstractions libraries which have been very stable.

Package mismatches are ok: specified versions are minimums by default, so you're generally safe picking lower versions since the end user can upgrade them in their own project as desired. Visual Studio now shows transitive dependencies in the package management UI, which helps with this.

dotnet/sdk-phase-1.md
nickwilliams-codynamic: I've had a go at writing up a quick attempt [in this gist](https://gist.github.com/nickwilliams-codynamic/e0a85b38a80f7d42ffd265f9096a7d5b). I think I couldn't hit the `0 or more args` requirement. I've got `0 or 1` but as you mentioned elsewhere

> Can technically add generic overloads for more parameters, but you have to stop somewhere

After going through the motions on this I end up with a very similar solution to the `Refs<>` you have. I think I slightly prefer the resulting API contract of my implementation but I know I have at least two major biases:

1. I don't like using attributes
2. I like the Temporal TypeScript SDK.

I suspect it mainly boils down to that first bias of mine, where I prefer to define things fluently than with annotations.

dotnet/sdk-phase-1.md
nickwilliams-codynamic: Cool, if the temporal worker implements some sort of StartAsync type method which has all the right handling of the cancellation token in it I think that's really good and there would be no need for a hosting package. i.e. something like:

```csharp
public Task StartAsync(CancellationToken stoppingToken)
{
    _temporalWorker.StartAsync(stoppingToken);
}
```

dotnet/sdk-phase-1.md
cretz: > Package mismatches are ok: specified versions are minimums by default, so you're generally safe picking lower versions since the end user can upgrade them in their own project as desired.

This isn't usually true with semver major version bumps. To clarify my original question just in case, I was asking about 7.x vs 6.x of that specific package, not  .NET versions.

Are you saying I should depend on https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/6.0.3, not https://www.nuget.org/packages/Microsoft.Extensions.Logging.Abstractions/7.0.0? And that the latter is compatible even though they did a major bump? I haven't checked their changelog yet.

It is usually very bad for a library to add a dependency to a package whose major version moves so frequently. I'm forcing the user to override my transitive version to upgrade.

dotnet/sdk-phase-1.md
cretz: (will review and get back, thanks!)

cretz: > I think I couldn't hit the 0 or more args requirement

That's no big deal, we can leave > 1 arg untyped currently. So long as it's technically possible on client and impl side.

> I don't like using attributes

Can you give some more justification? So many other libraries use them for decorating methods for these types of purposes but Temporal should not? We just need to mark a method as having a purpose and let it be configurable (e.g. name). We don't need to force users into runtime instantiation of several manually instantiated pieces IMO (though we can support it). Technically every ORM, JSON class, etc in C# could require runtime definition instead, but it's rough on the code reader.

> I like the Temporal TypeScript SDK.

I like it too, but I also recognize 1) top-level functions are a thing in that language and _not_ in C#, 2) function referencing with typing is much more advanced there, and 3) if they had a more appropriate way to provide function-level metadata, they'd probably use it (TypeScript is considering an impl that lets people use experimental JS decorators). In all of the languages we support that do offer class/method metadata facilities (Java, Python, PHP, and Python), all take advantage of them to do what function metadata does best - mark/color the class/function for its purpose and allow slight customization. 

> I suspect it mainly boils down to that first bias of mine, where I prefer to define things fluently than with annotations.

I definitely think we need to add technical reasoning for the bias. We want to reduce burden on our users. Having many more user-facing classes and forcing users to do more work due to such bias really needs good justification. This is a decent amount of effort here to conform to this bias. Not to mention we would be intentionally inconsistent with our other SDKs due to this bias.

> I've had a go at writing up a quick attempt in this gist

Notes (numbered for easy response):

1. Workflow inheritance is a very common thing. How can I build an abstract base workflow class with signal handling? Do I now have some `SetHandlers()` or something on my base class? C# has all of these great facilities for marking methods as virtual, overriding, requiring interface conformance, etc but we'd throw them out the window if we encouraged such lambda use everywhere (not that we can't support it of course).
2. A workflow is more than a function. Many workflows will want to store workflow-level state as class members, not define it as a lambda. You are making a workflow "entry point" the entire workflow which can be harder for users who want to properly use OO principles. This is the whole reason behind "workflow as a class" in all of our SDKs with good OO+metadata.
3. Why are we making users instantiate their workflow classes themselves? Does that code have to be deterministic like other workflow code? I can foresee many bugs caused by non-deterministic code in this "start workflow" lambda we are making them write themselves.
4. Where is the workflow name defined? We usually expect this to be the workflow class name in OO languages (though TS and Go, the languages which encourage top-level functions, we use function name).
5. Why force users to instantiate a workflow definition? Is there any value in this short of the limitation of this pattern? Can you clarify where you think it's best for this workflow definition instance to live considering many clients may use it? If it's just a static readonly property on the class itself, why make these extra hoops for users?
6. In OO languages, we prefer signal and query handlers to be methods instead of lambdas. It is clearer to a reader to see the declaration of a signal as a method instead of buried in the workflow. It also encourages general whole-workflow handling of these items instead of conditionally adding handlers at different times which is hard to reason about, can be error prone (signals get buffered which has surprising behavior for some, queries working and not working depending on how far in the workflow you are, signal with start, etc). Though we likely will _also_ offer manual signal/query handler addition.
7. It is much easier to use the name of a construct in the language as a signal name (i.e. the method name) than require a user hand type it every time they declare it. All of our SDKs with reflection capabilities use them by default to save users from doing something we can do for them.
8. Can I define an interface with no workflow body that defines its entry point, signals, and queries for client use in a typed way? Many people separate their definitions and don't include the workflow body in the code/library shared with clients.
9. Reading a workflow with annotations, I can easily tell what the entry point is, what signals/updates it accepts, etc. But I cannot here. It is important for the user to know they are under "workflow restrictions" (e.g. determinism) and so it is important for a reader to know it's workflow code not normal code. The example given, I wouldn't even know it's for Temporal much less clearly a Temporal workflow that has severe determinism restrictions.
10. One of the future hopes is to have a workflow sandbox. Granted .NET has done away with code access security, app domains, etc but there are other possible approaches. Having a workflow as a class and not as a lambda/instance is an important step in being able to provide this boundary. A fully qualified class name is serializable across a boundary in a way that an instance may not be in some cases.
11. Is there any way to enforce client-side, even at runtime, that a signal belongs to a workflow? (I don't think we plan on doing this, just wanted to get thoughts here)
12. I assume you would expect activities to be defined in this same way? (i.e. require the user to manually instantiate an "activity definition")

Attributes aside, the difference in the two proposals seems to be "workflow lambdas and signal/query lambdas" vs "workflow classes and signal/query methods". I think C# has much more composability at the interface/class/method level. Only languages without good OO support do we use top-level functions. With the approach from the gist I imagine we'd just end up saying, "we know we only accept `Func`s everywhere, but we encourage classes/interfaces/methods" anyways. They are much more composable as programs grow.

There is an alternative where we can have both abstractions and let attributes sit on top of this more manual approach. I will have to convince team having two ways of defining workflows, signals, queries, etc is valuable here. The downside is now I have to have 4 more overloads to every start, signal, query, etc call just due to fear of attributes. Of course most people will probably just use attributes anyways as that's how they prefer to put metadata on classes and methods in C# in my experience. I will discuss with team.

The primary benefit of your approach is you don't have to have "instance method references" using that "ref" since now all definitions are objects hand-made by users. I am not yet convinced this is easier for developers to have to separate the definitions and implementations of everything they build, but maybe it is. Or maybe we can do both. I will discuss with team.

Thanks for the suggestion! (sorry for long comment!)

dotnet/sdk-phase-1.md
cretz: This is the helper:

```csharp
public class WorkerBackgroundService : Microsoft.Extensions.Hosting.BackgroundService
{
  private readonly Worker worker;
  public WorkerBackgroundService(Worker worker) => this.worker = worker;
  protected override Task ExecuteAsync(CancellationToken stoppingToken) => worker.ExecuteAsync(stoppingToken);
}
```

dotnet/sdk-phase-1.md
nickwilliams-codynamic: Thanks very much for the detailed response @cretz! Reading back my initial comment I realize I could have been a lot clearer about some things, so apologies about that, hopefully I can clear up some of those bits:

# Attributes
> >   I don't like using attributes
>Can you give some more justification?

^ No, I should have been more explicit that my preferences are probably unjustified and not support by most of the dotnet community. I was meaning that I know I have this bias which probably impacts why I was looking for an alternative and probably means my opinion on the original is not particularly valid.

There are a few use cases where I've seen most companies move away from attributes, namely:
**[FluentValidation](https://docs.fluentvalidation.net/en/latest/) > Data Annotations** - Provides clearer syntax for complex rules, much easier to test for complex rules, particularly if you have polymorphic models being validated.
**Entity Framework Fluent** > **EF Attibutes** - There are a bunch of options not available via attibutes (like `WillCascadeOnDelete`, `FK mapping`)
**[Minimal API](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/minimal-apis/overview?view=aspnetcore-7.0) > Controllers** - Less ceremony / boiler plate and enables better code organization - if you like to organize your code by domain.

In all of these examples there are specific reasons why I don't particularly like the attribute equivalents. I can't think of anything the temporal SDK would lose in just using attributes and I don't think it would be worth supporting both methods. I think the only thing we'd need to ensure is that anything which could be decided at runtime is able to be, thought we may allow defaults set in an attribute. The reason I mention this is [Hangfire.NET](https://www.hangfire.io/) (a popular .NET background/scheduled worker) has a [lot of problems with this](https://github.com/HangfireIO/Hangfire/issues/680). That's more to do with poorly designed attributes than attributes in general though tbh.

# Questions

In general I think you've sold me on the approach you suggested in your PR. I've provided some answers to your questions below just in case there is anything useful it comes out of them. Please don't feel that you need to respond to them if there is nothing useful to be gained.

## 1
> Do I now have some SetHandlers() or something on my base class? 

I think that's what I would do, maybe I'd go with `ConfigureSignals` and `ConfigureQueries`, [which is a similar concept to EF's `OnModelCreating`](https://learn.microsoft.com/en-us/ef/core/modeling/#use-fluent-api-to-configure-a-model). Providing some hook to allow a fluent configuration.

## 2
> A workflow is more than a function. Many workflows will want to store workflow-level state as class members, not define it as a lambda.

Agreed, in the gist I instantiate a class which represents my workflow and can use instance methods and fields as part of it. By only allowing a lambda to be passed into the temporal SDK I was trying to make temporal ignorant of how my workflow is kicked off. I might instantiate a class, call a static method, or just leave it as an anonymous lambda.

## 3
> Why are we making users instantiate their workflow classes themselves? 

How would temporal know how to instantiate their workflow classes? And how would it actually do that?
- Does the class have to have a parameter-less constructor? 
  - How is that enforced? Compile-time or runtime?

> Does that code have to be deterministic like other workflow code?

Yes, it would have to be deterministic.  

In the model I suggest if we wanted to have better support for instantiating the class we could have an overload of the Create methods which instantiates the class for you and forces are parameter-less constructor at compile time, something like

```csharp
public static WorkflowDefinition Create<TWorkflow>(Func<TWorkflow, Task> action)
   where TWorkflow : new() // Require a parameterless constructor
{
    return new Empty(() => 
    {
        var workflow = new TWorkflow();
        return action(workflow);
    });
}

// x here is an instance of `MyWorkflowClass`
var workflowDefinition = WorkflowDefinition.Create<MyWorkflowClass>(x => x.StartWorkflow());
```
In any case, whatever code you have in the `MyWorkflowClass` constructor would have to be deterministic - regardless of where the instantiation occurs.

## 4
> Where is the workflow name defined? We usually expect this to be the workflow class name in OO languages (though TS and Go, the languages which encourage top-level functions, we use function name).

Yep, I missed this in the examples. It would have to be a user-defined parameter to the WorkflowDefinition. Again, I suspect I have some baseless preferences here. I prefer to explicitly declare these to having them automatically generated and have to refer to the docs to find the naming logic. But I suspect there are few who would agree with me on this one.

## 5
See answer to 3.

## 6
>In OO languages, we prefer signal and query handlers to be methods instead of lambdas.

That's fair. I've only ever used the go and TypeScript SDKs so I am probably fighting some language features to get something which looks closer to what I'm personally used to.

>It also encourages general whole-workflow handling of these items instead of conditionally adding handlers at different times which is hard to reason about, can be error prone

I completely agree with this.

> Though we likely will also offer manual signal/query handler addition.

This feels like it would be a good nice to have. It's a very minor thing but we use the Temporal UI for some debugging type queries and it's nice they only appear once it's possible to execute them. That being said having their implementation return some 'not ready yet' response isn't a problem.

## 7
> It is much easier to use the name of a construct in the language as a signal name (i.e. the method name) than require a user hand type it every time they declare it. All of our SDKs with reflection capabilities use them by default to save users from doing something we can do for them.

Fair enough. I tend to prefer declaring them myself regardless, but I suspect I'm in the minority.

## 8
> Can I define an interface with no workflow body that defines its entry point, signals, and queries for client use in a typed way? Many people separate their definitions and don't include the workflow body in the code/library shared with clients.

I don't see why not, that being said it does start to look a lot like your originally proposed `Refs` object. It's just all hand crafted rather than auto-generated (so worse I suppose, albeit with less magic). I don't think you'd need to declare an interface, just some empty lambda.

```csharp
public static class MyWorkflowRefs
{
    public static IWorkflowDefinition Workflow { get; } = WorkflowDefinition.Create("MyWorkflowType", () => { return Task.CompletedTask; });
    public static SignalDefinition<string> UnblockSignalWithMessage { get; } = new SignalDefinition<string>("unblock-with-message");
    public static QueryDefinition<bool> IsBlockedQuery { get; } = new QueryDefinition<bool>("isBlocked");
    public static QueryDefinition<string, bool> IsBlockedQueryWithArgs { get; } = new QueryDefinition<string, bool>("isBlocked-with-args");
}
```

## 9
> Reading a workflow with annotations, I can easily tell what the entry point is, what signals/updates it accepts, etc. But I cannot here. It is important for the user to know they are under "workflow restrictions"

This is a really good point and a big win with your approach. I was actively trying to make everything look more like 'normal code' and I was thinking from a code organization point of view this would live in some `.Workflows.` namespace so I'd know. I think what you've described here is a big downside of my approach, it creates more room for users to shoot themselves in the foot.

## 10
> A fully qualified class name is serializable across a boundary in a way that an instance may not be in some cases.

True, I think the class becoming fully serializable will add a lot of restrictions in and off itself, but you're right it would be far, far easier to do this if you have an instance of a class, I hadn't considered this.

## 11
> Is there any way to enforce client-side, even at runtime, that a signal belongs to a workflow? If

If everything in my gist became had a generic parameter of the workflow type I think you could get someway there.
```csharp
interface IWorkflowDefinition<in TWorkflow>
interface ISignalDefinition<in TWorkflow> {}
interface IQueryDefinition<in TWorkflow, TResult> {}

public interface IWorkflowHandle<in TWorkflow>
{
    public Task SignalAsync(ISignalDefinition<TWorkflow> signalDefinition, CancellationToken ct = default);
    public Task SignalAsync<TArgs>(ISignalDefinition<TWorkflow, TArgs> signalDefinition, TArgs args, CancellationToken ct = default);
    public Task<TResult> QueryAsync<TResult>(IQueryDefinition<TWorkflow, TResult> queryDefinition, CancellationToken ct = default);
    Task CancelAsync(CancellationToken ct = default);
}
```

I think this would ensure that the various signals / queries **should** belong to the workflow in question at compile time. But there would be nothing stopping you from creating a signal definition of the correct types which aren't wired up in the workflow. 

I suppose it could provide some type help for users who have a few inherited workflows and have some signals which are only particular to certain children, so you could have

```
SignalDefinition<MyBaseWorkflow>("allow-on-all-workflows");
SignalDefinition<MyDerviedWorkflow1>("allow-on-1");
SignalDefinition<MyDerviedWorkflow2>("allow-on-2");
```

So if you have a `IWorkflowHandle<MyDerivedWorkflow2>` you'd get a compile error if you tried to use `allow-on-1`. Which might be useful depending on where you're exposing your definitions.

## 12
> Can we, if we take this approach, just have overloads of the same WorkflowDefinition class w/ different results/argument like you did with signal there instead of class names like WithArgsAndResult.

I'll have another go to double check. That was the road I started on but I couldn't find a quick way to get the compiler to differentiate between `WorkflowDefinition<TArgs>` and `WorkflowDefinition<TResult>` without forcing some marker interface like `IWorkflowArgs` and `IWorkflowResult` :(

cretz: > Attributes

My primary issue is that these attributes are just metadata and improve code readability. Encouraging lambdas instead of requiring classes and forcing manual handler setting instead of methods is going backwards I believe. But I need to research all options. The key is to make sure code is easy to read and clarity is 

> I don't think it would be worth supporting both methods

It may be worth it, but yeah it's several overloads on the client side for each way. I do think if we support lambda entrypoints or other non-class-based workflows, we can add them later.

> I think the only thing we'd need to ensure is that anything which could be decided at runtime is able to be, thought we may allow defaults set in an attribute.

This is kinda what we're trying to avoid at least with workflows and activities. Technically the internals of all Temporal SDKs could never register a workflow at all and handle every raw workflow request at runtime. But we intentionally, for user safety, require _explicitly named_ workflows to be pre-registered, and we check their types and do other sanity checking things. There's no _technical_ reason we couldn't a full blown "workflow router" in code that a user could set, but we intentionally limit this flexibility due to pain it has caused people (though we may offer "dynamic workflows/activities" like you see in Java). We've found it is much easier to add this flexibility later than to try to limit it later.

With signal and query handlers however, there is a good use case for dynamic handler setting inside the workflow, so we'll likely support both attribute and non-attribute approaches, though the former is strongly encouraged for code readability and all of the method-level features C# provides (e.g. inheritance).

> Do I now have some SetHandlers() or something on my base class?

> > I think that's what I would do, maybe I'd go with ConfigureSignals and ConfigureQueries, which is a similar concept to EF's OnModelCreating. Providing some hook to allow a fluent configuration.

Why have a user go through all that when C# clearly defines a model for inheritance of callable things? At the very least by default we should encourage use of that. We have seen where we accept function references in other SDKs, people tend to favor lambdas over well structured methods, so we use method-metadata approaches and our samples to strongly encourage these things as methods.

> By only allowing a lambda to be passed into the temporal SDK I was trying to make temporal ignorant of how my workflow is kicked off. I might instantiate a class, call a static method, or just leave it as an anonymous lambda.

We don't accept lambdas in our SDKs for workflows for a reason. People use them then regret using them because they also need to reference them client side or they need to expand into state or whatever. We have explicitly chosen the less-flexible option for workflow definition. Workflows are a bit special in that it's heavy struggle to alter them after they're written. You can't just refactor in many cases. Also, we need workflows to not mutate external state or access mutable external state, something that a lambda doesn't do a good job of conveying. This class-based approach can also let us sandbox the entire workflow if we find a good sandboxing approach by just using the fully qualified class name to start the workflow in a sandbox. A lambda is often too dependent on its environment (or at least encourages that we support such a thing) which is a very very bad thing for workflows.

> How would temporal know how to instantiate their workflow classes? And how would it actually do that? Does the class have to have a parameter-less constructor? How is that enforced? Compile-time or runtime?

Yes, a parameterless constructor that is enforced at runtime. Specifically at workflow registration time though, not workflow invocation time, so still quite eagerly. I will probably also add an option to support _also_ accepting the workflow arguments in the constructor in addition to just the workflow entrypoint to help with those creating init-only/readonly properties off of them. But they'll have to opt-in to that.

> In the model I suggest if we wanted to have better support for instantiating the class we could have an overload of the Create methods which instantiates the class for you

If we support such a model, I think it'd be the only approach (for sandboxing and other reasons mentioned above) instead of an overload. But at that point it becomes silly to have them create a workflow definition at all right? Once we have resolved that a workflow must be a class, can we not say a workflow entry point must be a method? And once we have decided that, why not let clients calling the workflow just use the class instead of forcing instantiation of the definition manually?

> Where is the workflow name defined?
> > I prefer to explicitly declare these to having them automatically generated and have to refer to the docs to find the naming logic.

The vast majority of people would just type something like the class name. We're just making things tough on people unnecessarily. If we had some complicated naming logic, maybe, but it's just unqualified class name. And the vast majority of people don't care about naming logic anyways because they use the same types to instantiate the workflow as they do to define it, so there are no string literals anywhere (much less forced on them) which is nice.

> That's fair. I've only ever used the go and TypeScript SDKs

Our only two non-OO language SDKs. Neither have good class support with method metadata.

> I don't think you'd need to declare an interface, just some empty lambda.

Here's the use case our users have a lot: "I want to have an interface for my workflow that a client can use to invoke it and that an implementer must use to implement it". All of this manual definition of workflows instead of deriving from classes/interfaces really hurts the contracts the language gives us to use for better code.

> A fully qualified class name is serializable across a boundary in a way that an instance may not be in some cases.

> > True, I think the class becoming fully serializable will add a lot of restrictions in and off itself, but you're right it would be far, far easier to do this if you have an instance of a class, I hadn't considered this.

To clarify here, it's the _class name_ I need serializable, not the class itself. I would instantiate the class in the sandbox for each workflow run. The class instance itself does not need to be serializable.

> Is there any way to enforce client-side, even at runtime, that a signal belongs to a workflow?

> > I think this would ensure that the various signals / queries should belong to the workflow in question at compile time.

But it only kinda does. What it actually ensures is that the author defined them on there at the same time. It doesn't say that the workflow has a handler for them. It doesn't say that the author didn't miss a definition that they added a handler for. Languages like C# give us the ability to add named callable constructs as part of a larger stateful construct. Workflows need named callable constructs as part of larger stateful constructs. It'd be silly to not use the language facilities.

If C# generics were more powerful (like Java's kinda are and that definitely Python is), I could enforce a "this" type on method references at compile time. But since I can't, at least I can do it at client invocation time if I want (which I may not, undecided atm).

> That was the road I started on but I couldn't find a quick way to get the compiler to differentiate between `WorkflowDefinition<TArgs>` and `WorkflowDefinition<TResult>` without forcing some marker interface like IWorkflowArgs and IWorkflowResult :(

Yup, what you'd end up having I think is:

```csharp

class WorkflowDefinition<TResult> where TResult : Task
{
  private protected readonly MethodInfo method;

  private protected WorkflowDefinition(MethodInfo method)
  {
    this.method = method;
  }

  public WorkflowDefinition(Func<TResult> action) : base(action.GetMethodInfo()) { }
}

class WorkflowDefinition<TArg, TResult> : WorkflowDefinition<TResult> where TResult : Task
{
  public WorkflowDefinition(Func<TArg, TResult> action) : base(action.GetMethodInfo()) { }
}
```

That way a `TResult` can represent a `Task` or a `Task<SomeType>`.

> See answer to 3.

I think it's worth specifically addressing "Can you clarify where you think it's best for this workflow definition instance to live considering many clients may use it? If it's just a static readonly property on the class itself, why make these extra hoops for users?". We would have to encourage a best practice of putting these definitions somewhere. And once you follow this to its logical conclusion, you'll end up with a best practice like:

```csharp
class MyWorkflow
{
  public static WorkflowDefinition.WithArgsAndResult<string, string> Definition { get; } =
    WorkflowDefinition.WithArgsAndResult("MyWorkflow", arg => new MyWorkflow().Run(arg));

  public static SignalDefinition<string> MySignal { get; } = new SignalDefinition<string>("MySignal");

  public Task<string> RunAsync(string arg)
  {
    MySignal.SetHandler(arg => Task.CompletedTask);
    throw new NotImplementedException();
  }
}
```

Because we don't want to encourage lambdas or static entry points since we want to encourage workflow-instance state instead of local vars or static state. So once we arrive at this best practice, we say to ourselves "can we reduce this boilerplate and retain type safety". As you start whittling that down thinking:

* Would be nice if you didn't have to type name but could if you wanted to
* Would be nice not to have to manually put type args in so many places
* Would be nice to make sure I don't forget to set the handler
* Would be nice to support inheritance for those signals and such
* Would be nice not to have to instantiate these definitions every time since they're always just doing the same thing over and over

Etc. I think that's what my current proposal does. Changes it do:

```csharp
[Workflow]
class MyWorkflow
{
  public static readonly MyWorkflow Ref = new();

  [WorkflowRun]
  public Task<string> RunAsync(string arg)
  {
    throw new NotImplementedException();
  }

  [WorkflowSignal]
  public async Task MySignalAsync(string arg)
  {
  }
}
```

I think this is a significant readability improvement. Doubly typing all the workflow types, handler types, etc is a bit much. I would love to come up with other solutions for the `Ref` problem where I need to be able to reference a single instance method outside the class in C#. Note here I changed from using `Refs` to just `new()` because there is no need for proxying a class, only an interface or abstract class.

awright415:
> If you could show me a good way to reference a (non-virtual) instance method on a class in .NET without instantiating the class, I'd definitely welcome it. I need the method reference for proper typing. The fake instance via DynamicProxy using this `Ref` pattern is the best I could come up with.

Could an open delegate be useful here? Borrowing an example from https://blog.slaks.net/2011/06/open-delegates-vs-closed-delegates.html
```
Func<string, string> open = (Func<string, string>)
    Delegate.CreateDelegate(
        typeof(Func<string, string>),
        typeof(string).GetMethod("ToUpperInvariant")
    );
```

cretz:
> Could an open delegate be useful here? Borrowing an example from https://blog.slaks.net/2011/06/open-delegates-vs-closed-delegates.html

> ```
> Func<string, string> open = (Func<string, string>)
>     Delegate.CreateDelegate(
>         typeof(Func<string, string>),
>         typeof(string).GetMethod("ToUpperInvariant")
>     );
> ```

@awright415 - The concern is not that I can't get an instance method from a class, it's how the user provides it to the client. For example, you have `StartWorkflow`. That needs to accept a workflow entry point method so it can derive the workflow name, input type, and return type and generically return it. Say you have these two workflows:

```csharp
[Workflow]
public static class MyStaticWorkflow
{
    [WorkflowRun]
    public static async Task<string> RunAsync(string arg)
    {
        throw new NotImplementedException();
    }
}

[Workflow]
public class MyInstanceWorkflow
{
    [WorkflowRun]
    public async Task<string> RunAsync(string arg)
    {
        throw new NotImplementedException();
    }
}
```

Now if you wanted to "execute" the former (start + result), no problem, you can reference that method and everything:

```csharp
var result = await myClient.ExecuteWorkflowAsync(
    MyStaticWorkflow.RunAsync, "some arg", myOptions);
```

But how do you do it with an instance method? Yes, you can instantiate:

```csharp
var instance = new MyInstanceWorkflow();
var result = await myClient.ExecuteWorkflowAsync(
    instance.RunAsync, "some arg", myOptions);
```

But not only is instantiating something only for its method reference annoying, but constructors can do things for a workflow that make no sense here. Also, how do you do this for referencing an interface or an abstract class? It is perfectly normal to use an interface or abstract class on the client side to reference a workflow. So we landed at suggesting adding this to all classes/interfaces with `[WorkflowRun]`:

```csharp
public static readonly MyInstanceWorkflow Ref = Refs<MyInstanceWorkflow>.Instance;
```

For interfaces or abstract classes, that uses http://www.castleproject.org/projects/dynamicproxy/ to instantiate. So now you can:

```csharp
var result = await myClient.ExecuteWorkflowAsync(
    MyInstanceWorkflow.Ref.RunAsync, "some arg", myOptions);
```

We have also considered the proxy approach like so:

```csharp
var result = await myClient.ExecuteWorkflowAsync<MyInstanceWorkflow>(
    inst => inst.RunAsync("some arg"), myOptions);
```

But that has many problems including that we can't control the contents lambda, that the method has to be marked `virtual` (can't dynamic-proxy a non-virtual method), that statics need a separate overload, etc. We also considered post-sharp-like compilation steps, the newer .NET source generation features, etc and all had too many downsides. For F# I'll probably do type providers or similar.

Here's the problem: "What is the easiest way for a user to reference an instance method in C#?". What @nickwilliams-codynamic  suggested was to manually define all definitions, which (ignoring the lambda vs class approach) is basically what `Refs<MyInstanceWorkflow>.Instance` is a shortcut for doing. And remember, we need to apply this to signal and query methods too, so you can't just have some interface they implement, it needs to be free form. I am definitely open to any way to make this better, but we need to make the workflow read well to the user and be flexible at the same time. Any ideas I may have missed? Any other projects besides Azure Durable Functions/Entities I can look to for inspiration on how they handle "instance method references"?

cretz:
Merging as this discussion as run its course. We will likely be providing source generation as an option for users.

PR #73, opened 2023-01-07
Workflow Task Completion Metadata

Sushisource:
[Rendered](https://github.com/Sushisource/proposals/blob/wft-complete-metadata/api/wft-complete-metadata.md)

### DIFF
--- a/api/wft-complete-metadata.md
+++ b/api/wft-complete-metadata.md
@@ -1,0 +1,133 @@
+# Workflow Task Completion Metadata
+
+We have a need to attach various bits of metadata to workflow task completions. There are a number
+of usecases in which the SDK would like to save some information for itself, or tell server about
+something, but this information does not influence replay or the reconstitution of workfow state in
+any way. Thus, it shouldn't go in a marker or some new kind of history event, because this would
+violate the (useful) idea that history events are primarily concerned with recreating workflow
+state.
+
+## Use cases
+
+### Reporting currently-being-awaited commands & facts about them
+
+Right now, besides a query, there's no simple way for a user to understand what their workflow is
+actively waiting on. Solving this would make it much easier for users to have insight into why
+their workflow is or isn't making progress, and could be nicely exposed in the UI.
+
+In particular, understanding what's going on while your workflow is currently running local
+activities and performing WFT heartbeats is very opaque. This would make that much more clear.
+
+Imagine being able to get information like this (and do useful things with it in the UI):
+
+```
+waiting_on: [
+    { command_type: "local_activity", activity_name: "la_func", id: 1, retries: 0, location: "my_file.py:23" },
+    { command_type: "local_activity", activity_name: "la_func", id: 2, retries: 5, location: "my_file.py:23" },
+    { command_type: "timer", id: 3, duration: "5s", location: "my_file.py:42" },
+]
+```
+
+### SDK internal versioning
+
+We would like to be able to make SDK changes which affect how history is written, which means
+flagging our behavior based on what version of the SDK the history was produced with. Note: *not*
+relating to concepts of versioning a *user's* code, but rather the SDK itself.
+
+This internal version could easily slot into this metadata.
+
+Note: This could potentially go in `temporal.api.taskqueue.v1.VersionId` too, but that's more
+intended to relate to user code.
+
+### Dynamic config adjustments
+
+Allow the worker to set configurable things which don't directly impact determinism or workflow
+state. For example, changing the WFT timeout for the next task.
+
+### Various debugging / analysis information
+
+There's a whole world of little bits of info that could be useful to enable selectively to help
+users debug their workflows, help us learn about performance information, etc.
+
+Some examples:
+* Emit the OTel span id that was associated with the WFT - now the user can go easily look up any
+  logs/events that were emitted as part of that WFT.
+* Metrics work great for all sorts of latency measurements, but what if you wanted to know about
+  how long a _specific_ WFT took? You could emit that.
+* Interceptors could pipe whatever arbitrary data the user wants into there, for use cases where
+  memo / search attributes / side effects may not be appropriate.
+
+### Things we haven't thought of
+
+Having this metadata be suitably generic means we dont need to update the API repo for every
+possible usecase. Particularly ones where the SDK is the only one who needs to be able to read it.
+
+## Implementation Proposal
+
+We'd add a new metadata field to the
+[WFT response](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L274).
+And the [WFT completed event](https://github.com/temporalio/api/blob/f0350f8032ad2f0c60c539b3b61ea37f412f1cf7/temporal/api/history/v1/message.proto#L186).
+Though the completion event need not necessarily contain all the info sent in the completion. If
+it's known and understood that a particular metadata field is simply telling server something
+ephemeral, that could be stripped before writing the event.
+
+With the following proto as its type:
+
+```protobuf
+message StructuredMetadata {
+  // For JSON-esque data representations that can be easily decoded by anyone. Could contain user
+  // data that they'd like to show up the UI, for example.
+  map<string, google.protobuf.Struct> struct_metadata = 1;
+  // For better typed / more efficient binary serialization of types that are expected to be known
+  // by their consumers.
+  map<string, google.protobuf.Any> proto_metadata = 2;
+}
+```
+
+Having both fields seems useful here, since `Any` is useful for the SDK making notes to itself which
+it knows the type of. `Struct` is essentially just a more efficient JSON representation, and can
+be used for arbitrary data that can easily be rendered.
+
+Since `Any` can also be `Struct`, we could have just one field, but this wastes some bytes on the
+`type_url` field if we are throwing a lot of structs in there. Keeping them separate also makes
+it clear what you can decode without any type knowledge, rather than walking the whole map looking
+for struct types in case you want to decode them - that way you only look in the `Any` map when you
+know you're looking for something specific.
+
+### Why not just add more fields to WFT completion response?
+
+* Things that the SDK only needs to write down for itself (ex: the versioning usecase) can now be
+  made without requiring changes to the API repo, updating the server, releasing it, and ensuring
+  that that version is in use. That's real nice.
+* Supports arbitrary user-originated data (TBD if we actually want to allow this, but with this
+  approach it's easy to add if we do).
+
+### Why not use markers?
+
+This is mentioned above, but I want to reiterate it: 
+
+I think it's very important that history events stay focused on recreating workflow state. The
+usecases mentioned here don't impact workflow state, and so they shouldn't get dedicated events
+in history. Anything which *does* directly impact workflow state or the recreation of, should indeed
+be represented as an event.
+
+You might reasonable argue "Hey! The SDK internal versioning usecase you mentioned earlier in some
+sense does or can impact workflow state, why doesn't that get a marker?". I think it's worth
+drawing a distinction between "things that impact workflow state and are a consequence of user
+code", "things that indirectly impact workflow state but are not a consequence of user code" , and
+"things that don't impact state at all". I think only the first of those deserve top-level
+representation as history events. The others can be stored as metadata, or not at all.
+
+## Drawbacks
+* No dedicated fields means no docstrings for those fields, and less type safety. Though this is
+  much less of a concern when the writer and reader are the same person. Type safety can be achieved
+  with the `Any` valued map - and should be used for data that is expected to be interpreted by
+  more than one component.
+* Potential to bloat history.
+
+## Adoption / teaching / etc
+
+Not applicable to this proposal since developers will not *directly* interact with this, but rather
+it's just providing a mechanism for ourselves. The "what are we wating on" usecase has obvious and
+big value to users, but what exactly gets written there and how we expose it in the UI etc ought to
+be another proposal.
--- a/api/wft-complete-metadata.md
+++ b/api/wft-complete-metadata.md
@@ -71,30 +71,20 @@ Though the completion event need not necessarily contain all the info sent in th
 it's known and understood that a particular metadata field is simply telling server something
 ephemeral, that could be stripped before writing the event.
 
-With the following proto as its type:
+With the following type:
 
 ```protobuf
-message StructuredMetadata {
-  // For JSON-esque data representations that can be easily decoded by anyone. Could contain user
-  // data that they'd like to show up the UI, for example.
-  map<string, google.protobuf.Struct> struct_metadata = 1;
-  // For better typed / more efficient binary serialization of types that are expected to be known
-  // by their consumers.
-  map<string, google.protobuf.Any> proto_metadata = 2;
+message RespondWorkflowTaskCompletedRequest {
+  // ...
+  
+  map<string, temporal.api.common.v1.Payload> metadata = 11;
+  
+  // ...
 }
 ```
 
-Having both fields seems useful here, since `Any` is useful for the SDK making notes to itself which
-it knows the type of. `Struct` is essentially just a more efficient JSON representation, and can
-be used for arbitrary data that can easily be rendered.
 
-Since `Any` can also be `Struct`, we could have just one field, but this wastes some bytes on the
-`type_url` field if we are throwing a lot of structs in there. Keeping them separate also makes
-it clear what you can decode without any type knowledge, rather than walking the whole map looking
-for struct types in case you want to decode them - that way you only look in the `Any` map when you
-know you're looking for something specific.
-
-### Why not just add more fields to WFT completion response?
+### Why not just add more (specific) fields to WFT completion response?
 
 * Things that the SDK only needs to write down for itself (ex: the versioning usecase) can now be
   made without requiring changes to the API repo, updating the server, releasing it, and ensuring
@@ -121,8 +111,8 @@ representation as history events. The others can be stored as metadata, or not a
 ## Drawbacks
 * No dedicated fields means no docstrings for those fields, and less type safety. Though this is
   much less of a concern when the writer and reader are the same person. Type safety can be achieved
-  with the `Any` valued map - and should be used for data that is expected to be interpreted by
-  more than one component.
+  by serializing a known proto into the payload - and should be used for data that is expected to be
+  interpreted by more than one component.
 * Potential to bloat history.
 
 ## Adoption / teaching / etc
@@ -131,3 +121,6 @@ Not applicable to this proposal since developers will not *directly* interact wi
 it's just providing a mechanism for ourselves. The "what are we wating on" usecase has obvious and
 big value to users, but what exactly gets written there and how we expose it in the UI etc ought to
 be another proposal.
+
+Payload values which are deserializable using the existing set of deserializers the UI understands
+by default should be shown in the UI. Ones that can't should be listed as uninterpretable.


api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+actively waiting on. Solving this would make it much easier for users to have insight into why
+their workflow is or isn't making progress, and could be nicely exposed in the UI.
+
+In particular, understanding what's going on while your workflow is currently running local
+activities and performing WFT heartbeats is very opaque. This would make that much more clear.
+
+Imagine being able to get information like this (and do useful things with it in the UI):
+
+```
+waiting_on: [
```
bergundy, line 24: I don't think we want to add this type of data.
We already know which activities and timers are unresolved from workflow history.
We don't know what the workflow code is effectively blocked on and we can't know that.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+relating to concepts of versioning a *user's* code, but rather the SDK itself.
+
+This internal version could easily slot into this metadata.
+
+Note: This could potentially go in `temporal.api.taskqueue.v1.VersionId` too, but that's more
+intended to relate to user code.
+
+### Dynamic config adjustments
+
+Allow the worker to set configurable things which don't directly impact determinism or workflow
```
bergundy, line 44: I think this is a nice feature to have but I don't think this is the right mechanism for that since we need a tighter contract for the server to be able to interpret these config changes.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+
+### Various debugging / analysis information
+
+There's a whole world of little bits of info that could be useful to enable selectively to help
+users debug their workflows, help us learn about performance information, etc.
+
+Some examples:
+* Emit the OTel span id that was associated with the WFT - now the user can go easily look up any
+  logs/events that were emitted as part of that WFT.
+* Metrics work great for all sorts of latency measurements, but what if you wanted to know about
```
bergundy, line 55: Don't we have this information already?

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+With the following proto as its type:
+
+```protobuf
+message StructuredMetadata {
+  // For JSON-esque data representations that can be easily decoded by anyone. Could contain user
+  // data that they'd like to show up the UI, for example.
+  map<string, google.protobuf.Struct> struct_metadata = 1;
+  // For better typed / more efficient binary serialization of types that are expected to be known
+  // by their consumers.
+  map<string, google.protobuf.Any> proto_metadata = 2;
```
bergundy, line 83: It really depends how this data is made visible in the UI.
I think struct is probably more than we'd need and is efficient enough.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+  made without requiring changes to the API repo, updating the server, releasing it, and ensuring
+  that that version is in use. That's real nice.
+* Supports arbitrary user-originated data (TBD if we actually want to allow this, but with this
+  approach it's easy to add if we do).
+
+### Why not use markers?
+
+This is mentioned above, but I want to reiterate it: 
+
+I think it's very important that history events stay focused on recreating workflow state. The
```
bergundy, line 109: We're using this mechanism for versioning the SDK logic which effectively means it's for recreating workflow state.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+usecases mentioned here don't impact workflow state, and so they shouldn't get dedicated events
+in history. Anything which *does* directly impact workflow state or the recreation of, should indeed
+be represented as an event.
+
+You might reasonable argue "Hey! The SDK internal versioning usecase you mentioned earlier in some
```
bergundy, line 114: I just argued that above :)

I don't think the distinction matters.
To me it's more about leaving less traces when we make changes, especially if we charge users per marker generated in Cloud.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+code", "things that indirectly impact workflow state but are not a consequence of user code" , and
+"things that don't impact state at all". I think only the first of those deserve top-level
+representation as history events. The others can be stored as metadata, or not at all.
+
+## Drawbacks
+* No dedicated fields means no docstrings for those fields, and less type safety. Though this is
+  much less of a concern when the writer and reader are the same person. Type safety can be achieved
+  with the `Any` valued map - and should be used for data that is expected to be interpreted by
+  more than one component.
+* Potential to bloat history.
```
bergundy, line 126: This could imply slightly increased storage costs but I'd argue those are negligible.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
+# Workflow Task Completion Metadata
+
+We have a need to attach various bits of metadata to workflow task completions. There are a number
+of usecases in which the SDK would like to save some information for itself, or tell server about
+something, but this information does not influence replay or the reconstitution of workfow state in
```
Quinn-With-Two-Ns, line 5: For SDK internal versioning won't this feature effect replay/reconstitution of workfow state? Might be clearer to say user workfow state?

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+## Drawbacks
+* No dedicated fields means no docstrings for those fields, and less type safety. Though this is
+  much less of a concern when the writer and reader are the same person. Type safety can be achieved
+  with the `Any` valued map - and should be used for data that is expected to be interpreted by
+  more than one component.
+* Potential to bloat history.
+
+## Adoption / teaching / etc
+
+Not applicable to this proposal since developers will not *directly* interact with this, but rather
```
bergundy, line 130: I think we should definitely show this in the UI and make sure that the additional information is clearly visible.

api/wft-complete-metadata.md
cretz: Same. But I consider almost all of this as "could use metadata for one day" and not necessarily now.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,133 @@
...
+
+```protobuf
+message StructuredMetadata {
+  // For JSON-esque data representations that can be easily decoded by anyone. Could contain user
+  // data that they'd like to show up the UI, for example.
+  map<string, google.protobuf.Struct> struct_metadata = 1;
+  // For better typed / more efficient binary serialization of types that are expected to be known
+  // by their consumers.
+  map<string, google.protobuf.Any> proto_metadata = 2;
+}
```
cretz, line 84:
```suggestion
message RespondWorkflowTaskCompletedRequest {
  // Other fields

  map<string, temporal.api.common.v1.Payload> metadata = 11;
}

message WorkflowTaskCompletedEventAttributes {
  // Other fields

  map<string, temporal.api.common.v1.Payload> metadata = 6;
}
```

I think this has over-thought the issue. I think we can just use normal payloads and have reserved field names for ourselves. Our UI knows how to show user-structured payloads already and our codecs understand that payloads should be encrypted and users may want to encrypt this information.

If we wanted to keep the `StructuredMetadata` wrapper message, that's fine too. But I think we need to reuse our existing structured metadata mechanisms for structured metadata. We can discuss if certain keys should use the default converters (i.e. unencrypted).

api/wft-complete-metadata.md
Sushisource: How would we not want to add this? This is incredibly useful for a user trying to understand what their workflow is currently waiting on.

Local activities are the prime example, right there, and they are *not* already in history. What rationale is there for not adding this? 

It's also not true that we don't know what code is blocked on. We know exactly what it's blocked on, because that's the definition of when you finish an activation / workflow task - when all routines are at an await point-type thing. They don't all necessarily correspond 1:1 to commands, but you know exactly what you're blocked on, by definition.

api/wft-complete-metadata.md
Sushisource: For things that you just want to show in the UI, sure. For notes the SDK wants to make to itself, a defined proto makes more sense.

api/wft-complete-metadata.md
Sushisource: Heh, I need to make more "Read the whole thing before commenting" notes at the top. 

You don't - I do. These sorts of rules don't make a difference in any specific situation much, but over the long run they make the difference between maintainable, sensible systems, and hodge-podge.

api/wft-complete-metadata.md
Sushisource: Yeah, which is what I get at later on at the bottom.

api/wft-complete-metadata.md
bergundy: What is this code blocked on?

```ts
export async function myWorkflow() {
  // run in the background and ignore if workflow is completed
  someActivity();
  await CancellationScope.current().cancelRequested;
}
```

You can tell by the stack trace (especially the structured one where users can see the actual code) but the SDK would struggle knowing which promises are "blocking" the workflow from making progress.

Local activities are definitely more interesting here since they don't show up in history until completed but even then I think we can infer from the fact that a WFT heartbeat is present that local activities are running and instruct the user to go to the stack trace query to get more information.

api/wft-complete-metadata.md
bergundy: All of the history events have timestamps, is that not enough? Do you mean something else here?

api/wft-complete-metadata.md
bergundy: Even then I'd want the information to be clearly displayed in the UI.

api/wft-complete-metadata.md
bergundy: I actually like the use of structured data here assuming that this data doesn't need to be encrypted.

api/wft-complete-metadata.md
Sushisource: It's blocked on waiting for the current cancellation scope to be requested. It's right there, where you're awaiting.

Admittedly that's harder to determine in TS than some other languages, but it's still 100% possible and useful. In Rust it's trivial. In Python it's probably pretty easy too.

Why would I want to _infer_ that we're probably waiting on some local activities when I could know exactly how many are currently running and their types? I don't really follow what the argument is for _not_ doing it? Besides that, we need this for cloud, which is one of the motivating reasons.

bergundy: My point is that we already know this from the stack trace, no need to add this data to histories.

Sushisource: Stack trace is a point-in-time view. This is a progressive view that works for every WFT, not just wherever you currently are.

A good point is that "currently waiting on" and "currently outstanding" are worth distinguishing. We need to know currently outstanding LAs for cloud, and perhaps things that are being waiting on is fine to leave for stack traces, but seems to me this would still be remarkably useful for debugging the progress of a workflow and it's not going to take up much space in history.

api/wft-complete-metadata.md
mmcshane: Having both the Any and the Struct maps isn't great - it's hard to predict or document where a given field will be as semantically they're the same. But we're really hamstrung by the protobuf type system here so the options aren't great. Do we need the two layers of lookup provided by a map of structs? Since a struct is itself a map it feels like we could get away with just a single Struct instance. What do you think about a convention where a `meta` field is always a Struct and if the Value of one of the Struct fields has to be a Payload in the future then so be it.

Sushisource: I'd be fine with just one `Struct` which can potentially have `Payload` (or `Any`) values in it.

cretz: I still think payload works very well for structured data and the UI knows how to deal with structured payload data and you get the bonus of encryptability and easy user value decoding. There are so many wins to using the same thing we use for memo, failure attributes, etc.

Sushisource: It works... OK. Something like `Struct` internally probably would've made more sense in them as our default for things we want renderable in the UI rather than serialized JSON which is a bit wasteful, but that's in the past so w/e. 

I take your point about just being able to re-use it.

Honestly, none of this is terribly consequential. They're all equivalent and it's just a matter of what's easier in what situation. I think we should just pick something and move on. I like `Struct` because it sets us up to easily and efficiently represent what I think will be the case with all or the vast majority of the data represented here, which is data that won't ever be passed through a custom codec. If we'd rather just keep using Payload b/c it exists, I get that too. @bergundy if you can tie break, I'd say let's do that and move forward (or maybe we can grab someone from server team)

bergundy: The main question is if we want users to be able to put their (potentially sensitive) data in these fields.
If that's not a requirement, we could go with Structs which are easier for us to work with and don't require extra serialization. If we're unsure, I'd go with @cretz's suggestion of using `Payload`s.

cretz: It's not just sensitive data. You're missing the other benefits I mentioned. What about the ability to easily decode to language-defined types in all languages? What about the consistency with all other metadata forms we use?

bergundy: I'm convinced.

Sushisource: OK, let's go with Payloads then.

I just want to say though that
> There just isn't a good reason for untyped structs

This isn't true - there is a pretty good reason, which is serializing and encoding JSON strings is wasteful by comparison, and they are easier to work with for clients who don't have access to all our payload machinery. Now, I'm fine with saying that's not worth re-using all the payloads stuff for this case - but that doesn't mean there's literally no merit to the idea.

cretz: > when we have a format that is perfect for this

Agree there is value in structs and other WKT for proto. For example Envoy uses proto to define their config formats and accepts these types for unknown user input instead of JSON.

IME though, working with these types in code makes you long for the typed protos or language structs you use otherwise. I've seen people just convert these things to JSON just so they can convert them into their language-typed forms anyways. But yes, they do have generic value.

api/wft-complete-metadata.md
Sushisource: I think that's probably true, but it's an option.

api/wft-complete-metadata.md
Sushisource: It could be anything arbitrarily more granular than that as well.

cretz (approved):

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+waiting_on: [
+    { command_type: "local_activity", activity_name: "la_func", id: 1, retries: 0, location: "my_file.py:23" },
+    { command_type: "local_activity", activity_name: "la_func", id: 2, retries: 5, location: "my_file.py:23" },
+    { command_type: "timer", id: 3, duration: "5s", location: "my_file.py:42" },
+]
+```
+
+### SDK internal versioning
+
+We would like to be able to make SDK changes which affect how history is written, which means
```
MichaelSnowden, line 33: Could you explain more about the changes mentioned here? When do we want the SDK to drive how the server writes history?

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+flagging our behavior based on what version of the SDK the history was produced with. Note: *not*
+relating to concepts of versioning a *user's* code, but rather the SDK itself.
+
+This internal version could easily slot into this metadata.
+
+Note: This could potentially go in `temporal.api.taskqueue.v1.VersionId` too, but that's more
+intended to relate to user code.
```
MichaelSnowden, line 40: I think we should remove this note since it's not really an acceptable option

api/wft-complete-metadata.md
Sushisource: It won't affect how server writes history. It will affect how the SDK decides what to do, based on what version of an SDK a workflow was processed with, enabling or disabling certain features. SDKs may implement new features that affect how they produce commands, and on replay we need to do the same things that happened when the WF was first executed, hence a need to check our own version.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+
+### Dynamic config adjustments
```
MichaelSnowden, line 42: I don't see the use case for making dynamic config adjustments in the WFT completion event. There are already existing methods to modify the dynamic config. Is the idea here that you can make it more atomic somehow? If so, that's something users may end up relying on. So we'd need to maintain that contract on the server-side, which seems error-prone to me and difficult to future-proof.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+Allow the worker to set configurable things which don't directly impact determinism or workflow
+state. For example, changing the WFT timeout for the next task.
+
+### Various debugging / analysis information
+
+There's a whole world of little bits of info that could be useful to enable selectively to help
+users debug their workflows, help us learn about performance information, etc.
+
+Some examples:
+* Emit the OTel span id that was associated with the WFT - now the user can go easily look up any
```
MichaelSnowden, line 53: In my experience, the span id is attached to the request metadata, not stored within the request data itself. What's the rationale behind moving this into the request data?

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+  logs/events that were emitted as part of that WFT.
+* Metrics work great for all sorts of latency measurements, but what if you wanted to know about
+  how long a _specific_ WFT took? You could emit that.
```
MichaelSnowden, line 56: This is already done via tracing, so it seems like a duplicate of the note above

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+* Interceptors could pipe whatever arbitrary data the user wants into there, for use cases where
```
MichaelSnowden, line 57: Spans already allow you to store arbitrary key/value pairs. Why can't we just use that?

api/wft-complete-metadata.md
Sushisource: Maybe it wasn't clear, but these are all _possible_ use cases. The only one we know we need now is the SDK internal versioning ability, and probably the ability to track outstanding commands. Everything else is in "could be supported by this if we want" territory. Same applies for your subsequent comments, but I'll mention my thoughts on those too.

As mentioned here https://github.com/temporalio/proposals/pull/73#discussion_r1068742307 - yeah, maybe this one we'd avoid.

api/wft-complete-metadata.md
Sushisource: It's a simple matter of making it easily discoverable by a user looking at workflow history in the UI, who may very well not have initiated any of the requests associated with the workflow.

api/wft-complete-metadata.md
Sushisource: Fair enough.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+### Things we haven't thought of
+
+Having this metadata be suitably generic means we dont need to update the API repo for every
+possible usecase. Particularly ones where the SDK is the only one who needs to be able to read it.
+
+## Implementation Proposal
+
+We'd add a new metadata field to the
+[WFT response](https://github.com/temporalio/api/blob/master/temporal/api/workflowservice/v1/request_response.proto#L274).
+And the [WFT completed event](https://github.com/temporalio/api/blob/f0350f8032ad2f0c60c539b3b61ea37f412f1cf7/temporal/api/history/v1/message.proto#L186).
```
MichaelSnowden, line 69: Adding arbitrarily-sized attributes into history exposes us to resource exhaustion. For example, we had a major outage caused by storing too many child workflow executions in the mutable state for executions.

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+
+### Why not just add more (specific) fields to WFT completion response?
+
+* Things that the SDK only needs to write down for itself (ex: the versioning usecase) can now be
+  made without requiring changes to the API repo, updating the server, releasing it, and ensuring
+  that that version is in use. That's real nice.
+* Supports arbitrary user-originated data (TBD if we actually want to allow this, but with this
```
MichaelSnowden, line 92: I think that simple arbitrary data associated with WFTs can be accomplished easily with tracing, and more advanced use cases should be left up to users. They can store their own data associated with a WFT id.

MichaelSnowden (commented):
To continue with a review, I need to better understand the specific problem here. The solution here is very generic, but the only concrete use case I see is getting the version of the SDK, which produced a specific WFT, and I'm wondering why we even need that.

api/wft-complete-metadata.md
Sushisource: I resolved the other comments since they all largely resolve around the same concern w/ tracing.

Yep, bloat is called out in possible drawbacks. 

Not everyone uses tracing (in fact, I'd guess most users don't). This potentially gives them a way to write data into a place that we currently provide visibility most easily to them. Not everything makes sense to write as tracing data either.

Beyond that, this proposal isn't saying we'll definitely do that. It's something it could support if we ever needed to, now. No user data is going to be written here until if and when we decide we want to do that. If we did, almost certainly we'd put limits on that like we do w/ existing user-created fields.

api/wft-complete-metadata.md
Sushisource: See also https://github.com/temporalio/features/issues/200

api/wft-complete-metadata.md
```diff
@@ -0,0 +1,126 @@
...
+activities and performing WFT heartbeats is very opaque. This would make that much more clear.
+
+Imagine being able to get information like this (and do useful things with it in the UI):
+
+```
+waiting_on: [
+    { command_type: "local_activity", activity_name: "la_func", id: 1, retries: 0, location: "my_file.py:23" },
+    { command_type: "local_activity", activity_name: "la_func", id: 2, retries: 5, location: "my_file.py:23" },
+    { command_type: "timer", id: 3, duration: "5s", location: "my_file.py:42" },
+]
```
yiminc, line 28: Will those payload be encoded by user's custom data converter? I would assume some of them would not if server needs to understand them, but others would because it could be sensitive info.

Sushisource: This specific example of outstanding commands should use the standard data converter. They may need to be understood for cloud billing purposes (the local activity stuff, depending on our choices there). They also need to be UI-renderable to be useful for debugging info.

But, regardless, again, that's not really required to move forward with this proposal. The specifics of any piece of metadata is up to if and when we add it. This proposal is just the mechanism of transport.

Sushisource:
So, we talked more within the SDK team and we decided that the flexibility here comes at a bit too high of a cost in terms of size. We don't actually have any *concrete* use cases at the moment that require writing completely arbitrary data. Everything we want to write in here is well defined for at least one consumer (even if it's only the SDK consuming its own notes).

Hence, everything we know we want to add here is already a proto. The only question is how it will get encoded. (Hence the desire for `Any` in my original version). But, both `Any` and `Payload` are byte-wasteful.

The main motivation for adding this was speculation about needing to add generic stuff (but we don't know what that is) and decoupling SDK-only stuff from the long approval process / lifecycle in this repo. The first we can deal with if/when it comes up. The second we can deal with by simply introducing an `sdk` package in this repo with different codeowners. We'll do that.

PR #74, opened 2023-02-10
Typed search attributes in SDK

cretz:
Rendered (until merged): https://github.com/cretz/temporal-proposals/blob/sdk-typed-search-attributes.md/sdk-typed-search-attributes.md

Basically just tried to come up with a non-breaking way to add typed, single-value search attributes to our SDKs.

### DIFF
--- a/sdk-typed-search-attributes.md
+++ b/sdk-typed-search-attributes.md
@@ -55,6 +55,10 @@ Below is just a proposal of ideas on how better to type search attributes.
 * Setting invalid search attributes in workflow never returns a failure that can be handled by workflow code
 * Users have no way to clarify the difference between keyword and keyword-list
 
+# Solution 1
+
+(see "Solution 2" for a separate solution)
+
 ## Goals
 
 Immediate:
@@ -77,6 +81,8 @@ Non-goals:
 * Support for predefined search attributes in any form of query building
 
 ## Proposed solution
+
+See "Solution 2" for comparison of approaches
 
 ### General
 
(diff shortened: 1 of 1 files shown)


sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+* Upsert will be deprecated
+
+### Go
+
+```go
+// In client
+type SearchAttributes map[string]any
+
+// Yes these already exist today with maps themselves, but clear methods may
+// help discourage people from using the map at all. Can remove though.
```
Sushisource, line 98: If you're gonna do this, why not just hide the map inside a struct like you have with the in-workflow version? IMO it makes sense for it to not look like a normal map, because you can't just directly set stuff on it.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+
+* Deprecate `WorkflowOptions.Builder.setSearchAttributes`
+* Add overload `WorkflowOptions.Builder.setSearchAttributes` that accepts this map type
+  * The ability to need to instantiate this is why it's a class and not an interface
+    * Could have this be an interface if it was also easy to create somehow
+* Deprecate `Workflow.getSearchAttributes`, `Workflow.getSearchAttribute`, and `Workflow.getSearchAttributeValues`
+  * Can't think of a better way here, we can't reuse these from my POV and still be compat
+* Add `Workflow.getSearchAttributeMap` that returns this
+  * This is mutable! Mutations follow general rules laid out before
+* Should we log a runtime warning if they use the Map accessors directly?
```
Sushisource, line 264: Same Q here - why implement `Map` at all if you can't actually use it like a map? Applies to TS/Python too.

I get the idea of wanting to iterate over them or something - but in that case maybe these should just define a `as_collection()` or `iterator()` or similar method which returns a read-only view for use in those cases.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+  * Also, since impl time only, I can put the name in the annotation instead of having a search attribute constructor
+  * Is this ok to do for us to populate this field reflectively after construction and promise always non-null, or
+    should we require that they instantiate,
+    e.g. `private final SearchAttributeInt mySearchAttribute = new SearchAttributeInt();`?
+  * Is there a good use case to have this definition usable externally via static field on interface instead?
+* Do same thing for memos but allow looser types
+
+### TypeScript
+
+Being la
```
Sushisource, line 315: Truncated sentence

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+
+Non-immediate:
+
+* Provide ways to define predefined search attributes that are typed
+* Also provide ways to define predefined memo values that are typed
+
+Non-goals:
+
+* Better validation of actual search attribute type from server
+* Better way for server to tell SDK the upsert failed
```
Sushisource, line 74: My concern with ignoring this for now is this design would mean we'd have to make another different API if we ever want users to be able to upsert-and-confirm.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+* Support for user-explicit upsert of multiple typed search attributes in one statement
+* Support for typed registration of search attributes
+* Support for predefined search attributes in any form of query building
+
+## Proposed solution
+
+### General
+
+* All SDK workflow APIs will now have a mutable search attribute map with typed getters and setters
```
Sushisource, line 83: Not at all a problem with this proposal - but we will want to be clear in docs/docstrings that the search attribute map you see in the workflow may not necessarily contain a complete representation of actual search attributes in persistence if users ever modify them outside of a workflow.

Sushisource (commented):
In general makes good sense to me. My main questions are around looking like native map types.

sdk-typed-search-attributes.md
cretz: I have to continue to accept a map type in `StartWorkflowOptions.SearchAttributes`. So if I change that type from `map[string]interface{}` to this type, this code still works:

```go
myOptions.SearchAttributes = map[string]interface{}{"foo": "bar"}
```

sdk-typed-search-attributes.md
cretz: Have to support setting existing values (so we have to accept maps in start options) and returning this extension. Java can have overloads, so we don't have to accept a map type on start, but we do have to return this from getter or else we need to have a different name.

sdk-typed-search-attributes.md
cretz: Yes, we will have to decide whether returning errors on search attribute mutation is worth it in the long run.

If we do decide that it is, I need to change a lot of things here. Basically we'd be back to "upsert search attributes" which is fine, just need to get more input that that's the ideal direction (and we'd probably call it "apply search attributes" or something since we can't mess w/ that existing function).

sdk-typed-search-attributes.md
cretz: We have an event for external workflow SA mod now right? I think the only way they can mod them outside of workflow is with that which will be put into history. And if there's race, the workflow will get an unhandled command and retry. I sure hope the mutate-search-attribute API if/when it gets created has some form of conflict token if it loses the race.

The only way that this may not be accurate I think is if there is a server validation error with the search attribute (since we don't get that feedback).

bergundy (commented):
I'm wondering why you prefer splitting the proposal into 2 different phases with distinct APIs.
We typically like to have a single way to do things in our SDKs and this proposal goes against that principle.

sdk-typed-search-attributes.md
cretz: Will bring up in discussion. Lots of stuff in here is invalid if we need to error on invalid search attribute setting, so we can decide to invalidate here in proposal or invalidate later in code.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+  * Search attributes getting in workflow - on info as `Record<string, scalar[]>` (i.e. object)
+  * Search attributes setting in workflow - via `upsertSearchAttributes` accepting `Record<string, scalar[]>`
+  * Search attributes setting on start - via `Record<string, scalar[]>` on start options
+  * Search attributes getting on list/describe - as `Record<string, scalar[]>`
+* Python
+  * Search attributes getting in workflow - on info as `Mapping[str, List[scalar]]`
+  * Search attributes setting in workflow - via `upsert_search_attributes` accepting `Mapping[str, List[scalar]]`
+  * Search attributes setting on start - via `Mapping[str, List[scalar]]` on start options
+  * Search attributes getting on list/describe - as `Mapping[str, List[scalar]]`
+* .NET
```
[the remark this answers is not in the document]
cretz: Shrug, maybe PHP too, didn't check

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+  * Still in alpha, can decide whatever we want
+
+### Server
+
+* ElasticSearch
+  * Multi-value search attribute for non-keyword-list - accepted, but discouraged, and returned as given
```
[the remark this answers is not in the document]
cretz: This is under the heading "Current behavior" (and I don't think this note affects the proposal anyways)

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+Non-immediate:
+
+* Provide ways to define predefined search attributes that are typed
+* Also provide ways to define predefined memo values that are typed
+
+Non-goals:
+
+* Better validation of actual search attribute type from server
+* Better way for server to tell SDK the upsert failed
+* Support for user-explicit upsert of multiple typed search attributes in one statement
```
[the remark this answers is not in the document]
cretz: For relevance to this proposal, discussion at https://github.com/temporalio/proposals/pull/74/files/feeea0fcdfe4da697f1828167f716e608e64fd41#r1104572876

bergundy:
> > I'm wondering why you prefer splitting the proposal into 2 different phases with distinct APIs.
> > We typically like to have a single way to do things in our SDKs and this proposal goes against that principle.
> 
> I need an API for accessing multiple and an optional API for accessing single. Those are two separate things, not two ways of the same thing. I can remove the one for accessing single if needed (which we may need to do anyways if we want to prepare for accepting errors as a result of failed upsert inside workflow).

That's not what I meant, you have a "<Lang>" and "<Lang> later" mentioned here, which are 2 completely separate API approaches.
I'd rather figure out a single long term solution that we're happy with now and avoid introducing multiple ways to access search attributes.

sdk-typed-search-attributes.md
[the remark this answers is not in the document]
bergundy: Probably PHP too, yes.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+* Support for predefined search attributes in any form of query building
+
+## Proposed solution
+
+### General
+
+* All SDK workflow APIs will now have a mutable search attribute map with typed getters and setters
+  * When a mutation on a search attribute occurs:
+    * If the most recent command is an upsert search attribute, the single value is merged
+    * If the most recent command is not an upsert search attribute, make one
```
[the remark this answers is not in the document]
cretz: >  Other approaches are - not merged and preserve the correlation between the call site and the command or flush the entire batch at the end of each "activation".

I don't like either of these other approaches. The correlation per single value is not needed, and flushing at the end is hiding too much, but you do want _some_ kind of merging just because we are no longer providing a way to update multiple in one statement (unless dicussion about.

> If we're going for batching I'd rather defer the flushing to the end to get the most compact command form.

Not looking for most compact, looking for ability to replace upsert w/ multiple synchronous single-value setters and retain same event counts. If I used to upsert 10 keys in a single call (or 1 in a loop without commands interspersed) and I change to this single-value upsert, I don't need 10 events.

> Also, please note that this method will not be history compatible with the older APIs and since we plan to deprecate that older APIs, I'm afraid users might rush to change their code without understanding the implications.

I hope not. Nobody should use new command-causing workflow APIs and expect compatibility. But we can call it out.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+
+```go
+const MySearchAttribute = SearchAttributeInt("MySearchAttribute")
+```
+
+And then used from the workflow. Alternatively we could have something like:
+
+```go
+type searchAttributeValueConstraint interface {
+	string | int | float64 | time.Time | []string
```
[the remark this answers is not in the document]
cretz: > if we ever decide to expose a query builder

Alternative suggestion here?

sdk-typed-search-attributes.md
bergundy: I don't think we want to error on failure here, I think that makes a worse API and I don't see much benefit to it.
How would a user be expected to handle upsert failure? We might as well just fail the workflow task like we do today.

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,497 @@
...
+
+But the former is more explicit probably. This same thing would need to be done for memos but be looser on accepted
+types.
+
+### Java
+
+Visibility and method bodies left out for brevity.
+
+```java
+public class SearchAttributeMap implements Map<String, Object> {
```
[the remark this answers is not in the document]
cretz: How is it any different than the Go API? Get and put are just methods on top of an untyped map in both. The only benefit is you can deprecate this and not Go map accessors.

sdk-typed-search-attributes.md
cretz: Right, the potential future would be to tell a user it failed as a normal failure they can react to in code. But if we don't think we'd ever add that, then we can not concern ourselves with it.

sdk-typed-search-attributes.md
cretz: > How would a user be expected to handle upsert failure?

Swallow it because I don't care whether the search attribute is present on the server.

> let user deploy a fixed version of the workflow or define the missing search attributes.

This will become a much less viable option now that we restrict search attribute count to a small number.

Note, I don't mind never erroring, I just want to present all sides here.

cretz:
I have added a second solution that retains the ability to get/set/upsert search attributes as a whole with a new "search attributes" collection in each language that relies on user-defined keys and restricts types. As expected, the devil is in the backwards compatibility details.

sdk-typed-search-attributes.md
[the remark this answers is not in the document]
bergundy: The alternative would be what you had above this where we don't use Go types and instead we use the SA types.

cretz:
After some team discussion we have decided a typed search attribute collection with typed key definitions makes the most sense. I will be closing this PR (but leaving branch around for a bit) and making a new PR/branch with these updates. We have also decided to focus on a single language at first, so we chose Java since its challenges represent the other languages' challenges reasonably well.

cretz:
Closing in favor of #75

PR #75, opened 2023-02-17
Typed search attributes in SDK

cretz:
Rendered (until merged): https://github.com/cretz/temporal-proposals/blob/sdk-typed-search-attributes2/sdk-typed-search-attributes.md

This is a followup after discussion from #74 and supersedes that one.

### DIFF
--- a/sdk-typed-search-attributes.md
+++ b/sdk-typed-search-attributes.md
@@ -1,0 +1,340 @@
+# Typed Search Attributes in SDKs
+
+Today most languages just do arbitrary lists for search attribute values or half-implementations of some-list some not.
+This is a problem now with the server having multiple different forms of search attributes and preferring non-lists.
+Below is just a proposal of ideas on how better to type search attributes.
+
+## Current behavior
+
+### SDK
+
+* Go
+  * Search attributes getting in workflow - on info as raw proto `*commonpb.SearchAttributes`
+  * Search attributes setting in workflow - via `UpsertSearchAttributes` accepting `map[string]interface{}`
+  * Search attributes setting on start - via `map[string]interface{}` on start options
+  * Search attributes getting on list/describe - not available in high-level
+* Java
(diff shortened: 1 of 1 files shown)


bergundy (approved):

sdk-typed-search-attributes.md
```diff
@@ -0,0 +1,340 @@
...
+  * This accepts a vararg `SearchAttributeUpdate` that is either a "set" or a "delete"
+    * We accept this is ugly, but there is no better way to represent deletes
+    * We considered a mutable `SearchAttributes` collection but individual key/value mutations would be too hard to
+      reason about if buffered/batched and if single upsert commands people would ask for bulk as soon as individual
+      upsert events appeared
+      * If we change back to representing deletes as null/None, we can accept mutable form of this collection
+    * Each `SearchAttributeKey` has a `valueSet(value)` and a `valueUnset()` that return `SearchAttributeUpdate`s
+      * So Java may look like `Workflow.updateSearchAttributes(MY_ATTR1.valueSet("foo"), MY_ATTR2.valueUnset())`
+      * Ok that we call it "unset" instead of "delete"?
+      * Ok that we call it `valueSet` and `valueDelete` instead of just `set` and `delete`?
```
Sushisource, line 142: I think these are fine but I'd just switch it around to be `setValue` and `unsetValue`

Sushisource (approved):
Definitely prefer this to first proposal :+1:

sdk-typed-search-attributes.md
cretz: The reason I put the noun first is because the verb makes you think you're doing something instead of just building an update

cretz:
I have made updates based on discussion. Namely, `SearchAttributes` collection is immutable and a builder pattern is required to build it for client, child, and continue-as-new need.

cretz:
Merging now that discussion is through. However, we're already learning things on the implementation side that are changing some of this (i.e. that we know the type of every key from the server payload).

PR #76, opened 2023-02-21
.NET SDK - Phase 2 - Workflows - Initial Comment Period

cretz:
# .NET SDK - Phase 2 - Workflows - Initial Comment Period

This is the proposal for the second phase of the .NET SDK.

* See it [rendered](https://github.com/cretz/temporal-proposals/blob/dotnet-workflows/dotnet/sdk-phase-2.md) (or can use the "rich diff" link in upper right; rendered link will disappear on PR merge)

### DIFF
--- a/dotnet/sdk-phase-2.md
+++ b/dotnet/sdk-phase-2.md
@@ -1,0 +1,484 @@
+# .NET SDK - Phase 2
+
+The .NET SDK will be implemented in three loosely defined phases:
+
+* Phase 1 - Initial project, activity worker, client, and test framework
+* Phase 2 - Workflow worker and replayer
+* Phase 3 - Deterministic workflow sandbox and/or static analyzer and source generation
+
+This proposal is for phase 2. It is intentionally loosely defined and all aspects are subject to discussion/change.
+
+## "Workflow" Class and Namespace Rename
+
+For activities it is decided that, like Java, users can get a static reference to a
+`Temporalio.Activity.ActivityContext` instance for things like heartbeating and cancellation token. This is because this
+context may need to be passed around in ways that `AsyncLocal` does not propagate (same with Java `ThreadLocal`).
+
+For workflows it is decided that, like Java, users can make static calls on a `Workflow` class to do things like
+starting activities and getting info. The reason for say, getting info, is `Workflow.Info` instead of
+`WorkflowContext.Current.Info` is:
+
+* We always have access to our environment in a workflow, so there is no benefit to a context instance class over static
+  calls
+  * Also added benefit of no extension methods which could confuse devs on which primitives are supported
+* In .NET we often see contexts as instances in workflow engines and just in general, but this doesn't work for Temporal
+  because:
+  * There is no benefit to mocking or abstracting the workflow context since Temporal has full control of its underlying
+    representation even in tests
+  * We prefer interceptors over wrapping/inheriting a context for custom behavior
+  * It is so often used its presence is implied and therefore forcing every user to manually DI makes no sense
+  * We do not want to support constructor-based DI which ruins `[WorkflowInit]` behavior
+  * We do not want to support function-based DI which ruins typing of method references
+* It matches Java more clearly
+* It is less wordy
+
+We are intentionally taking a different tack in Temporal .NET SDK over other workflow engines with static calls and no
+context. However, since the `Workflow` class can't reasonably be in the `Temporalio.Workflow` namespace due to clash, we
+are renaming two namespaces:
+
+* `Temporalio.Activity` renamed to `Temporalio.Activities`
+* `Temporalio.Workflow` renamed to `Temporalio.Workflows`
+
+TODO(cretz): Alternative options for discussion:
+
+* Call it `WorkflowContext` instead of `Workflow` but keep as static
+  * Mismatches Java
+  * Inconsistent with `ActivityContext`
+  * Not really a context instance if static
+* Require `WorkflowContext.Current` everywhere
+  * Mismatches Java
+  * No benefit since it will always be present in a workflow
+  * Wordy
+* Find some way to DI the `WorkflowContext`
+  * Mismatches Java
+  * Basically required DI since 99% of workflows will use it, so unnecessary user work
+  * No good place to DI it and still keep reasonably typed method references
+* Rename `ActivityContext` to just `Activity`?
+  * Mismatches Java
+  * May be strange that `Temporalio.Activities.Activity` is an instance to access with `Current` whereas
+    `Temporalio.Workflows.Workflow` is static
+
+## Workflow Definition Overview
+
+In phase 1 there was an example of an interface defining a workflow. But you can use a class instead and you can even
+have a constructor with the init params. Example:
+
+```csharp
+namespace MyNamespace;
+
+using System;
+using System.Threading.Tasks;
+using System.Threading.Tasks.Dataflow;
+using Temporalio.Activities;
+using Temporalio.Workflows;
+
+public record GreetingInfo
+{
+    [JsonPropertyName("salutation")]
+    public string Salutation { get; init; } = "Hello";
+
+    [JsonPropertyName("name")]
+    public string Name { get; init; } = "<unknown>";
+}
+
+public static class Activities
+{
+    [Activity]
+    public static string CreateGreeting(GreetingInfo info) => return $"{info.Salutation}, {info.Name}!";
+}
+
+[Workflow]
+public class GreetingWorkflow
+{
+    public static readonly GreetingWorkflow Ref = Refs.Create<GreetingWorkflow>();
+
+    private readonly BufferBlock<GreetingInfo> updates = new();
+    private readonly string name;
+    private string currentGreeting = "<unset>";
+
+    [WorkflowInit]
+    public GreetingWorkflow(string name)
+    {
+        this.name = name;
+        updates.Post(new(Name: name));
+    }
+
+    [WorkflowRun]
+    public async Task<string> RunAsync(string _)
+    {
+        // Continually update greetings while updates made available
+        while (await updates.OutputAvailableAsync())
+        {
+            // Store current greeting
+            currentGreeting = await Workflow.ExecuteActivityAsync(
+                Activities.CreateGreeting,
+                await updates.ReceiveAsync(),
+                new(StartToCloseTimeout: TimeSpan.FromSeconds(5)));
+            Workflow.Logger.LogDebug("Greeting set to {Greeting}", currentGreeting);
+        }
+        return currentGreeting;
+    }
+
+    [WorkflowSignal]
+    public Task UpdateSalutationAsync(string salutation) =>
+        updates.SendAsync(new(Salutation: salutation, Name: name));
+
+    [WorkflowSignal]
+    public Task CompleteWithGreetingAsync()
+    {
+        updates.Complete();
+        return Task.CompletedTask;
+    }
+
+    [WorkflowQuery]
+    public string CurrentGreeting() => currentGreeting;
+}
+```
+
+Some things to not about the above code (much copied from phase 1):
+
+* Every workflow must have the non-inheritable `[Workflow]` attribute
+  * Optional `Name` positional property can be given. Default is the unqualified type name with the "I" prefix removed
+    if it's an interface and the next letter is also capital
+    * So what does `IDWorkflow` become? Yes, `DWorkflow` :-( Weighed as an acceptably rare cost for obvious benefit
+  * Non-inheritable means just because the interface or base class defines `[Workflow]`, the registered one must too
+* Workflows can have a single public constructor with `[WorkflowInit]` that take the exact same parameters as the
+  `[WorkflowRun]` method
+  * This is not required and most workflows probably won't use it but it is nice for those wanting to build readonly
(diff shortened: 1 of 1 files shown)


dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+  * Can be in base class, but if overridden, must also have the attribute
+  * Method must return a `Task` and can accept arguments
+    * It is a bit annoying to force Task on people that may not be doing an async thing but:
+      * They may change it later so might as well require it (especially since the "Async" suffix is usually present)
+      * Causes too many overloads if we accept `Task` and `void`
+  * Optional `Name` positional property can be given. Default is the unqualified method name with "Async" trimmed off if
+    present.
+* Workflow queries are public methods with the non-inheritable `[WorkflowQuery]` attribute
+  * Can be in base class, but if overridden, must also have the attribute
+  * Method cannot return a `Task`, but can accept arguments
```
bergundy, line 166: How hard would it be to add support for `Task` returning queries? I'm asking since we're talking about allowing starting local activities from update validators and queries.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+
+    public IList<Type> Workflows { get; set; } = new List<Type>();
+
+    public TemporalWorkerOptions AddWorkflow(Type type) { /*... */ }
+
+    public int MaxCachedWorkflows { get; set; } = 10000;
+
+    public int MaxConcurrentWorkflowTasks { get; set; } = 100;
+
+    public double NonStickyToStickyPollRatio { get; set; } = 0.2;
```
bergundy, line 216: I'd avoid exposing this for now since we might remove this option in the future from Core.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+
+    internal static bool DebugModeEnvironmentVariable { get; } = /*...*/;
+
+    public bool DebugMode { get; set; } = DebugModeEnvironmentVariable;
+}
+```
+
+TODO(cretz): For discussion:
+
+* Intentionally leaving off `MaxConcurrentWorkflowTaskPolls`, right?
```
bergundy, line 230: I would for now, let's add it later if we see there's demand and we decide not to do poller autoscaling.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+        Func<T, TResult> activity, T arg, ActivityStartOptions options) { /*...*/ }
+
+    public static Task<TResult> ExecuteActivityAsync<T, TResult>(
+        Func<T, Task<TResult>> activity, T arg, ActivityStartOptions options) { /*...*/ }
+
+    public static Task ExecuteActivityAsync(
+        string activity, object[] args, ActivityStartOptions options) { /*...*/ }
+
+    public static Task<TResult> ExecuteActivityAsync<TResult>(
+        string activity, object[] args, ActivityStartOptions options) { /*...*/ }
```
bergundy, line 268: I like how you avoided the ambiguity and only do multiple args with stringly typed activities.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+
+    public void RemoveDynamicSignalHandler() { /*...*/ }
+
+    // Search attributes and memo
+
+    public static SearchAttributes SearchAttributes { get; }
+
+    public static void UpdateSearchAttributes(params SearchAttributeUpdate[] updates) { /*...*/ }
+
+    public static IDictionary<string, Payload> RawMemo { get; }
```
bergundy, line 344: Can we have a memo getter that doesn't return raw payloads, something that will embed the configured data converter so users don't have to decode payloads themselves?

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+* A `CancellationToken` can be provided on every async call made, but if not set, by default workflow cancellation token
+  is the token used. This allows it to be overridden with a token not related to workflow cancellation.
+
+TODO(cretz): For discussion:
+
+* `ChildWorkflowHandle` has `GetResultAsync()` to match `WorkflowHandle` in client, but this could just be a `Task`
+  property. Should it be? Other .NET APIs tend to do this.
+* In client I call options `WorkflowStartOptions`, not only for `StartWorkflowAsync` but also for
+  `ExecuteWorkflowAsync`. I have a similar pattern here with activities and children. Should I continue this or should I
+  remove the `Start` word?
```
bergundy, line 449: Both Java and TS call this `WorkflowOptions`, I don't see why we need to mention `Start`, there's nothing to disambiguate (ATM).

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+* Should I rename `Now` to `UtcNow` to match .NET since it's non-local? Or does matching `DateTime.UtcNow` not matter
```
bergundy, line 450: I'd vote for matching .NET here and trading off verbosity for clarity.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+
+* We call this `WorkflowReplayer` instead of `Replayer` like we do in Python
+  * TODO(cretz): Should we call this `Replayer` instead? I think I still like `ReplayWorkflowAsync` instead of just
+    `ReplayAsync` regardless of class name.
+* `WorkflowReplayerOptions` shares many options with `TemporalWorkerOptions`, but not enough (and a few extra from
+  client) to justify any code sharing
+* `ReplayWorkflowAsync` is thread safe and we'll document this instead of making an overload of this that accepts an
+  iterator
+* Unlike Python we choose not to provide the history on the `WorkflowReplayResult` because we don't aggregate for the
+  user (they have LINQ ways to run many in parallel and they can correlate as desired)
```
bergundy, line 484: As long as the experience without accepting an iterator is nice, I'm all of less APIs.

bergundy (approved):
LGTM, didn't have any major comments.

dotnet/sdk-phase-2.md
cretz: :+1: This was also done client side for workflows, signals, and queries. This was also done in Python (kinda, in a slightly different way).

dotnet/sdk-phase-2.md
cretz: Yes, from the notes after this code:

```
* We may add helpers for getting individual memo values given a type to deserialize to
  * We don't expose a data converter to the user in a workflow at this time, so it is wise to make these helpers
```

dotnet/sdk-phase-2.md
cretz: Ok, I may make this change everywhere then.

dotnet/sdk-phase-2.md
cretz: It's less about verbosity and more about that "now" is just a clearer term across our SDKs. But I can change to `UtcNow` which may be a tad less clear/discoverable but that should be ok.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+    public string CurrentGreeting() => currentGreeting;
+}
+```
+
+Some things to not about the above code (much copied from phase 1):
+
+* Every workflow must have the non-inheritable `[Workflow]` attribute
+  * Optional `Name` positional property can be given. Default is the unqualified type name with the "I" prefix removed
+    if it's an interface and the next letter is also capital
+    * So what does `IDWorkflow` become? Yes, `DWorkflow` :-( Weighed as an acceptably rare cost for obvious benefit
```
Sushisource, line 143: I still maintain just having to type the name is better than the inevitable set of users that will run into this and be surprised.

dotnet/sdk-phase-2.md
Sushisource: I agree with Chad. I don't think we should even call such things "local activities" if we allow it. We should just say "you can disable the sandbox and do stuff from queries if you really want, just don't mutate state, pinky swear". I don't think making it a "real" local activity really buys people anything, but does wildly complicate internals.

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+  * Would love to put this on properties, but they aren't referenceable as delegates
+* Users are encouraged to set a `static` `readonly` `Ref` property for use by clients
+  * `Refs.Create<T>()` is how it is obtained
+    * `Refs` is in `Temporalio` not `Temporalio.Workflows`
+      * TODO(cretz): This acceptable?
+    * TODO(cretz): Should we do validation at this time on the attributes?
+      * Problem is instance activities will want to use `Refs` too.
+      * If I put `Refs` in `Temporalio.Workflows` then at least I can know that all uses are for workflows
+        * But then it'd clash with a `Temporalio.Activities.Refs` or do we want something like either
+          `Refs.ForWorkflow<T>` or `WorkflowRefs.Create<T>`?
```
Sushisource, line 177: I think `WorklowRefs.Create<T>` is best

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+
+    // Signal and query handlers
+
+    public bool TryGetSignalHandler(string name, out Delegate handler) { /*...*/ }
+
+    public void SetSignalHandler(string name, Func<Task> handler) { /*...*/ }
+
+    public void SetSignalHandler<T>(string name, Func<T, Task> handler) { /*...*/ }
+
+    public void RemoveSignalHandler(string name) { /*...*/ }
```
Sushisource, line 316: What happens with this and tryget (as well as for queries) if the handler was defined by `[WorkflowSignal]` ?

Sushisource (approved):

dotnet/sdk-phase-2.md
cretz: It's pretty normal to use the derived name by default (in .NET and in our SDKs). I don't think this is ever gonna matter in practice. This one gotcha is only for those that are intentionally violating .NET standards (something I've found is very rare for iface names).

dotnet/sdk-phase-2.md
cretz: Better than `Refs.Create<T>()`?

dotnet/sdk-phase-2.md
cretz: The handler map is built from the attributes but technically is mutable through these methods. So yes, you can technically remove a handler added via an attribute. And of course getting a handler by name works no matter how that handler was added. In that sense attribute-based handlers (our preference) are just sugar on this. It doesn't make much sense for a user to remove an attribute-added handler IMO, but at least it's a simple to understand concept. This is the same way it works in Python atm too. I hope nobody ever calls dynamic handler things FWIW.

dotnet/sdk-phase-2.md
bergundy: I think the reason why we do want to leverage activities here is for dependency injection.
I don't know if we'll want to call the API local activity, I agree that is confusing.
I do see a reason to eventually support returning tasks and also using a mechanism supported by the worker so we can avoid triggering the deadlock detector or in generally blocking the workflow thread.
We don't need to design this now though.
I was just pointing out that this might need to change and wanted to understand how hard it would be to change when needed.

dotnet/sdk-phase-2.md
cretz: Was decided to separate this into `WorkflowRefs` and `ActivityRefs` that will also validate the workflow annotations.

cretz:
Added updates based on design session. Specifically:

* Fixed typos
* For clarity use an instance-method for an activity
* Change `Refs` to `ActivityRefs` and `WorkflowRefs`
* Confirm `NonStickyToStickyPollRatio` and `MaxConcurrentWorkflowTaskPolls` will not be user-visible for now
* Add typed memo getters
* Conform to new typed search attribute proposal
* Changed `Workflow.Now` to `Workflow.UtcNow`
* Updated `ContinueAsNewException` to show it has to be created with a static function for type safety reasons

Will merge early next week if no more feedback comes

dotnet/sdk-phase-2.md
cretz: Fixed

dotnet/sdk-phase-2.md
cretz: Fixed

dotnet/sdk-phase-2.md
cretz: Added explicit API for this

dotnet/sdk-phase-2.md
cretz: Fixed

dotnet/sdk-phase-2.md
```diff
@@ -0,0 +1,484 @@
...
+        updates.Complete();
+        return Task.CompletedTask;
+    }
+
+    [WorkflowQuery]
+    public string CurrentGreeting() => currentGreeting;
+}
+```
+
+Some things to not about the above code (much copied from phase 1):
```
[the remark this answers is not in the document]
cretz: Fixed

dotnet/sdk-phase-2.md
cretz: Fixed

PR #78, opened 2023-03-11
WASM Core

Sushisource:
[Rendered](https://github.com/Sushisource/proposals/blob/wasm-core/core/WASM_core.md)

I encourage reading the doc in its entirety before adding review comments.

### DIFF
--- a/core/WASM_core.md
+++ b/core/WASM_core.md
@@ -19,6 +19,8 @@ A few problems a WASM Core Can solve, in no particular order:
   a bit easier (and, necessary) with it.
 * Auto-updates to core logic. This isn't detailed in this doc, but is a potentially very compelling
   feature. Imagine a worker that can update core logic without needing to be restarted.
+* Allow the use of lang-native gRPC clients (though core can still provide things like retry and 
+  backoff logic, etc.)
 
 
 # Detailed design
@@ -93,7 +95,9 @@ The more we can lean on WASI, the less we have to expand the Core<-->Lang interf
 thing in terms of both immediate work and long-term maintenance. There are some problems though:
 
 * Assuming we need to use `wazero` for non-CGo support, it doesn't appear to be able to handle
-  network calls. We'll test this, though.
+  network calls. We'll test this, though. Even if we intend to use native gRPC clients, being able
+  to make network calls in Core is still pretty desirable for things like auto-updates, test server
+  downloads, etc.
 * Calling back into lang gives us opportunities for more user customization, which can be nice.
   Especially when it comes to the metrics and tracing stuff. gRPC interceptors are also a potential
   plus.
@@ -134,6 +138,9 @@ fn func_name(receiver: ExternRef, arg: ArgType) -> ResultType;
 
 // -------------- Grpc -----------------
 
+// Would be passed into core along with client config options when creating instantiating a client.
+// Core can then augment the client with retry behavior, and the client can then be used when
+// instantiating Core workers.
 struct GrpcClientRef {
     /// Points to the host's gRPC client
     host_ref: ExternRef,
@@ -163,14 +170,9 @@ struct GrpcClientConfig {
     // continue to be handled in core.
 }
 
-// Error is stringly typed since matching on it doesn't appear immediately useful
-fn grpc_client_new(
-    config: GrpcClientConfig
-) -> Result<GrpcClientRef, String> {}
-
 impl GrpcClientRef {
     /// Make a unary grpc call
-    fn grpc_unary(
+    async fn grpc_unary(
         &self,
         req: GrpcRequest,
     ) -> Result<GrpcResponse, GrpcStatus> {}
--- a/core/WASM_core.md
+++ b/core/WASM_core.md
@@ -176,9 +176,6 @@ impl GrpcClientRef {
         &self,
         req: GrpcRequest,
     ) -> Result<GrpcResponse, GrpcStatus> {}
-
-    // Will be called by a `Drop` impl, but must exist and be exported by the WASM host.
-    fn close(self) {}
 }
 
 // -------------- Metrics --------------
--- a/core/WASM_core.md
+++ b/core/WASM_core.md
@@ -1,0 +1,350 @@
+# Summary
+
+This proposal intends to outline the shape of the interfaces and implementation work necessary
+to make Core WASM-compilable.
+
+# Motivation
+
+A few problems a WASM Core Can solve, in no particular order:
+
+* Allow Go and Java to be ported to run on top of Core. *Potentially* without the downsides of CGo
+  and JNI. More on that later.
+* No longer any need to ship compiled Core binaries of various architectures with our existing SDKs.
+  This is both a nice plus for us, in terms of reducing CI complexity, but also a substantial size
+  reduction in our packages.
+* Workers in the browser
+* Hosted workers (also possible without, but with, there are some nice benefits)
+* A simpler way to manage callbacks into the lang layer, enabling things like user-defined metrics
+  exporters, lower-level interceptors, etc. This is certainly doable without WASM, but is probably
+  a bit easier (and, necessary) with it.
+* Auto-updates to core logic. This isn't detailed in this doc, but is a potentially very compelling
+  feature. Imagine a worker that can update core logic without needing to be restarted.
+
+
+# Detailed design
+
+The first and largest chunk of work involved in implementation is more or less pure grunt work. 
+I need to switch Core to compiling with the wasm32-unknown-unknown target, and then fix all the
+compiler errors.
+
+There are a few predictable problems we'll definitely have to deal with:
+
+The largest hiccup is the use of network (gRPC) calls. There is more than one way we could skin this
+cat. Luckily (intentionally), all of this is contained to the `temporal-client` crate.
+
+The next major source of side effects come from the tracing and metrics exporting facilities for
+OTel and Prometheus.
+
+Additionally, there is the code dealing with downloading and starting the CLI / Java test server.
+
+Finally, there may be some issues with other things that in native world use syscalls like atomics,
+locks, random number gen, etc. Luckily my experience so far is that those pretty much all just work
+without any, or fairly minor, changes.
+
+Ultimately we have two strategies available to us. We can create interfaces for Core to call back
+into lang to perform all these functions, or we can attempt to leverage WASI for this functionality.
+WASI is quite immature, but *in theory* it can support these use cases.
+
+## WASM Runtime
+
+Regardless of the strategy, we need to choose a WASM runtime which each Lang SDK will use, so I'll
+address that first. The two major players in the space are
+[Wasmer](https://github.com/wasmerio/wasmer) and
+[Wasmtime](https://github.com/bytecodealliance/wasmtime). It's also not impossible that we might
+use one library from Rust for defining things like the host imported functions, but use different
+libraries for the runtime in the host langs.
+
+### Wasmer
+Wasmer's biggest pro is it's language support. Every language we currently have an SDK for (and
+more beyond that) has a Wasmer SDK for embedding and running WASM in that language. It's docs are
+decent, but not great.
+
+It has an MIT license.
+
+It's controlled by a startup (but hey, so are we) and has somewhat less active open-source
+development than Wasmtime.
+
+### Wasmtime
+Wasmtime supports all our languages _except_ Java, which is unfortunate, though it looks like
+there's an unofficial repo. It's also not immediately clear how usable JS/Node support is, 
+particularly when it comes to using WASI. It does have quite excellent docs, and built-in support
+for calling async functions from WASM, which is pretty awesome.
+
+It has an Apache 2.0 license.
+
+It's owned and maintained by the bytecode alliance, which also owns the WASM/WASI spec. It's backed
+by Mozilla.
+
+### Oh no CGo (and JNI)
+There's a big downside to both Wasmer and Wasmtime. Both of their Go implementations use CGo, and
+Wasmer's Java implementation uses JNI. This blows one of the bigger pluses out of the water.
+Luckily, [wazero](https://github.com/tetratelabs/wazero) exists, which we can use to get a pure Go
+environment. It supports WASI as well, but seemingly [not what we would need for network
+calls](https://wazero.io/specs/).
+
+I can't seem to find anything reasonable for Java. Maybe if we're blessed, such a thing
+will exist by the time we get around to porting Java, but that's a huge if.
+
+## Implementation Strategy
+
+### How much, if any WASI to use?
+
+The more we can lean on WASI, the less we have to expand the Core<-->Lang interface. That's a good
+thing in terms of both immediate work and long-term maintenance. There are some problems though:
+
+* Assuming we need to use `wazero` for non-CGo support, it doesn't appear to be able to handle
+  network calls. We'll test this, though.
+* Calling back into lang gives us opportunities for more user customization, which can be nice.
+  Especially when it comes to the metrics and tracing stuff. gRPC interceptors are also a potential
+  plus.
+
+My inclination for now is to at least *investigate* how much we can get out of WASI. Particularly
+with respect to running from Go.
+
+Unfortunately, my assumption is that we will need to callback into Lang for all the usecases
+mentioned at the top of the doc. The only sys interface things it appears we'll be able to reliably
+use are clocks (which is good, that does come up) and possibly file I/O, which currently does not
+happen in Core (and probably never should) except for in the test server usecase.
+
+
+### Core API changes
+Operating under the assumption that all the major syscalls will need to be handled through callbacks
+into lang. In WASM, calling functions in the host is supported via `import`ing those host functions.
+Ultimately these functions may be bundled together into a trait or something on the Core side, but
+this doc will focus on the individual functions.
+
+Here are the functions I think we'll need to add to the Core<-->Lang interface, beyond the existing
+ones that lang already calls. They're defined as they would appear in Rust, after being imported.
+This is definitely a rough draft, as it will be influenced by runtime selection and whatever
+practical problems we run into along the way.
+
+```rust
+// In reality, everything in these functions is going to need to be serialized and shuttled back
+// and forth as bytes. That makes for uninteresting signatures, though, so instead I show what
+// those signatures would look like if we were using deserialized types directly. The exact details
+// of how that serialization happens are intentionally not included here.
+//
+// WASM has an `ExternRef` type which is a pointer to a host-owned opaque object.
+
+// For all the types which wrap an `ExternRef`, the implementations on those types which take
+// `self` will be implemented as calls to a host-defined function which looks like:
+fn func_name(receiver: ExternRef, arg: ArgType) -> ResultType;
+// There may be a more clever way to manage some of this by using `Table` and `Func` types along with
+// `ExternRef` but I need to play around more to determine that.
+
+// -------------- Grpc -----------------
+
+struct GrpcClientRef {
+    /// Points to the host's gRPC client
+    host_ref: ExternRef,
+}
+
+struct GrpcRequest {
+    uri: String,
+    body: Vec<u8>,
+    metadata: Metadata,
+}
+
+struct GrpcResponse {
+    body: Vec<u8>,
+    metadata: Metadata,
+}
+
+struct GrpcStatus {
+    // Serializable equivalent of https://docs.rs/tonic/latest/tonic/struct.Status.html
+}
+
+struct Metadata {
+    data: HashMap<String, Vec<u8>>,
+}
+
+struct GrpcClientConfig {
+    // Details elided. See `ClientConfig` in core. Would *not* include `RetryConfig`, as that will
+    // continue to be handled in core.
+}
+
+// Error is stringly typed since matching on it doesn't appear immediately useful
+fn grpc_client_new(
+    config: GrpcClientConfig
+) -> Result<GrpcClientRef, String> {}
+
+impl GrpcClientRef {
+    /// Make a unary grpc call
+    fn grpc_unary(
+        &self,
+        req: GrpcRequest,
+    ) -> Result<GrpcResponse, GrpcStatus> {}
+
+    // Will be called by a `Drop` impl, but must exist and be exported by the WASM host.
+    fn close(self) {}
+}
+
+// -------------- Metrics --------------
+
+// This is mostly a simplified version of the OTel API, which, ideally we can use to implement on the
+// lang side. https://opentelemetry.io/docs/reference/specification/metrics/
+// Given this is probably our easiest common denominator, it makes sense to stick close to it.
+// At the moment, all metrics are sync rather than async (callback based).
+
+struct MeterProviderRef {
+    host_ref: ExternRef,
+}
+
+// I anticipate that each worker will live in its own module instance, so there's no need to pass
+// some kind of worker reference here. If that doesn't work, we'll need one.
+fn get_meter_provider() -> MeterProviderRef {}
+
+impl MeterProviderRef {
+    fn new_counter(&self, name: String) -> CounterRef {}
+    // We deviate from OTel here by providing a histogram's buckets at instrument instantiation.
+    // Adhering to the OTel spec would seem to overcomplicate things substantially here. See
+    // https://opentelemetry.io/docs/reference/specification/metrics/sdk/#aggregation
+    fn new_histogram(&self, name: String, buckets: Buckets) -> HistogramRef {}
+    fn new_gauge(&self, name: String) -> GaugeRef {}
+}
+
+
+struct MetricsAttributesRef {
+    host_ref: ExternRef,
+}
+
+struct MetricsAttributesOptions {
+    attributes: Vec<MetricKeyValue>,
+}
+
+struct MetricKeyValue {
+    key: String,
+    value: MetricValue,
+}
+
+enum MetricValue {
+    String(String),
+    Int(i64),
+    Float(f64),
+    Bool(bool),
+    // Maybe array type, but I don't use that anywhere currently.
+}
+
+impl MetricsAttributesRef {
+    fn new(options: MetricsAttributesOptions) -> Self {}
+    fn with_kvs(&self, kvs: Vec<MetricKeyValue>) -> MetricsAttributesRef {}
+}
+
+struct CounterRef {
+    host_ref: ExternRef,
+}
+
+struct HistogramRef {
+    host_ref: ExternRef,
+}
+
+struct GaugeRef {
+    host_ref: ExternRef,
+}
+
+impl CounterRef {
+    fn add(&self, value: u64, attributes: MetricsAttributesRef) {}
+}
+
+impl HistogramRef {
+    // When referring to durations, this value is in millis
+    fn record(&self, value: u64, attributes: MetricsAttributesRef) {}
+}
+
+impl GaugeRef {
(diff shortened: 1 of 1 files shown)


core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+locks, random number gen, etc. Luckily my experience so far is that those pretty much all just work
+without any, or fairly minor, changes.
+
+Ultimately we have two strategies available to us. We can create interfaces for Core to call back
+into lang to perform all these functions, or we can attempt to leverage WASI for this functionality.
+WASI is quite immature, but *in theory* it can support these use cases.
+
+## WASM Runtime
+
+Regardless of the strategy, we need to choose a WASM runtime which each Lang SDK will use, so I'll
```
bergundy, line 50: For TS, V8 can run WASM directly, there's no need for another WASM runtime.
For Go, I'm leaning towards wazero because of the CGo argument.
We'll need to assess what to do for Java but that will come in a later stage, by which the landscape may change.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+will exist by the time we get around to porting Java, but that's a huge if.
+
+## Implementation Strategy
+
+### How much, if any WASI to use?
+
+The more we can lean on WASI, the less we have to expand the Core<-->Lang interface. That's a good
+thing in terms of both immediate work and long-term maintenance. There are some problems though:
+
+* Assuming we need to use `wazero` for non-CGo support, it doesn't appear to be able to handle
```
bergundy, line 95: I think a goal for Go and TS is to be able to use the native gRPC clients instead of adding a Rust backed client implementation.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+  Especially when it comes to the metrics and tracing stuff. gRPC interceptors are also a potential
+  plus.
+
+My inclination for now is to at least *investigate* how much we can get out of WASI. Particularly
+with respect to running from Go.
+
+Unfortunately, my assumption is that we will need to callback into Lang for all the usecases
+mentioned at the top of the doc. The only sys interface things it appears we'll be able to reliably
+use are clocks (which is good, that does come up) and possibly file I/O, which currently does not
+happen in Core (and probably never should) except for in the test server usecase.
```
bergundy, line 107: I'm okay not having support for WASM-Core based ephemeral servers initially. Don't think it's worth the complexity.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+
+struct GrpcClientRef {
+    /// Points to the host's gRPC client
+    host_ref: ExternRef,
+}
+
+struct GrpcRequest {
+    uri: String,
+    body: Vec<u8>,
+    metadata: Metadata,
```
bergundy, line 145: What about deadline?

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+// -------------- Grpc -----------------
+
+struct GrpcClientRef {
+    /// Points to the host's gRPC client
+    host_ref: ExternRef,
+}
+
+struct GrpcRequest {
+    uri: String,
+    body: Vec<u8>,
```
bergundy, line 144: I'd verify that gRPC interceptors in the different languages (JS / Go for now) will work with raw bodies instead of messages.

core/WASM_core.md
Sushisource: The `grpc-timeout` header specifies the deadline per https://github.com/grpc/grpc/blob/master/doc/PROTOCOL-HTTP2.md

core/WASM_core.md
bergundy: I checked, it seems like bytes is what we need here.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+}
+
+// Error is stringly typed since matching on it doesn't appear immediately useful
+fn grpc_client_new(
+    config: GrpcClientConfig
+) -> Result<GrpcClientRef, String> {}
+
+impl GrpcClientRef {
+    /// Make a unary grpc call
+    fn grpc_unary(
```
bergundy, line 173: nit: This will need to be async or accept a callback.

bergundy: In `grpc-js` the signature of unary request is:

```ts
function makeUnaryRequest<RequestType, ResponseType>(
  method: string,
  serialize: (value: RequestType) => Buffer,
  deserialize: (value: Buffer) => ResponseType,
  argument: RequestType, metadata: Metadata,
  options: CallOptions,
  callback: UnaryCallback<ResponseType>
): ClientUnaryCall;

```

```ts
export interface CallOptions {
    deadline?: Deadline;
    host?: string;
    parent?: ServerUnaryCall<any, any> | ServerReadableStream<any, any> | ServerWritableStream<any, any> | ServerDuplexStream<any, any>;
    propagate_flags?: number;
    credentials?: CallCredentials;
    interceptors?: Interceptor[];
    interceptor_providers?: InterceptorProvider[];
}
```

Let's make sure we can use this and the Go / Java gRPC libs.
Can adjust post proposal when we get to implementation, but I'd make compatibility with those clients a goal.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+
+Flatbuffer's design is more oriented towards what we're doing, whereas Cap'n'Proto is more concerned
+with sending things over the wire, but is perhaps a bit more ergonomic.
+
+In either case, the primary pain point would be figuring out how to deal with the annoyance of
+having lifetime annotations in the Rust code. Cap'n'Proto has built-in support for creating owned
+versions of the structs which makes that pretty easy, but at that point it's unclear if whatever
+speedup we'd get is worth it compared to Protobuf.
+
+Ultimately, it may makes sense to defer this work until we have measured performance and decided
```
bergundy, line 317: It's already going to be painful to move away from protobuf having 4 Lang implementations, I'd love for us to do this optimization but I feel like it's not time yet and shouldn't be coupled to this effort.

core/WASM_core.md
bergundy: Lang impls would have to pop from the metadata since both Go and JS specify the deadline separately.
That's fine but not optimal from the lang perspective.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+    body: Vec<u8>,
+    metadata: Metadata,
+}
+
+struct GrpcStatus {
+    // Serializable equivalent of https://docs.rs/tonic/latest/tonic/struct.Status.html
+}
+
+struct Metadata {
+    data: HashMap<String, Vec<u8>>,
```
bergundy, line 158: You won't be able to use HashMaps in the boundary.
Might be worth using a newline delimited byte array as Chad did in the .NET->Core bridge.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+
+## Implementation Strategy
+
+### How much, if any WASI to use?
+
+The more we can lean on WASI, the less we have to expand the Core<-->Lang interface. That's a good
+thing in terms of both immediate work and long-term maintenance. There are some problems though:
+
+* Assuming we need to use `wazero` for non-CGo support, it doesn't appear to be able to handle
+  network calls. We'll test this, though.
```
bergundy, line 96: Also note that if we want to use WASM Core in the browser we can't make direct gRPC calls and need a grpc-web or grpc-gateway based client.

core/WASM_core.md
Sushisource: Can you show me the APIs you're thinking of using? This looks like it https://nodejs.dev/en/learn/nodejs-with-webassembly/

Which also mentions there's no WASI implementation. We may very well need at least basic WASI support for basic syscalls - so we may need to pick Wasmtime/Wasmer here anyway.

core/WASM_core.md
Sushisource: Yep, that's sort of hinted at behind:

> A simpler way to manage callbacks into the lang layer

But yes that's an explicit plus & I'll add a bullet about it

core/WASM_core.md
Sushisource: This is mentioned above.
```
// In reality, everything in these functions is going to need to be serialized and shuttled back
// and forth as bytes. That makes for uninteresting signatures, though, so instead I show what
// those signatures would look like if we were using deserialized types directly.
```

core/WASM_core.md
Sushisource: That doesn't really get at my more important second point, though, which is I think we should at least try this out and see what the perf difference is. If it's hugely substantial, we probably want to just pay this cost now, because changing later is more expensive than doing it together.

core/WASM_core.md
bergundy: Node has experimental WASI support: https://nodejs.org/api/wasi.html
Browser has this polyfill: https://wasi.dev/polyfill/
Using wasmtime or wasmer will require CGo, which is a no go (pun intended) IMHO.

I think it's worth exploring if we can avoid any syscalls other than memory allocation. We can easily handle memory allocation from the host.

core/WASM_core.md
bergundy: I think this requires discussion because of the required level of investment.

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+Here are the functions I think we'll need to add to the Core<-->Lang interface, beyond the existing
+ones that lang already calls. They're defined as they would appear in Rust, after being imported.
+This is definitely a rough draft, as it will be influenced by runtime selection and whatever
+practical problems we run into along the way.
+
+```rust
+// In reality, everything in these functions is going to need to be serialized and shuttled back
+// and forth as bytes. That makes for uninteresting signatures, though, so instead I show what
+// those signatures would look like if we were using deserialized types directly. The exact details
+// of how that serialization happens are intentionally not included here.
```
Quinn-With-Two-Ns, line 125: There is a lot of work in this area for calling into and out of the wasm runtime, but I guess non of it is mature enough?

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+
+Finally, there may be some issues with other things that in native world use syscalls like atomics,
+locks, random number gen, etc. Luckily my experience so far is that those pretty much all just work
+without any, or fairly minor, changes.
+
+Ultimately we have two strategies available to us. We can create interfaces for Core to call back
+into lang to perform all these functions, or we can attempt to leverage WASI for this functionality.
+WASI is quite immature, but *in theory* it can support these use cases.
+
+## WASM Runtime
```
Quinn-With-Two-Ns, line 48: is the idea we use one runtime for every lang or use different runtimes for different languages?

core/WASM_core.md
```diff
@@ -0,0 +1,350 @@
...
+    data: HashMap<String, Vec<u8>>,
+}
+
+struct GrpcClientConfig {
+    // Details elided. See `ClientConfig` in core. Would *not* include `RetryConfig`, as that will
+    // continue to be handled in core.
+}
+
+// Error is stringly typed since matching on it doesn't appear immediately useful
+fn grpc_client_new(
```
Quinn-With-Two-Ns, line 167: Why does core need to create a client? Shouldn't core be given a client users have created?

core/WASM_core.md
Sushisource: Ideally, one, so there's less cognitive load. The reality is it seems clear that at minimum Go will be an exception.

core/WASM_core.md
Sushisource: Excellent point. That is indeed the case. I will revise.

I'm trying to think if there might be some exception to that, but even stuff like checking server capabilities can happen at worker startup time or whatever.

core/WASM_core.md
Sushisource: Yeah, honestly there's so much stuff going on and it's so new it's hard to even keep track of what's implemented / just an idea / etc.

Like there's this new proposal here https://github.com/WebAssembly/component-model/blob/main/design/mvp/Explainer.md which apparently supersedes interface types, which was the old thing.

I think, at the end of the day, we're going to have to do stuff the manual way until something mature (and by mature, I mean even implemented at all) emerges from the soup.

core/WASM_core.md
Sushisource: Yeah allocation is a given, and IME works everywhere without issue so far. The open questions for me are things like "Do locks and atomics work?" etc.

core/WASM_core.md
Quinn-With-Two-Ns: Yeah Go SDK checks most capabilities at startup ( I think there is one specifically clients need to know)

core/WASM_core.md
Sushisource: Yeah, of course. If we need to lift more things up to be explicit we can do that as-needed, but keeping it as thin as possible is a sensible goal I think.

core/WASM_core.md
Sushisource: Yep we can adjust as needed once we start digging in. Closing this thread since it effectively is the same as https://github.com/temporalio/proposals/pull/78#discussion_r1136328810

core/WASM_core.md
bergundy: I think it’s worth considering if we actually need multithreaded WASM core. Based on some initial digging it might not be straightforward to set up consistently across the different hosts.
V8 gives atomics that can be used to implement the other synchronization primitives and by leveraging multiple worker threads with shared memory you can build a multi threaded runtime.

core/WASM_core.md
bergundy: Looks like our first two targets are Go and TS, we should use the WASM runtime that works best for each use case.

core/WASM_core.md
Sushisource: We probably don't really - and core is already largely not single threaded. The problem is actually expressing that in the type system means converting a bunch of things that are still `&self` currently to `&mut self`. Which means splitting the polling and completing halves of core into two, etc.

IE: Atomics and locks are needed not because of multi-threading, but because it is not encoded into the types that the _caller_ is single threaded, and even then you sometimes need synchronization between multiple mutable borrows on one thread. That never needs atomics (it could be RefCell instead), but a lot of the time it's easier to just use an atomic.

Regardless we'll just need to see what the support is like.

Quinn-With-Two-Ns: What support do we need from our wasm runtime for atomics if it isn't actually being used and is just for the type system? It isn't super obvious to me if `wazero` supports threads/atomics

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+impl GrpcClientRef {
+    /// Make a unary grpc call
+    async fn grpc_unary(
+        &self,
+        req: GrpcRequest,
+    ) -> Result<GrpcResponse, GrpcStatus> {}
+
+    // Will be called by a `Drop` impl, but must exist and be exported by the WASM host.
+    fn close(self) {}
```
Quinn-With-Two-Ns, line 181: What would core use this for?

core/WASM_core.md
Sushisource: If I'm using them, even if it's just for the type system, then they're needed and those calls are made.

I'm pretty sure wazero supports them, it seems fairly complete. Probably even the v8 one too, we just need to check. Typically they'll just be implemented with spinlocks or something if the actual underlying runtime isn't turning them into syscalls or the right CPU instruction.

bergundy: If we use a single threaded runtime all of the atomic operations are effectively noops apart from notify / signal.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+This proposal intends to outline the shape of the interfaces and implementation work necessary
+to make Core WASM-compilable.
+
+# Motivation
+
+A few problems a WASM Core Can solve, in no particular order:
+
+* Allow Go and Java to be ported to run on top of Core. *Potentially* without the downsides of CGo
+  and JNI. More on that later.
+* No longer any need to ship compiled Core binaries of various architectures with our existing SDKs.
```
cretz, line 12: While there may no longer be a "need", there's surely still considerable performance improvement to doing so (at least for .NET and Python). For any existing core-based SDK the tradeoffs of course should be discussed because I don't think we want to maintain two approaches in any SDK.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+  This is both a nice plus for us, in terms of reducing CI complexity, but also a substantial size
+  reduction in our packages.
```
cretz, line 14: > but also a substantial size reduction in our packages

Will be curious what the size of the WASM blob is. I think we may find ourselves wanting to reduce/externalize some third party Rust dependencies.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+There are a few predictable problems we'll definitely have to deal with:
+
+The largest hiccup is the use of network (gRPC) calls. There is more than one way we could skin this
+cat. Luckily (intentionally), all of this is contained to the `temporal-client` crate.
+
+The next major source of side effects come from the tracing and metrics exporting facilities for
+OTel and Prometheus.
+
+Additionally, there is the code dealing with downloading and starting the CLI / Java test server.
```
cretz, line 40: I think this can be removed from the WASM version of core. If it needs to be moved out of crate or put behind a "feature" so be it. The only main benefit of sharing this code is for the HTTP and disk and process features which don't apply to WASM core. Only langs using Rust get benefit from this, other langs (e.g. Go) have their own download/extract/run code.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+Finally, there may be some issues with other things that in native world use syscalls like atomics,
+locks, random number gen, etc. Luckily my experience so far is that those pretty much all just work
+without any, or fairly minor, changes.
+
+Ultimately we have two strategies available to us. We can create interfaces for Core to call back
+into lang to perform all these functions, or we can attempt to leverage WASI for this functionality.
+WASI is quite immature, but *in theory* it can support these use cases.
```
cretz, line 48: I vote no WASI. A set of well-defined C-style externs for each direction is going to give us the smallest, nicest integration I think.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+## WASM Runtime
+
+Regardless of the strategy, we need to choose a WASM runtime which each Lang SDK will use, so I'll
```
cretz, line 52: This can all be up to lang, no need to be in proposal IMO (though I know this is more of a "landscape survey")

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+### Oh no CGo (and JNI)
+There's a big downside to both Wasmer and Wasmtime. Both of their Go implementations use CGo, and
+Wasmer's Java implementation uses JNI. This blows one of the bigger pluses out of the water.
+Luckily, [wazero](https://github.com/tetratelabs/wazero) exists, which we can use to get a pure Go
+environment. It supports WASI as well, but seemingly [not what we would need for network
+calls](https://wazero.io/specs/).
+
+I can't seem to find anything reasonable for Java. Maybe if we're blessed, such a thing
+will exist by the time we get around to porting Java, but that's a huge if.
```
cretz, line 88: Best developer I ever worked with wrote [asmble](https://github.com/cretz/asmble) which compiles WASM to literal `.class` files in Java. He doesn't maintain it much, but it's high quality and likely will work well for us barring benchmarks.

All non-V8 languages should shoot for AOT-based WASM IMO.

core/WASM_core.md
cretz: Concur (mentioned above)

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+* Assuming we need to use `wazero` for non-CGo support, it doesn't appear to be able to handle
+  network calls. We'll test this, though. Even if we intend to use native gRPC clients, being able
+  to make network calls in Core is still pretty desirable for things like auto-updates, test server
+  downloads, etc.
+* Calling back into lang gives us opportunities for more user customization, which can be nice.
+  Especially when it comes to the metrics and tracing stuff. gRPC interceptors are also a potential
+  plus.
+
+My inclination for now is to at least *investigate* how much we can get out of WASI. Particularly
+with respect to running from Go.
```
cretz, line 106: I don't think there's enough win for WASI. WASI has benefit if you need a POSIX-like abstraction which I don't think we do. Can just send extern/repr(C) structs in each direction as I'm doing with .NET. Many of the things we're doing are byte arrays or UTF-8 strings anyways.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+// Would be passed into core along with client config options when creating instantiating a client.
+// Core can then augment the client with retry behavior, and the client can then be used when
+// instantiating Core workers.
+struct GrpcClientRef {
+    /// Points to the host's gRPC client
+    host_ref: ExternRef,
+}
+
+struct GrpcRequest {
+    uri: String,
```
cretz, line 150: What is URI here? I think this should be two strings, "service" and "method" in gRPC parlance. Most users of this will have to have a big switch statement for this to make typed calls from their lang (though many gRPC impls have a https://pkg.go.dev/google.golang.org/grpc#ClientConn.Invoke like dynamic call, you often still need to know the response type). And since this would be one big switch, if you wanted to instead make gRPC request service+method one enum of needed methods, that may work better than string.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+    // Details elided. See `ClientConfig` in core. Would *not* include `RetryConfig`, as that will
+    // continue to be handled in core.
+}
+
+impl GrpcClientRef {
+    /// Make a unary grpc call
+    async fn grpc_unary(
+        &self,
+        req: GrpcRequest,
+    ) -> Result<GrpcResponse, GrpcStatus> {}
```
cretz, line 178: I think you'll be simplifying these types when you start making them "extern" to be exported from your WASM module. Specifically, just put the gRPC status into the response.

And of course you'll end up needing a two-function approach so you'll need some kind of request identifier. Langs will impl a "submit gRPC request" and core will impl a "on gRPC response" to be called by lang.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+amusing.
+
+The options here are Cap'n'Proto or Flatbuffers. Both appear to have support for all the languages
+we care about, except for Flatbuffers not having a Ruby implementation.
+
+Flatbuffer's design is more oriented towards what we're doing, whereas Cap'n'Proto is more concerned
+with sending things over the wire, but is perhaps a bit more ergonomic.
+
+In either case, the primary pain point would be figuring out how to deal with the annoyance of
+having lifetime annotations in the Rust code. Cap'n'Proto has built-in support for creating owned
```
cretz, line 315: > In either case, the primary pain point would be figuring out how to deal with the annoyance of having lifetime annotations in the Rust code

Byte array and object lifetime in general are the biggest problem with this issue unrelated to Rust (though Rust puts the problem front and center). Since gRPC obviously deals only in proto, this only has benefits in the worker activation/task/complete side. And then you even reference some existing API protos you'd have to copy. But the proto bytes and the flat/capn bytes for activations/tasks are still owned by Rust until freed, and lang will need to do things like convert to strings and ints from the binary format anyways, so I don't think you are saving as much as you think delaying the conversion to scalars as needed during activation/task vs upfront into protos. You are only saving lang a bit of memory (and only kinda because it is a potentially very delayed "byte array free" call on flat/capn that we can make immediately after conversion for proto).

While this may be an interesting experiment, I am not sure even the POC would have much value compared to the effort heh.

core/WASM_core.md
```diff
@@ -0,0 +1,352 @@
...
+
+Ultimately, it may makes sense to defer this work until we have measured performance and decided
+it's worth it. Of course, the rub there is adding this support later is a pretty monster change.
+
+What I do think we should take some time to do is make a toy application with the runtime we land on
+(say, with Go as the host), and then just shuffle back and forth some Workflow/Activity tasks using
+proto and using Cap'n/Flatbuffer and measure performance. Unless there's a striking difference, we
+can stick with proto.
+
+# Drawbacks
```
cretz, line 327: Another big one to add/discuss: "fixed memory size". WASM modules are given a fixed-size memory to work with by the WASM runtime (they have pre-defined fixed-size memory blocks and can "import" memory I believe which the runtime can provide, but both are a fixed size IIRC). This number basically has to get set properly from the beginning and we may/should expose it to users, but it's definitely not dynamic. Guess to high and it's unused memory (most WASM runtimes prealloc this, it's not dynamic), guess too low and core will probably panic I figure when it tries to alloc.

cretz (approved):
LGTM as is. But I think for a real substantive proposal, you're gonna have to develop that "core<-->lang" WASM iface. That's where the real value/challenge is.

core/WASM_core.md
bergundy: I wasn't sure if we could use raw bytes for the inputs and outputs, seems like you're saying we can't.

cretz: We absolutely can use raw bytes to/from core (and should, we only want a single gRPC invoke call). But most lang sides probably will have a big switch statement on the method name to convert to protos they are aware of. I think most (all?) gRPC client libraries intentionally hide the byte-level communication so they have complete control over the protocol.

core/WASM_core.md
bergundy: Rust allows override the `#[global_allocator]`.
I've seen `buddy_alloc` used for that purpose with pre-allocated memory.
But as you've mentioned, the runtime may also provide memory blocks.
AFAICT, externally provided memory can be use for dynamic allocation. See for reference https://developer.mozilla.org/en-US/docs/WebAssembly/JavaScript_interface/Memory.

cretz: :+1: I was misremembering here based on a JVM limitation I had with `ByteBuffer` (which cannot dynamically grow)

core/WASM_core.md
Sushisource: Well the main point, the same point of all of core, is to just have fewer places we're duping logic. Where to download from, already downloaded checking, etc etc. In the long run, it's definitely worth centralizing to the extent we can actually save code by doing so. It's also not much of a priority.

core/WASM_core.md
Sushisource: We *can*, of course, and probably will because WASI isn't very mature, but I think it's pretty much impossible to disagree with the statement that if I could compile core to WASM more or less as it is today and all the networking stuff worked, then we'd likely just do that. It saves on nearly all the work that needs to be done.

Now, that isn't the case, and it certainly seems like we'll have to go with passing back and forth, but if that assumption is wrong it's something worth knowing and worth spending a day to find out.

core/WASM_core.md
cretz: But some things we remove duplication of "logic" that is WASM-able and some we remove duplication of "logic" that is just system stuff. This download, extract, run thing is good for those using Rust, not good for anyone else like WASM users.

core/WASM_core.md
cretz: :+1: There is no way for gRPC networking to work and for your goal of "Workers in the browser" to be accomplished, so the intermediary is required anyways. I am also hoping for a much smaller WASM blob when tonic, hyper, etc are not included.

core/WASM_core.md
cretz: I think this big switch is fine (some langs can employ code gen or reflection if they must). Benefit of using the gRPC client from the outside with its expected methods and req/resp types is that you get all the other gRPC client benefits such as interceptors and advanced dial/call options.

bergundy (approved):

Sushisource:
Investigatory work on the implementation is started. There are some internal notion docs about that.

PR #79, opened 2023-05-19
HTTP API

cretz:
HTTP API on top of gRPC API.

See [rendered](https://github.com/cretz/temporal-proposals/blob/http-api/api/http-api.md).

Please review.

### DIFF
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -1,0 +1,176 @@
+# HTTP API
+
+## Overview
+
+Temporal has a gRPC API but no HTTP API. We need an HTTP API.
+
+**Goals**
+
+* Create HTTP API that governs all reasonable user-facing calls
+* Ensure new API calls are automatically adopted
+* Expose HTTP API in server by default for all to use like gRPC, but have option to disable
+* Support all existing interceptors and auth methods
+
+**Non-Goals**
+
+* Design a high-quality REST API from first principles
+* grpc-web (browser-only gRPC workaround that requires server-side support)
+* Make an easy call to get workflow result
+
+## Approach
+
+This project will use https://github.com/grpc-ecosystem/grpc-gateway and annotations inspired by work already done on
+the UI server for similar purpose at https://github.com/temporalio/ui-server. After review of the UI server's
+annotations, they so closely match our goals that many things won't change.
+
+Since namespace is required for most calls and is usually URL safe, we have decided to put the namespace in the URL
+instead of as an HTTP header. This makes URL's large and a bit unwieldy, but calls like from `curl` would have to have
+`-H Temporal-Namespace: <my-ns>` which is even more unwieldy.
+
+The bodies of requests are proto JSON of the request messages and responses are full proto JSON of the response
+messages. While we try to somewhat follow REST principles, this is not a full RESTful API necessarily. Developers adding
+new API will need to add the gRPC gateway annotations for their RPC calls.
+
+For the initial version, we will only be doing workflow service. Operator service and cloud APIs may come soon after
+though.
+
+### Rules
+
+* Naming
+  * Multi-word URL paths use kebab-case
+  * Multi-word query parameters use camelCase (only first letter of abbreviations capitalized if not first word)
+  * Do not assume any names reserved. So we can't have `/workflows/{workflow_id}` and `/workflows/general-call` in case
+    there's a workflow named `general-call`.
+  * Use plural nouns where an identifier may be provided to differentiate
+    * But singular nouns are not _required_ in other places, can still be plural
+* Response codes
+  * 201 success if a creation
+  * 202 if action eventually consistent
+  * 200 other all other successes
+  * Do not use 204 for no response, even if there's no response because we may add response later, just 200 everything
+* Methods:
+  * `GET` if read-only
+    * Noun as last URL term
+    * Use query parameters for commonly set items to avoid relying on request bodies (but still rely on bodies as
+      needed). This may require some gRPC gateway customization.
+  * `PUT` if strongly consistent add/update on fixed-URL w/ idempotency
+    * Noun as last URL term
+  * `DELETE` if item is actually deleted (i.e. unreachable from `GET`)
+    * Noun as last URL term
+  * `POST` for everything else
+    * Noun as last URL term if for a create, verb if for an action
+  * `PATCH` is not needed because we don't adhere to common CRUD
+    * TODO(cretz): Needs discussion
+
+### RPCs
+
+Below is list of gRPC calls to HTTP calls so far and some notes/questions beneath each. Some notes/questions beneath one
+actually apply to many but we are not repeating ourselves too much here.
+
+* Namespaces:
+  * `RegisterNamespace` - `POST /api/v1/namespaces`
+    * 202 on success since eventually consistent
+  * `DescribeNamespace` - `GET /api/v1/namespaces/{namespace}`
+  * `ListNamespaces` - `GET /api/v1/namespaces`
+  * `UpdateNamespace` - `PUT /api/v1/namespaces/{namespace}`
+    * TODO(cretz): Confirm strongly consistent and idempotent or not
+  * `DeprecateNamespace` - `DELETE /api/v1/namespaces/{namespace}`
+    * Is this eventually consistent? Will determine response code
+    * Does this prevent returning from `GET`? Will determine HTTP method
+* Workflows:
+  * `StartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
+    * 204 on success
+    * Should accept header as `identity` on these calls?
+    * Ok to be at root URL here instead of appending `/start`?
+  * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
+    * Accept `?nextPageToken` as query param? I'm thinking so
+  * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
+  * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
+    * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal`
+    * Should we put signal name in the URL? Can be optional (e.g. in URL or body)
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
+  * `ResetWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/reset`
+  * `TerminateWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/terminate`
+  * `ListOpenWorkflowExecutions` - Intentionally not exposed
+  * `ListClosedWorkflowExecutions` - Intentionally not exposed
+  * `ListWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflows`
+    * Accepts `?query=` query parameter
+  * `ListArchivedWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/archived-workflows`
+    * Intentionally did not do `/workflows/archived` due to ambiguity
+  * `CountWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflow-count`
+  * `QueryWorkflow` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/query`
+    * Should we put query name in the URL? Can be optional (e.g. in URL or body)
+  * `DescribeWorkflowExecution` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
+  * `UpdateWorkflowExecution`- `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}`
+    * Decided to make this an identity so we can have "describe update" or "cancel update" or whatever one day
+  * `PollWorkflowExecutionUpdate` - Intentionally not exposed
+    * Not sure if best to have long HTTP API here. Would prefer non-blocking describe and maybe one day the difference
+      between describing an update and polling for an update is a "wait" boolean.
+* Activities:
+  * `RecordActivityTaskHeartbeat` - `POST /api/v1/namespaces/{namespace}/activities/heartbeat`
+    * We consider "heartbeat" a verb for these purposes
+  * `RecordActivityTaskHeartbeatById` - `POST /api/v1/namespaces/{namespace}/activities/heartbeat-by-id`
+    * We are intentionally not doing `/activities/{activity_id}/heartbeat` or similar due to URL ambiguity and that we
+      want to encourage the task token approach
+  * `RespondActivityTaskCompleted` - `POST /api/v1/namespaces/{namespace}/activities/complete`
+  * `RespondActivityTaskCompletedById` - `POST /api/v1/namespaces/{namespace}/activities/complete-by-id`
+  * `RespondActivityTaskFailed` - `POST /api/v1/namespaces/{namespace}/activities/fail`
+  * `RespondActivityTaskFailedById` - `POST /api/v1/namespaces/{namespace}/activities/fail-by-id`
+  * `RespondActivityTaskCanceled` - `POST /api/v1/namespaces/{namespace}/activities/cancel`
+  * `RespondActivityTaskCanceledById` - `POST /api/v1/namespaces/{namespace}/activities/cancel-by-id`
+* Search attributes:
+  * Intentionally not exposed high-level search attribute calls from workflow service
+* Task polling/completing:
+  * Intentionally not exposed (except for activity task stuff needed for async activities)
+* Task queues:
+  * `ResetStickyTaskQueue` - Intentionally not exposed
+  * `DescribeTaskQueue` - `GET /api/v1/namespaces/{namespace}/task-queues/{task_queue_name}`
+  * `ListTaskQueuePartitions` - Intentionally not exposed
+* Schedules:
+  * `CreateSchedule` - `POST /api/v1/namespaces/{namespace}/schedules/{schedule_id}`
+  * `DescribeSchedule` - `GET /api/v1/namespaces/{namespace}/schedules/{schedule_id}`
+  * `UpdateSchedule` - `POST /api/v1/namespaces/{namespace}/schedules/{schedule_id}/update`
+    * This is not a `PUT` because update different from create and not `PATCH` to avoid confusion with below
+  * `PatchSchedule` - `POST /api/v1/namespaces/{namespace}/schedules/{schedule_id}/patch`
+    * This is not a `PATCH` to avoid confusion with update above
+  * `ListScheduleMatchingTimes` - `GET /api/v1/namespaces/{namespace}/schedules/{schedule_id}/matching-times`
+  * `DeleteSchedule` - `DELETE /api/v1/namespaces/{namespace}/schedules/{schedule_id}`
+    * Confirm whether eventually consistent
+  * `ListSchedules` - `GET /api/v1/namespaces/{namespace}/schedules`
+* Batch operations:
+  * `StartBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `DescribeBatchOperation` - `GET /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `ListBatchOperations` - `GET /api/v1/namespaces/{namespace}/batch-operations`
+  * `StopBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}/stop`
+* Worker build IDs:
+  * Intentionally not exposed until they settle
+* Other calls:
+  * `GetClusterInfo` - Intentionally not exposed
+  * `GetSystemInfo` - Intentionally not exposed
+
+### Embedding into Server
+
+Current suggested approach, subject to discussion:
+
+* Generated gRPC gateway code will be auto-generated at the same time as api-go is for gRPC and live alongside it
+* Service config has new `httpPort`
+  * Only allowed on "frontend" service, presence elsewhere is config failure
+  * Default value is `grpcPort` + 10
+    * Makes `7243` the default - unassigned in IANA, unused in server
+    * Cannot think of case where +10 clashes, e.g. if they use 80 or 443 or 8080 or whatever, 90, 453, and 8090 are fine
+* Service config has new `disableHttp`
+  * Default enabled, as harmless as exposing gRPC by default, and will encourage use
+  * This is separate from `httpPort` because we don't want to support some `false` on an int option
+* Full support for same TLS options as frontend
+* Since it goes through gRPC, all metadata (i.e. HTTP headers) and interceptors are respected
+* Will test all of the above
+* Can discuss approach for cloud in separate place
+
+### Documentation
+
+* Need to simply document its presence at first, OpenAPI spec _can_ come later though Temporal may consider a more
+  intentionally-designed API in the future
+  * Maybe a snippet or two in docs showing how to use `curl` to start, signal, cancel, and maybe even get history with
+    filter type of close event to get workflow result
+* https://api-docs.temporal.io/ already exists, may be worth adapting for this use too
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -103,8 +103,10 @@ actually apply to many but we are not repeating ourselves too much here.
   * `CountWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflow-count`
   * `QueryWorkflow` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/query/{query_name}`
   * `DescribeWorkflowExecution` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
-  * `UpdateWorkflowExecution`- `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}`
-    * Decided to make this an identity so we can have "describe update" or "cancel update" or whatever one day
+  * `UpdateWorkflowExecution`- `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/update/{update_name}`
+    * We also support `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}` which will
+      expect an update name in the body but gives a stable URL to do future things like "describe update" or
+      "cancel update" one day (note the plural `updates`)
   * `PollWorkflowExecutionUpdate` - Intentionally not exposed
     * Not sure if best to have long HTTP API here. Would prefer non-blocking describe and maybe one day the difference
       between describing an update and polling for an update is a "wait" boolean.
@@ -168,6 +170,97 @@ Current suggested approach, subject to discussion:
 * Will test all of the above
 * Can discuss approach for cloud in separate place
 
+#### Payload Formatting
+
+One of the major drawbacks of payloads is that if we accept them directly in JSON, they look like:
+
+```json
+{
+  "payloads": [{
+    "metadata": {
+      "encoding": "anNvbi9wbGFpbg=="
+    },
+    "data": "IkhlbGxvLCBXb3JsZCEi"
+  }]
+}
+```
+
+This is for the simple JSON message of `"Hello, World"`. With larger JSON objects it gets even more ugly like if we
+wanted: `{ "foo": "bar" }`, it becomes `"data": "eyAiZm9vIjogImJhciIgfQ=="`. This is because by default proto byte
+arrays are serialized/deserialized as base64.
+
+But we may be able to solve this using a "shorthand" approach. Some research has been done yet, here are the notes:
+
+* grpc-gateway uses `jsonpb` and supports custom marshalling choice (so `gogoproto`'s `jsonpb` could be used instead)
+* We can make custom JSON unmarshalling _per proto_ on these older forms of unmarshalling, e.g. see
+  [here](https://pkg.go.dev/github.com/gogo/protobuf/jsonpb#JSONPBUnmarshaler)
+  * But newer `protojson` does not support this, see [this issue](https://github.com/golang/protobuf/issues/1322) and
+    others. So this could harm our ability to upgrade.
+* We can make `go.temporal.io/api/common/v1.Payloads` impl JSON marshal/unmarshal to assume if it is given a JSON
+  collection, it can assume it's for the "payloads" field
+* We can make `go.temporal.io/api/common/v1.Payload` impl JSON marshal/unmarshal to assume if it not an object that can
+  deserialize into payload using strict rules (e.g. exact `metadata` and `data` format), then treat it as a `json/plain`
+  payload for non `null` or null payload for null value.
+  * We could just say "without metadata assume JSON/null" but what if someone has an object with a "metadata" field?
+  * The only way it will clash with the raw form is if someone happens to have a `metadata` and/or `data` field with
+    proper base64
+* Should we assume that a payloads shorthand collection means all payloads within are shorthand?
+  * If we do, it reduces clash likelihood, but won't allow raw payloads to come through
+  * Answer: probably not
+* Sure on unmarshal we can accept either type, but on marshal can we assume the shorthand notation is best?
+  * I think for ease of use, the answer is "yes" for JSON/null payloads
+* Upon JSON marshal however, how do we know that we are in the HTTP handler and therefore should use shorthand?
+  * This is a tough challenge. We don't want to change what this looks like on history dump or something, right?
+  * There's no real way to pass contextual I-am-HTTP-handler state to the underlying 
+  * The best way is probably to use the [payload visitor](https://pkg.go.dev/go.temporal.io/api/proxy#VisitPayloads) to
+    set metadata on every payload saying it should be shorthand on marshal
+* What do we do about `json/protobuf` input/output?
+  * On the way out we can unmarshal in shorthand mode, no problem
+  * On the way in, how can we know that it is possibly proto form? Should we have a new shorthand for proto JSON that is
+    a bit different than regular JSON? Probably and should apply to output too.
+  * This new proto JSON form needs to have some way to make it clear it's a proto JSON payload. Since protos always have
+    to be objects, a `_protoMessageType` key will have to be present that is the qualified "messageName"
+
+So now, with the above shorthand, that same payloads set now looks like:
+
+```json
+["Hello, World!"]
+```
+
+And that works in either direction. Nice. And if your memo before was:
+
+```json
+{
+  "fields": {
+    "my-key1": {
+      "metadata": {
+        "encoding": "anNvbi9wbGFpbg=="
+      },
+      "data": "Im15LXZhbHVlMSI="
+    },
+    "my-key2": {
+      "metadata": {
+        "encoding": "anNvbi9wbGFpbg=="
+      },
+      "data": "Im15LXZhbHVlMiI="
+    }
+  }
+}
+```
+
+It is now:
+
+```json
+{
+  "fields": {
+    "my-key1": "my-value1",
+    "my-key2": "my-value2"
+  }
+}
+```
+
+Nice.
+
 ### Documentation
 
 * Need to simply document its presence at first, OpenAPI spec _can_ come later though Temporal may consider a more
@@ -175,3 +268,169 @@ Current suggested approach, subject to discussion:
   * Maybe a snippet or two in docs showing how to use `curl` to start, signal, cancel, and maybe even get history with
     filter type of close event to get workflow result
 * https://api-docs.temporal.io/ already exists, may be worth adapting for this use too
+
+## Usage Examples
+
+These are `curl` examples that are based on documentation but are not tested because that would require actually
+implementing the proposal. So there may be some mistakes.
+
+Note, responses are shown pretty printed though they would not be by default. But we can make this a header/query
+option.
+
+### Start a workflow - no input
+
+Request:
+
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id -d '{
+  "workflowType": { "name": "MyWorkflow" },
+  "taskQueue": { "name": "my-task-queue" }
+}'
+```
+
+Response:
+
+```json
+{
+  "runId": "7124ede7-fef5-4e15-8411-baa23d6beefd"
+}
+```
+
+### Start a workflow with input
+
+Request:
+
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id -d '{
+  "workflowType": { "name": "MyWorkflow" },
+  "taskQueue": { "name": "my-task-queue" },
+  "input": ["Hello, World!"]
+}'
+```
+
+This assumes that the approach from "Payload Formatting" above is applied and works.
+
+Response:
+
+```json
+{
+  "runId": "7124ede7-fef5-4e15-8411-baa23d6beefd"
+}
+```
+
+### Signal workflow
+
+Request:
+
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id/signal/MySignal
+```
+
+Response:
+
+```json
+{}
+```
+
+### Query workflow
+
+Request:
+
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id/query/MyQuery
+```
+
+Response:
+
+```json
+{
+  "queryResult": ["Hello, World!"]
+}
+```
+
+We could set `queryResult` to be the implied body, but it is best we stick with the gRPC protocol we've bad for better
+or worse. So `queryRejected` comes back as a success. This is just how we chose to implement query rejection (not to be
+confused with query failure).
+
+### Update workflow without ID
+
+Request:
+
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id/update/MyUpdate -d '{
+  "request": {
+    "input": {
+      "args": ["Hello, World!"]
+    }
+  }
+}'
+```
+
+Response:
+
+```json
+{
+  "updateRef": {
+    "workflowExecution": {
+      "workflowId": "my-workflow-id",
+      "runId": "7124ede7-fef5-4e15-8411-baa23d6beefd"
+    },
+    "updateId": "6dad5439-40dd-4b01-9f15-f5723a450776"
+  },
+  "outcome": {
+    "success": ["Hello, World!"]
+  }
+}
+```
+
+### Update workflow with ID
+
+Request:
+​
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id/updates/my-update-id -d '{
+  "request": {
+    "input": {
+      "name": "MyUpdate",
+      "args": ["Hello, World!"]
+    }
+  }
+}'
+```
+​
+Response:
+​
+```json
+{
+  "updateRef": {
+    "workflowExecution": {
+      "workflowId": "my-workflow-id",
+      "runId": "7124ede7-fef5-4e15-8411-baa23d6beefd"
+    },
+    "updateId": "my-update-id"
+  },
+  "outcome": {
+    "success": ["Hello, World!"]
+  }
+}
+```
+
+So `/update/<update-name>` is sending to a name, but `/updates/<update-id>` is also sending to a name but an ID url too.
+The reason for this is to support a stable ID-based URL for creating and describing (i.e. non-blocking-result fetch) one
+day.
+
+### Cancel workflow
+
+Request:
+​
+```bash
+curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id/cancel -d '{
+  "reason": "some reason"
+}'
+```
+​
+Response:
+​
+```json
+{ }
+```
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -16,8 +16,12 @@ Temporal has a gRPC API but no HTTP API. We need an HTTP API.
 * Design a high-quality REST API from first principles
 * grpc-web (browser-only gRPC workaround that requires server-side support)
 * Make an easy call to get workflow result
+* Alter existing gRPC API
 * gRPC stream support (maybe later)
 * Poll calls (maybe later)
+
+The reason for all non-goals is to have a reasonably scoped project and to not harm/effect our existing gRPC API in any
+way. This proposal is for a limited-effort approach.
 
 **Unresolved Research**
 
@@ -67,6 +71,7 @@ though.
     * Noun as last URL term if for a create, verb if for an action
   * `PATCH` is not needed because we don't adhere to common CRUD
     * TODO(cretz): Needs discussion
+* `?pretty` can be added to any URL for pretty printing of response
 
 ### RPCs
 
@@ -220,6 +225,7 @@ But we may be able to solve this using a "shorthand" approach. Some research has
     a bit different than regular JSON? Probably and should apply to output too.
   * This new proto JSON form needs to have some way to make it clear it's a proto JSON payload. Since protos always have
     to be objects, a `_protoMessageType` key will have to be present that is the qualified "messageName"
+* Shorthand payload parsing/formatting can be opted out via a `?noPayloadShorthand` query parameter
 
 So now, with the above shorthand, that same payloads set now looks like:
 
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -69,7 +69,6 @@ actually apply to many but we are not repeating ourselves too much here.
 
 * Namespaces:
   * `RegisterNamespace` - `POST /api/v1/namespaces`
-    * 202 on success since eventually consistent
   * `DescribeNamespace` - `GET /api/v1/namespaces/{namespace}`
   * `ListNamespaces` - `GET /api/v1/namespaces`
   * `UpdateNamespace` - `PUT /api/v1/namespaces/{namespace}`
@@ -79,17 +78,15 @@ actually apply to many but we are not repeating ourselves too much here.
     * Does this prevent returning from `GET`? Will determine HTTP method
 * Workflows:
   * `StartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
-    * 204 on success
-    * Should accept header as `identity` on these calls?
-    * Ok to be at root URL here instead of appending `/start`?
+    * May accept header as `identity` and `request_id` on these calls
   * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
-    * Accept `?nextPageToken` as query param? I'm thinking so
+    * `?nextPageToken` can be a query param even though it's big
   * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
   * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
     * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
-  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal`
-    * Should we put signal name in the URL? Can be optional (e.g. in URL or body)
-  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal[/{signal_name}]`
+    * Note signal name on the URL is optional but must be escaped. Must be in body otherwise.
+  * `SignalWithStartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
   * `ResetWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/reset`
   * `TerminateWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/terminate`
   * `ListOpenWorkflowExecutions` - Intentionally not exposed
@@ -99,8 +96,8 @@ actually apply to many but we are not repeating ourselves too much here.
   * `ListArchivedWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/archived-workflows`
     * Intentionally did not do `/workflows/archived` due to ambiguity
   * `CountWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflow-count`
-  * `QueryWorkflow` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/query`
-    * Should we put query name in the URL? Can be optional (e.g. in URL or body)
+  * `QueryWorkflow` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/query[/{query_name}]`
+    * Note query name on the URL is optional but must be escaped. Must be in body otherwise.
   * `DescribeWorkflowExecution` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
   * `UpdateWorkflowExecution`- `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}`
     * Decided to make this an identity so we can have "describe update" or "cancel update" or whatever one day
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -6,7 +6,7 @@ Temporal has a gRPC API but no HTTP API. We need an HTTP API.
 
 **Goals**
 
-* Create HTTP API that governs all reasonable user-facing calls
+* Expose existing API over HTTP in addition to gRPC
 * Ensure new API calls are automatically adopted
 * Expose HTTP API in server by default for all to use like gRPC, but have option to disable
 * Support all existing interceptors and auth methods
@@ -16,6 +16,12 @@ Temporal has a gRPC API but no HTTP API. We need an HTTP API.
 * Design a high-quality REST API from first principles
 * grpc-web (browser-only gRPC workaround that requires server-side support)
 * Make an easy call to get workflow result
+* gRPC stream support (maybe later)
+* Poll calls (maybe later)
+
+**Unresolved Research**
+
+* Can payloads be made in non-base64 form without over-complicating the effort?
 
 ## Approach
 
@@ -144,7 +150,7 @@ actually apply to many but we are not repeating ourselves too much here.
   * Intentionally not exposed until they settle
 * Other calls:
   * `GetClusterInfo` - Intentionally not exposed
-  * `GetSystemInfo` - Intentionally not exposed
+  * `GetSystemInfo` - `GET /api/v1/system-info`
 
 ### Embedding into Server
 
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -94,8 +94,8 @@ actually apply to many but we are not repeating ourselves too much here.
   * `SignalWithStartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start/{signal_name}`
   * `ResetWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/reset`
   * `TerminateWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/terminate`
-  * `ListOpenWorkflowExecutions` - Intentionally not exposed
-  * `ListClosedWorkflowExecutions` - Intentionally not exposed
+  * `ListOpenWorkflowExecutions` - Intentionally not exposed (deprecated)
+  * `ListClosedWorkflowExecutions` - Intentionally not exposed (deprecated)
   * `ListWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflows`
     * Accepts `?query=` query parameter
   * `ListArchivedWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/archived-workflows`
--- a/api/http-api.md
+++ b/api/http-api.md
@@ -434,3 +434,207 @@ Response:
 ```json
 { }
 ```
+
+## Rejected Alternative Approach - Full API Redesign
+
+An alternative to the approach to this proposal is to make a quality Temporal HTTP API from scratch instead of just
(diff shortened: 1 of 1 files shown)


mbernier:
Is it possible to generate an openAPI schema along with the API interface?

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+    * 204 on success
+    * Should accept header as `identity` on these calls?
+    * Ok to be at root URL here instead of appending `/start`?
+  * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
+    * Accept `?nextPageToken` as query param? I'm thinking so
+  * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
+  * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
+    * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal`
+    * Should we put signal name in the URL? Can be optional (e.g. in URL or body)
```
tminusplus, line 91: What if the entire signal could be sent within either the URL or body? It makes linking to interact with human-in-the-loop workflows nice and simple. But also allows for more complex signals to be sent in the body of the POST request.

I was going to post a similar question for async activity completion but will keep the conversation here since it is a similar purpose.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
```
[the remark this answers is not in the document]
cretz: Yes, typo, thanks! Will fix.

api/http-api.md
cretz: > What if the entire signal could be sent within either the URL or body?

Yes, this was the subject of the question and the parenthetical. I am thinking optional signal name in URL makes sense. Note that doesn't help signal args which I think goes a step too far. This can apply to queries too. But sadly not updates that have their own stable identifier.

Also note this is a POST and a thin layer over gRPC. Simplicity for humans is not as large of a factor (part of the non-goal of designing a better API).

> I was going to post a similar question for async activity completion but will keep the conversation here since it is a similar purpose.

This is a different deal. To refer to an activity you need a task token or an ID. The former is what we recommend but is too large and unwieldy to put on a URL and therefore we don't want to make the latter easier. Also, we discourage async activities unless they are really needed, so we're not trying to design easy HTTP API approaches for their use beyond surfacing the gRPC API.

cretz:
> Is it possible to generate an openAPI schema along with the API interface?

@mbernier - From the near the bottom:

> Need to simply document its presence at first, OpenAPI spec _can_ come later though Temporal may consider a more intentionally-designed API in the future 

It's not a part of this initial project. We can do it. But OpenAPI/swagger often suck anyways and can lead to many developers over-compensating for their suckiness. I am trying to avoid some of the stuff seen at https://grpc-ecosystem.github.io/grpc-gateway/docs/mapping/customizing_openapi_output/ from invading our currently-simple-to-read proto spec (e.g. [this](https://github.com/grpc-ecosystem/grpc-gateway/blob/f30ed26235580b6fccfd6eadbbe83eb15f9d6c3a/examples/internal/proto/examplepb/a_bit_of_everything.proto#L20-L191) is a bit too gross to embed so I am punting on OpenAPI until a later date where it'll probably be curated/generated external to the existing API spec).

feedmeapples:
> We can do it. But OpenAPI/swagger often sucks anyways..

We have OpenAPI in ui-server, i think it can stay there unless we decide to move it server. Try accessing it by `temporal server start-dev and navigating to http://localhost:8233/openapi/ cc @mbernier

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+* Response codes
+  * 201 success if a creation
+  * 202 if action eventually consistent
+  * 200 other all other successes
+  * Do not use 204 for no response, even if there's no response because we may add response later, just 200 everything
+* Methods:
+  * `GET` if read-only
+    * Noun as last URL term
+    * Use query parameters for commonly set items to avoid relying on request bodies (but still rely on bodies as
+      needed). This may require some gRPC gateway customization.
```
bergundy, line 55: It will also be hard to generate type safe clients if we break the request params out of proto messages.

bergundy: Looks like we're in for a lot of hand coding.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+    * TODO(cretz): Needs discussion
+
+### RPCs
+
+Below is list of gRPC calls to HTTP calls so far and some notes/questions beneath each. Some notes/questions beneath one
+actually apply to many but we are not repeating ourselves too much here.
+
+* Namespaces:
+  * `RegisterNamespace` - `POST /api/v1/namespaces`
+    * 202 on success since eventually consistent
```
bergundy, line 72: No longer eventually consistent since 1.20.

api/http-api.md
bergundy: I'm wondering how complex are read only request types are, might not be a big issue.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `DescribeNamespace` - `GET /api/v1/namespaces/{namespace}`
+  * `ListNamespaces` - `GET /api/v1/namespaces`
+  * `UpdateNamespace` - `PUT /api/v1/namespaces/{namespace}`
+    * TODO(cretz): Confirm strongly consistent and idempotent or not
+  * `DeprecateNamespace` - `DELETE /api/v1/namespaces/{namespace}`
+    * Is this eventually consistent? Will determine response code
+    * Does this prevent returning from `GET`? Will determine HTTP method
+* Workflows:
+  * `StartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
+    * 204 on success
```
bergundy, line 82: You said no 204, is this a mistake? I think you meant 201.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+    * Should accept header as `identity` on these calls?
```
bergundy, line 83: Yes, and also request ID IMHO.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+    * Ok to be at root URL here instead of appending `/start`?
```
bergundy, line 84: Don't have a strong preference, root seems to fit here.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
+    * Accept `?nextPageToken` as query param? I'm thinking so
```
bergundy, line 86: Yes

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
```
bergundy, line 87: Can this be history `?reverse=true` or `orderBy=...`?

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
+    * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
```
bergundy, line 89: I prefer POST requests to use only the body if possible.

api/http-api.md
bergundy, line 91: Signal name in the path makes sense to me.

api/http-api.md
bergundy, line 92:
```suggestion
  * `SignalWithStartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
```

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal`
+    * Should we put signal name in the URL? Can be optional (e.g. in URL or body)
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start`
+  * `ResetWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/reset`
+  * `TerminateWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/terminate`
+  * `ListOpenWorkflowExecutions` - Intentionally not exposed
+  * `ListClosedWorkflowExecutions` - Intentionally not exposed
```
bergundy, line 96: I think both are (unofficially?) deprecated.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `ListWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflows`
+    * Accepts `?query=` query parameter
+  * `ListArchivedWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/archived-workflows`
+    * Intentionally did not do `/workflows/archived` due to ambiguity
```
bergundy, line 100: Can this just be `/workflows?archived=true` or `Archived = true` in the query? I know it deviates from the gRPC API (I wish we'd combined those there).

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `CountWorkflowExecutions` - `GET /api/v1/namespaces/{namespace}/workflow-count`
```
bergundy, line 101: We're also getting a `CountGroupBy` API, I wish we'd just defined a more powerful query language instead of URL per type of query.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `QueryWorkflow` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/query`
+    * Should we put query name in the URL? Can be optional (e.g. in URL or body)
```
bergundy, line 103: I vote yes.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+  * `DescribeWorkflowExecution` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
+  * `UpdateWorkflowExecution`- `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}`
+    * Decided to make this an identity so we can have "describe update" or "cancel update" or whatever one day
+  * `PollWorkflowExecutionUpdate` - Intentionally not exposed
+    * Not sure if best to have long HTTP API here. Would prefer non-blocking describe and maybe one day the difference
```
bergundy, line 108: I vote for polling, it's a useful enough API and we'll definitely want that in our UI.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+      between describing an update and polling for an update is a "wait" boolean.
+* Activities:
+  * `RecordActivityTaskHeartbeat` - `POST /api/v1/namespaces/{namespace}/activities/heartbeat`
+    * We consider "heartbeat" a verb for these purposes
+  * `RecordActivityTaskHeartbeatById` - `POST /api/v1/namespaces/{namespace}/activities/heartbeat-by-id`
+    * We are intentionally not doing `/activities/{activity_id}/heartbeat` or similar due to URL ambiguity and that we
+      want to encourage the task token approach
```
bergundy, line 115: I'm fine with that.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
...
+* Batch operations:
+  * `StartBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `DescribeBatchOperation` - `GET /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `ListBatchOperations` - `GET /api/v1/namespaces/{namespace}/batch-operations`
+  * `StopBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}/stop`
+* Worker build IDs:
+  * Intentionally not exposed until they settle
+* Other calls:
+  * `GetClusterInfo` - Intentionally not exposed
+  * `GetSystemInfo` - Intentionally not exposed
```
bergundy, line 150: I want this one and the rest of the worker APIs so we can use HTTP to power the TS SDK in browser or CF worker.

api/http-api.md
```diff
@@ -0,0 +1,176 @@
+# HTTP API
+
+## Overview
+
+Temporal has a gRPC API but no HTTP API. We need an HTTP API.
+
+**Goals**
```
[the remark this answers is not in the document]
cretz: I think that's a bit different. I was leaving polling out of this. Do we have to infect this project w/ that requirement?

api/http-api.md
cretz: > It will also be hard to generate type safe clients if we break the request params out of proto messages.

This is already happening with URL path members coming from proto messages.

> Looks like we're in for a lot of hand coding.

gRPC gateway mostly supports this out of the box. Though they default to letting _every_ field of a `GET` message be in the query params if no body is specified. There may be some hand-coding here, but we only have to do it once and how I would do it is via Temporal-based proto annotations. Not per RPC code.

> I'm wondering how complex are read only request types are, might not be a big issue.

Not too bad, but I have to investigate whether we can choose specific query params or if we have to choose "query or body" and that's all. Either way, anything I do here will be generally applicable, so worst case you'd have to specify which body members can be part of the query.

api/http-api.md
cretz: Yes, typo! Will fix.

api/http-api.md
cretz: Concur

api/http-api.md
cretz: Not really without actually designing an alternative. We need to keep HTTP call;s 1:1 with gRPC calls. I asked this years ago at https://github.com/temporalio/api/pull/147#discussion_r796948865. It was at that time we should have decided we wanted a single call for these if we did.

api/http-api.md
cretz: That becomes a bit inconsistent though to specify run ID one way for one type of operation but another way for another type.

api/http-api.md
cretz: Starting to lean that way, unfortunately it makes us inconsistent with workflows and workflow updates

api/http-api.md
cretz: Yes, pointed out by previous commenter, will fix

api/http-api.md
cretz: Right, which is why they are intentionally not exposed

api/http-api.md
cretz: No. Like history reverse, we as a company chose to make these separate calls, we accept that choice. We need a thin HTTP API here on top of what exists, not a redesign. If you go back and add that bool to the non-archived form for gRPC, I can support it in HTTP too.

api/http-api.md
cretz: So long as it's understood that our HTTP API represents the same fragmentation decisions we made with our gRPC API, whether that's here or worker versioning APIs :-)

api/http-api.md
cretz: Leaning this way, but Inconsistent with workflow and workflow update

api/http-api.md
cretz: This needs to shake out better. It's too young. Until we have a non-blocking poll-update call I don't think it's ready for HTTP-ification.

api/http-api.md
cretz: Ok

api/http-api.md
cretz: The goal was for user benefit instead of worker benefit. But will consider, though easily added later instead of infecting initial project.

api/http-api.md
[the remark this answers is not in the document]
paulnpdev: Strong vote for no mid-url optional parameters

api/http-api.md
paulnpdev: Since it wasn’t mentioned and I’m feeling very ocd today, there’s another 204 here

api/http-api.md
paulnpdev: I vote yes too. Query is weird any ways

api/http-api.md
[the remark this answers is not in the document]
bergundy: Absolutely required IMHO.

api/http-api.md
bergundy: > gRPC gateway mostly supports this out of the box
We'll have to make sure we keep the TS client in sync with each of these deviations from the proto definitions.

api/http-api.md
bergundy: Yeah, both of those have IDs, so we'll want to use those in the path.
I think it's fine to be inconsistent for the sake of usability.

api/http-api.md
bergundy: I would make it part of the query, not a bool, ideally.

api/http-api.md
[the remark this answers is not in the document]
cretz: Fixed

api/http-api.md
cretz: Updated to allow optional signal name on URL. Made optional because names can get unwieldy and not URL friendly, but I'm starting to consider just making them required.

api/http-api.md
bergundy: Can wait, but I would definitely want this in the HTTP API eventually.

api/http-api.md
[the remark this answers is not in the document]
cretz: Required for _you_ or required for people that want to use curl to start a workflow? I don't see a big reason why not I guess.

api/http-api.md
bergundy: Agree those can come later.

api/http-api.md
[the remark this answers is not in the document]
bergundy: I want the UI to use the TS SDK client instead of them having to write their own wrapper.

api/http-api.md
cretz: Yes, the HTTP API is not 100% machine derived from the gRPC API. People (internal or users) that want to use the APIs will need to know what to call and with what options. Also notice the URL names don't match the gRPC names. You'll have to keep that in sync too. Basically we're going to API maintenance simplicity (i.e. not a bunch of custom HTTP-server-layer code) while still doing the best we can with HTTP constructs.

api/http-api.md
cretz: Fixed

api/http-api.md
cretz: Fixed

api/http-api.md
```diff
@@ -0,0 +1,173 @@
...
+### Embedding into Server
+
+Current suggested approach, subject to discussion:
+
+* Generated gRPC gateway code will be auto-generated at the same time as api-go is for gRPC and live alongside it
+* Service config has new `httpPort`
+  * Only allowed on "frontend" service, presence elsewhere is config failure
+  * Default value is `grpcPort` + 10
+    * Makes `7243` the default - unassigned in IANA, unused in server
+    * Cannot think of case where +10 clashes, e.g. if they use 80 or 443 or 8080 or whatever, 90, 453, and 8090 are fine
```
bergundy, line 158: I don't think these are commonly used ports for Temporal. I bet nobody customizes the grpc port apart from our devs that want to run multiple servers for testing purposes.

api/http-api.md
cretz: Fixed. Mentioned that identity and request ID we may support as headers. Some more thought is needed there, but not at proposal stage I don't think.

api/http-api.md
cretz: Removed this question.

api/http-api.md
cretz: Cemented this as how we'll handle it.

api/http-api.md
cretz: Concur. Fixed to clarify optional signal name. Though I am now starting to think about requiring them.

api/http-api.md
cretz: Fixed.

api/http-api.md
cretz: :+1: When done on the gRPC side, will automatically apply to this side.

api/http-api.md
cretz: Done.

api/http-api.md
cretz: So even a stronger statement that default of `grpcPort + 10` is ok :-)

bergundy: Now that I'm thinking of it, maybe 8233 is a better port since we already use that for the single binary UI server.

cretz: That's what makes it a bad port. We use it for something else. And this is gonna be embedded in the server, so the +10 still keeps the port in the small range of ports we open there.

api/http-api.md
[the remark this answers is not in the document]
cretz: The UI should use their own server tailored to their needs and that server can make backend-to-backend calls to the API and translate as needed :-)

api/http-api.md
yiminc: The GetWorkflowHistory is too complicated, it had the long poll thing which really should be a separate API.

api/http-api.md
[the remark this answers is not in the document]
yiminc: optional ?runId= sounds good.

api/http-api.md
yiminc: Yes, they are deprecated.

api/http-api.md
yiminc: Agree with Chad, should try to keep it consistent with gRPC side.

api/http-api.md
yiminc: I vote no for polling. Keep it minimal until we see enough interest with valid use cases.

api/http-api.md
yiminc: I strongly disagree that we need to expose worker APIs now. I need to see enough interests with real use cases before we do that.

api/http-api.md
```diff
@@ -0,0 +1,173 @@
+# HTTP API
+
+## Overview
+
+Temporal has a gRPC API but no HTTP API. We need an HTTP API.
+
+**Goals**
+
+* Create HTTP API that governs all reasonable user-facing calls
```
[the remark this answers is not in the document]
cretz: > Can we make sure this HTTP API work does not make it harder to support grpc streaming in future

It can make it harder depending on the shape of the stream. https://github.com/grpc-ecosystem/grpc-gateway now handles streams as newline-delimited JSON streams of request/response body but there are surely caveats.

> Most likely the polling API and the get history API could be candidate for streaming.

While they _could_, we wouldn't get rid of existing API surely because there are many relying on it across many SDK versions. So does streaming really need to stop the HTTP API project?

There is also a lot of detail that goes into proper failure handling and LB/HA of streams anyways that has to be discussed when we get there. I don't think this project needs to concern itself w/ future streaming possibility (nor can it really do anything about it short of contrib'ing something if we think it's messing). Maybe we want to make a rule "no streaming API without a non-stream equivalent".

We're using a third party library. I don't think there's anything to do here concerning some hypothetical feature if/until that feature has a more concrete spec.

api/http-api.md
cretz: True, but making better RPC is unrelated to this project.

api/http-api.md
cretz: :+1: It is so easy to add that @bergundy can alter a few lines here during the development of the thing that needs it and we can merge it in then.

api/http-api.md
bergundy: I think it's fine, it should be fairly straightforward to translate between HTTP and gRPC.

api/http-api.md
bergundy: Can make a note here to explain why they're not exposed.

api/http-api.md
bergundy: 👍

api/http-api.md
bergundy: We need to make a call if this API is going to serve the UI or not.
Seems like you're against it for some reason.

api/http-api.md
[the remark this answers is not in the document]
cretz: >  We can choose to maintain 2 separate sets of annotations or double down on what you're proposing here.

They can use the this API no problem or they can choose to maintain two sets or they can choose to use normal gRPC and build HTTP APIs for their use case. I neither want to tell them the best way nor alter any of this user HTTP API for that.

You can write a TS client with this today I believe, but I can't think of any benefit of making that a goal. I don't want to make OpenAPI or auto-generated TS part of this initial project.

api/http-api.md
cretz: > We need to make a call if this API is going to serve the UI or not.

I don't think we need to make this call. If needed I can add as a non-goal "decide how UI is going to access server" to make it more clear.

> Seems like you're against it for some reason

I'm not, I just don't think it needs to infect any work here. But if they want to be a consumer of this like anyone else, no problem. No UI specific stuff needed in this proposal though.

api/http-api.md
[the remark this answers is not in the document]
cretz: Right, so you'll be able to use this API in TypeScript no problem. Same with UI (directly or through TS SDK or whatever). Y'all don't have to be special consumers. Doesn't affect this proposal.

api/http-api.md
bergundy: I think we should let someone on the UI team review this proposal and comment here.

> Nothing in this proposal precludes potential UI use or potential TypeScript use, they are just distractions from an initial version.

TS is for sure going to be a heavy consumer of this API and maybe UI too.
I don't think these are distractions, they're requirements.
We don't want to add another HTTP API if we realize that this one can't be used for purposes that we've foreseen.

cretz: What in this proposal is stopping TS or UI use? You can answer for TS and will wait for UI review.

bergundy: For TS I'd want to know how the translation of HTTP bodies <-> gRPC messages would work.
If it's standard proto3 JSON (with the acceptable exception of our custom enum encoding) you have a go from the TS SDK perspective.

Looks like the encoding is [configurable](https://github.com/grpc-ecosystem/grpc-gateway#mapping-grpc-to-http) so we should be good, but let's please clarify in the proposal.

api/http-api.md
```diff
@@ -0,0 +1,173 @@
...
+  * `ListSchedules` - `GET /api/v1/namespaces/{namespace}/schedules`
+* Batch operations:
+  * `StartBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `DescribeBatchOperation` - `GET /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `ListBatchOperations` - `GET /api/v1/namespaces/{namespace}/batch-operations`
+  * `StopBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}/stop`
+* Worker build IDs:
+  * Intentionally not exposed until they settle
+* Other calls:
+  * `GetClusterInfo` - Intentionally not exposed
```
stevekinney, line 146: We won't be able to use this API if this is not exposed.

api/http-api.md
```diff
@@ -0,0 +1,173 @@
...
+  * `GetSystemInfo` - Intentionally not exposed
```
stevekinney, line 147: We won't be able to use this API if this is not exposed.

api/http-api.md
```diff
@@ -0,0 +1,173 @@
...
+  * `RecordActivityTaskHeartbeatById` - `POST /api/v1/namespaces/{namespace}/activities/heartbeat-by-id`
+    * We are intentionally not doing `/activities/{activity_id}/heartbeat` or similar due to URL ambiguity and that we
+      want to encourage the task token approach
+  * `RespondActivityTaskCompleted` - `POST /api/v1/namespaces/{namespace}/activities/complete`
+  * `RespondActivityTaskCompletedById` - `POST /api/v1/namespaces/{namespace}/activities/complete-by-id`
+  * `RespondActivityTaskFailed` - `POST /api/v1/namespaces/{namespace}/activities/fail`
+  * `RespondActivityTaskFailedById` - `POST /api/v1/namespaces/{namespace}/activities/fail-by-id`
+  * `RespondActivityTaskCanceled` - `POST /api/v1/namespaces/{namespace}/activities/cancel`
+  * `RespondActivityTaskCanceledById` - `POST /api/v1/namespaces/{namespace}/activities/cancel-by-id`
+* Search attributes:
```
stevekinney, line 119: We won't be able to use this API if this is not exposed.

api/http-api.md
cretz: Resolving...

api/http-api.md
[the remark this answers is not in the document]
cretz: That's not user friendly and our user interfaces for workflow interaction do not require run ID. We don't require it in SDKs and we shouldn't in the APIs. We intentionally decided to make programmatic interfaces to workflow interaction work on workflow ID only by default.

api/http-api.md
cretz: There are multiple places where our API is not the cleanest and/or won't translate as well to JSON as it would if we _designed_ a REST API. But I think that is out of scope for this simple project which is just make HTTP and gRPC both available for the same API.

api/http-api.md
cretz: I will definitely expose it then. Would y'all be willing to switch to `GetSystemInfo`?

api/http-api.md
cretz: :+1: Will definitely expose then.

api/http-api.md
cretz: :+1: Which calls specifically are being used here? I wonder if I should expose `operatorservice` ones that are better for this I believe.

api/http-api.md
lorensr: +1 to either body or accepting either

mmcshane:
Generally speaking my preference is to discriminate operations on the same resource by content-type rather than by tacking on extra path segments. E.g. I'd prefer for updates and signals both to be posted to  $namespace/workflows/$workflow_id[?run_id=$run_id] and be differentiated on the server side based on the content type `application/vnd.io.temporal.signal.v1+json` versus `application/vnd.io.temporal.update.v1.Request+json` etc. A similar pattern would apply in a number of cases. I'm not saying the current approach doesn't work - it does. This is just a matter of preference.

cretz:
> Generally speaking my preference is to discriminate operations on the same resource by content-type rather than by tacking on extra path segments

@mmcshane - Unfortunately this is getting into design of a REST API instead of thin layer over gRPC calls where URLs are 1:1 with gRPC calls. Also, it's important to think about user friendliness and API approach familiarity across the ecosystem for users. Many users have tools to post JSON with implied content types. Many standard library HTTP clients have shortcuts for this. Many (most?) HTTP APIs at companies don't accept multiple JSON content types.

Also, I intentionally call this an HTTP API instead of a REST API because it's a thin HTTP layer on gRPC and is only as resource-based as the gRPC layer is.

> I just saw a comment elsewhere that 'We need to keep HTTP call;s 1:1 with gRPC calls" Is this a limitation of gRPC gateway? Or is there some other reason?

Mostly a limitation of gRPC gateway yes, though of course we can manually do anything in HTTP we wanted if we have to. The other reason is simplicity/consistency.

api/http-api.md
```diff
@@ -0,0 +1,173 @@
...
+* grpc-web (browser-only gRPC workaround that requires server-side support)
+* Make an easy call to get workflow result
+
+## Approach
+
+This project will use https://github.com/grpc-ecosystem/grpc-gateway and annotations inspired by work already done on
+the UI server for similar purpose at https://github.com/temporalio/ui-server. After review of the UI server's
+annotations, they so closely match our goals that many things won't change.
+
+Since namespace is required for most calls and is usually URL safe, we have decided to put the namespace in the URL
```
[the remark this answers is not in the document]
cretz: Hence "most", "usually", etc

api/http-api.md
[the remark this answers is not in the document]
cretz: Confirmed TS client can use this out of the box

api/http-api.md
cretz: Resolving

api/http-api.md
cretz: Resolving as beyond goals of this project.

api/http-api.md
cretz: Resolving as accepted in either.

api/http-api.md
cretz: Resolving

api/http-api.md
cretz: Resolving as not actionable

api/http-api.md
cretz: Added poll calls as non-goal currently

api/http-api.md
cretz: Resolving as no actionable suggestions and/or beyond the goals

api/http-api.md
cretz: Exposed

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+    * May accept header as `identity` and `request_id` on these calls
+  * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
+    * `?nextPageToken` can be a query param even though it's big
+  * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
+  * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
+    * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
+  * `SignalWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal/{signal_name}`
+  * `SignalWithStartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/signal-with-start/{signal_name}`
+  * `ResetWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/reset`
+  * `TerminateWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/terminate`
```
josh-berry, line 96: I see this is a `POST`—is it not idempotent? What happens if `terminate` is called twice for the same workflow? I dug through the server code a bit and it seems like it will just fail the second time (please correct me), but the actual state of the system will not change.

(I have similar questions about why some other things are `POST`s instead of `PUT`s, but I suspect there's something I may be missing here so I'll just ask about this one for now.)

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+
+Current suggested approach, subject to discussion:
+
+* Generated gRPC gateway code will be auto-generated at the same time as api-go is for gRPC and live alongside it
+* Service config has new `httpPort`
+  * Only allowed on "frontend" service, presence elsewhere is config failure
+  * Default value is `grpcPort` + 10
+    * Makes `7243` the default - unassigned in IANA, unused in server
+    * Cannot think of case where +10 clashes, e.g. if they use 80 or 443 or 8080 or whatever, 90, 453, and 8090 are fine
+* Service config has new `disableHttp`
```
josh-berry, line 163: Maybe this clashes with existing config conventions, but I don't generally like having negatives in variable names, because setting something like `disableHttp=true` might parse strangely especially for non-native English speakers.

I suggest naming this `enableHttp`, and have the default be `true`. Then if a user wants to disable HTTP, they can set `enableHttp=false` explicitly.

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+  * This is separate from `httpPort` because we don't want to support some `false` on an int option
+* Full support for same TLS options as frontend
+* Since it goes through gRPC, all metadata (i.e. HTTP headers) and interceptors are respected
+* Will test all of the above
+* Can discuss approach for cloud in separate place
+
+### Documentation
+
+* Need to simply document its presence at first, OpenAPI spec _can_ come later though Temporal may consider a more
+  intentionally-designed API in the future
```
josh-berry, line 174: Can the OpenAPI spec be generated automatically? I think it might be a bit unfriendly to provide the API, tell people it exists, and then not document how to use it.

josh-berry (commented):
Most (if not all) of these are newbie questions. Nothing in here should be construed as blocking or in any way holding up this proposal.

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+    * Use query parameters for commonly set items to avoid relying on request bodies (but still rely on bodies as
+      needed). This may require some gRPC gateway customization.
+  * `PUT` if strongly consistent add/update on fixed-URL w/ idempotency
+    * Noun as last URL term
+  * `DELETE` if item is actually deleted (i.e. unreachable from `GET`)
+    * Noun as last URL term
+  * `POST` for everything else
+    * Noun as last URL term if for a create, verb if for an action
+  * `PATCH` is not needed because we don't adhere to common CRUD
+    * TODO(cretz): Needs discussion
```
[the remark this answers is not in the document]
cretz: But we don't have anything that conforms to that reasonably well. If we were designing a REST API from scratch, we definitely would use `PATCH`.

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+**Goals**
+
+* Expose existing API over HTTP in addition to gRPC
+* Ensure new API calls are automatically adopted
+* Expose HTTP API in server by default for all to use like gRPC, but have option to disable
+* Support all existing interceptors and auth methods
+
+**Non-Goals**
+
+* Design a high-quality REST API from first principles
```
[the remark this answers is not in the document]
cretz: I cannot say. We haven't thus far. There is discussion that if we build this we will never be incentivized to redesign a REST API from scratch. But in the meantime we have nothing, hence the proposal to just add a decent-enough HTTP iface on our existing API.

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+
+* Namespaces:
+  * `RegisterNamespace` - `POST /api/v1/namespaces`
+  * `DescribeNamespace` - `GET /api/v1/namespaces/{namespace}`
+  * `ListNamespaces` - `GET /api/v1/namespaces`
+  * `UpdateNamespace` - `PUT /api/v1/namespaces/{namespace}`
+    * TODO(cretz): Confirm strongly consistent and idempotent or not
+  * `DeprecateNamespace` - `DELETE /api/v1/namespaces/{namespace}`
+    * Is this eventually consistent? Will determine response code
+    * Does this prevent returning from `GET`? Will determine HTTP method
```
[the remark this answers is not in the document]
cretz: This assumes you have more control over these APIs than this proposal is saying. This proposal is saying reuse the existing APIs and just add HTTP iface. There is no custom code for any of these calls in this proposal to support certain approaches.

api/http-api.md
```diff
@@ -0,0 +1,177 @@
...
+* Workflows:
+  * `StartWorkflowExecution` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}`
+    * May accept header as `identity` and `request_id` on these calls
+  * `GetWorkflowExecutionHistory` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history`
+    * `?nextPageToken` can be a query param even though it's big
+  * `GetWorkflowExecutionHistoryReverse` - `GET /api/v1/namespaces/{namespace}/workflows/{workflow_id}/history-reverse`
+  * `RequestCancelWorkflowExecutionRequest` - `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/cancel`
+    * Intentionally not taking the run ID in the URL, but should it be `?runId=` query param?
```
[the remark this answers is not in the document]
cretz: Yes, but it is optional. I think of it more like a filter than a resource. Having said that, yes, we could have a separate `runs` endpoint, but with these APIs I am trying not to have two URL endpoints mean the same thing. I could even have a `/latest` for absent run. Also not trying to have `/runs` for all of these.

But it's subjective and it isn't that ugly if others agree. Always easy to add later too.

api/http-api.md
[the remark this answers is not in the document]
josh-berry: Right, I understand we can't change the behavior of the existing API, but do we _also_ not have control over HTTP-specific things like the error codes used in the response?

api/http-api.md
cretz: It _can_ be generated automatically. I am not making it a part of the initial project because OpenAPI sucks and maintaining it sucks and there are a lot of little pieces we may want to add based on auth, docs, etc. I don't think the project needs that at first.

> I think it might be a bit unfriendly to provide the API, tell people it exists, and then not document how to use it.

No more unfriendly than what happens today. I think we should provide the API first, we can discuss the rest later.

api/http-api.md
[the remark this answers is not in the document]
josh-berry: I don't really like `/latest` because I think it violates `Do not assume any names [are] reserved`, but that's fair. If it's something that we expect people to use very often, I'd even be fine with leaving it out entirely and deciding later if we want `?runId=...` or a separate route.

api/http-api.md
josh-berry: D'oh, you're right.

api/http-api.md
josh-berry: hrm. If we're already consistently using `disableX` names for config values, then yes, much as I hate to say it, it's probably better to be consistently confusing than inconsistently confusing (which just adds to the confusion).

api/http-api.md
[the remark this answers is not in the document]
cretz: > I don't really like /latest because I think it violates Do not assume any names [are] reserved

It would be `/api/v1/namespaces/{namespace}/workflows/{workflow_id}/latest/cancel`, not in place of workflow ID or run ID which would have violated that assumption. But yes, supporting that and `/api/v1/namespaces/{namespace}/workflows/{workflow_id}/runs/{run_id}/cancel` for every endpoint related to workflows is a bit ugly.

> I'd even be fine with leaving it out entirely and deciding later if we want ?runId=... or a separate route.

I can kinda agree `POST` calls (they can just set it in the body). For getters I think we're leaning to all RPC fields being available as query params, so describing a workflow would probably have this query supported implicitly. But we often offer this as a clear optional value when programmatically interacting with workflows, so it would make sense to be present on the URL.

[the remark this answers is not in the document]
josh-berry: Yikes, my comment above should have said, "If it's [not] something that we expect people to use very often, ..."

Anyway, I'm fine either way; I'll leave this open though in case anyone else wants to chime in.

api/http-api.md
josh-berry: Yeah that's fair. I was thinking about this from the perspective of "what do we need to do to get this into a place where we can call it 'GA'?" (by which I mean, unlikely to change in incompatible ways and reliable/polished enough for production use). And per offline discussion, I understand that this is probably not something we will say is production-quality, etc. right off the bat.

The question behind my question is: How can we make the HTTP API accessible enough that people will actually use it and give us feedback? If that question is already answered / there's another way to answer it, I agree we don't have to do OpenAPI to start with.

cretz: > "what do we need to do to get this into a place where we can call it 'GA'?"

In the interest of incremental implementation, I don't think that logic needs to be applied to this proposal or the initial implementation.

> How can we make the HTTP API accessible enough that people will actually use it and give us feedback?

To be honest, I don't really think of this in that way. It's mostly too late for feedback as our API already exists and is GA. And I don't think we need to heavily promote just adding a second protocol to our existing API. This is an existing API, not a new API.

josh-berry: > > "what do we need to do to get this into a place where we can call it 'GA'?"
> 
> In the interest of incremental implementation, I don't think that logic needs to be applied to this proposal or the initial implementation.

Yes, we agree here.

> > How can we make the HTTP API accessible enough that people will actually use it and give us feedback?
> 
> To be honest, I don't really think of this in that way. It's mostly too late for feedback as our API already exists and is GA. The only feedback we can get is URL path and that is mostly subjective and dependent upon what exists. And I don't think we need to heavily promote just adding a second protocol to our existing API. This is an existing API, not a new API. This isn't like developing a new product IMO, just a tiny feature on an existing product.

Ah, so my point was more about accessibility than feedback per se. (Though I do think we will want feedback, even if only so that we can consider it if/when we ever decide to do something more RESTful.)

More specifically: If not via OpenAPI, then how will people figure out how gRPC maps to HTTP—e.g. what URL paths are available to use and what those paths accept and return?

cretz: > If not via OpenAPI, then how will people figure out how gRPC maps to HTTP

OpenAPI can be added later. How do people today figure out what the RPC calls are? If the answer is "look at the proto" and that is good enough for us then the same answer applies to HTTP. Having said that, I would expect docs could be auto-generated. This is what many gRPC+HTTP companies do today, mostly following https://aip.dev/ like https://cloud.google.com/service-infrastructure/docs/apis (which are gRPC-to-HTTP APIs like this proposal proposes IIRC).

cretz:
New updates:

* A new "Payload Formatting" section has been written with research on how to clean-up the JSON-ification of payloads
* A new "Usage Examples" section has been added with some `curl` calls
* `UpdateWorkflowExecution` now has a `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/update/{update_name}` endpoint with auto-generated ID to match signal/query, though the `POST /api/v1/namespaces/{namespace}/workflows/{workflow_id}/updates/{update_id}` still exists

There is another update coming with pros/cons compared to redesigning an HTTP API from scratch.

cretz:
By request, added two rejected alternatives. One is for API redesign on top of our existing RPC (with some samples), and another is for entirely new frontend service (not in very much detail).

api/http-api.md
```diff
@@ -168,10 +170,471 @@ Current suggested approach, subject to discussion:
...
+}'
+```
+​
+Response:
+​
+```json
+{ }
+```
+
+## Rejected Alternative Approach - Full API Redesign
```
josh-berry, line 438: I think there is a "middle path" here that we haven't considered, which is to extend the gRPC API to be more RESTful (where needed; seems like it's already close?), make that an explicit design goal of gRPC going forward, and then build the HTTP API on top of that.

We still get the benefits of having HTTP be "merely" a layer on top of gRPC (reduced time to ship, easier to maintain parity between gRPC and HTTP), and we also get something that looks nice HTTP-wise.

And yes, I'm aware this is just shifting the problem around, code-wise, but I think it might also reduce some of the cons you mention below re. re-reviewing every single gRPC API, and being at the mercy of gRPC.

I'm not advocating we have to clean up gRPC right now, necessarily—just that we give it a bit of thought, capture pros/cons here, and see if this is something we can work towards.

api/http-api.md
```diff
@@ -168,10 +170,471 @@ Current suggested approach, subject to discussion:
...
+  * But newer `protojson` does not support this, see [this issue](https://github.com/golang/protobuf/issues/1322) and
+    others. So this could harm our ability to upgrade.
+* We can make `go.temporal.io/api/common/v1.Payloads` impl JSON marshal/unmarshal to assume if it is given a JSON
+  collection, it can assume it's for the "payloads" field
+* We can make `go.temporal.io/api/common/v1.Payload` impl JSON marshal/unmarshal to assume if it not an object that can
+  deserialize into payload using strict rules (e.g. exact `metadata` and `data` format), then treat it as a `json/plain`
+  payload for non `null` or null payload for null value.
+  * We could just say "without metadata assume JSON/null" but what if someone has an object with a "metadata" field?
+  * The only way it will clash with the raw form is if someone happens to have a `metadata` and/or `data` field with
+    proper base64
```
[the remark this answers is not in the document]
cretz: > I think any distinction between "shorthand" and "non-shorthand" should be explicit and out-of-band, and I don't think we should try to be clever about detecting one or the other. Otherwise we will inevitably run into corner cases like this one.

I defined the cases fairly clearly here. This edge case will likely never happen and therefore should not justify killing the shorthand approach. Do you have a proposal for how someone should provide a "shorthand" JSON payload that is not base64 encoded?

Are you saying instead of something like:

```bash
curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id -d '{
  "workflowType": { "name": "MyWorkflow" },
  "taskQueue": { "name": "my-task-queue" },
  "input": ["Hello, World!"]
}'
```

Something like:

```bash
curl -X POST http://localhost:7243/api/v1/namespaces/my-namespace/workflows/my-workflow-id -d '{
  "workflowType": { "name": "MyWorkflow" },
  "taskQueue": { "name": "my-task-queue" },
  "input": [{
    "_jsonPayload": true,
    "data": "Hello, World!"
  }]
}'
```

I think that's a big leap. I think just because I pointed out an extreme edge case doesn't mean we should assume it will ever be hit. Granted more research is required to even see if I can do this custom decoding at all.

api/http-api.md
cretz: > I think there is a "middle path" here that we haven't considered, which is to extend the gRPC API to be more RESTful (where needed; seems like it's already close?), make that an explicit design goal of gRPC going forward, and then build the HTTP API on top of that.

It was never considered it because I don't think altering or adding more gRPC calls is acceptable for this project.

> seems like it's already close?

Can you give a concrete example here?

> And yes, I'm aware this is just shifting the problem around, code-wise

Not just shifting the problem around, we would be adding gRPC API surface area just to satisfy this little difference.

> I'm not advocating we have to clean up gRPC right now, necessarily—just that we give it a bit of thought, capture pros/cons here, and see if this is something we can work towards.

I think that can be done external to this proposal.

api/http-api.md
```diff
@@ -0,0 +1,640 @@
...
+    * This is not a `PATCH` to avoid confusion with update above
+  * `ListScheduleMatchingTimes` - `GET /api/v1/namespaces/{namespace}/schedules/{schedule_id}/matching-times`
+  * `DeleteSchedule` - `DELETE /api/v1/namespaces/{namespace}/schedules/{schedule_id}`
+    * Confirm whether eventually consistent
+  * `ListSchedules` - `GET /api/v1/namespaces/{namespace}/schedules`
+* Batch operations:
+  * `StartBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `DescribeBatchOperation` - `GET /api/v1/namespaces/{namespace}/batch-operations/{job_id}`
+  * `ListBatchOperations` - `GET /api/v1/namespaces/{namespace}/batch-operations`
+  * `StopBatchOperation` - `POST /api/v1/namespaces/{namespace}/batch-operations/{job_id}/stop`
```
josh-berry, line 148: Why is this not a `DELETE`?

cretz: From rules:

> `DELETE` if item is actually deleted (i.e. unreachable from `GET`)

This is a request to stop a job, not delete it.

josh-berry: Ah okay. I was confused because `DELETE /api/v1/namespaces/{namespace}` is a DELETE, but then I forgot about your comment right below it until just now. I assume namespace-delete would also change to a `POST` if it's eventually-consistent.

api/http-api.md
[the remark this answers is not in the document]
josh-berry: Sure, let's find out if we can do custom decoding first, and once there is a more concrete proposal, we can revisit this.

In general, though, I'm still very suspicious of mixing arbitrary properties with properties that have special behaviors together in the same object because, given enough time and enough users, someone who isn't aware eventually WILL hit it and have a bad day. My suspicion goes even deeper when we are talking about both magic keys AND magic values (e.g. "values must be base64-encoded or a different behavior applies")—those tend to be even harder to debug.

api/http-api.md
cretz: Exactly, and it probably is eventually consistent.

api/http-api.md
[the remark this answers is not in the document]
cretz: I am not planning on a another, more concrete proposal.

> In general, though, I'm still very suspicious of mixing

That makes sense, but we have to consider all options. Can you put forth an alternative proposal for how a payload should look?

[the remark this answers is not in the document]
josh-berry: Roughly what I am thinking is: make "shorthand mode" be something that is turned on or off entirely on a per-request basis, either by an HTTP header or a top-level JSON property in the request body. It would therefore not be possible to mix shorthand and non-shorthand payloads.

This does have the downside of making it harder for users to compose an individual request out of different parts, since the fully-formed request would have to be consistent in its use of shorthand or not. I think this is a worthwhile trade-off because a failure due to "mix-and-match" is more obvious—a user can look at the failed request and see that it mixes shorthand and non-shorthand. Whereas if the server tries to guess, it's not immediately obvious what's happening if nothing in the request itself clues in the user—who didn't read the docs, by the way—that "hey, shorthand is a thing that exists".

I have no opinion on:
- Whether to use an HTTP header or a property in the request body. Some options:
  - Baked into Content-Type: `Content-Type: application/json; temporal-payloads=shorthand`
  - Custom HTTP header: `X-Temporal-Payloads: shorthand`
  - JSON property in request body: `{"_payloads": "shorthand", ...}`
- Whether shorthand mode is on or off by default (i.e. what the behavior is if the payload mode is unspecified)
- The name of the parameter used to enable/disable shorthand mode

If you want more details than this, I'll probably need your help to get a deeper understanding of payloads, the context/places they're used, etc. given that all of this is new to me. :)

api/http-api.md
josh-berry: > > I think there is a "middle path" here that we haven't considered, which is to extend the gRPC API to be more RESTful (where needed; seems like it's already close?), make that an explicit design goal of gRPC going forward, and then build the HTTP API on top of that.
> 
> It was never considered it because I don't think altering or adding more gRPC calls is acceptable for this project.

Why would it not be acceptable?

> > seems like it's already close?
> 
> Can you give a concrete example here?

From a cursory glance, Schedules, Namespaces, and Batch Operations all seem to have very CRUD-like gRPC interfaces. I haven't gone through the protobuf in detail, though, so I'm happy to be corrected if there are some hidden details I'm missing.

(And I will point out workflows as a good counter-example—the gRPC API for workflows does not seem to be very RESTful, based on my reading of this proposal alone.)

> > And yes, I'm aware this is just shifting the problem around, code-wise
> 
> Not just shifting the problem around, we would be adding gRPC API surface area just to satisfy this little difference.

Right, we would be. If this is a "little difference", I don't see that it would add very much surface area to gRPC, though? Again, happy to be corrected.

> > I'm not advocating we have to clean up gRPC right now, necessarily—just that we give it a bit of thought, capture pros/cons here, and see if this is something we can work towards.
> 
> I think it's an unreasonable rejected alternative, but then again I thought that about the ones I was forced to add here and any that are "what ifs" beyond the goals. I think if we're going to work towards it, it should be a real proposal not a rejected alternative (I don't think we'll ever work towards it though).
> 
> Having said that, if we need a "Rejected Alternative Approach - More RPC Methods" I can add it.

I know a decision has already been made internally and I'm not going to hold up the train on this any further, but I do think it would be good to briefly document what we decided and why (which I'm still not clear on).

api/http-api.md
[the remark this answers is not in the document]
cretz: I think opting in to the clearer approach to payloads is a very bad idea and ruins the benefit. I think you are overestimating the potential for clash. I regret saying there is possibility for ambiguity. Can you give the example of the situation where you would see a clash to justify forcing everyone do to this opt-in? I think you're harming everyone to solve a hypothetical that in reality will never occur. If, in some very strange reality, ambiguity did occur I can show how it would be handled.

I think it's very clear to demonstrate the problem you are seeing before trying to solve it here.

api/http-api.md
cretz: > Why would it not be acceptable?

I guess I should have made it clearer in the prelude that it is a goal of the project not to alter or add to the existing API. But to clarify - this project does not want to alter or add to the existing API. I would suggest putting forth an alternative proposal if this is a new goal.

> From a cursory glance, Schedules, Namespaces, and Batch Operations all seem to have very CRUD-like gRPC interfaces. I haven't gone through the protobuf in detail, though, so I'm happy to be corrected if there are some hidden details I'm missing.

Namespaces is rarely used (not even available on cloud) so to alter its API would likely be a waste of effort. Batch doesn't have things like update. Schedules very much does not follow basic CRUD wrt updates. To clarify my request, "can you give a concrete example of proto file alterations you would like to see?"

> If this is a "little difference", I don't see that it would add very much surface area to gRPC, though? 

Can you give a concrete example of proto file alterations you would like to see? It will be much easier if we can discuss practicalities instead of hypotheticals. I think you'll find you'll either be suggesting changes to API that is rarely used anyways, or significant alterations to the commonly used API.

> I do think it would be good to briefly document what we decided and why (which I'm still not clear on).

That was the goal of this document. I assumed I wasn't going to have to document obvious things like "why HTTP?" and "why not change the existing API?" but if I need to, ok.

api/http-api.md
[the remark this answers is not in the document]
josh-berry: > I think opting in to the clearer approach to payloads is a very bad idea and ruins the benefit. [...]

We can make it opt-out (i.e. turn on shorthand by default) instead of opt-in. As I said earlier, I have no opinion on whether it's opt-out or opt-in.

> I think it's very clear to demonstrate the problem you are seeing before trying to solve it here.

You yourself gave an example earlier:

>  * The only way it will clash with the raw form is if someone happens to have a `metadata` and/or `data` field with proper base64

api/http-api.md
josh-berry: > > Why would it not be acceptable?
> 
> I guess I should have made it clearer in the prelude that it is a goal of the project not to alter or add to the existing API. But to clarify - this project does not want to alter or add to the existing API. I would suggest putting forth an alternative proposal if this is a new goal.

That's been clear from the discussion; I'm asking why it's a non-goal.

To be super clear/explicit: given the decision that was already made internally, I would be happy / my concern here would be addressed if we add one more bullet point under [Non-Goals](https://github.com/temporalio/proposals/pull/79/files#diff-7e381adf57b23564610f84e9ffa150a2cf7a558a0e0c6114eb83c3eff42e7487R14) that says something to the effect that we have decided not to touch gRPC because: time/we'll do it later/we don't think it will ever be necessary because $REASON/whatever the actual reason is.

I think the rest of your comment is moot given that gRPC is out of scope. Happy to discuss further if/when we get to that point, though.

api/http-api.md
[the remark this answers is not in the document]
cretz: > We can make it opt-out (i.e. turn on shorthand by default) instead of opt-in. As I said earlier, I have no opinion on whether it's opt-out or opt-in.

This is hard to do because at deserialization time, we don't have contextual data to make this decision. https://pkg.go.dev/github.com/gogo/protobuf/jsonpb#JSONPBUnmarshaler is what we will implement. There is no external "am I in shorthand mode". We have a possible approach to do this in the https://pkg.go.dev/github.com/gogo/protobuf/jsonpb#JSONPBMarshaler direction because I can https://pkg.go.dev/go.temporal.io/api/proxy#VisitPayloads all the payloads and mark them as "in shorthand mode". I am open to options here though. I can probably cheat by making some kind of custom any resolver and checking that type, but it is ugly.

> You yourself gave an example earlier:

I meant an actual curl call that shows an HTTP call with the clash. I think you'll find it's rare. But at this point, I have explained all I can here.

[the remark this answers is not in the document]
josh-berry: > > We can make it opt-out (i.e. turn on shorthand by default) instead of opt-in. As I said earlier, I have no opinion on whether it's opt-out or opt-in.
> 
> This is hard to do because at deserialization time, we don't have contextual data to make this decision. https://pkg.go.dev/github.com/gogo/protobuf/jsonpb#JSONPBUnmarshaler is what we will implement. There is no external "am I in shorthand mode". We have a possible approach to do this in the https://pkg.go.dev/github.com/gogo/protobuf/jsonpb#JSONPBMarshaler direction because I can https://pkg.go.dev/go.temporal.io/api/proxy#VisitPayloads all the payloads and mark them as "in shorthand mode". I am open to options here though. I can probably cheat by making some kind of custom any resolver and checking that type, but it is ugly.

I understand implementation may be hard/ugly, and if that turns out to be the case, we can always revisit whether it's worth the effort. Do you agree this is ideally what we should be doing, though, or do you have any other concerns?

> I meant an actual curl call that shows an HTTP call with the clash. I think you'll find it's rare. But at this point, I have explained all I can here.

100% agree it would be very rare. My worry is that when someone does run into it, it's going to be extremely frustrating and non-obvious and may cost someone a bunch of time and effort to figure out, and that will be a bad user experience.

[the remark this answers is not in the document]
cretz: > Do you agree this is ideally what we should be doing, though, or do you have any other concerns?

I think the devil is in the details. Do you have a suggested way to implement `UnmarshalJSONPB`? How in that method do you know whether you're in "shorthand" mode or not?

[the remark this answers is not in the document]
josh-berry: > > Do you agree this is ideally what we should be doing, though, or do you have any other concerns?
> 
> I think the devil is in the details. Do you have a suggested way to implement `UnmarshalJSONPB`?

I'm intentionally not thinking about implementation details until we are aligned on what the goal actually is. If we agree that, in principle, a per-request flag is preferable to auto-detecting the format for each payload, then we can talk about whether/what/how much implementation complexity is an acceptable trade-off.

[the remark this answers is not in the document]
cretz: The goals are clearer when you see how they manifest. Having said that, ok I agree, regardless of implementation we'll add a flag. I don't think it'll ever be used, but at this point it's less work than continuing this thread.

api/http-api.md
josh-berry: > > I think the rest of your comment is moot given that gRPC is out of scope.
> 
> What do you mean by "out of scope"? You're suggesting gRPC alterations correct? Or are you saying "my suggestions are out of scope so justifying them or giving concrete samples is moot"?

I mean that the decision has already been made to exclude any gRPC changes from the scope of this project. So I don't think it's worth either of us spending any more time on this, other than to make sure we document that the decision was made and the reason for that decision.

You and Ryland have already reached an agreement here, and I don't want to get in the way of that (especially as someone who is new to the party).

> I think it's as reasonable for me to ask for concrete examples of API alterations you want to make as it is for you to ask why I don't want to make API alterations. It is much easier to respond to the API alterations you are suggesting if I know what API alterations you are suggesting. Can you provide these so we can discuss them instead of hypotheticals? Then it may be clear the harm of your suggestions are going to vastly outweigh their benefits.

I'd be happy to do that if/when we ever decide we actually want to make gRPC and HTTP align with each other. I don't think it's worth writing a proposal right now given that we have already decided not to pursue it.

cretz: > I don't think it's worth writing a proposal right now given that we have already decided not to pursue it.

:+1: In the future, it will make discussions like this thread easier. If we're willing to discuss alternatives, I think we should be willing to show exactly what we mean if it's otherwise unclear and/or would help the discussion.

cretz:
New updates:

* Add non-goal for "Alter existing gRPC API"
* Clarified non-goals are for project scope and to not affect what exists in any way and that this proposal is specifically a limited-effort approach
* Noted that `?pretty` query parameter can pretty print JSON output
* Noted that `?noPayloadShorthand` query parameter can disable payload shorthand parsing/formatting

josh-berry (approved):

MaiHoangViet1809:
> > We can do it. But OpenAPI/swagger often sucks anyways..
> 
> We have OpenAPI in ui-server, i think it can stay there unless we decide to move it server. Can be accessed by `temporal server start-dev` and navigating to http://localhost:8233/openapi/ cc @mbernier

404 not found for the latest version when I tried to access "http://localhost:8233/openapi/", either it have not finished yet or someone document way too soon.

cretz:
@MaiHoangViet1809 - this is a PR for the proposal from a while back. Can ask on Slack or community forums if needing assistance.

PR #80, opened 2023-05-19
Error Curation

cretz:
Error curation to catalog and organize our errors/rules

See [rendered](https://github.com/cretz/temporal-proposals/blob/error-curation/error-curation.md).

Please review.

### DIFF
--- a/error-curation.md
+++ b/error-curation.md
@@ -1,0 +1,87 @@
+# Error Curation
+
+## Overview
+
+Many errors occur in Temporal yet pithy error messages alone are not enough to give context. We need to provide
+easy-to-reference identifiers to each and pages on each to help.
+
+**Goals**
+
+These are goals for the error curation project, not this proposal in particular though there is overlap.
+
+* Document requirements for rule numbering/identification
+* Document requirements for rule documenting
+* Provide home for all rule documents
+* Build initial set of rules for use in SDKs
+
+**Non-Goals**
+
+* Document every error
+* Make the error documents available on docs.temporal.io at first
+* Provide any UX linking errors from UI, CLI, etc
+* Extremely strict documentation guidelines
+
+## Approach
+
+This approach is inspired by .NET analyzers which have coded values and dedicated pages for each error. For example,
+see:
+
+* https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/categories (e.g.
+  https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca1001)
+* https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/DOCUMENTATION.md (e.g.
+  https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1200.md)
+* https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/index.md (e.g.
+  https://github.com/microsoft/vs-threading/blob/main/doc/analyzers/VSTHRD003.md)
+
+Solution:
+
+* A "rule" is referenceable, documented error/warning/note/guideline
+* We are calling these "rules" not "errors"
+  * They may be found eagerly, may just be disable-able warnings, or may even be just best practice rules
+* Rules can be for many things:
+  * Errors (e.g. SDK errors, static analysis errors, server errors, etc)
+  * Guidelines (e.g. best practices, warnings, etc)
+* Every rule has an identifier in the form of `TMPRLXXXX`
+  * "Categories" of rules are given a 100 block as rarely is more needed
+    * Category 100 blocks should try not to be next to another so space can be given in some rare case we exceed 100 in
+      a category
+    * There are no rules on what a category is, just a grouping of like things
+  * The rules within a category do not have to be ordered in any manner. Usually you just use the next available.
+  * `0` is never used in the first two digits
+  * Can discuss the first thousands-block digit though initial thought for first two is:
+    * `TMRPL1XXX` block is for SDK-specific and workflow/activity authoring things
+    * `TMPRL2XXX` block is for server-specific and operation things
+* Rule/error messages across Temporal should be prepended with `"[TMPRLXXXX] "` (sans quotes)
+  * Note the brackets and space after.
+  * We are intentionally not providing a stricter requirement than this. There is intentionally no requirement to make
+    these available for programmatic consumption and we are not even going to guarantee these are even present or there
+    is a stable URL to learn more.
+  * SDK, gRPC API and other programmatic places where errors occur do not need to go out of their way to separate this
+    from the message string. There is limited value. From POV of code, this might as well be the beginning of an English
+    sentence and no more. Forcing stricter format/abstraction is more cost than benefit.
+* All rules are housed in a single repository
+  * Maybe something like https://github.com/temporalio/temporal-rules https://github.com/temporalio/rule-library
+  * `rules/README.md` will be the primary index of all categories
+  * `rules/TMRPL11XX.md` with the literal `XX.md` suffix but the two-digit category will be the index of all rules in
+    that category
+  * `rules/TMRPL1234.md` will be the literal page
+    * Consideration was given to breaking these out into subdirectories, but experience has shown a nice, common
+      `https://github.com/temporalio/temporal-rules/blob/main/rules/<RULE_NUM>.md` URL for _all_ rules has a lot of
+      benefits
+  * This project does not concern itself with the eventual promotion of this content to regular docs, just that it needs
+    to be in a repository for review/submission by Temporal developers
+* The `rules/TMPRL1234.md` file will contain rule details
+  * No strict metadata/frontmatter required at this time
+  * Ideally the rules should have at least the following headings:
+    * `# <TMPRL1234> - <one-sentence-description>`
+    * `## Cause`
+    * `## Description`
+    * `## Solution`
+  * More headings can be added
+  * None of the contents/headings are strict/rigid/required, they are just suggestions
+  * Authors are encouraged to get as detailed as they'd like with code snippets, stable links, hypotheticals, use cases,
+    etc
+* The SDKs will be crawled for error/warning messages, and have ones that are categorizable/documentable errors will be
+  cataloged as rules, be given identifiers, have rule pages created, and have the strings prepend the identifiers
+  * Note, initial version of the project may only do some of these
+* Static analyzers (of which we only have one at the moment) will be updated with rules similar to SDKs above


error-curation.md
```diff
@@ -0,0 +1,87 @@
...
+  * "Categories" of rules are given a 100 block as rarely is more needed
+    * Category 100 blocks should try not to be next to another so space can be given in some rare case we exceed 100 in
+      a category
+    * There are no rules on what a category is, just a grouping of like things
+  * The rules within a category do not have to be ordered in any manner. Usually you just use the next available.
+  * `0` is never used in the first two digits
+  * Can discuss the first thousands-block digit though initial thought for first two is:
+    * `TMRPL1XXX` block is for SDK-specific and workflow/activity authoring things
+    * `TMPRL2XXX` block is for server-specific and operation things
+* Rule/error messages across Temporal should be prepended with `"[TMPRLXXXX] "` (sans quotes)
```
MichaelSnowden, line 54: For implementing this in Go, I think we can leverage golangci-lint, which has things like enforced error-wrapping via goerr113

cretz: I think that's overthinking at first, but if that's what y'all want to do in your codebase, ok. Not only would I not expect these identifiers on every error (that's silly, not all errors deserve a rule), there's no need to add friction. I've seen these projects get out of hand with overthink/overabstraction or too-rigid of requirements on developers. Just keep it simple, prepend your error string literals on errors that deserve to be cataloged, and move on.

paulnpdev: How (if at all) are we going to try to maintain a correspondence between these error codes and (future) error codes in our source?  A small group of us are considering improvements to how we code error logging, and one popular idea is to assign unique error codes to each log message.  Many of these code locations are also the “source” of the errors you are delineating in this proposal. Maybe we should just admit that trying for any synchrony here is a fool’s game, but I wanted to at least raise the question.  Also note that there will be a lot more than 1000 of the other error codes. I wonder if you shouldn’t add one more digit just on general principles

cretz: This is one of the reasons I avoided calling these error codes. They are more "rules" and only applicable for a reasonable subset of situations worth documenting. Some deep down software error may not really be worth having its own page with mitigation and everything. And sometimes many errors are really violations of the same rule (e.g. `TMPRL1322` may be for invalid input/output data type and put on lots of errors in many SDKs).

The goal here is more to have a reasonably sized set of identifiers that users can reference and read about. If there is an effort to add error codes at a lower level or more granular, I think it should be a bit different than these identifiers.

> Maybe we should just admit that trying for any synchrony here is a fool’s game

Yes! This is exactly where I'm at. If we keep a reasonably sized enough list, not only can a developer go see if a situation applies to the error they are about to throw, but a user can literally scroll the list to see bad things they could do. That's why I linked the .NET rules as inspiration because they fit more of my mental model of rules vs error codes.

I have no doubt we'll miss some, have to go add later, etc. But I am hoping we don't get too granular with them.

> Also note that there will be a lot more than 1000 of the other error codes. I wonder if you shouldn’t add one more digit just on general principles

8099 in this case I think or something like that (`1100-9999` with no `0`s in second digit), but really for me it was about identifiers vs string length and the category blocks. For how I was hoping they'd be used, I didn't figure a category would need more than 100. Honestly don't even care that much about first digit, I have ~80 categories to give out which I think is ok.

If we do thing we need more, I would support alternative prefixes besides `TMPRL` which is already a mouthful. And maybe decimals, e.g. `TMPRL3563.12` but I am hoping not tbh. Or we could even get into letters for digit parts. I'd be willing to make this 5 digits if we can take a character off `TMPRL` somewhere.

paulnpdev: TMPRL => TIO? (aka t.io)
I keep hoping we'll actually buy that domain!

cretz: We do own `t.mp` and I fully expect something like `t.mp/rules/TMPRL1234` does the proper redirection. Also unsure if `.io` is something we want to make part of the company abbreviation name here. `TPRL` is an option. Will let others weigh in.

error-curation.md
```diff
@@ -0,0 +1,87 @@
...
+      `https://github.com/temporalio/temporal-rules/blob/main/rules/<RULE_NUM>.md` URL for _all_ rules has a lot of
+      benefits
+  * This project does not concern itself with the eventual promotion of this content to regular docs, just that it needs
+    to be in a repository for review/submission by Temporal developers
+* The `rules/TMPRL1234.md` file will contain rule details
+  * No strict metadata/frontmatter required at this time
+  * Ideally the rules should have at least the following headings:
+    * `# <TMPRL1234> - <one-sentence-description>`
+    * `## Cause`
+    * `## Description`
```
lorensr, line 78: wonder if usually the cause has ~all the description?

error-curation.md
```diff
@@ -0,0 +1,87 @@
...
+  * "Categories" of rules are given a 100 block as rarely is more needed
+    * Category 100 blocks should try not to be next to another so space can be given in some rare case we exceed 100 in
+      a category
+    * There are no rules on what a category is, just a grouping of like things
+  * The rules within a category do not have to be ordered in any manner. Usually you just use the next available.
+  * `0` is never used in the first two digits
+  * Can discuss the first thousands-block digit though initial thought for first two is:
+    * `TMPRL1XXX` block is for SDK-specific and workflow/activity authoring things
+    * `TMPRL2XXX` block is for server-specific and operation things
+* Rule/error messages across Temporal should be prepended with `"[TMPRLXXXX] "` (sans quotes)
```
lorensr, line 54: What about one of these?

```
[TMP1234] Context deadline exceeded. For more info, see: t.mp/1234
[TMP1234] Context deadline exceeded. More info: t.mp/1234
[TMP1234] Context deadline exceeded t.mp/1234
[t.mp/1234] Context deadline exceeded
Context deadline exceeded t.mp/1234
```

My favorite is 4 or 5

lorensr (approved):

error-curation.md
cretz: Yeah, I could see that for simple stuff. I mention this is not a strict guideline for things like this. The reason for description separate is to separate "why" from "where". But can be combined if simple enough, sure.

error-curation.md
cretz: I think we need to filename-safe number and I wasn't wanting to alter all of our English messages, just a simple prefix

lorensr: If the ID is a URL, then user doesn't need to Google it. Why filename safe? t.mp/1234 could point to a `1234.md` file.

My top goal is having a URL somewhere in error messages. Secondary goal is brevity—of message, code, and URL. If we only have a prefix, then having the URL as the prefix meets both goals.

If we have both a code and a URL, and the URL is at the end of a potentially long message and possibly missed, then it would be useful for the code to be uniquely googleable, and could still keep the URL short:

```
[TMPRL1234] Context deadline exceeded t.mp/1234
```

(I just personally find the longer message and maybe-not-seeing-link-and-googling to be a less good DX)

cretz: > If the ID is a URL, then user doesn't need to Google it

I'll take the .NET-style/RFC-style/etc-style rule identifier over URL, even if it costs someone a search. It stands out as an error code like others before it when developers glance at the string.

> Why filename safe? t.mp/1234 could point to a 1234.md file.

I have found this easiest when using the existing .NET rules I cited as inspiration. It is easy to reference that string in filename, title, error string, etc. You find this a lot with static analysis tools, which is admittedly where I plan on leveraging this heavily.

> My top goal is having a URL somewhere in error messages.

Wasn't one of my goals listed at the top of the project. Would prefer not to add goals/requirements if we can.

> If we have both a code and a URL

I don't think we need both. But I support easy to reach rules via URL though (and easy to reach via search, and easy to recognize as error codes in strings, etc).

lorensr: >> If the ID is a URL, then user doesn't need to Google it
>
> I'll take the .NET-style/RFC-style/etc-style rule identifier over URL, even if it costs someone a search. It stands out as an error code like others before it when developers glance at the string.

Would all caps stand out enough?

```
[T.MP/1234] Context deadline exceeded
```

>> My top goal is having a URL somewhere in error messages.
>
> Wasn't one of my goals listed at the top of the project. Would prefer not to add goals/requirements if we can.

Since it's so similar and also would involve changing error messages, I think it makes sense to discuss here? Versus eg I open another proposal with the goal of "make it easy for users to get from error message to error page" and we change all the error messages again.

lorensr: When there's a short fix, I like reading it in the message, like:

```
No workflows found in src/workflow.ts. Make sure to export your workflow functions.
```

I'd argue against long explanations since they're harder to read in the place you're reading errors versus a formatted markdown page and more likely to need to be updated, which is easier to do in one place than many.

>> open another proposal with the goal of "make it easy for users to get from error message to error page" and we change all the error messages again.
> 
> I think this proposal is good enough for that use case and we are over-optimizing going any further. 

I think it's good enough for some, but not all, and think it might be worth going further. Would be curious to hear if @mbernier @tlalfano has an opinion or process for answering user questions like this. I don't know the best answer, I think it depends on how many people are in each con category.

### Option A: traditional error codes

```
[TMPRL1234] Context deadline exceeded
```

Con: some won't know to google for TMPRL1234

### Option B: URL error codes

```
[t.mp/1234] Context deadline exceeded
or
[T.MP/1234] Context deadline exceeded
```

Cons: 
- those used to traditional error codes may not see/recognize/pay attention to them
- some may not realize it's a URL

### Option C: traditional error codes + URLs

```
[TMPRL1234] Context deadline exceeded t.mp/1234
```

Con: some may not realize there's a URL, particularly if it's a long error message

### Option D: traditional error codes + explicit URLs

```
[TMPRL1234] Context deadline exceeded. More info: t.mp/1234
```

Cons: 
- longer
- some may not get to the URL if it's a long error message

Pro: `. More info: ` makes it more clear that it's a URL with more info

cretz: I was afraid of this. There's just too much subjectivity here and it seems like the principled approach just isn't going to win over discussion, voting, options, etc. I was hoping for simple identifiers I could e.g. `#pragma warning disable TMPRL1234` in code and just follow some of that inspiration I linked. But alas it may not be. We have seen this with other projects that just end up taking months instead of the few days they should have (and often end up at a bad answer anyways).

I think it's becoming clear this project isn't ready for a proposal. I will bring it up with team internally and see if we can close this pending more research and/or delegation to another team.

error-curation.md
```diff
@@ -0,0 +1,87 @@
...
+      benefits
+  * This project does not concern itself with the eventual promotion of this content to regular docs, just that it needs
+    to be in a repository for review/submission by Temporal developers
+* The `rules/TMPRL1234.md` file will contain rule details
+  * No strict metadata/frontmatter required at this time
+  * Ideally the rules should have at least the following headings:
+    * `# <TMPRL1234> - <one-sentence-description>`
+    * `## Cause`
+    * `## Description`
+    * `## Solution`
```
Sushisource, line 79: I might call this "Remediations" rather than "Solution" since not everything is going to be totally solvable, and perhaps there's more than one.

error-curation.md
Sushisource: I do think every genuinely _error_-level event that becomes user visible should get one of these. However, finding those situations statically isn't suuuuper doable.

Probably, the best approach here is to be able to run the linter and then just mark a bunch of stuff as ignored. I think it'd be good to have that turned on, so that devs adding new errors are forced to choose between adding the code and saying "no it's ok this gets an ignore". It's a slight annoyance but one that will ensure we keep up with this effort long run.

error-curation.md
Sushisource: I think whether or not you have a URL really depends on the context of what you're doing. I wouldn't want a whole bunch of them showing up in log messages for example. Rust compiler errors just have `error[E0609]: blah` for example, and then more about the location but clippy lints look like this:

```
error: useless conversion to the same type: `temporal_sdk_core_protos::temporal::api::workflowservice::v1::PollActivityTaskQueueResponse`
   --> core/src/worker/activities.rs:632:20
    |
632 |                   Ok(PollActivityTaskQueueResponse {
    |  ____________________^
633 | |                     task_token: vec![2],
634 | |                     activity_id: "act2".to_string(),
635 | |                     ..Default::default()
636 | |                 }
637 | |                 .into())
    | |_______________________^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#useless_conversion
help: consider removing `.into()`
    |
632 ~                 Ok(PollActivityTaskQueueResponse {
633 +                     task_token: vec![2],
634 +                     activity_id: "act2".to_string(),
635 +                     ..Default::default()
636 ~                 })
```

So, I don't think we really need super hard rules here, or maybe we add them over time. Logs / errors / panics maybe just have the code, our static analysis tools maybe behave more like clippy, etc.

I wouldn't let this stop the proposal at all. It's clearly something that can evolve as we implement.

Sushisource (approved):
I think this is fairly standard and straightforward. I think we should just start doing this, and this is a pretty obvious way to do it. We can always tweak as we move forward.

error-curation.md
lorensr: I'm very pro this proposal, and would much rather have it implemented now as written than get held up for months in user studies. I don't know the right amount of time to put toward the discussion, versus when to call it and go with what we have. If this is the last pending discussion item and we don't see a quick resolution (internal and/or external slack poll gathering data?), then moving forward with implementation and continuing to discuss this point later and maybe adding a URL at some point SGTM.

Quinn-With-Two-Ns (approved):

error-curation.md
cretz: :+1: FWIW I am not against URL adding in error messages, I just don't want to make it a part of this project and I don't want the rule identifiers to have to be URLs.

error-curation.md
cretz: Updated

cretz:
Some scratch notes as I begin toying w/ implementation of the proposal:

* Additional notes/requirements that we'll add to rules repo's contributing docs:
  * Don't be tempted to create an rule for every error you see. Remember this needs to be a list of rules that can be read, not a list of all possible rules that can be created. If the list cannot be viewed by a human, it's too big. It is easier to break a rule up than it is to combine later or copy/paste a bunch of shared docs across multiple. Make sure it's a reasonably sized, documentable topic you are creating a rule for.
  * Don't overthink about the order/grouping of the rules. They are going to come in as needed, so just take the next index in the high-level category and move on.
  * Rules are for Temporal use not user use. We do not encourage users to reuse our identifiers and we should not make rules that expect to have users use them.
  * Identifiers do not have to be set in all cases that Do not wrap errors just to get a rule identifier in. For example, if existing code just bubbles out a user error, say, for payload conversion failure, don't catch and wrap just to get this identifier in. The identifier is only needed on existing error messages.
  * Don't hesitate to create a rule for a best practice that you can't programmatically enforce yet (of course not all rules are for errors). We can still link to it.
  * Making a rule that only applies to a certain specific language is fine, but don't concern yourself with language-specific variants of the same rule. Language-specific causes/remedies can be mentioned/organized on the rule documentation. Language-specific concerns are only an SDK concern and currently we are not breaking out rule documentation by language for simplicity reasons.
* Initial set of rules:
  * TMPRL1100-1199 - SDK rules
    * TMPRL-11XX - Non-determinism detected on replay
      * Covers basically every SDK-side detection
    * TMPRL-11XX - Calling known non-deterministic call
      * More for static analyzer use in Go, but definitely useable in Python but kinda overlaps with above
    * TMPRL-11XX - Using non-deterministic concurrent construct
      * More for static analyzer use
    * TMPRL-11XX - Payload conversion failure
      * Includes conversion from JSON and conversion to values
      * TODO: break this up? Most not needed
    * TMPRL-11XX - Invalid workflow definition
    * TMPRL-11XX - Invalid activity definition
    * TMPRL-11XX - Invalid signal definition
      * Includes "__" reserved
    * TMPRL-11XX - Invalid query definition
      * Includes "__" reserved
    * TMPRL-11XX - Invalid update definition
      * This includes invalid validator definition
      * Includes "__" reserved
    * TMPRL-11XX - Invalid timer duration
    * TMPRL-11XX - Invalid query result size
      * TODO(cretz): Update result size too? General payload size?
    * TMPRL-11XX - Invalid activity options
      * Includes requiring start-to-close or schedule-to-close set
    * TMPRL-11XX - Invalid local activity options
    * TMPRL-11XX - Top-level workflow should have ID set when starting
      * This is just a static analyzer warning, not an error
    * TMRPL-11XX - Unknown workflow
    * TMRPL-11XX - Unknown activity
      * Includes local activity
    * TMRPL-11XX - Unknown signal
    * TMRPL-11XX - Unknown query
    * TMRPL-11XX - Unknown update
    * TMPRL-11XX - Unable to build binary checksum
    * TMPRL-11XX - Connection establishment failed
      * Needs to cover bad auth/mTLS too
      * Only needed for initial connection establishment (can use for reconnect too, but not required)
    * TMPRL-11XX - Unexpected event type
      * New event on server but SDK too old
    * TMPRL-11XX - Invalid search attribute type in SDK
      * Governs setting and decoding
      * This is not the same as it failing on server for invalid type
    * TMPRL-11XX - Session failed
  * TMPRL2100-2199 - General server-side rules
    * TODO
* Errors not worth rules at this time
  * TMPRL-11XX - Go client must have been created using new client - too detailed
  * TMPRL-11XX - Header propagation failure - internal validation issue
  * TMPRL-11XX - Invalid side effect marker - internal only
  * TMPRL-11XX - Invalid version marker - internal only
  * TMPRL-11XX - Child workflow already exists - more control flow than error
  * TMPRL-11XX - Child workflow cancel fail - more control flow than error
  * TMPRL-11XX - External workflow cancel fail - more control flow than error
  * TMPRL-11XX - External workflow signal fail - more control flow than error
  * TMPRL-11XX - Basic schedule validation errors - like other basic validation errors, just a language expressiveness issue
  * TMPRL-11XX - Not all errors handled - this is a language thing
  * TMPRL-11XX - Invalid history format on replay - not sure deserves a "rule"
  * TMPRL-11XX - Worker stopping - state, not a rule

lorensr:
> External workflow signal fail - more control flow than error

Wondering about the control flow thing. Where is this thrown from and why? 

> Invalid history format on replay - not sure deserves a "rule"

Like I give an invalid history JSON to a replay worker? I think it would be nice to have a page that explains reasons why it might be invalid, ways to get a valid history file.

cretz:
> Wondering about the control flow thing. Where is this thrown from and why?

This is in a section called "Errors not worth rules at this time" and this specific one is meant to say that a normal Temporal error for external workflow signal failure (e.g. workflow does not exist) is not worth a rule or documented practice at this time. Same with normal client errors.

> Like I give an invalid history JSON to a replay worker? I think it would be nice to have a page that explains reasons why it might be invalid, ways to get a valid history file.

That page may be valuable as general documentation, but it's not really a "rule" or practice worth documenting when a user just gives a badly formatted parameter like this. Same with, say, the `host:port` being bad on connection.

I was concerned we would fall in the trap of wanting to put codes and documents on even programming errors. We need a short, easy-to-view-as-a-whole list of "rules" that document practices in Temporal.

error-curation.md
josh-berry: I like having identifiers that are somewhat self-describing (e.g. `[TMPRL.sdk.whatever]`) but also agree with the need for brevity. Just noting it here as another option should we ever find ourselves with too many of these things (which I agree is unlikely—more than even ~1,000 rules means we probably need to rethink our system >.>).

josh-berry (approved):
Generally love it. One small suggestion on the rule explanation files, and one note about rule IDs (mainly for the future).

djmagee (approved):
I'm all for this proposal. A user-facing message that includes a distinct, searchable prefix not only lets us provide easily discoverable documentation but also provides a dead-simply way for community members to unambiguously reference a rule in our forum, Stack Overflow, or blog posts.

error-curation.md
```diff
@@ -0,0 +1,89 @@
...
+    to be in a repository for review/submission by Temporal developers
+* The `rules/TMPRL1234.md` file will contain rule details
+  * No strict metadata/frontmatter required at this time
+  * Ideally the rules may have at the following headings (but license is given to author to format as needed):
+    * `# <TMPRL1234> - <one-sentence-description>`
+    * `## Cause`
+    * `## Description`
+      * Only if cause is not enough information
+    * `## Remediations`
+      * Chosen over "Solutions" as not everything will be solvable
```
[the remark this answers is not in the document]
cretz: Can clarify (though I am considering removing this section as it is meant to be a loose idea). It is important to note the part above:

> Ideally the rules may have at the following headings (but license is given to author to format as needed)

So in those different-approach situations, the author can split.

error-curation.md
cretz: Not just brevity, but people are bad at naming those identifiers IME

cretz:
I am merging this, there's enough approval

PR #82, opened 2023-07-06
Add proposal for Nexus operation IDs

bergundy:
[Rendered](https://github.com/temporalio/proposals/blob/nexus-operation-ids/nexus/operation-ids.md)

### DIFF
--- a/nexus/operation-ids.md
+++ b/nexus/operation-ids.md
@@ -3,28 +3,31 @@
 
 - [Nexus RPCs - short recap](#nexus-rpcs---short-recap)
 - [Operation identifiers](#operation-identifiers)
+  - [Identifier requirements](#identifier-requirements)
+    - [Sharing identifiers between clients](#sharing-identifiers-between-clients)
   - [Caller generated identifiers](#caller-generated-identifiers)
     - [Handler addressability](#handler-addressability)
-      - [Mapping from client to handler identifiers](#mapping-from-client-to-handler-identifiers)
   - [Handler generated identifiers](#handler-generated-identifiers)
     - [Cancel before started](#cancel-before-started)
   - [Comparing the approaches](#comparing-the-approaches)
-    - [Handler Id qualification](#handler-id-qualification)
-    - [Ids as inputs](#ids-as-inputs)
+    - [Handler ID qualification](#handler-id-qualification)
+    - [IDs as inputs](#ids-as-inputs)
     - [Caller friendliness](#caller-friendliness)
     - [Data sensitivity](#data-sensitivity)
 - [Proposal](#proposal)
+  - [API draft](#api-draft)
+  - [Notes](#notes)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
 
 ## Nexus RPCs - short recap
 
-At the fundamental level, the Nexus protocol is a synchronous RPC protocol. Asynchronous operations are modelled on top
-of a set of pre-defined RPCs.
+At the fundamental level, the Nexus protocol is a synchronous RPC protocol. Arbitrary length operations are modelled on
+top of a set of pre-defined synchronous RPCs.
 
 A Nexus **caller** calls a **handler**. The handler may respond inline or return a reference for a future, asynchronous
-operation. These asynchronous operations can be cancelled and checked for their outcomes and current state. The caller
-can also specify a callback URL, which the handler uses to asynchronously deliver the result of an operation's when it
+operation. The caller can cancel an asynchronous operation, check for its outcome, or fetch its current state. The caller
+can also specify a callback URL, which the handler uses to asynchronously deliver the result of an operation when it
 is ready.
 
 ## Operation identifiers
@@ -32,141 +35,179 @@ is ready.
 Since an asynchronous operation is modelled on top of synchronous RPCs, it must be addressable (identifiable) by both
 the caller and the handler. There are two approaches for generating identifiers, caller side and handler side.
 
+### Identifier requirements
+
+#### Sharing identifiers between clients
+
+The may be situations where it's desirable to safely share access between different clients, e.g. start an operation
+with one client and get the result with another using the same identifier.
+In those cases identifiers should be unguessable, so that they can behave as an object capability.
+
+This is a non-goal and out of scope for Nexus at the moment but can optionally be implemented on either the caller or
+handler sides.
+
+Note that it's the handler's responsibility to qualify identifiers of different tenants in the system and they may
+choose to allow cross-tenant sharing.
+
 ### Caller generated identifiers
 
-With this approach, the caller provides an Id to each of the pre-defined RPCs.
+With this approach, the caller provides an operation ID to each of the pre-defined RPCs.
 
 The handler _should_ qualify an identifier when dealing with untrusted sources and multi-tenant applications.
 
-The following set of RPCs would be needed to model asynchronous operations using client generated Ids:
+The following set of RPCs would be needed to model asynchronous operations using client generated IDs:
 
-- `Start(operation, id, input, callbackURL)` - handler responds inline or indicates that an async operation was started
-- `Cancel(operation, id)` - cancel operation by Id
-- `GetResult(operation, id)` - get operation result by Id
-- `Describe(operation, id)` - get operation status by Id
-- `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
+- `Start(service, operation, id, input, callbackURL)` - handler responds inline or indicates that an async operation was started
+- `Cancel(service, operation, id)` - cancel operation by ID
+- `GetResult(service, operation, id)` - get operation result by ID
+- `Describe(service, operation, id)` - get operation status by ID
+- `DeliverResult(token, result)` - deliver a result of an async operation using a token in the Start request's callback URL
 
 Example invocations:
 
 ```js
 Start(
-  'charge',
+  "payments",
+  "charge",
   transactionId,
-  { customerId: 'someCustomerId', amount: 5000 },
-  'nexus://caller.example.com/callback?token=' + token
+  { customerId: "someCustomerId", amount: 5000 },
+  "nexus://caller.example.com/callback?token=" + token
 );
-Cancel('charge', transactionId);
+Cancel("payments", "charge", transactionId);
 ```
 
 #### Handler addressability
 
 In some cases, the handler might require more than just a single identifier for a given operation, e.g. when an
-operation is backed by a workflow update, which requires both a workflow Id and an update Id. In order to avoid
-maintaining a handler side persistent mapping between caller and handler identifiers, the handler may choose to "leak"
-this detail to the caller, which would need to concatinate multiple identifiers for a given call. The example above
-becomes:
+operation is backed by a workflow update, which requires both a workflow ID and an update ID.
+The handler may deal with this in a couple ways:
 
-```js
-Start(
-  'charge',
-  customerId + '/' + transactionId,
-  { amount: 5000 },
-  'nexus://caller.example.com/callback?token=' + token
-);
-Cancel('charge', customerId + '/' + transactionId);
-```
+1. Maintain a persistent mapping between caller and handler identifiers incurring overhead for every call.
+   Temporal may provide a strongly consistent store out of the box for this mapping or require wrapping these operations in another workflow execution.
 
-> NOTE: This looks a lot like a nested REST style URL, e.g: `/customers/{customerId}/charges/{transactionId}`.
+   A single workflow execution may be mapped to many client identifiers to account for workflow starts and updates.
 
-> To take this approach to the extreme, `operation` and `id` could be combined into a single `path` argument, as shown
-> below:
->
-> - `Start(path, input, callbackURL)`
-> - `Cancel(path)`
-> - `GetResult(path)`
-> - `Describe(path)`
+2. "Leak" this detail to the caller, which would need to concatenate multiple identifiers for a given call. The example
+   above becomes:
 
-##### Mapping from client to handler identifiers
+   ```js
+   Start(
+     "payments",
+     "charge",
+     customerId + "/" + transactionId,
+     { amount: 5000 },
+     "nexus://caller.example.com/callback?token=" + token
+   );
+   Cancel("payments", "charge", customerId + "/" + transactionId);
+   ```
 
-To avoid leaking the identifier construction logic to the client, the handler would need a mapping of client identifiers
-to handler identifiers. Temporal can provide a strongly consistent store out of the box for this mapping.
+   > NOTE: This looks a lot like a nested REST style URL, e.g: `/payments/customers/{customerId}/charges/{transactionId}`.
 
-A single execution would need to map to multiple client identifiers to account for workflow starts and updates.
-
-Looking up an identifier in persistent storage added latency, and should be opted in to.
+   > To take this approach to the extreme, `service`, `operation` and `id` could be combined into a single `path` argument, as shown
+   > below:
+   >
+   > - `Start(path, input, callbackURL)`
+   > - `Cancel(path)`
+   > - `GetResult(path)`
+   > - `Describe(path)`
 
 ### Handler generated identifiers
 
-With this approach, the handler returns a reference Id in response to a start call, which the caller should persist for
+With this approach, the handler returns a reference ID in response to a start call, which the caller should persist for
 use with the pre-defined RPCs as shown below:
 
-- `Start(operation, input, callbackURL, cancelBeforeStarted)` - handler responds inline or returns a reference Id
+- `Start(operation, input, callbackURL, cancelBeforeStarted)` - handler responds inline or returns a reference ID
   ([`*`](#cancel-before-started))
-- `Cancel(reference_id)` - cancel operation by handler provided reference Id
-- `GetResult(reference_id)` - get operation result by handler provided reference Id
-- `Describe(reference_id)` - get operation status by handler provided reference Id
+- `Cancel(reference_id)` - cancel operation by handler provided reference ID
+- `GetResult(reference_id)` - get operation result by handler provided reference ID
+- `Describe(reference_id)` - get operation status by handler provided reference ID
 - `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
 
-The handler _should_ either sign or qualify caller-provided reference Ids when dealing with untrusted sources and
+The handler _should_ either sign or qualify caller-provided reference IDs when dealing with untrusted sources and
 multi-tenant applications.
 
 In order to support stateless clients, the handler may provide an additional set of APIs to resolve client input to a
-reference Id. This may be something that is built into the protocol or implemented ad-hoc. To support this at the
+reference ID. This may be something that is built into the protocol or implemented ad-hoc. To support this at the
 protocol level, we would need to add the following APIs:
 
 - `Resolve(operation, input) -> reference_id`
 
 -- OR --
 
-- `Cancel(operation, input)` - cancel operation by handler derived Id
-- `GetResult(operation, input)` - get operation result by handler derived Id
-- `Describe(operation, input)` - get operation status by handler derived Id
+- `Cancel(operation, input)` - cancel operation by handler derived ID
+- `GetResult(operation, input)` - get operation result by handler derived ID
+- `Describe(operation, input)` - get operation status by handler derived ID
 
-NOTE: This looks a lot like the APIs for the "client generated Ids" approach, except the ID is structured instead of a
+NOTE: This looks a lot like the APIs for the "client generated IDs" approach, except the ID is structured instead of a
 single string.
 
 #### Cancel before started
 
-A caller may cancel an operation before it obtains a reference Id, e.g. if it is uncertain that the operation has been
-already started due to crashes and timeouts. Because the caller doesn't have a referencable Id or knowledge of which
-input fields are used to derive the Id, it would need to pass in the entire Start call input.
+A caller may cancel an operation before it obtains a reference ID, e.g. if it is uncertain that the operation has been
+already started due to crashes and timeouts. Because the caller doesn't have a referencable ID or knowledge of which
+input fields are used to derive the ID, it would need to pass in the entire Start call input.
 
 This is especially important when we want to automate cancellations of workflow initiated operations, Temporal server
-cannot derive the Id from opaque input.
+cannot derive the ID from opaque input.
 
 ### Comparing the approaches
 
-#### Handler Id qualification
+#### Handler ID qualification
 
-With handler generated Ids, to support stateless clients, the handler would need to be able to map an operation and
-input to a reference Id. It would also need to qualify a reference Id or securely issue the reference Ids when callers
-cannot be trusted. With caller generated Ids, the handler would only need to qualify an Id based on the caller context
-(e.g by concatinating it with a tenant Id).
+With handler generated IDs, to support stateless clients, the handler would need to be able to map an operation and
+input to a reference ID. It would also need to qualify a reference ID or securely issue the reference IDs when callers
+cannot be trusted. With caller generated IDs, the handler would only need to qualify an ID based on the caller context
+(e.g by concatenating it with a tenant ID).
 
-#### Ids as inputs
+#### IDs as inputs
 
-With the caller generated Id approach, if the Id is used for handler side addressability, the handler would need to
+With the caller generated ID approach, if the ID is used for handler side addressability, the handler would need to
 parse it and merge with the request's input.
 
 #### Caller friendliness
 
-Shifting the responsibility for creating a handler addressable Id to the caller makes an API harder to use.
+Shifting the responsibility for creating a handler addressable ID to the caller makes an API harder to use.
 
-If we can map arbitrary client Ids to handler addressable resources, the client generated Id approach is easiest to use
-from stateless clients. The SDK can generate deterministic Ids for calls started from workflows.
+If we can map arbitrary client IDs to handler addressable resources, the client generated ID approach is easiest to use
+from stateless clients. The SDK can generate deterministic IDs for calls started from workflows.
 
-Handler generated Ids is the least friendly approach for stateless clients without additional APIs for mapping input to
+Handler generated IDs is the least friendly approach for stateless clients without additional APIs for mapping input to
 a handler addressable resources.
 
 #### Data sensitivity
 
-If a client is expected to provide a business meaningful Id that encodes sensitive information to address an operation,
-the Id would need to be encrypted making it harder to use.
+If a client is expected to provide a business meaningful ID that encodes sensitive information to address an operation,
+the ID would need to be encrypted making it harder to use.
 
-Both handler generated Ids and caller generated Ids with handler mapping options avoid the data sensitivity issue.
+Both handler generated IDs and caller generated IDs with handler mapping options avoid the data sensitivity issue.
 
 ## Proposal
 
-- Require client generated Ids for any call that starts an asynchronous operation.
-- API will be based on the rough draft in [Caller generated identifiers](#caller-generated-identifiers).
-- Temporal exposes a strongly consistent mapping of caller to handler identifiers that handlers may choose to leverage.
+### API draft
+
+```
+Start(service, operation, (optional) operation_id, input, ...) -> operation_id | result
+Cancel(service, operation, operation_id)
+GetResult(service, operation, operation_id)
+Describe(service, operation, operation_id)
+...
+```
+
+### Notes
+
+- Caller identifiers are optional. If not provided with an ID, the handler must generate one. Handlers may reject (with
+  an error) caller generated IDs for certain operations (e.g. low latency use cases with complex identifiers or when
+  multiple calls are backed by a single operation).
+- Operations are addressed by service, operation name, operation ID, and caller information (e.g. tenant ID).
+- Temporal _may_ at some point expose a strongly consistent mapping of caller to handler identifiers that handlers may
+  choose to leverage, this will be required to alleviate the caller burden of generating addressable IDs for workflow
+  updates.
+  We may start by requiring wrapping update requests with another workflow execution, which achieves the same effect.
+  Note that there are still open questions around using workflow updates over Nexus (e.g. guarateed callback delivery in
+  case of workflow termination).
+- When starting operations from a Temporal workflow, there's no need to provide an ID, Temporal will persist the handler
+  generated ID on the workflow's behalf. If an ID is provided from a workflow, the handler can use it for deduping
+  purposes.
+- `GetResult` may be used to long poll for the result or eagerly return if not ready yet.
+- `Describe` and `GetResult` may be merged into a single request.
+- `DeliverResult`, which was mentioned above, may be added in the full API proposal.
--- a/nexus/operation-ids.md
+++ b/nexus/operation-ids.md
@@ -67,13 +67,13 @@ Example invocations:
 
 ```js
 Start(
-  "payments",
-  "charge",
+  'payments',
+  'charge',
   transactionId,
-  { customerId: "someCustomerId", amount: 5000 },
-  "nexus://caller.example.com/callback?token=" + token
+  { customerId: 'someCustomerId', amount: 5000 },
+  'nexus://caller.example.com/callback?token=' + token
 );
-Cancel("payments", "charge", transactionId);
+Cancel('payments', 'charge', transactionId);
 ```
 
 #### Handler addressability
@@ -92,13 +92,13 @@ The handler may deal with this in a couple ways:
 
    ```js
    Start(
-     "payments",
-     "charge",
-     customerId + "/" + transactionId,
+     'payments',
+     'charge',
+     customerId + '/' + transactionId,
      { amount: 5000 },
-     "nexus://caller.example.com/callback?token=" + token
+     'nexus://caller.example.com/callback?token=' + token
    );
-   Cancel("payments", "charge", customerId + "/" + transactionId);
+   Cancel('payments', 'charge', customerId + '/' + transactionId);
    ```
 
    > NOTE: This looks a lot like a nested REST style URL, e.g: `/payments/customers/{customerId}/charges/{transactionId}`.
@@ -186,7 +186,7 @@ Both handler generated IDs and caller generated IDs with handler mapping options
 ### API draft
 
 ```
-Start(service, operation, (optional) operation_id, input, ...) -> operation_id | result
+Start(service, operation, input, request_id, ...) -> operation_id | result
 Cancel(service, operation, operation_id)
 GetResult(service, operation, operation_id)
 Describe(service, operation, operation_id)
@@ -195,19 +195,15 @@ Describe(service, operation, operation_id)
 
 ### Notes
 
-- Caller identifiers are optional. If not provided with an ID, the handler must generate one. Handlers may reject (with
-  an error) caller generated IDs for certain operations (e.g. low latency use cases with complex identifiers or when
-  multiple calls are backed by a single operation).
-- Operations are addressed by service, operation name, operation ID, and caller information (e.g. tenant ID).
-- Temporal _may_ at some point expose a strongly consistent mapping of caller to handler identifiers that handlers may
-  choose to leverage, this will be required to alleviate the caller burden of generating addressable IDs for workflow
-  updates.
-  We may start by requiring wrapping update requests with another workflow execution, which achieves the same effect.
-  Note that there are still open questions around using workflow updates over Nexus (e.g. guarateed callback delivery in
-  case of workflow termination).
-- When starting operations from a Temporal workflow, there's no need to provide an ID, Temporal will persist the handler
-  generated ID on the workflow's behalf. If an ID is provided from a workflow, the handler can use it for deduping
-  purposes.
+- Handler generates operation IDs.
+- To cancel or get the status or result of an operation, a caller must provide the operation name, and handler provided
+  operation ID.
+  - Callers must be store operation identifiers in order to use these APIs
+  - We may later add the ability for callers to provide operation IDs to remove the requirement for caller state.
+  - This may require some form of seconary index when mapping mulitple client IDs to the same handler operation.
+    Alternatively, we may expose a handler API to map caller input to an operation ID, as mentioned in the [Handler
+    generated identifiers](#handler-generated-identifiers) above.
+- Callers may provide a request ID used to dedupe the start request.
 - `GetResult` may be used to long poll for the result or eagerly return if not ready yet.
-- `Describe` and `GetResult` may be merged into a single request.
-- `DeliverResult`, which was mentioned above, may be added in the full API proposal.
+- The handler _should_ either sign or qualify caller-provided operation IDs when dealing with untrusted sources and
+  multi-tenant applications.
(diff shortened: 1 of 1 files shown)


nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+    - [Caller friendliness](#caller-friendliness)
+    - [Data sensitivity](#data-sensitivity)
+- [Proposal](#proposal)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Nexus RPCs - short recap
+
+At the fundamental level, the Nexus protocol is a synchronous RPC protocol. Asynchronous operations are modelled on top
+of a set of pre-defined RPCs.
```
cretz, line 23: To confirm, "operations" in this proposal refers to what we have called ALOs?

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+
+The handler _should_ qualify an identifier when dealing with untrusted sources and multi-tenant applications.
+
+The following set of RPCs would be needed to model asynchronous operations using client generated Ids:
+
+- `Start(operation, id, input, callbackURL)` - handler responds inline or indicates that an async operation was started
+- `Cancel(operation, id)` - cancel operation by Id
+- `GetResult(operation, id)` - get operation result by Id
+- `Describe(operation, id)` - get operation status by Id
+- `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
```
cretz, line 47: Not sure this one is a Nexus RPC, but that can be discussed elsewhere

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+With this approach, the caller provides an Id to each of the pre-defined RPCs.
+
+The handler _should_ qualify an identifier when dealing with untrusted sources and multi-tenant applications.
+
+The following set of RPCs would be needed to model asynchronous operations using client generated Ids:
+
+- `Start(operation, id, input, callbackURL)` - handler responds inline or indicates that an async operation was started
+- `Cancel(operation, id)` - cancel operation by Id
+- `GetResult(operation, id)` - get operation result by Id
+- `Describe(operation, id)` - get operation status by Id
```
cretz, line 46: Debatable on whether `operation` should be required in these calls

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+);
+Cancel('charge', transactionId);
+```
+
+#### Handler addressability
+
+In some cases, the handler might require more than just a single identifier for a given operation, e.g. when an
+operation is backed by a workflow update, which requires both a workflow Id and an update Id. In order to avoid
+maintaining a handler side persistent mapping between caller and handler identifiers, the handler may choose to "leak"
+this detail to the caller, which would need to concatinate multiple identifiers for a given call. The example above
```
cretz, line 66:
```suggestion
this detail to the caller, which would need to concatenate multiple identifiers for a given call. The example above
```

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+-- OR --
+
+- `Cancel(operation, input)` - cancel operation by handler derived Id
+- `GetResult(operation, input)` - get operation result by handler derived Id
+- `Describe(operation, input)` - get operation status by handler derived Id
+
+NOTE: This looks a lot like the APIs for the "client generated Ids" approach, except the ID is structured instead of a
+single string.
+
+#### Cancel before started
```
cretz, line 128: I don't think this should be supported in Nexus. I think an operation does not exist until `Start` returns and _no_ calls can be made on an operation that does not exist yet. If the Temporal workflow side of Nexus wants to implement cancel-before-start by never sending the start, so be it. But a cancel after start should always wait on start to return before issuing the cancel.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+input fields are used to derive the Id, it would need to pass in the entire Start call input.
+
+This is especially important when we want to automate cancellations of workflow initiated operations, Temporal server
+cannot derive the Id from opaque input.
+
+### Comparing the approaches
+
+#### Handler Id qualification
+
+With handler generated Ids, to support stateless clients, the handler would need to be able to map an operation and
```
cretz, line 141: This is for client generated IDs correct? In the client generated ID section, you have:

> the handler would need a mapping of client identifiers to handler identifiers. Temporal can provide a strongly consistent store out of the box for this mapping.

For handler generated IDs, the handler can be stateless and just ask the client to provide the same ID every time.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+If a client is expected to provide a business meaningful Id that encodes sensitive information to address an operation,
+the Id would need to be encrypted making it harder to use.
+
+Both handler generated Ids and caller generated Ids with handler mapping options avoid the data sensitivity issue.
+
+## Proposal
+
+- Require client generated Ids for any call that starts an asynchronous operation.
+- API will be based on the rough draft in [Caller generated identifiers](#caller-generated-identifiers).
+- Temporal exposes a strongly consistent mapping of caller to handler identifiers that handlers may choose to leverage.
```
cretz, line 172: I was a bit torn here. Usually I would say allowing the clients to identify their operation they want to interact with statelessly is a good practice. It's the nice thing to do to let clients be stateless. This is what Temporal does with workflows and such. But looking around at https://google.aip.dev/151 and general I can understand that this is not resource identification, this is operation identification. If a handler wants to support a "create resource with ID" and then a "affect resource with ID" then that makes sense, but does that need to apply to operation? In traditional programming languages, one might expect a promise or future to be returned from an async op. The caller doesn't create the promise/future, rather the handler does and the caller is expected to use that promise/future.

I would like a hybrid approach of optional client identifier but handler generated if not present (i.e. how an auto generated sequence PK might be in a DB upon insert, but you can provide it) allowing the handler to remain stateless. But I can understand it's difficult to require clients to provide a _globally unique_ identifier while also letting them use a meaningful one.

So I could be convinced of either direction, but I slightly lean to the handler generated identifiers.

But maybe there's a hybrid we can take. Maybe there is some way we can uniquely identify clients same way as we do namespaces with workflows today? Then we can have client generated identifiers and tell handlers that we guarantee "client name + operation identifier are globally unique". Then both the client and handler can be stateless and the true identifier is a tuple of client and identifier.

> Temporal exposes a strongly consistent mapping of caller to handler identifiers that handlers may choose to leverage.

If we're really gonna support this, it needs to be more general purpose operation metadata the handler can set on start response and is delivered to the handler on each other invocation.

nexus/operation-ids.md
Sushisource: +1

nexus/operation-ids.md
Sushisource: > hybrid approach of optional client identifier but handler generated if not present ... allowing the handler to remain stateless. 

Does that actually allow the handler to remain stateless in the not-provided case? Regardless, I like that flexibility too.

Generally though, I agree with Chad.

>it's difficult to require clients to provide a globally unique identifier while also letting them use a meaningful one.

Is definitely true. Either you require some format like GUID, and rob clients of the opportunity to make the ID meaningful, or you don't, and have to spend cycles checking it's unique which could've been used to just return them an ID.

The analogy to futures makes sense to me as well

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+- `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
+
+The handler _should_ either sign or qualify caller-provided reference Ids when dealing with untrusted sources and
+multi-tenant applications.
+
+In order to support stateless clients, the handler may provide an additional set of APIs to resolve client input to a
+reference Id. This may be something that is built into the protocol or implemented ad-hoc. To support this at the
+protocol level, we would need to add the following APIs:
+
+- `Resolve(operation, input) -> reference_id`
```
Sushisource, line 117: I don't see how this can work generally. There's no guarantee that op/input is a unique combo. If it was we could just be hashing that and using that as the ID.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+In some cases, the handler might require more than just a single identifier for a given operation, e.g. when an
+operation is backed by a workflow update, which requires both a workflow Id and an update Id. In order to avoid
+maintaining a handler side persistent mapping between caller and handler identifiers, the handler may choose to "leak"
+this detail to the caller, which would need to concatinate multiple identifiers for a given call. The example above
+becomes:
+
+```js
+Start(
+  'charge',
+  customerId + '/' + transactionId,
```
Sushisource, line 72: I don't really follow how this is distinguished from the previous example. Customer ID moves from being a field in the input, to part of this ID - and I guess the point there is so it can be passed to things like `Cancel` which don't accept input.

IMO if we did end up doing something along these lines, I think you'd probably want the `path` argument to become a properly defined type rather than doing some concatenation. That way operations can explicitly define one-or-more typed IDs they care about.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+
+Since an asynchronous operation is modelled on top of synchronous RPCs, it must be addressable (identifiable) by both
+the caller and the handler. There are two approaches for generating identifiers, caller side and handler side.
+
+### Caller generated identifiers
+
+With this approach, the caller provides an Id to each of the pre-defined RPCs.
+
+The handler _should_ qualify an identifier when dealing with untrusted sources and multi-tenant applications.
+
```
antlai-temporal, line 40: This is not enough when you are trying to share access with others, e.g.,  give the id to someone else so that they can GetResult(id) safely. In those cases you want the identifiers to be **unguessable**, so that they can behave as an object capability. The client is hopeless for that, e.g., it may not have enough entropy in the RNG, so if sharing is important we need handler ids...

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+  'nexus://caller.example.com/callback?token=' + token
+);
+Cancel('charge', transactionId);
+```
+
+#### Handler addressability
+
+In some cases, the handler might require more than just a single identifier for a given operation, e.g. when an
+operation is backed by a workflow update, which requires both a workflow Id and an update Id. In order to avoid
+maintaining a handler side persistent mapping between caller and handler identifiers, the handler may choose to "leak"
```
antlai-temporal, line 65: I don't see the benefit of this "middle-of-the-ground". What the server returns should be opaque to the client, so that we can easily change things without breaking clients, or support different backend implementations.  Leaking implementation details is a bad idea. Forcing the client to do extra work is even worse. And you have to maintain the mappings in the handler anyway...

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+
+Looking up an identifier in persistent storage added latency, and should be opted in to.
+
+### Handler generated identifiers
+
+With this approach, the handler returns a reference Id in response to a start call, which the caller should persist for
+use with the pre-defined RPCs as shown below:
+
+- `Start(operation, input, callbackURL, cancelBeforeStarted)` - handler responds inline or returns a reference Id
+  ([`*`](#cancel-before-started))
```
antlai-temporal, line 104: We need to add a client generated transaction id to `Start(transaction_id,...)` to allow "exactly-once" semantics for the operations. This transaction id may, or may not, ended up as part of the handler id, this is implementation dependent and not visible to the client (but may simplify the handler implementation...). The other calls do not need the transactional ID because they are naturally idempotent, or side effect free...

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+- `Cancel(reference_id)` - cancel operation by handler provided reference Id
+- `GetResult(reference_id)` - get operation result by handler provided reference Id
+- `Describe(reference_id)` - get operation status by handler provided reference Id
+- `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
+
+The handler _should_ either sign or qualify caller-provided reference Ids when dealing with untrusted sources and
+multi-tenant applications.
+
+In order to support stateless clients, the handler may provide an additional set of APIs to resolve client input to a
+reference Id. This may be something that is built into the protocol or implemented ad-hoc. To support this at the
```
antlai-temporal, line 114: Stateless clients will need to create aliases to handler ids and "the service" should help them to keep these mappings persistent, e.g., a persistent BOP (Box of Promises). When we share handlers ids, different clients may have their own (different) aliases. And a single client may want to scope these aliases in an ad-hoc manner, e.g., managing multiple similar workflows with the same aliases. So "the service" may ended up being a different service from this one...

nexus/operation-ids.md
antlai-temporal: +1

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+This is especially important when we want to automate cancellations of workflow initiated operations, Temporal server
+cannot derive the Id from opaque input.
+
+### Comparing the approaches
+
+#### Handler Id qualification
+
+With handler generated Ids, to support stateless clients, the handler would need to be able to map an operation and
+input to a reference Id. It would also need to qualify a reference Id or securely issue the reference Ids when callers
+cannot be trusted. With caller generated Ids, the handler would only need to qualify an Id based on the caller context
```
antlai-temporal, line 143: The handler ids are opaque for the client, as long as they are unique and unguessable, how this is done is handler implementation dependent. There are several ways to do that, but this should not be part of the spec...

nexus/operation-ids.md
cretz: > Does that actually allow the handler to remain stateless in the not-provided case? Regardless, I like that flexibility too.

Not if they need to translate to some internal ID. But if you take that namespace-ish identifier + client-provided operation identifier concatenated to be your workflow ID in the handler, then you can just reuse that for each operation on that workflow. But otherwise if there is a stateful translation needed, then yes some kind of meta store is needed.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+With the caller generated Id approach, if the Id is used for handler side addressability, the handler would need to
+parse it and merge with the request's input.
+
+#### Caller friendliness
+
+Shifting the responsibility for creating a handler addressable Id to the caller makes an API harder to use.
+
+If we can map arbitrary client Ids to handler addressable resources, the client generated Id approach is easiest to use
+from stateless clients. The SDK can generate deterministic Ids for calls started from workflows.
+
```
antlai-temporal, line 157: This breaks down when you start to share the identifiers, so that others can GetResult(). The structure that makes sense for your program, could be very different to whoever created (and named) the identifier...

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+If a client is expected to provide a business meaningful Id that encodes sensitive information to address an operation,
+the Id would need to be encrypted making it harder to use.
+
+Both handler generated Ids and caller generated Ids with handler mapping options avoid the data sensitivity issue.
+
+## Proposal
+
+- Require client generated Ids for any call that starts an asynchronous operation.
+- API will be based on the rough draft in [Caller generated identifiers](#caller-generated-identifiers).
```
antlai-temporal, line 171: I think this is a bad idea. It will make safe sharing much more difficult because the ids are not unguessable. And, in most cases, they cannot be completely client generated unless the handler leaks abstractions, or maintains bindings. 
I'm all for handler ids but  we need to help stateless clients maintain persistent aliases with "a service" (that may or not be this service).
However, we do need a client-created transaction id in Start() for "exactly-once", but that is not needed in other calls, and may not be part of the opaque identifier...

nexus/operation-ids.md
cretz: > In those cases you want the identifiers to be unguessable

I am not sure we want that though. Often letting a user choose an identifier that is meaningful to them so they can reuse that is ideal.

nexus/operation-ids.md
cretz: > Forcing the client to do extra work is even worse

Forcing the client to store an identifier that can only be generated server side is cording the client to do extra work

nexus/operation-ids.md
cretz: Why can't IDs be guessable? We allow it with workflow IDs today and in general letting a record creator have the ability to set their own valuable-to-them primary key has value.

nexus/operation-ids.md
antlai-temporal: Unguessable helps with fine-grained access control. If they are unguessable you can use them as a capability, so if you have it, you also have the right to see the result, and that simplifies safe sharing.

But you are right, you cannot expect clients to directly use these ids, they are not user-friendly, you need to provide local, persistent aliases to access them, so that they can choose names meaningful to them.

nexus/operation-ids.md
antlai-temporal: Point taken. But I think we can help stateless clients to persist and manage these aliases with the right service, and if we get the right client abstraction, it won't be too painful...

nexus/operation-ids.md
mfateev: I think it is up to the implementer of the operation to map the operation arguments to the id. For example, it can return the same id for two separate calls if their arguments point to the same implementation workflow.

nexus/operation-ids.md
antlai-temporal: I think the use case is a bit different. Sharing workflows IDs with others for collaboration is not that common today (I guess), with security an obvious problem, but when you go "higher-up-the-stack", people will start interacting with temporal in new ways, if we get the security right...
And I do agree with you, we need creators and consumers to make their own aliases to make this usable...

nexus/operation-ids.md
bergundy: Agree, I had similar thoughts but decided to include it so it to help the handler construct an addressable workflow ID.

In the `Start` request below, the backing workflow ID would be `charge-{tenantId}-{transactionId}`.

nexus/operation-ids.md
bergundy: I think we're going to want this for using the handler as a response interceptor, let's discuss when we get to defining the full API.

nexus/operation-ids.md
bergundy: That's what the persistent mapping is for, but the handler would decide when to use it, not the caller.

nexus/operation-ids.md
bergundy: That's definitely something that I've considered but decided to keep it flat for simplicity (modeled after URL paths).
I'm not ruling it out though, agree that forcing the handler and client to encode and decode input to/from strings is suboptimal.

nexus/operation-ids.md
bergundy: Transaction ID would be part of the input here. The handler would need to interpret it as an idempotency token.

nexus/operation-ids.md
bergundy: This would assume that the input includes some form of unique identifier, agree that it is not always the case.

nexus/operation-ids.md
bergundy: You're forcing the caller to issue an additional call and delaying the delivery of the cancellation. We can discuss how this would be implemented with workflows.

nexus/operation-ids.md
bergundy: This is for handler generated IDs to make `Start(operation, input)` idempotent. (See also the comment about `Resolve(operation, input) -> reference_id`).

nexus/operation-ids.md
bergundy: It would need to include one to make the start call idempotent.

nexus/operation-ids.md
bergundy: I will add something to the proposal about sharing identifiers between different consumers (multi-tenant apps).

bergundy (commented):
I had to go before I had the chance to respond to all of the comments.
Thank you everyone for the feedback.

nexus/operation-ids.md
cretz: That a handler needs to build some addressable Temporal thing is not a general Nexus concern IMO. Caller should only have to provide exactly what is needed, the identifier, to interact with uniquely identified thing.

nexus/operation-ids.md
cretz: I just want to make sure from a Nexus perspective, Nexus servers do not have to accept cancellation from operations that have not returned from start.

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+
+> To take this approach to the extreme, `operation` and `id` could be combined into a single `path` argument, as shown
+> below:
+>
+> - `Start(path, input, callbackURL)`
+> - `Cancel(path)`
+> - `GetResult(path)`
+> - `Describe(path)`
+
+##### Mapping from client to handler identifiers
```
cretz, line 89: Can we document the option to have some unique "namespace"-like value for a client? Then we can tell handlers that this "namespace" value + client-provided unique identifier is globally unique?

nexus/operation-ids.md
bergundy: I've been thinking about this some more and a few discussions resonated with me:

1. If a caller identifier is constrained to a single string, and is used to address a resource on the handler side, the caller may end up concatenating multiple identifiers to generate a single string, which the handler then has to parse. This is suboptimal experience IMHO.
2. Max's comment that with caller generated IDs, separate calls cannot map to the same workflow implementation.
3. Chad's comment (which has been mentioned in previous discussions), and the reference to Google's longrunning spec, where the caller doesn't "create" a promise, it can only get a reference to one from calling a method.
4. As I mentioned in this proposal, with caller generated IDs, extra care should be taken to encrypt any sensitive information that make up the ID.

The proposal to use client generated IDs came to me from writing this part of the spec:

> - `Cancel(operation, input)` - cancel operation by handler derived Id
> - `GetResult(operation, input)` - get operation result by handler derived Id
> - `Describe(operation, input)` - get operation status by handler derived Id
>
> NOTE: This looks a lot like the APIs for the "client generated IDs" approach, except the ID is structured instead of a single string.

Writing this made me think that we might as well go with caller generated IDs for stateless client support as it would be easier to implement the handler. All you would need is to map a caller context+ID to an internal resource ID, no need to sign or qualify reference IDs.
But this approach is less flexible. For full flexibility, I'm starting to think we'll want to expose methods for addressing an operation by both a handler provided reference ID and by an operation and structured input.

I'm now leaning more toward the API I mentioned in the handler generated IDs section:

```
Start(operation, input, ...)
CancelByID(reference_id)
GetResultByID(reference_id)
DescribeByID(reference_id)
// Option 1:
Cancel(operation, partial_input)
GetResult(operation, partial_input)
Describe(operation, partial_input)
// Option 2:
GetReferenceID(operation, partial_input)
```

A few things to note here:

1. To get a reference to an operation, the caller only needs to send the input fields that are needed to resolve the reference (very much like the caller generated IDs proposal, except this API doesn't require encoding input as string). In practice this input will often times be a single, caller unique, identifier.
2. The handler doesn't **have** to support resolving a reference from operation + input for all operation types, some APIs may not need to be idempotent (e.g. they may be read only). The handler will necessarily have the ability to map operation + input to an internal identifier (workflow ID) for any idempotent operation.
3. `Cancel` is equivalent to `GetReferenceID` + `CancelByID` but it requires one less round trip, I prefer that.
4. We can choose not to include either option 1 or option 2 in the API, and tell handler implementors to expose this functionality per use case with custom operations. For example a handler implementing a `charge` operation, would also implement `cancelCharge`, `describeCharge`, etc... but I would argue that supporting stateless callers is a goal for Nexus and I would vote to build this functionality into the protocol. 
5. Names aren't final, I'll finalize them when I do the API proposal.

I'll wait for a round of responses here and update the proposal based on the feedback.

nexus/operation-ids.md
cretz: Would rather handle that as a Temporal thing in the server than say it's optional in Nexus. I assume the server has similar logic for child workflows and activity workflows when it receives a cancel for a not-started/asynchronously-waiting-to-start thing.

nexus/operation-ids.md
cretz: :+1: I forgot to mention that client-provided unique identifier should be optional

nexus/operation-ids.md
bergundy: > Accept optional caller identifiers and namespace the caller and tell caller that their identifier must be unique for their namespace

Would those IDs be used to reference async operations in the cancel, describe, and get result APIs? Do they replace handler provided reference IDs? If that's your intention, the handler will likely need a persistent mapping from client ID to some addressable resource because it doesn't know which operation an ID is referring to and if it's backed by a workflow/activity/update. With the options I provided, the handler can statelessly map client input to a reference ID.

> GetReferenceID should not be part of Nexus

I think we want some stateless method of mapping client input to a handler resource.

cretz: > Would those IDs be used to reference async operations in the cancel, describe, and get result APIs?

Yes. Same as workflow IDs.

> Do they replace handler provided reference IDs?

They are the same. The handler should only provide an ID if it is not given one.

> If that's your intention, the handler will likely need a persistent mapping from client ID to some addressable resource because it doesn't know which operation an ID is referring to and if it's backed by a workflow/activity/update.

Depends on the situation. Maybe. But if I were a handler, I'd use client's namespace + client-provided ID tuple as the workflow ID and not use a mapping. If it wasn't for workflow ID but instead for, say update on a workflow, my handler is smart enough to know what each thing is for (maybe I prepend the word "update" or something).

I think I understand your issue now. Yes, I would say that every call on an "operation" is qualified by its service name and operation. You don't need the separate by ID. The way I read it earlier, when I saw `Cancel(operation, id)` I thought you were burdening the user with operation, but a lot of the time a user will use a handle which will have that information.

So now that I have my head right, I think of "operations" as REST resources. Just having the resource identifier is not enough, you need the resource name too. I think we should call it "operation name". So technically it's more like:

* `Start(namespace, service name, operation name, (optional) ID, input, ...)`
* `Cancel(namespace, service name, operation name, ID)`
* `GetResult(namespace, service name, operation name, ID)` + `Describe(namespace, service name, operation name, ID)`
  * I think these are probably the same thing, but whatever is decided, I shouldn't have to make two calls to "get status or result"

Sorry for my confusion. I am on team client-provided-optional-identifier + operation name on every call.

rylandg:
I generally love how this has turned out. My one question is whether we've considered supporting a client side dict key type, where uniqueness of the keypairs is required but ordering isn't. It requires processing but also is very easy to reason about and removes the nastiness with the current concatenation.

nexus/operation-ids.md
bergundy: > Would rather handle that as a Temporal thing in the server than say it's optional in Nexus

How would Temporal be able to achieve this without it being a part of the Nexus protocol?

> I assume the server has similar logic

Not sure I understand what you mean here.

The server (history) would be the one issuing these cancel-maybe-started requests, if the handler is backed by Temporal, it would be trivial to implement for most operations.

nexus/operation-ids.md
bergundy: > `Start(namespace, service name, operation name, (optional) ID, input, ...)`

What's `namespace` here? There's no concept of namespaces in Nexus.

I would rather avoid requiring the client to concatenate multiple identifiers into a single string (and the handler parse and extract those identifiers) when an operation is backed by a resource that is addressed using a complex identifier.
The last API I suggested in this thread avoids this issue.

The only reason I'd want the handler to return a reference id is so Temporal (or any other platform) performing calls on behalf of the user wouldn't need to understand what part of the input is an identifier.

The other alternative might be something like @rylandg mentioned, where part of the input is provided in a dedicated structured identifier field. But I think we want client identifiers to be optional and don't want to burden every client (especially stateful ones, e.g. workflows) to care about what the handler considers an identifier. This approach would also make coding the handlers annoying, as they would have to extract the input from multiple fields.

bergundy: If the identifier is optional, how does temporal call the cancel and describe apis on behalf of the workflow?

cretz: Because the identifier is always returned, it's only optional for the client to provide. The handler is expected to make the identifier if the client doesn't provide it. The handler must always return an identifier for an operation it starts. Sorry I wasn't clear. Identifier is only optional from a client POV.

antlai-temporal: So what happens if the client does not provide an id, and crashes after performing the operation but before seeing the reply. Do we have an "orphan" operation that nobody can cancel or describe?

bergundy: @cretz I think we (almost) want the same thing then.
I just prefer to not have the client id encoded in the protocol and no string concatenation and parsing, that’s why I’m suggesting that we substitute the string id with structured input.

bergundy: I also have the http use case in mind (see the proposal and the reference to http paths), urls can be mapped to structured input fairly easily. I would want an http gateway to be responsible for this mapping

cretz: >  I just prefer to not have the client id encoded in the protocol and no string concatenation and parsing, that’s why I’m suggesting that we substitute the string id with structured input.

I completely agree wrt structured input. I think you need 4 structured things to uniquely identify an operation: some kind of client/caller-level namespacing, service name, operation name, operation ID (but this can be optional on client start, but required on handler return and all other operation identification uses).

bergundy: I'll clarify my comment above so we'll see how far off we are.
In the **start** call, the caller sends arbitrary input to the handler, which may contain identifiers, but there is no string ID field on that call.

Here's a more condensed version of my proposal above:

```
Start(service, operation, input, ...) -> reference_id
Cancel(service, reference_id | structured_id)
GetResult(service, reference_id | structured_id)
Describe(service, reference_id | structured_id)
```

The `structured_id` field can take any payload-like form. Here are some examples values:
```
"some-string-identifier"
{ "customerId": "...", "transactionId": "..." }
```

The start call may optionally include a structured identifier as part of the input, such as:
```
Start("some-service", "some-op", { "id": "some-string-identifier", "arg1": 3 })
Start("payments", "charge", { "customerId": "foo", "transactionId": "bar", "amount": 50 })
```

If the caller was responsible for generating complex string identifiers the call above would look something like this:
```
Start("payments", "charge", "foo/bar", { "amount": 50 })
```

When we expose HTTP gateways on top of Nexus, this complex identifier is likely going to be part of the path in the form of: `/payments/charges/customers/foo/charges/bar`, but I don't this string mess in the basic Nexus protocol.

Again, IDs are usually simple values, but there are cases where they are complex and require multiple fields to address a specific resource on the handler side. A stateless client would still be required to provide all of these fields so a handler can address a specific resource but there shouldn't need to be any string encoding or parsing.
Stateful clients (e.g. workflows), do not have to deal with identifiers at all, Temporal takes care of that under the hood, since it can use the handler returned `reference_id`.

bergundy: I'll just add that if the handler wants to hide complex addressability details from stateless clients, it can use a persistent mapping of client ID to internal resource, allowing the client to use simple identifiers in the `Cancel`, `Describe`, and `GetResult` APIs.

I have changed my mind since originally writing this proposal, and I don't think Temporal should provide this persistent mapping capability, at least not initially.

cretz: I think this is overcomplicating it. We don't need some arbitrary identifier that can be any JSON object of any size. I think there should be one ID to reference an operation and that's it. Very simple. Here's how I would represent your API (assuming namespace was implicit somewhere):

```
Start(service, operation, (optional) operation_id, input, ...) -> operation_id
Cancel(service, operation, operation_id)
GetResult(service, operation, operation_id)
Describe(service, operation, operation_id)
```

Start without ID:

```
Start(service="my-service", operation="my-operation", input={"some-key": "some-val"})
```

Start with ID:

```
Start(service="my-service", operation="my-operation", operation_id="my-operation-id", input={"some-key": "some-val"})
```

That simple. We really shouldn't make this some complex thing. Also, this translates to HTTP well.

> When we expose HTTP gateways on top of Nexus, this complex identifier

Just get rid of the notion of a complex identifier. Get rid of the idea of multiple ways to refer to an operation. Keep it simple.

> handler wants to hide complex addressability

Just get rid of the notion of a complex addressability. Get rid of the idea of multiple ways to refer to an operation. Keep it simple.

> I don't think Temporal should provide this persistent mapping capability, at least not initially.

I'm kinda with you, but I think there should be metadata. Every handler call for an operation should be able to return metadata (i.e. memo) outbound and get it back inbound. They could use this for mapping if they wanted.

cretz: Same as it would for any other operation (if they wanted to provide an ID). Why does the client need to know it's a workflow update? An operation is an operation to be identified.

Now, if you're asking "how does the handler know what workflow the update needs to go to", that is an operation-specific piece so that is operation-specific parameters. If I were a handler that needed to write an user-facing operation that interacted with another user-facing operation, I would ask the user to provide me as part of input the identifier of that other operation to be interacted with. But if I were a handler that needed to write a user-facing operation that interacted with an internal workflow, then I would require the user as part of input to give whatever was needed to address that workflow.

An operation identifier is a simple concept. Operations that needs to address _other things_ require that the user provide what is needed to address those other things as parameters to my operation. But it sure isn't part of the Nexus protocol.

bergundy: I buy that ids are simple for the client side, and TBH what you're proposing is almost identical to my last option except that you restrict an ID to be a single string which is separate from the rest of the input.

If we go with:
```
Start(service, operation, (optional) operation_id, input, ...) -> operation_id
Cancel(service, operation, operation_id)
GetResult(service, operation, operation_id)
Describe(service, operation, operation_id)
```

There's a few things to note:
1. There's now a single concept of an operation ID, there's no structured way to differentiate between caller and handler generated IDs, requiring the handler to qualify all IDs as if they are client generated.
2. With a handler generated ID, the `Cancel`, `Describe` and `GetResult` calls do not need an operation name in addition to the ID to qualify an operation. This has some benefits.
3. We may want to provide the metadata store for operations with Temporal to avoid burdening callers with handler resource addressable identifiers, this could come at a later stage.
4. Caller/handler may agree on some encoded ID format so callers can directly reference handler resources to avoid the overhead of persistent metadata lookup per operation. We may want to document this.

Overall, I think this API works well.

nexus/operation-ids.md
lorensr: Why not backing WF ID: `{tenantId}-{transactionId}`? Are operation IDs unique per tenant or per [operation, tenant]?

From Caller DX POV, +1 to only requiring ID, even if that means Temporal handler needs to remember mapping or we need to impose tenant-wide uniqueness constraint.

nexus/operation-ids.md
lorensr, line 47:
```suggestion
- `DeliverResult(token, result)` - deliver a result of an async operation using a token in the Start request's callback URL
```

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+- `DeliverResult(token)` - deliver a result of an async operation using a token in the Start request's callback URL
+
+Example invocations:
+
+```js
+Start(
+  'charge',
+  transactionId,
+  { customerId: 'someCustomerId', amount: 5000 },
+  'nexus://caller.example.com/callback?token=' + token
```
lorensr, line 56: curious what the benefit of the token is here, vs the below, which simplifies the API by not having a token:

- provide `nexus://caller.example.com/callback`
- call `DeliverResult(transactionId, result)`
- callback is made to `nexus://caller.example.com/callback` with transactionId in the query string and/or body

nexus/operation-ids.md
```diff
@@ -0,0 +1,172 @@
...
+> - `Cancel(path)`
+> - `GetResult(path)`
+> - `Describe(path)`
+
+##### Mapping from client to handler identifiers
+
+To avoid leaking the identifier construction logic to the client, the handler would need a mapping of client identifiers
+to handler identifiers. Temporal can provide a strongly consistent store out of the box for this mapping.
+
+A single execution would need to map to multiple client identifiers to account for workflow starts and updates.
```
lorensr, line 94:
```suggestion
A single workflow execution would need to map to multiple client identifiers to account for workflow starts and updates.
```

to differentiate from operation execution

nexus/operation-ids.md
lorensr: So the caller lib could be something like this, with service provided at client creation and operation/id in a handle?

```ts
const client = await NexusClient.connect({ 
  url: '<url>', 
  service: 'loren-staging' 
})

const handle = await client.start({ 
  operation: 'charge',  
  operationId: transactionId,
  input: { customerId: 'someCustomerId', amount: 5000 },
  callback: 'nexus://caller.example.com/callback?token=' + token
})

const description = await handle.describe()
await handle.cancel()  
const result = await handle.getResult()  
```

Benefit of not requiring `operationId` is that caller doesn't have to do the work of creating it? If caller doesn't create it, they should have an idempotency key in the input? Is idempotency key much easier to create than an operationId? If yes, then the only benefit is they have the ease of providing neither operationId nor idempotency key (and thus risking multiple operations, which may be a fine risk for their use case)?

nexus/operation-ids.md
bergundy: This is actually operation execution, I'll clarify

nexus/operation-ids.md
bergundy: Added a section in the proposal (will push soon)

nexus/operation-ids.md
bergundy: Agree that it's desirable from the DX POV to only require an ID, but as mentioned in one of the threads below, this is not enough to fully identify an operation with caller generated IDs.

nexus/operation-ids.md
bergundy: Token can be used for security in case transaction ID is guessable.
Token is not part of the spec but it will likely be generated in the Temporal implementation.

nexus/operation-ids.md
bergundy: Can I resolve this thread?

nexus/operation-ids.md
bergundy: I'm changing the proposal to accept optional caller generated IDs, those IDs may be random idempotency keys, addressable paths, etc...

I think the addressability concern in caller generated IDs is minor, if we provide a persistent mapping from caller ID to handler operation, it becomes a non issue at the cost of an extra persistence hit.
Handler for low latency operations that require complex identifiers for addressability could choose to disallow caller IDs or require data to be encoded in the ID.

nexus/operation-ids.md
bergundy: I mentioned this above in "Sharing identifiers between clients" (will push in a bit)

nexus/operation-ids.md
lorensr: ah i see. an alternative would be signing the message

antlai-temporal:
It may belong somewhere else, but do we also want to make public the API for fulfiling the result (promise) of the operation, or this is just some internal API for the handler only. I can see it useful for communication between workflows if user code could directly resolve them, but this could also open the security pandora box...

nexus/operation-ids.md
cretz: > Benefit of not requiring operationId is that caller doesn't have to do the work of creating it?

Yes. Many calls they just care about the created operation ID. This becomes a lot clearer when looking at a potential HTTP `POST` impl of this which allows one to leave off the ID of the URL.

> If caller doesn't create it, they should have an idempotency key in the input?

There is no requirement for whole-operation idempotency I don't believe. But if there is an operation for _this request_ idempotency to support client retries, that I think is a separate concern.

bergundy:
> It may belong somewhere else, but do we also want to make public the API for fulfiling the result (promise) of the operation, or this is just some internal API for the handler only. I can see it useful for communication between workflows if user code could directly resolve them, but this could also open the security pandora box...

I mentioned this in the proposal, it'll be fully specced in the next proposal where I'll introduce the entire API.

cretz (approved):
I disagree with not allowing the client to provide the identifier, but I am capitulating. I will just remain jealous that Temporal allows this for workflows.

PR #83, opened 2023-08-08
HTTP API: Clarify which part of the system takes HTTP requests

josh-berry:
There was some internal confusion around whether HTTP requests are handled by a separate process/service or the same process as gRPC.


cretz:
I think we want to avoid retroactively updating the proposal. There are other things on this proposal that are now inaccurate and/or underspecified (i.e. learned based on implementation) and we don't usually see these as actual specs, more of just accepted proposals.

PR #84, opened 2023-09-07
Temporal Go SDK - Nexus Handler

bergundy:
See rendered version [here](https://github.com/bergundy/proposals/blob/nexus-handler/nexus/go-handler.md).

### DIFF
--- a/nexus/go-handler.md
+++ b/nexus/go-handler.md
@@ -251,6 +251,13 @@ func GetMetricsHandler(context.Context) metrics.Handler
 
 All of these methods panic if called outside of an operation handler.
 
+### WorkerOptions
+
+The following worker options will be added to apply limits:
+
+- `MaxConcurrentNexusTaskPollers`
+- `MaxConcurrentNexusExecutionSize`
+
 [temporal-go-sdk]: https://github.com/temporal/sdk-go
 [nexus-go-sdk]: https://github.com/nexus-rpc/sdk-go
 [nexus-http-api]: https://github.com/nexus-rpc/api/blob/main/SPEC.md
--- a/nexus/go-handler.md
+++ b/nexus/go-handler.md
@@ -61,7 +61,7 @@ type OperationRegistry interface {
 package operation // go.temporal.io/sdk/operation
 
 type Handler interface {
-	StartOperation(context.Context, *StartOperationRequest) (StartOperationResponse, error)
+	StartOperation(context.Context, *StartOperationRequest) (OperationResponse, error)
 	CancelOperation(context.Context, *CancelOperationRequest) error
 	// Later we will add GetResult and GetInfo as well
 }
--- a/nexus/go-handler.md
+++ b/nexus/go-handler.md
@@ -253,6 +253,6 @@ All of these methods panic if called outside of an operation handler.
 
 [temporal-go-sdk]: https://github.com/temporal/sdk-go
 [nexus-go-sdk]: https://github.com/nexus-rpc/sdk-go
-[nexus-http-api]: https://github.com/nexus-rpc/api/TODO
+[nexus-http-api]: https://github.com/nexus-rpc/api/blob/main/SPEC.md
 [activity-package]: https://pkg.go.dev/go.temporal.io/sdk@v1.24.0/activity#pkg-functions
 [nexus-start-request]: https://pkg.go.dev/github.com/nexus-rpc/sdk-go@v0.0.0-20230906182323-c6769ebc0c28/nexus#StartOperationRequest
--- a/nexus/go-handler.md
+++ b/nexus/go-handler.md
@@ -1,0 +1,307 @@
+# Temporal Go SDK - Nexus Handler
+
+## Overview
+
+An addition to the [Temporal Go SDK][temporal-go-sdk] that exposes the ability for users to handle Nexus API requests.
+
+**Goals**
+
+- Expose a simple, high level API for binding Nexus operations to common Temporal constructs
+- Support for starting and canceling operations
+- Support multi-tenant applications with minimal boilerplate
+
+**Non-Goals**
+
+- Support for getting an operation's result or info
+- API definitions for other SDKs
+- API for invoking Nexus calls - refer to the [Nexus Go SDK][nexus-go-sdk] for an out-of-workflow reference client
+- Server implementation details
+- Worker / server API contract
+- Allow advanced users to implement Nexus inbound and outbound request proxies for cross-organization calls where
+  data converters are not shared and data is not exposed to Temporal server.
+
+  This will quickly become a requirement for production use cases, especially with Temporal Cloud's zero trust model. It
+  will be addressed shortly after Nexus becomes available for preview.
+
+> NOTE: These are non-goals for _this_ proposal, but they will be defined in follow up proposals / PRs.
+
+## Background
+
+Nexus is a thin protocol on top of HTTP, defined in [this spec][nexus-http-api]. Temporal serves as an HTTP proxy,
+translating Nexus requests into tasks that can be polled on by an SDK worker on its configured task queue.
+Due to the synchronous nature of HTTP, the generated tasks are not persisted, and require an available worker
+to be processed.
+
+The process of translating requests into tasks and delivering them to workers is out of scope for this proposal.
+
+## SDK Additions
+
+The Temporal Go SDK exposes a `Worker` interface that embeds a `Registry` interface defined as:
+
+```go
+type Registry interface {
+	WorkflowRegistry
+	ActivityRegistry
+}
+```
+
+Add and embed a third registry type: `OperationRegistry` with the following method:
+
+```go
+type OperationRegistry interface {
+	RegisterOperation(operation string, handler operation.Handler)
+	// Alternative: get the operation name from handler.Name()
+	// RegisterOperation(handler operation.Handler)
+}
+```
+
+`operation.Handler` is an interface with a minimal definition of:
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type Handler interface {
+	StartOperation(context.Context, *StartOperationRequest) (StartOperationResponse, error)
+	CancelOperation(context.Context, *CancelOperationRequest) error
+	// Later we will add GetResult and GetInfo as well
+}
+
+// Maps almost 1:1 to the Nexus Go SDK StartOperationRequest, exposing HTTP header and body directly instead of the HTTP
+// Request object.
+type StartOperationRequest struct {
+	Operation   string
+	RequestID   string
+	CallbackURL string
+
+	URL    *url.URL
+	Header http.Header
+	Body   []byte
+}
+
+// Maps almost 1:1 to the Nexus Go SDK CancelOperationRequest, exposing HTTP header and body directly instead of the
+// HTTP Request object.
+type CancelOperationRequest struct {
+	Operation   string
+	OperationID string
+
+	URL    *url.URL
+	Header http.Header
+}
+
+type OperationResponse interface {
+	// private methods
+}
+
+// Indicates that an operation completed successfully. Implements OperationResponse.
+type OperationResponseSync struct {
+	// Header to deliver in the HTTP response.
+	Header http.Header
+	// Body conveying the operation result.
+	// If it is an io.Closer, it will be automatically closed by the framework.
+	// Note that we accept a reader instead of bytes here to keep the future possiblity of supporting streaming of large
+	// payloads.
+	Body io.Reader
+}
+
+// Indicates that an operation has been accepted and will complete asynchronously. Implements OperationResponse.
+type OperationResponseAsync struct {
+	OperationID string
+}
+```
+
+To minimize the amount of boilerplate required to map operations to common Temporal constructs, the SDK exposes default
+handler implementations.
+
+### Workflow Run
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type WorkflowStartOptions struct {
+	// Options to start the workflow with. TaskQueue defaults to the current worker's task queue.
+	// The workflow ID used here will be returned as the operation ID by default.
+	client.WorkflowStartOptions
+	// Type of workflow to start.
+	WorkflowType string
+	// Input to provide to the workflow.
+	Input        []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+// Returns a Handler that can maps and operation to a workflow that can be started and canceled.
+// Handles HTTP request to payload deserialization for input of type `T`.
+func NewWorkflowRunHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowStartOptions, error)) Handler
+```
+
+**Basic usage:**
+
+```go
+w.RegisterOperation("provision-cell", NewWorkflowRunHandler(
+	func(ctx context.Context, input Input, request *StartOperationRequest) (*WorkflowStartOptions, error) {
+		return &WorkflowStartOptions{
+			Input:        []any{input},
+			WorkflowType: "provision-cell",
+			StartWorkflowOptions: StartWorkflowOptions{
+				ID: fmt.Sprintf("%s-%s", request.Operation, input.CellID),
+			},
+		}, nil
+	}))
+```
+
+### Workflow Query
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type WorkflowQueryOptions struct {
+	WorkflowID           string
+	// Optionally expose RunID too
+	QueryType            string
+	QueryRejectCondition enums.QueryRejectCondition
+	// Input to provide to the query.
+	Input                []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+func NewWorkflowQueryHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowQueryOptions, error)) Handler
+```
+
+**Basic usage:**
+
+```go
+w.RegisterOperation("get-cell-status", NewWorkflowQueryHandler(
+	func(ctx context.Context, input Input, request *StartOperationRequest) (*WorkflowQueryOptions, error) {
+		return &WorkflowQueryOptions{
+			Input:      []any{input},
+			QueryType:  "get-cell-status",
+			WorkflowID: fmt.Sprintf("%s-%s", "provision-cell", input.CellID),
+		}, nil
+	}))
+```
+
+### Workflow Signal
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type WorkflowSignalOptions struct {
+	WorkflowID           string
+	// Optionally expose RunID too
+	SignalType            string
+	// Input to provide to the query.
+	Input                []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+func NewWorkflowSignalHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowSignalOptions, error)) Handler
+```
+
+**Basic usage:**
+
+```go
+w.RegisterOperation("update-cell-config", NewWorkflowSignalHandler(
+	func(ctx context.Context, input Input, request *StartOperationRequest) (*WorkflowSignalOptions, error) {
+		return &WorkflowSignalOptions{
+			Input:      []any{input},
+			SignalType: "update-cell-config",
+			WorkflowID: fmt.Sprintf("%s-%s", "provision-cell", input.CellID),
+		}, nil
+	}))
+```
+
+### Workflow Update
+
+Not defined yet. We can easily support sync update, but async update would require more thought on the guarantees
+Temporal provides for callback delivery on update completion.
+
+### Multi-tenant handlers
+
+This section introduces a couple of additional concepts, failing handler requests, and operation interceptors.
+
+Handlers should be able to fail requests in a meaningful way, providing the proper HTTP response code to convey errors
+such as bad request and unauthorized. To control request failure, handlers can return a `nexus.HandlerError` with an
+HTTP status code and failure object.
+Returning any other error from a handler should result in an internal server error and the error details omitted from
+the response.
+
+Interceptors are used to share logic across all registered operations. i.e, interceptors can enforce authorized access
+to operations making them ideal for low boilerplate multi-tenant application development.
+
+```go
+package interceptor // go.temporal.io/sdk/interceptor
+
+type OperationInboundInterceptor interface {
+	StartOperation(ctx context.Context, request *StartOperationRequest) (StartOperationResponse, error)
+	CancelOperation(ctx context.Context, request *CancelOperationRequest) error
+
+	mustEmbedOperationInboundInterceptorBase()
+}
+```
+
+**Example:**
+
+```go
+import (
+  "github.com/nexus-rpc/go-sdk/nexus"
+  "go.temporal.io/sdk/interceptor"
+  "go.temporal.io/sdk/operation"
+)
+
+type operationAuthorizationInterceptor struct {
+	interceptor.OperationInboundInterceptorBase
+}
+
+func (*operationAuthorizationInterceptor) CancelOperation(ctx context.Context, request *CancelOperationRequest) error {
+	// Use the request headers to extract and authenticate the tenant.
+	// Temporal may alter the received HTTP request, adding headers such as Temporal-Source-Namespace and X-Forwarded-For
+	// that handlers can trust.
+	tenantID, err := getTenantID(ctx, request.Header)
+	if err != nil {
+		return &nexus.HandlerError{StatusCode: http.StatusUnauthorized, Failure: &nexus.Failure{Message: "unauthorized"}}
+	}
+	// Enforce a specific pattern.
+	if !strings.HasPrefix(request.OperationID, fmt.Sprintf("%s-%s-", request.Operation, tenantID)) {
+		return &nexus.HandlerError{StatusCode: http.StatusUnauthorized, Failure: &nexus.Failure{Message: "unauthorized"}}
+	}
+	return nil
+}
+
+w.RegisterOperation("provision-cell", NewWorkflowRunHandler(
+	func(ctx context.Context, input Input, request *StartOperationRequest) (*WorkflowStartOptions, error) {
+		tenantID, err := getTenantID(ctx, request.Header)
+		if err != nil {
+			return nil, &nexus.HandlerError{StatusCode: http.StatusUnauthorized, Failure: &nexus.Failure{Message: "unauthorized access"}}
+		}
+
+		return &WorkflowStartOptions{
+			Input:        []any{input},
+			WorkflowType: "provision-cell",
+			StartWorkflowOptions: StartWorkflowOptions{
+				// Add the tenant ID as part of the workflow ID.
+				// The returned ID is mapped directly to the operation ID, an interceptor can be used to manipulate it.
+				ID: fmt.Sprintf("%s-%s-%s", request.Operation, tenantID, input.CellID),
+			},
+		}, nil
+	}))
+```
+
+### Operation context
+
+The context provided to `operation.Handler` methods can be used to extract the client, logger, metrics handler, and data
+converter from the worker, similiarly to how the [activity package][activity-package] exposes similar methods.
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+func GetClient(context.Context) client.Client
+func GetDataConverter(context.Context) converter.DataConverter
+func GetLogger(context.Context) log.Logger
+func GetMetricsHandler(context.Context) metrics.Handler
+```
+
+All of these methods panic if called outside of an operation handler.
+
+[temporal-go-sdk]: https://github.com/temporal/sdk-go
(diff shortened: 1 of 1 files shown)


nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+func GetDataConverter(context.Context) converter.DataConverter
+func GetLogger(context.Context) log.Logger
+func GetMetricsHandler(context.Context) metrics.Handler
+```
+
+All of these methods panic if called outside of an operation handler.
+
+[temporal-go-sdk]: https://github.com/temporal/sdk-go
+[nexus-go-sdk]: https://github.com/nexus-rpc/sdk-go
+[nexus-http-api]: https://github.com/nexus-rpc/api/TODO
```
bergundy, line 306: Still need to merge the API PR, I'll update the link when merged.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+package operation // go.temporal.io/sdk/operation
+
+type WorkflowStartOptions struct {
+	// Options to start the workflow with. TaskQueue defaults to the current worker's task queue.
+	// The workflow ID used here will be returned as the operation ID by default.
+	client.WorkflowStartOptions
+	// Type of workflow to start.
+	WorkflowType string
+	// Input to provide to the workflow.
+	Input        []any
```
bergundy, line 127: Went with multiple arguments since all of the Temporal APIs support this.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+
+**Goals**
+
+- Expose a simple, high level API for binding Nexus operations to common Temporal constructs
+- Support for starting and canceling operations
+- Support multi-tenant applications with minimal boilerplate
+
+**Non-Goals**
+
+- Support for getting an operation's result or info
```
cretz, line 15: Hrmm, doesn't seem that hard to do

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+The Temporal Go SDK exposes a `Worker` interface that embeds a `Registry` interface defined as:
+
+```go
+type Registry interface {
+	WorkflowRegistry
+	ActivityRegistry
+}
+```
+
+Add and embed a third registry type: `OperationRegistry` with the following method:
```
cretz, line 48: Embed? So a Nexus handler has to run in a Temporal worker? I wonder if a separate Nexus worker makes sense. I would like to see details on how Nexus handlers are getting work to do.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+// HTTP Request object.
+type CancelOperationRequest struct {
+	Operation   string
+	OperationID string
+
+	URL    *url.URL
+	Header http.Header
+}
+
+type OperationResponse interface {
```
cretz, line 91: You called this `StartOperationResponse` above

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+	RegisterOperation(operation string, handler operation.Handler)
+	// Alternative: get the operation name from handler.Name()
+	// RegisterOperation(handler operation.Handler)
+}
+```
+
+`operation.Handler` is an interface with a minimal definition of:
+
+```go
+package operation // go.temporal.io/sdk/operation
```
cretz, line 61: > go.temporal.io/sdk/operation

This should either be in a Nexus-specific area or have Nexus-specific prefixes IMO

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+// Maps almost 1:1 to the Nexus Go SDK StartOperationRequest, exposing HTTP header and body directly instead of the HTTP
+// Request object.
+type StartOperationRequest struct {
+	Operation   string
+	RequestID   string
+	CallbackURL string
+
+	URL    *url.URL
+	Header http.Header
+	Body   []byte
```
cretz, line 78: Hrmm, I wonder if a user might want HTTP method or TLS details. Can wait, but alternatively could reconstruct an `*http.Request` via gRPC response and add things on the gRPC response to build it up as needed.

In fact, I think just reusing the `nexus.Handler` interface is cleaner, even if you are proxying through the server.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+	// Type of workflow to start.
+	WorkflowType string
+	// Input to provide to the workflow.
+	Input        []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+// Returns a Handler that can maps and operation to a workflow that can be started and canceled.
+// Handles HTTP request to payload deserialization for input of type `T`.
+func NewWorkflowRunHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowStartOptions, error)) Handler
```
cretz, line 133: Why `T`? I don't think it's too much to ask users to deserialize the request body themselves. You can add helpers on the start operation request if needed. But people need to be able to do this without args too.

cretz, line 133: I think this makes more sense as:

```go
// This always uses the workflow ID as the operation ID
type WorkflowRunHandlerFunc func(context.Context, *StartOperationRequest) (client.WorkflowRun, error)
var _ Handler = (WorkflowRunHandlerFunc)(nil)

type SyncHandlerFunc func(context.Context, *StartOperationRequest) (*OperationResponseSync, error)
var _ Handler = (SyncHandlerFunc)(nil)
```

It's not asking too much for a user to make a client call themselves IMO. And the fixed workflow-id-as-operation-id makes cancel guaranteed to work. People need two Temporal handler helpers:

* Single-function async handler based on workflow ID that lets me do anything with Temporal client in starter
* Single-function sync handler that lets me do anything with Temporal client

It's easy for users to make a one-liner for workflows using the first and one-liner for signals/queries using the second. But I also get a lot of flexibility without being forced into returning options and such. Just let the user make the client call how they want. So they could put something on the context as gRPC metadata, or they could make a different client with different client options, or whatever.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+	WorkflowID           string
+	// Optionally expose RunID too
+	QueryType            string
+	QueryRejectCondition enums.QueryRejectCondition
+	// Input to provide to the query.
+	Input                []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+func NewWorkflowQueryHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowQueryOptions, error)) Handler
```
cretz, line 166: Mentioned above, I don't think it's too much to ask users to make the client calls themselves

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+type WorkflowSignalOptions struct {
+	WorkflowID           string
+	// Optionally expose RunID too
+	SignalType            string
+	// Input to provide to the query.
+	Input                []any
+	// TODO: Should Temporal Header be settable here?
+}
+
+func NewWorkflowSignalHandler[T any](func(context.Context, T, *StartOperationRequest) (*WorkflowSignalOptions, error)) Handler
```
cretz, line 196: Same here as query, a simple `SyncHandlerFunc` can do this

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+Returning any other error from a handler should result in an internal server error and the error details omitted from
+the response.
+
+Interceptors are used to share logic across all registered operations. i.e, interceptors can enforce authorized access
+to operations making them ideal for low boilerplate multi-tenant application development.
+
+```go
+package interceptor // go.temporal.io/sdk/interceptor
+
+type OperationInboundInterceptor interface {
```
cretz, line 233: I don't think this is how interceptors should be handled. Handlers are interfaces, so it is very easy to create a wrapper handler that intercepts and does things. That's how most Go middleware works (e.g. how we do data converters, how HTTP does HTTP handler middleware, etc).

nexus/go-handler.md
bergundy: Definitely not hard to do, just trying to focus on getting cross namespace workflow calls working E2E and removing any unnecessary pieces.

nexus/go-handler.md
bergundy: You can run your own Nexus handlers over HTTP but if you want to use Temporal, might as well hook this into a worker.

This is roughly what I have in mind:

image

nexus/go-handler.md
bergundy: I actually like that we don't expose Nexus in the SDK. Operation is a new concept, like an activity or a workflow and deserves a top level namespace IMHO.

nexus/go-handler.md
bergundy: HTTP method is always POST for start, that's what we defined in the Nexus HTTP spec.
Not sure what you mean by gRPC response, Nexus doesn't use gRPC, gRPC is just for delivering Nexus tasks and completing tasks between frontend and worker (and internally matching).

But I may reuse the nexus.Handler interface and construct an http.Request as you mentioned, although I'm not 100% there's too much value there.

nexus/go-handler.md
bergundy: Thanks, that's the downside of markdown, you don't get static analysis. I'll fix above to match the nexus API.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+}
+```
+
+`operation.Handler` is an interface with a minimal definition of:
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type Handler interface {
+	StartOperation(context.Context, *StartOperationRequest) (StartOperationResponse, error)
```
bergundy, line 64:
```suggestion
	StartOperation(context.Context, *StartOperationRequest) (OperationResponse, error)
```

nexus/go-handler.md
bergundy: I don't want users to have to set the `CallbackURL` and `RequestID`, I want the SDK to do that for them, also think it's nice to have the task queue default to the current worker's task queue.

Users can implement the Handler interface for more advanced functionality.

> And the fixed workflow-id-as-operation-id makes cancel guaranteed to work

Agree, that's what I have too, they can intercept `StartOperation` to manipulate that or implement their own handler.

It's important to know what the operation represents so we can generate handlers for cancel, get result, and get info.

I like that there SDK does http <-> payload conversion and fills in the gaps, I wouldn't want every user to have to implement that for every handler binding, that's what the handler interface is for.

nexus/go-handler.md
bergundy: But the SDK exposes an API to register a handler for an operation, not a generic handler for any operation and there's no concept of chaining handlers.

Do you think that should change?

nexus/go-handler.md
cretz: While it can be Temporal specific, hooking into an existing worker can be rough in some languages. You may have to do some annoying work to force yourself to be a third poller type in some of the languages as opposed to a separate Nexus worker. I also think the mapping of Temporal task queue to Nexus service may also be a bit rough. But it's ok w/ me if we want to combine these.

bergundy: I think this is better experience for Temporal users.

nexus/go-handler.md
cretz: I think we should expose this as Nexus. Nexus is the new concept, not just Nexus operations. Nexus handlers, calling non-Temporal Nexus servers, etc. Unless you plan on supporting non-Nexus operation forms, there's no value in hiding IMO.

f we _must_ hide that we're using Nexus while also expecting them, I think we need a new word or at least an additional adjective besides "operation" which is fairly overloaded.

nexus/go-handler.md
cretz: > HTTP method is always POST for start, that's what we defined in the Nexus HTTP spec.

True, but what about TLS details? May not be needed now, but we should probably not just build our own form of HTTP request for every language if the language has a perfectly good one

> Not sure what you mean by gRPC response

I mean it's not a real HTTP request coming to the handler, it's built based on things that come from Temporal server over gRPC

> But I may reuse the nexus.Handler interface and construct an http.Request as you mentioned, although I'm not 100% there's too much value there.

I think there's plenty of value in developing a Nexus handler that can work with and without Temporal. So you should also probably provide a way to set the client in the context too.

nexus/go-handler.md
cretz: > I don't want users to have to set the CallbackURL and RequestID

They don't have to with what I propose. My proposal in this comment is handler side.

> Users can implement the Handler interface for more advanced functionality.

Exactly, but I think the high-level shortcuts I propose are cleaner and more flexible

> Agree, that's what I have too

Except the comment on WorkflowStartOptions says "The workflow ID used here will be returned as the operation ID by default.", specifically the "by default" part.

> It's important to know what the operation represents so we can generate handlers for cancel, get result, and get info.

Right, my proposal does not change that

> I like that there SDK does http <-> payload conversion and fills in the gaps

But it may be dependent upon things in the request. If you really wanted to go that far with the shortcut, you wouldn't even force people to implement a whole new function, you'd just ask for a workflow function to map to an operation and that's it. Maybe you do want that shortcut, unsure, but this half shortcut of forcing a function implementation but not letting them start the workflow and/or work with args I think has limited value.

nexus/go-handler.md
cretz: Right, just like Go has the ability to register a handler for an HTTP call. Users are expected to wrap it before registering in these languages. That's how the handler-based middleware works.

> Do you think that should change?

Not necessarily. Anyone can do `RegisterOperation("my-operation", NewAuthHandlerMiddleware(actualOperationHandler))`.

nexus/go-handler.md
bergundy: > I mean it's not a real HTTP request coming to the handler, it's built based on things that come from Temporal server over gRPC

Is there anything specific that you would add to the request from the PollNexusTaskResponse or the gRPC metadata? I was going to reflect the incoming HTTP request here, not the internal server-worker protocol.

I'll switch to use the nexus handler interface.

cretz: Nah, I think we can start basic with the URL, header, and body like you have now and grow as needed

nexus/go-handler.md
bergundy: Yeah, but they'd have to repeat this wrapping for every operation, I don't like that.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+  will be addressed shortly after Nexus becomes available for preview.
+
+> NOTE: These are non-goals for _this_ proposal, but they will be defined in follow up proposals / PRs.
+
+## Background
+
+Nexus is a thin protocol on top of HTTP, defined in [this spec][nexus-http-api]. Temporal serves as an HTTP proxy,
+translating Nexus requests into tasks that can be polled on by an SDK worker on its configured task queue.
+Due to the synchronous nature of HTTP, the generated tasks are not persisted, and require an available worker
+to be processed.
```
antlai-temporal, line 33: In an async start operation,  before we can return a "201 created" we have to make sure that either the Nexus request+id is persisted, or the generated tasks+id are persisted. Otherwise, we force a stateless client to do retries, and that defeats the purpose of nexus imho...

nexus/go-handler.md
cretz: > Your proposal would force every user to remember to map these fields from StartOperationRequest to the ExecuteWorkflow client call.

"these fields" being ID and body, yes. That's two fields, and can provide helpers if that's too much. But at least I could set a gRPC auth header, or a start timeout, or react to a start error a certain way. Basically I want the simplicity of tying a started workflow to an operation. I'll deal with how to start the workflow and any other client calls I may want to make. We can have higher levels on top, but forcing a user to implement every handler method just because they want to control the start call is a bit rough.

> We haven't asked users to deal directly with payload conversion in their application logic in other APIs, why would we want to do that here, and in addition bother them with converting http requests to payloads?

Because this is a remote request. Data converters are specific to you, not to a third party calling you. Only those that never customize converters on either side can use the default, and that's still available for use. And we can make helpers still.

> It still needs to be a function because there's input manipulation, ID generation, and setting of workflow options based on the request.

Right, so let me do more than just those things.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+
+The process of translating requests into tasks and delivering them to workers is out of scope for this proposal.
```
antlai-temporal, line 35: Yes, but it would help to clarify the properties that handler code need to respect.

For example, handlers need to be idempotent, if we want to trigger nexus operation from activities.
Handlers will also need to be deterministic if we want to do the triggering from a workflow, possibly by just storing the outcome, similar to an activity...
Also, if handlers are doing complex coordination we will quickly lose some of the nicest properties of temporal, it is best to use a single workflow to do the orchestration...

So all that means that handlers are going to be really simple, mostly a one-to-one mapping to workflow operations, and we should be clear about that upfront...

nexus/go-handler.md
cretz: But everyone else writing middleware does. If you'd like, you can accept this worker option:

```go
NexusHandlerMiddleware: []func(Handler) Handler
```

nexus/go-handler.md
bergundy: > "these fields" being ID and body

No, those are `RequestID` and `CallbackURL`. I'm just worried that users will forget to propagate these, it's better that we do it for them IMHO than explain that they need to do it themselves, it's just to easy to forget or overlook.
If you think of a good way to enforce this while letting users control the start request, I'll change the return type to `client.WorkflowRun`, maybe we can inject interceptors into the provided client to automatically propagate these fields from the request if they're not set by the user.

> Because this is a remote request. Data converters are specific to you, not to a third party calling you. Only those that never customize converters on either side can use the default, and that's still available for use. And we can make helpers still.

I want to optimize for the Temporal->Temporal case, maybe there's a case for interceptors that can set the data converter on the context.

> Right, so let me do more than just those things while also letting me tie a single workflow ID/run to an operation.

What else do you need to do? Apart from sending the start request, you can do anything in the handler.

nexus/go-handler.md
bergundy: Cool, I'll change then.

nexus/go-handler.md
```diff
@@ -0,0 +1,307 @@
...
+			Input:      []any{input},
+			SignalType: "update-cell-config",
+			WorkflowID: fmt.Sprintf("%s-%s", "provision-cell", input.CellID),
+		}, nil
+	}))
+```
+
+### Workflow Update
+
+Not defined yet. We can easily support sync update, but async update would require more thought on the guarantees
```
antlai-temporal, line 214: For sync update, and all the issues about ordering guarantees between signals and updates, and updates from a single source, will nexus affect the ordering? 
Nexus could return a 201 for an operation in which the handler is doing a sync update, making a sync operation async...
I think that eventually nexus will need to serialize updates per workflow if we want to 
use sync updates in a sensible manner

nexus/go-handler.md
bergundy: They'll be persisted if the client gets a response.

nexus/go-handler.md
bergundy: Sure, I'll find a way to clarify some of that.

nexus/go-handler.md
bergundy: I don't have super strong opinion here, I like that operations are a top level concept and it's not like we're hiding Nexus completely especially if we're reusing some of the Nexus Go SDK types.
`go.temporal.io/sdk/worker/nexus`?

nexus/go-handler.md
bergundy: Not sure what you mean here, the handler is only responsible for sending the update request, ordering guarantees is out of scope for this proposal. The challenge with async update is ensuring that the callback is delivered even if the update doesn't complete or the workflow is terminated midway.

nexus/go-handler.md
bergundy:
```
NexusHandlerMiddleware: []func(Handler) Handler
```

That's essentially what the interceptors do, just a different way of exposing this, I think interceptors are more consistent with how we approach middleware in the the other parts of the SDK.

nexus/go-handler.md
antlai-temporal: I was thinking ordering end-to-end: a client triggers nexus op1 and gets a 201, then it triggers nexus op2 and gets another 201 (or 200). If the handler is doing updates on the same workflow, people may assume that op1 update happened before op2 update, but is that something we can guarantee with Nexus? Is the handler going to reorder requests?

nexus/go-handler.md
antlai-temporal: That's great, we need to make that explicit in the text, this is important...

nexus/go-handler.md
bergundy: Handler is basically a stateless HTTP request handler, it's not going to be able to do much, and it's not the handler's job.

antlai-temporal: I understand... I'm just thinking how we are going to get the end-to-end ordering back for updates, and it is not going to be easy... 
And if people assume that this is implementing "cross-namespace updates", ordering is the first thing they are going to be surprised about...

nexus/go-handler.md
cretz: Maybe. I think it's worth showing what a program that only handles operations looks like. Today in the Go SDK for example, an activity worker only has to explicitly disable workflow worker. But other SDKs are more sane here.

nexus/go-handler.md
cretz: At the least you could add them later and not concern yourself so early with the advanced interception feature.

> I think interceptors are more consistent with how we approach middleware in the the other parts of the SDK.

Just know that many people are going to ignore interceptors and do middleware the way they do for other interface-based handlers in Go.

nexus/go-handler.md
cretz: > No, those are RequestID and CallbackURL. I'm just worried that users will forget to propagate these, it's better that we do it for them IMHO than explain that they need to do it themselves, it's just to easy to forget or overlook.

Ah, I see now. Ok, can we at least have both options? I need a way to tie a handler to a workflow ID but still control the client calls.

> What else do you need to do? Apart from sending the start request, you can do anything in the handler.

Sure I can. I can set gRPC headers, I can set context timeout, I can react to one start error differently than another, I can signal with start, etc.

bergundy: Hmm.. you may not be able to use signal with start to back an operation. It depends if we want to support multiple callback URLs for a single workflow. I still need to spec that out.

But yes, you have a point that you may want more control, there may be a middle ground where you don't have to implement the entire handler and still get to start the workflow yourself. I just think most users won't need this and for reasons I've stated above, I think this abstraction will address the common case and let's the framework take care of most of the boilerplate.

In the PoC that I ran for this, I originally had a public struct that is created by `NewWorkflowRunHandler`, I removed it from this proposal but I can bring it back.

```go
type WorkflowRunHandler struct {
	StartWorkflow func(context.Context, *StartOperationRequest) (client.WorkflowRun, error)
}
```

cretz: > a middle ground where you don't have to implement the entire handler and still get to start the workflow yourself

Exactly.

> I removed it from this proposal but I can bring it back.

Or better, from the original comment in this thread:

```go
type WorkflowRunHandlerFunc func(context.Context, *StartOperationRequest) (client.WorkflowRun, error)
var _ Handler = (WorkflowRunHandlerFunc)(nil)

type SyncHandlerFunc func(context.Context, *StartOperationRequest) (*OperationResponseSync, error)
var _ Handler = (SyncHandlerFunc)(nil)
```

Single function implementations of the Nexus handler interface make sense. So people can:

```
RegisterOperation("my-operation", WorkflowRunHandlerFunc(
  func(ctx context.Context, req *StartOperationRequest) (client.WorkflowRun, error) {
    // do stuff with client on context
  }
))
```

And you can implement your higher level one on top of this too. And the same for `SyncHandlerFunc` so they can do whatever they want in the function.

bergundy: > ```
> type WorkflowRunHandlerFunc func(context.Context, *StartOperationRequest) (client.WorkflowRun, error) var _ Handler = (WorkflowRunHandlerFunc)(nil)
> ```

How can a single function implement the entire handler interface? Maybe you know some Go trick that I don't.

I still think we want the higher level API I suggested to be what users first encounter as it addresses the common case requirements.

I thought about another option:

```go
w.RegisterOperation("my-operation", NewWorkflowRunHandler(
	// Input is deserialized from payload extracted from request body 
	func(ctx context.Context, input Input, request *StartOperationRequest) (client.WorkflowRun, error) {
	}
))

w.RegisterOperation("my-operation", NewWorkflowRunHandler(
	// Input is the raw request body
	func(ctx context.Context, input io.Reader, request *StartOperationRequest) (client.WorkflowRun, error) {
	}
))
```

I think that mostly works but we're still left with the problem of attaching the request ID and callback URL to the workflow run.
We could solve that using interceptors and augment the StartWorkflowExecution request in this context but I don't love that this is implicit and may be attached to more than one workflow if started from that context.

nexus/go-handler.md
```diff
@@ -0,0 +1,258 @@
...
+
+## Overview
+
+An addition to the [Temporal Go SDK][temporal-go-sdk] that exposes the ability for users to handle Nexus API requests.
+
+**Goals**
+
+- Expose a simple, high level API for binding Nexus operations to common Temporal constructs
+- Support for starting and canceling operations
+- Support multi-tenant applications with minimal boilerplate
```
Quinn-With-Two-Ns, line 11: should I read this as "support cross namespace calls?"

nexus/go-handler.md
Quinn-With-Two-Ns: will users have a way to disable the workflow and activity worker so they can only run a nexus handler?

bergundy: They should be able to IMHO.

nexus/go-handler.md
bergundy: That's part of it but we want to allow calling from different namespaces possibly from different organizations.

nexus/go-handler.md
cretz: > How can a single function implement the entire handler interface? Maybe you know some Go trick that I don't.

By making the function the receiver, e.g. [http.HandlerFunc](https://pkg.go.dev/net/http#HandlerFunc):

```go
type SyncHandlerFunc func(context.Context, *StartOperationRequest) (*OperationResponseSync, error)

func (s SyncHandlerFunc) StartOperation(ctx context.Context, req *StartOperationRequest) (StartOperationResponse, error) {
   run, err := s(ctx, req)
   // stuff
}
```

Granted `WorkflowHandlerFunc` may not be able to function as a full handler with cancel, so it may be worth bringing back `WorkflowRunHandler`.

nexus/go-handler.md
bergundy:
```suggestion
[nexus-http-api]: https://github.com/nexus-rpc/api/blob/main/SPEC.md
```

nexus/go-handler.md
dandavison: +1, I don't understand this sentence:

> Due to the synchronous nature of HTTP, the generated tasks are not persisted, and require an available worker to be processed.

My understanding is that the whole point of Nexus is that it is adding the possibility for asynchronicity to HTTP. But I read the sentence as saying that it only supports synchronous responses.

nexus/go-handler.md
```diff
@@ -0,0 +1,265 @@
...
+handler implementations.
+
+### Workflow Run
+
+```go
+package operation // go.temporal.io/sdk/operation
+
+type WorkflowStartOptions struct {
+	// Options to start the workflow with. TaskQueue defaults to the current worker's task queue.
+	// The workflow ID used here will be returned as the operation ID by default.
```
bergundy, line 72: Workflow ID should be required here.

nexus/go-handler.md
```diff
@@ -0,0 +1,265 @@
...
+	client.WorkflowStartOptions
+	// Type of workflow to start.
```
bergundy, line 74: Optional, defaults to operation name.

nexus/go-handler.md
```diff
@@ -0,0 +1,265 @@
...
+
+**Basic usage:**
+
+```go
+w.RegisterOperation("provision-cell", NewWorkflowRunHandler(
+	func(ctx context.Context, input Input, request *StartOperationRequest) (*WorkflowStartOptions, error) {
+		return &WorkflowStartOptions{
+			Input:        []any{input},
+			WorkflowType: "provision-cell",
+			StartWorkflowOptions: StartWorkflowOptions{
```
dandavison, line 94: Having both `WorkflowStartOptions` and `StartWorkflowOptions` is going to be confusing.

nexus/go-handler.md
bergundy: I'm resolving this. Feel free to unresolve if you think it's worth continuing this thread.

bergundy:
Now that I'm close to working on SDK again, I'll close and open a new proposal with the knowledge I've gained in the last few months.

PR #85, opened 2023-10-24
Propose translation of Temporal Payloads to/from Nexus requests and responses

bergundy:
[Rendered](https://github.com/bergundy/proposals/blob/nexus-temporal-payloads/nexus/temporal-payloads.md)

### DIFF
--- a/nexus/temporal-payloads.md
+++ b/nexus/temporal-payloads.md
@@ -1,0 +1,69 @@
+# Temporal Payloads over Nexus+HTTP
+
+## Motivation
+
+For ideal integration of Temporal applications with Nexus, Temporal users should be able to leverage Temporal Payloads
+and not have to concern themselves with Nexus implementation details.
+
+When a workflow completes, we'd like Temporal server to be able to deliver the completion result to any registered Nexus
+callbacks and the callback receiver (a Temporal server) to translate the completion to a Temporal Payload.
+
+Inherently, we need a fixed set of translation rules from Payloads to Nexus+HTTP and back. These rules should be
+executed server-side.
+
+## Background
+
+Nexus reserves the entire HTTP request and response bodies (for input and successful results) for data and uses HTTP
+headers (a mapping of string to an array of strings) for transmitting metadata.
+
+> The main motivation for this design was to support streaming and avoid requiring HTTP proxies to inspect request
+> bodies.
+
+A Temporal Payload is defined as:
+
+```proto3
+message Payload {
+    map<string,bytes> metadata = 1;
+    bytes data = 2;
+}
+```
+
+## Translation rules
+
+To unmarshal a Temporal Payload from a Nexus request or response:
+
+1. Extract all `Content-Temporal-` prefixed headers into the payload metadata using the following rules:
+   1. Ignore multiple values (not supported)
+   2. Strip the key prefix and transform header case to camel case. E.g: `Content-Temporal-Foo-Bar` becomes `fooBar`
+   3. Read the values as [data URLs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs)
+2. Set the `encoding` Payload metadata field as follows:
+   1. If `Content-Length` is 0 or body is empty → `binary/null`
+   2. If `Content-Type` media type is `application/json` →
+      1. If the content type parameter contains a `format=protobuf` → `json/protobuf` and set `messageType` to the corresponding parameter value.
+         Example: `Content-Type: application/json; format=protobuf; messageType=com.example.Message`
+      2. Otherwise → `json/plain`
+   3. If `Content-Type` media type is `application/x-protobuf` → `binary/protobuf` and set the `messageType` parameter as given in the header.
+      Example: `Content-Type: application/x-protobuf; messageType="com.example.Message"`
+   4. Otherwise → `binary/plain`
+3. Extract all `Content-Encoding` values into a `codecs` metadata key with a comma separator. Note that the server may
+   decompress gzipped content and other popular encoding formats.
+4. Read the entire body into `Payload.data` and unset `Content-Length`
+5. Extract any other interesting `Content-` prefixed headers (none seem relevant ATM)
+
+To unmarshal from a Nexus request or response to a Temporal payload:
+
+1. Set the `Content-Length` header to the `Payload.data` length
+2. Set body to `Payload.data`
+3. Set `Content-Type` based on the `Payload.metadata` `encoding` field:
+   1. `binary/null` → unset
+   2. `json/plain` → `application/json`
+   3. `json/protobuf` → `application/json; proto=$Payload.metadata.messageType`
+   4. `binary/protobuf` → `application/x-protobuf; messageType=$Payload.metadata.messageType`
+   5. `binary/plain` → `application/octet-stream`
+   6. Anything else → `application/octet-stream; temporalEncoding=$Payload.metadata.encoding`
+4. Set `Content-Encoding` to `Payload.metadata.codecs`
+5. Other metadata fields are set as `Content-Temporal-` headers, transforming camel case to header case. Values are encoded as base64 data URLs.
+
+## Custom data converters
+
+Users implementing custom data converters will need to add a codec that converts to and from their non-standard payload format.
--- a/nexus/temporal-payloads.md
+++ b/nexus/temporal-payloads.md
@@ -67,3 +67,8 @@ To unmarshal from a Nexus request or response to a Temporal payload:
 ## Custom data converters
 
 Users implementing custom data converters will need to add a codec that converts to and from their non-standard payload format.
+
+## References
+
+- https://stackoverflow.com/a/48051331
+- https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#directives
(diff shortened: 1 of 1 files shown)


nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+
+## Motivation
+
+For ideal integration of Temporal applications with Nexus, Temporal users should be able to leverage Temporal Payloads
+and not have to concern themselves with Nexus implementation details.
+
+When a workflow completes, we'd like Temporal server to be able to deliver the completion result to any registered Nexus
+callbacks and the callback receiver (a Temporal server) to translate the completion to a Temporal Payload.
+
+Inherently, we need a fixed set of translation rules from Payloads to Nexus+HTTP and back. These rules should be
```
bergundy, line 11: We may want to also expose this at a later stage in the Temporal SDKs to allow using Nexus clients from application code but that's not required in the immediate future.

nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+## Translation rules
+
+To unmarshal a Temporal Payload from a Nexus request or response:
+
+1. Extract all `Content-Temporal-` prefixed headers into the payload metadata using the following rules:
+   1. Ignore multiple values (not supported)
+   2. Strip the key prefix and transform header case to camel case. E.g: `Content-Temporal-Foo-Bar` becomes `fooBar`
+   3. Read the values as [data URLs](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/Data_URLs)
+2. Set the `encoding` Payload metadata field as follows:
+   1. If `Content-Length` is 0 or body is empty → `binary/null`
```
cretz, line 40: I don't think null is the same as absent. We support default/zero values for parameters that are not specified, but null will fail to deserialize to some types. I think if there is no body, it's no payload. But if there is a body with just the text `null` that's `application/json`, it can be converted to `binary/null`.

nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+   2. If `Content-Type` media type is `application/json` →
+      1. If the content type parameter contains a `format=protobuf` → `json/protobuf` and set `messageType` to the corresponding parameter value.
```
cretz, line 42: I think it makes more sense to have `Content-Temporal-MessageType` than to put it on the content type

nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+         Example: `Content-Type: application/json; format=protobuf; messageType=com.example.Message`
+      2. Otherwise → `json/plain`
+   3. If `Content-Type` media type is `application/x-protobuf` → `binary/protobuf` and set the `messageType` parameter as given in the header.
+      Example: `Content-Type: application/x-protobuf; messageType="com.example.Message"`
+   4. Otherwise → `binary/plain`
+3. Extract all `Content-Encoding` values into a `codecs` metadata key with a comma separator. Note that the server may
```
cretz, line 48: I think this is premature and doesn't need to exist today (nor does `Accept-Encoding`). I think we can decide who converts HTTP content later from these headers (IMO it would be the Nexus side not the Temporal side reading these new metadata keys).

nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+   decompress gzipped content and other popular encoding formats.
+4. Read the entire body into `Payload.data` and unset `Content-Length`
+5. Extract any other interesting `Content-` prefixed headers (none seem relevant ATM)
+
+To unmarshal from a Nexus request or response to a Temporal payload:
+
+1. Set the `Content-Length` header to the `Payload.data` length
+2. Set body to `Payload.data`
+3. Set `Content-Type` based on the `Payload.metadata` `encoding` field:
+   1. `binary/null` → unset
```
cretz, line 58: Hrmm, null is different than absent

nexus/temporal-payloads.md
```diff
@@ -0,0 +1,74 @@
...
+   2. `json/plain` → `application/json`
+   3. `json/protobuf` → `application/json; proto=$Payload.metadata.messageType`
+   4. `binary/protobuf` → `application/x-protobuf; messageType=$Payload.metadata.messageType`
+   5. `binary/plain` → `application/octet-stream`
+   6. Anything else → `application/octet-stream; temporalEncoding=$Payload.metadata.encoding`
+4. Set `Content-Encoding` to `Payload.metadata.codecs`
```
cretz, line 64: Disagree that HTTP `Content-Encoding` should be used for this. I think that's reusing the HTTP layer too much for the codec. Proxies set/remove these headers, this is protocol-layer encoding not application-layer encoding.

nexus/temporal-payloads.md
bergundy: What's the difference between no payload and null?
There's not really a concept of 0 argument operations.

nexus/temporal-payloads.md
bergundy: Agree this could be deferred but I'm already thinking ahead to provide some information to users that want to know which codecs to apply.
This will also apply to encryption in my mind, e.g: you may see the following sequence `gzip, aes` and you'd need to deflate and decrypt.

nexus/temporal-payloads.md
bergundy: It's not, there's always a result and an input but we may need to construct the default value in some cases.

nexus/temporal-payloads.md
cretz: So how does one not send a payload at all?

nexus/temporal-payloads.md
cretz: Meh, even if you've seen that way around I still think metadata about the content is a bit different than its content type. But I don't care that much since it's just protos. Arguably you don't even need even need to support protos at first anyways, but this is fine.

nexus/temporal-payloads.md
cretz: I don't think we should confuse application codecs w/ HTTP ones. Let the HTTP middleware/proxies handle known HTTP codecs, and let applications handle known application codecs.

nexus/temporal-payloads.md
cretz: But if you have blind Temporal workflow mapping, there is not always an input

nexus/temporal-payloads.md
bergundy: Fair point.

But I do think that we'll want to come up with some convention for "application" codecs considering that these applications are going to be handling requests by arbitrary callers.

nexus/temporal-payloads.md
bergundy: There's a blind Temporal Payload mapping, these inputs map to operations, operations map to workflows at the application level.

nexus/temporal-payloads.md
cretz: I think you'll want to affect the content type. Not only would one expect to not leak the content type of their encoded body (so it should be embedded in the encoded body), but content type can't really be JSON if it's not JSON. This can sound like an argument for reusing `Content-Encoding` but I still think that deserves to be separate and that still leaks the internal content type.

Yes, it can be annoying deciding how to embed the original content type in the body, but that's the encoder authors problem since they have to extract it back out. Can also discuss whether other metadata deserves to be able to be encoded-and-hidden.

nexus/temporal-payloads.md
cretz: So how do you map operation input, which is required but may be null, to workflow input, which is not required and would be disallowed if given but not accepted.

nexus/temporal-payloads.md
bergundy: Nexus to Temporal goes through the user's operation handler.
The handler can specify that it expects nil input and invoke the workflow however it wants.

nexus/temporal-payloads.md
bergundy: I don't think they want to embed the original content type in the body, they can use headers for that.
Other metadata binary and encoded as base64, it's up to the the custom data converter implementation to put that metadata there in the first place.

nexus/temporal-payloads.md
bergundy: Not sure I follow the last point.

Basically I would expect a handler to look like this:

```go
var startWorkflowOp = temporalnexus.NewWorkflowRunOperation("some-op", temporalnexus.WorkflowRunOptions[nexus.Void, MyOutput]{
	Start: func(ctx context.Context, c client.Client, input nil, options nexus.StartOperationOptions) (temporalnexus.WorkflowHandle[MyOutput], error) {
		return temporalnexus.StartWorkflow(ctx, c, client.StartWorkflowOptions{
			ID: "some-id",
		}, MyHandlerWorkflow)
	},
})
```

nexus/temporal-payloads.md
cretz: Hrmm, I forget if we decided on a form that will auto-map the workflow without requiring a client call. If that exists, are we telling users that they cannot use it for parameterless workflows?

nexus/temporal-payloads.md
bergundy: Agreed offline to have the HTTP server handle content encoding.

nexus/temporal-payloads.md
cretz: Confirmed off-PR that the only workflow shortcuts where a user doesn't invoke the workflow themselves _requires_ a single parameter on the workflow for generic reasons. So I'm ok with the idea that every Nexus has some kind of input.

Can you show an example of a high-level Nexus operation handler for an operation with no user input? Must users type `*struct{}`?

nexus/temporal-payloads.md
bergundy: sgtm

cretz (approved):

PR #88, opened 2024-02-24
Worker Autotuning

Sushisource:
Rendered: https://github.com/Sushisource/proposals/blob/worker-autotune/all-sdk/autotuning.md


all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+users at the moment, and autotuning it is subject to the same lack of knowledge we have as with slot tuning.
+
+## Constraints
+
+There are some invariants that must be maintained:
+
+* The current number of active polls for a task type must be <= the number of free slots for that task type at all
+  times. Otherwise, we can end up receiving a task that won't have anywhere to go until a slot frees. For example,
+  in Core this is expressed by using a permit reservation system.
+* Max outstanding workflow tasks must be <= max cached workflows.
```
Quinn-With-Two-Ns, line 51: I don't think this is true? You can run a worker with no cache today

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+```rust
+trait SlotSupplier {
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
+  /// Returned errors will bubble up and cause the worker to shut down.
+  async fn reserve_slot(&self) -> Result<(), anyhow::Error>;
+
+  /// Tries to immediately reserve a slot, returning true if a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
```
Quinn-With-Two-Ns, line 77: For my understanding what sort of `SlotSupplier` could error?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// These functions would be provided to the user as top level statics they can access to get information about how
+// all outstanding slots are currently being used. Ostensibly they could've stored this in their own implementation
+// too, but this is to make it convenient for them to answer questions like "How many slots are being used by 
+// activities of type X right now?". 
+//
+// These could potentially be provided as parameters to the trait methods - but at least in the case of the 
+// async/blocking reserve, the information may change over time, so it may be best to have these be explicit calls.
+pub fn used_workflow_slots_info() -> WorkflowSlotsInfo { /* ... */ }
```
Quinn-With-Two-Ns, line 103: `SlotSupplier` is configured per worker right ? Should these be per worker?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+a more dynamic one, if we find that it substantially outperforms in most scenarios. If and when we do make that change,
+we'll want to give users some advance warning, since although the change is not technically breaking, it could have 
+knock-on effects in terms of the external load on the systems their workers are interacting with.
+
+Docs & education will both need content about what the interface is for, what default implementations are provided &
+what their limitations are, and how to implement a custom one.
+
+## Open questions
+
+### Async/fallible `mark_slot_used` and `release_slot`
```
Quinn-With-Two-Ns, line 393: Is this questions relevant to all SDKs or just a Rust SDK implementation of a `SlotSupplier`?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+/// Users might want to be able to pause the handing-out of slots as an effective way of pausing their workers.
+/// We can provide an implementation for this that wraps their implementation, or one of the defaults we provide.
+struct PauseableSlotSupplier<T: SlotKind> {
+  inner: Arc<dyn SlotSupplier<SlotKind=T>>,
+  paused: AtomicBool,
+  // ... details
+}
+impl<T: SlotKind> PauseableSlotSupplier<T> {
+  pub fn new(supplier: impl SlotSupplier<SlotKind=T>) -> Self { /* ... */ }
+  pub fn pause(&self) { /* ... */ }
```
Quinn-With-Two-Ns, line 164: There is a interesting question on when `pause` returns. Do you return when you stop handing out new slots or return when you have no more slots in "inflight" ie reserved but not marked in use

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+there is room in the cache (either additional capacity, or an idle cached workflow that can be evicted). This avoids
+a potential problem where we ask for a slot, they provide one, but there's nothing we can evict at the moment.
+
+If and when we do introduce user-configurable caching, it's likely that they would want the same thing to be responsible
+for both slots and cache management, since the two are closely related.
+
+A possible future:
+
+```rust
+trait WorkflowCacheSizer {
```
Quinn-With-Two-Ns, line 188: I would like us to explore the idea of fully pluggable workflow cache instead of just a pluggable cache sizer. I suspect there is some performance wins for certain workflows to use different caching strategy then we currently use. I see you cover some of it with the `eviction_hint `, but I think just letting users provide a cache implementations would be simpler .

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  /// Marks a slot as actually now being used. This is separate from reserving one because the pollers need to
+  /// reserve a slot before they have actually obtained work from server. Once that task is obtained (and validated)
+  /// then the slot can actually be used to work on the task.
+  /// 
+  /// Users' implementation of this can choose to emit metrics, or otherwise leverage the information provided by the
+  /// `info` parameter to be better able to make future decisions about whether a slot should be handed out.
+  fn mark_slot_used(&self, info: &Self::SlotKind::Info);
+
+  /// Frees a slot.
+  fn release_slot(&self, info: &Self::SlotKind::Info);
```
Quinn-With-Two-Ns, line 90: We could also consider taking an error here. If processing the task resulted in an error the provider may want to use that information to provide less slots.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+[see the branch
+here](https://github.com/temporalio/sdk-go/compare/master...Quinn-With-Two-Ns:sdk-go:poller-autottune-v0.2). Hence this
+doc will not focus on this area in detail.
+
+It is important to note that there is a distinction between the active number of poller *routines* and the active number
+of *open actual RPC calls*. The number of routines may be scaled independently of the number of open RPC calls at any
+time. The former may have a floor, whereas the latter must always be able to go to 0. Users are unlikely to care about
+the number of routines as they are relatively cheap. They *do*, however, care about how quickly we will change the
+number of active RPCs. The floor on the number of routines effectively determines how quickly we are willing to consume
+available slots and thus make actual RPC calls. IE: If the floor on the number of routines is 10, and 10 slots open up
```
dandavison, line 24: This is the first time the term "slots" is used -- a definition would be helpful.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+Workers have a configurable number of slots for task types. In Core: `max_outstanding_workflow_tasks`,
+`max_outstanding_activities`, and `max_outstanding_local_activities`. These are set by the user and can't change at
+runtime. Autotuning of these values seeks to allow them to change at runtime according to available resources on
+the worker. It is not feasible for us to choose *optimal* values for these at runtime, because we cannot know a-priori
+about how much resources any given workflow or activity task will consume - only the user can (maybe) know this.
+However, we can provide reasonable default autotuning implementations while also allowing users to provide their own.
+
+Additionally, there is a maximum cache size for cached workflows. In Core: `max_cached_workflows`. This is also set by
+users at the moment, and autotuning it is subject to the same lack of knowledge we have as with slot tuning.
+
```
dandavison, line 43: Somewhere before this point, I feel that I'm missing an explicit statement of what the problem to be solved is (expressed in terms of the technical concepts introduced in the paragraphs above, but "slot management" needs some background explanation).

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+is provided. These guarantees can't be compile-time enforced across the language boundary, but are at least useful after
+runtime validation.
+
+```rust
+trait SlotSupplier {
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
+  /// Returned errors will bubble up and cause the worker to shut down.
+  async fn reserve_slot(&self) -> Result<(), anyhow::Error>;
```
Quinn-With-Two-Ns, line 74: I think `reserve_slot` should take some sort of cancellation token (what is very lang dependent) and also return a bool if it did not acquire a slot. `reserve_slot` needs to be cancellable for the worker shutdown case, if the number of pollers is scaling down due to poller autotuning, and the whole eager workflow/activity use case. It should return a bool as well since in the later two cases a slot was not reserved, but I do not want the whole worker to shutdown

all-sdk/autotuning.md
Quinn-With-Two-Ns: In my Go implementation I split `release_slot` into two methods `mark_sot_unused` and `release` I did this because not all reserved slots will be used. If we just want one method to release used and unused slots maybe we can make `info` optional and if it is not present that indicates the slot was never used since if a slot wasn't used it will have no info?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+However, we can provide reasonable default autotuning implementations while also allowing users to provide their own.
+
+Additionally, there is a maximum cache size for cached workflows. In Core: `max_cached_workflows`. This is also set by
+users at the moment, and autotuning it is subject to the same lack of knowledge we have as with slot tuning.
+
+## Constraints
+
+There are some invariants that must be maintained:
+
+* The current number of active polls for a task type must be <= the number of free slots for that task type at all
```
Quinn-With-Two-Ns, line 48: > number of free slots 

Here "free slots" actually means slots not in use and not reserved by some eager request

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+}
+struct ActivitySlotsInfo {
+  used_slots: Vec<ActivitySlotInfo>,
+}
+struct LocalActivitySlotsInfo {
+  used_slots: Vec<LocalActivitySlotInfo>,
+}
+
+struct WorkflowSlotInfo {
+  workflow_type: String,
```
Quinn-With-Two-Ns, line 124: Would we expose if these are eager workflows/activities here? I think it would be useful for some metrics

dandavison:
I really appreciate how well-written this proposal is. (I'm just personally lacking a bit of necessary background.)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+
+
+### Slot tuning & Cache Size
+
+Workers have a configurable number of slots for task types. In Core: `max_outstanding_workflow_tasks`,
+`max_outstanding_activities`, and `max_outstanding_local_activities`. These are set by the user and can't change at
+runtime. Autotuning of these values seeks to allow them to change at runtime according to available resources on
+the worker. It is not feasible for us to choose *optimal* values for these at runtime, because we cannot know a-priori
+about how much resources any given workflow or activity task will consume - only the user can (maybe) know this.
+However, we can provide reasonable default autotuning implementations while also allowing users to provide their own.
```
paulnpdev, line 39: It might be nice to say *something* here about what this might look like.  For instance, we could take the approach that we tune for maximum concurrency (maintaining OOM or other error avoidance), or total throughput per worker, or minimum latency (for some measure of latency and some form of "minimum").  Any (or even even each) of these seem like valid strategies, as might some utility function that combined them.  Do we plan to offer the user a choice?  Pick one of these (or something I didn't mention) and do just that for now?  What's our true goal here?

Even though I have placed my comment here in the slot/cache tuning section, it's really more of a global comment.  We can probably do some stuff that improves worker performance in ways that are pareto optimal ... i.e. improve some dimension of performance without any cost in other dimensions.  Maybe that should be our first goal.  But most (I might even say all) performance tuning eventually requires a definition of performance that expresses the specific goal directly because there are always tradeoffs.

all-sdk/autotuning.md
paulnpdev: my earlier comment seems to be a long-winded expression of what Dan is saying here.

all-sdk/autotuning.md
paulnpdev: sorry if I'm late to the game here, but what do you mean by "fully pluggable"?  Do you (basically) mean user-written, or that we should provide multiple implementations?  If so could you give a hint what a different (from current) version might look like?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+shut down the worker.
+
+The downsides seem substantial, and users who want to do this can "offload" the async-ness/fallibility to the
+`reserve_slot` function as implied in my toy, which seems reasonable.
+
+## Future considerations
+
+This proposal has focused specifically on worker autotuning and user customization thereof. However, it's worth
+considering how this effort will fit in to future work we know we want to do, namely around autoscaling and on-demand
+workers.
```
paulnpdev, line 409: Another future option worth considering (imo) would be where some workers (at least new ones) would learn from already running workers about their learned configuration to speed up the learning process.

all-sdk/autotuning.md
Quinn-With-Two-Ns: By fully pluggable I mean user written and we provide a few implementation. Right now we have a bounded LRU cache, a different version I would like to experiment with is a cache that accounts for how much work it would take to replay the workflow. One way I image it could do that is when it needs to evict a workflow weight the history size in what workflow we evict.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+// These functions would be provided to the user as top level statics they can access to get information about how
+// all outstanding slots are currently being used. Ostensibly they could've stored this in their own implementation
+// too, but this is to make it convenient for them to answer questions like "How many slots are being used by 
+// activities of type X right now?". 
+//
+// These could potentially be provided as parameters to the trait methods - but at least in the case of the 
+// async/blocking reserve, the information may change over time, so it may be best to have these be explicit calls.
+pub fn used_workflow_slots_info() -> WorkflowSlotsInfo { /* ... */ }
+pub fn used_activity_slots_info() -> ActivitySlotsInfo { /* ... */ }
+pub fn used_local_activity_slots_info() -> LocalActivitySlotsInfo { /* ... */ }
```
cretz, line 105: I think this stuff can come later. The slot supplier impl can already keep general counts (and if they need to differentiate local from non-local activities, we should provide the contextual information at reservation time). As for count per workflow/activity type, I think we can expand to that later.

all-sdk/autotuning.md
cretz: IMO it should return immediately, pause is a state not an actionable request for the most part.

all-sdk/autotuning.md
cretz: Concur, the difference between providing a cache interface and a can-cache interface is minimal, but the former has lots of upside. Let's just expose a get/put/delete cache interface that also has a "can allow" call.

For core langs however, they will only expose the "can" (and eviction callbacks if they want) because the stateful cache impl has to be in Rust IMO w/ just hooks for checking.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  /// task and wait for another to complete so we can evict it and make room for the new one.
+  fn can_allow_workflow(&self, slots_info: &WorkflowSlotsInfo, new_task: &WorkflowSlotInfo) -> bool;
+  /// Called when deciding what workflow should be evicted from the cache. May return the run id of a workflow to evict.
+  /// If None, or an invalid run id is returned, the default LRU eviction / strategy is used. The id returned must
+  /// be one of the workflows provided in `evictable` (the ones for which there is no actively processing WFT).
+  fn eviction_hint(&self, evictable: &[WorkflowSlotInfo]) -> Option<&str>;
+  /// Called when a workflow is evicted from the cache
+  fn evicted_workflow(&self, evicted: &WorkflowSlotInfo);
+  /// If there is a fixed maximum number of cached workflows, return it.
+  fn max(&self) -> Option<usize>;
```
cretz, line 203: What is the value of this call on the interface? Isn't a fixed-max an implementation of this interface not a part of every implementation?

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  W->>S: `mark_slot_used`
+  W->>W: Process workflow task
+  W->>S: `release_slot`
+```
+
+## Default implementations
+
+We should provide a few default implementations:
+
+* Semaphore based SlotSupplier: This is the current implementation. There's a fixed number of slots, that's it.
```
cretz, line 252: We should agree on a name for this. I like "fixed" over "semaphore" because the latter implies algorithm details.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+* Resource based SlotSupplier: This implementation can look at how close the worker is to some resource limit
+  (% of CPU and/or memory used), and hand out slots if it's still below that limit. This only makes sense for
+  activity slot suppliers, since workflows use little CPU, and the memory they use is the concern of cache
+  management, not the slot supplier.
+* Memory based WorkflowCacheSizer: Like above, but for the sizing of the workflow cache.
```
cretz, line 257: We should get into the details somewhere about these impls, e.g. whether it's process mem used or system-free-mem based, encouragement of cross-worker use, which CPU load number is referenced and how it's aggregated across cores, what config values are exposed, etc.

I am ok deferring post-proposal, but I would like SDKs to be consistent in behavior and config.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+
+## Metrics
+
+Users can add their own metrics behind their implementations now, which is great - but we can also provide some out of
+the box. We can always provide `slots_in_use` by type. Right now we emit the inverse of this,
+`worker_task_slots_available` - we can keep emitting that with a bundled implementation that replicates the old
```
cretz, line 263: > we can keep emitting that with a bundled implementation

A little point: I vote we don't even emit this if they change from the default fixed impl (i.e. the emitting of this should happen in the fixed-slot impl, not part of the general "system" anymore)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  /// 
+  /// Users' implementation of this can choose to emit metrics, or otherwise leverage the information provided by the
+  /// `info` parameter to be better able to make future decisions about whether a slot should be handed out.
+  fn mark_slot_used(&self, info: &Self::SlotKind::Info);
+
+  /// Frees a slot.
+  fn release_slot(&self, info: &Self::SlotKind::Info);
+  
+  /// If this implementation knows how many slots are available at any moment, it should return that here.
+  fn available_slots(&self) -> Option<usize>;
```
cretz, line 93: I don't think this is necessary on this interface. That an implementation is has this information is not shouldn't matter to the system (it can emit its own available slot metric, we don't need to abstract it).

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+## Where it hooks up and the language barrier
+
+Staying in Rust for a moment, the slot supplier implementations can be hooked up on a per-worker basis through
+`WorkerConfig`, as expected. We would add:
+
+```rust
+pub struct WorkerConfig {
+  // ... existing fields
+  #[builder(default)]
+  pub workflow_task_slot_and_cache: Option<Arc<dyn WorkflowSlotAndCacheManager>>,
```
cretz, line 279: I think slot supplier and cache should be provided as separate option, even if the implementer chooses to have the same struct impl both

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+We can also of course keep emitting the current number of cached workflows.
+
+
+## Where it hooks up and the language barrier
+
+Staying in Rust for a moment, the slot supplier implementations can be hooked up on a per-worker basis through
+`WorkerConfig`, as expected. We would add:
+
+```rust
+pub struct WorkerConfig {
```
cretz, line 276: Feel free to not make these options as `Option` but instead require them and make the default/current impls easy to instantiate. (this is just for Core interfacing with langs, for users they absolutely should be optional with us falling back to today's behavior)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
+  /// Returned errors will bubble up and cause the worker to shut down.
+  async fn reserve_slot(&self) -> Result<(), anyhow::Error>;
+
+  /// Tries to immediately reserve a slot, returning true if a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
+  /// Returned errors will bubble up and cause the worker to shut down.
+  fn try_reserve_slot(&self) -> Result<bool, anyhow::Error>;
```
cretz, line 79: Just to understand, can I get some examples where we'd use non-blocking reserve attempt vs blocking? I can only think about blocking use cases.

cretz (commented):
Left some comments. Like with many of these proposals, two recommendations (take em or leave em):

* Setup meeting to knock out each contentious point
* Close (without merge) early, noisier PRs where there's a lot of chatter once many things are resolved and make a final PR for final +1's for merge

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+
+  /// Marks a slot as actually now being used. This is separate from reserving one because the pollers need to
+  /// reserve a slot before they have actually obtained work from server. Once that task is obtained (and validated)
+  /// then the slot can actually be used to work on the task.
+  /// 
+  /// Users' implementation of this can choose to emit metrics, or otherwise leverage the information provided by the
+  /// `info` parameter to be better able to make future decisions about whether a slot should be handed out.
+  fn mark_slot_used(&self, info: &Self::SlotKind::Info);
```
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I think it is an important distinction even the basic `FixedSlotSupplier` or `ResourceBoundSlotSupplier` need this information. `FixedSlotSupplier` for metrics and `ResourceBoundSlotSupplier` can use it for part of its logic. If we want to support passing any information on the reserved slot then the distinction is important.

all-sdk/autotuning.md
Quinn-With-Two-Ns: Eager workflows/activities

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: > FixedSlotSupplier for metrics
Which metrics? If we're talking about new metrics, we can give their definitions (and I consider a "used" slot to be a reserved one in that case).

> ResourceBoundSlotSupplier can use it for part of its logic

I don't believe it needs to. I think this complicates the interface and logic. Having said that, I don't mind it that much.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: >which metrics

`worker_task_slots_available`

I think it is important for `SlotSuppliers` to have the ability to know if a slot is just reserved vs actually being used.  Consider also a slot can be reserved, but never used. I don't think it complicates the interface, and if an implementation doesn't need to distinguish they just have a noop implementation so i don't see how it complicates the logic. In langs that support it we can provide a default noop implementation if we want.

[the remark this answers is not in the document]
Quinn-With-Two-Ns: `worker_task_slots_available` is unused slots. I think that is the correct definition since pollers reserve a slot before making the poll request so there may be no more unreserved slots, but the worker can still accept `unused slot` amount of work. Eager stuff does make it awkward I agree.

all-sdk/autotuning.md
Sushisource: Yeah that's true - but practically speaking it means your cache's peak size will equal the number of concurrent workflow tasks. In a real-world setting you probably care about cache size as it relates to peak memory usage, and since users are (should) always using the cache, the constraint makes sense in that context.

all-sdk/autotuning.md
Sushisource: All of them

all-sdk/autotuning.md
Sushisource: Yeah, i am imagining it returns immediately. You're saying "become paused" (like initiating shutdown of a worker)

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: > worker_task_slots_available is unused slots [...] Eager stuff does make it awkward I agree.

More than awkward, it makes it wrong. We may have to change the definition of "used" to "started a task or requested eager" instead of just "started a task" to make it be compatible. I am ok w/ that, but we should probably make sure to `mark_slot_used` as soon as we get a slot for eager use then.

all-sdk/autotuning.md
Sushisource: That's why it's optional - implementations that know it return it, if not they return None. We can emit a metric about total cache size if it's returned here.

all-sdk/autotuning.md
Sushisource: Sure, the point was to express what a user is going to see

all-sdk/autotuning.md
Sushisource: I don't want a provided implementation to be literally responsible for storing the data, though, that seems potentially easy to create confusing-to-diagnose bugs & more work for the implementor.

What would actual storage provide that `eviction_hint` doesn't? The example you gave @Quinn-With-Two-Ns seems to be serviced by the eviction hint.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Quinn-With-Two-Ns: I don't think it is wrong, I don't think there is a material difference between eager requests and polling a task queue. Those eager requests are just asking for a new task with a filter for what types they accept. I don't think we need to change any existing metric definition here. I would support adding a new metric to capture the unserved slots or capture the number of "pending"

all-sdk/autotuning.md
Sushisource: yes that's the intention

all-sdk/autotuning.md
cretz: But I think the impl can expose its own metric about total cache size if it cares to, the system doesn't need to care whether a max exists or not.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Sushisource: Sure, that's a good point. It can't (well, shouldn't) be exactly the same type as in the other places though, as it'd mean 90% of the fields on it become optional. Perfectly reasonable to have a new type here though with the fields we know like you mentioned, and then room to add more fields later on.

Quinn-With-Two-Ns: >I don't totally follow how the number of pollers scaling down works with that though

So if the number of pollers is scaling down, a `reserve_slot` call may be cancelled and not reserve a slot, but also not error (since that would kill the worker)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+There is some mild generic magic going on to represent different types of slots and constrain that the right info type
+is provided. These guarantees can't be compile-time enforced across the language boundary, but are at least useful after
+runtime validation.
+
+```rust
+trait SlotSupplier {
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available.
+  /// May also return an error if the backing user implementation encounters an error.
+  /// Returned errors will bubble up and cause the worker to shut down.
```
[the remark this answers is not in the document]
Sushisource: I don't really see what the difference is. A user calling shutdown is going to do exactly the same thing as returning an error, but with a somewhat less clear path of origin.

all-sdk/autotuning.md
cretz: The only reason is flexibility. It's basically a question of allowing replacement of the cache impl vs just the cache size. The former is more flexible I think and easy enough to build a sizer on top of, but I guess you may have to abstract the sizer anyways then. So a double-abstraction could exist: a worker accepts a custom cache implementation, and the fixed-size implementation could accept a dynamic sizer.

I don't have a strong opinion and can understand why just the sizer is enough. But I would change the names on some of these things a bit.

all-sdk/autotuning.md
Sushisource: I think it's nice to have that metric be standardized when it is available. It's another thing support can look at when available, etc.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: The only difference is whether the interface makes it seem like shutting down the worker is a normal thing for a slot supplier to do. I suggest that it should not be a normal thing to do.

If there's really a use case to shutdown the worker trying to reserve a slot from the slot supplier, I wonder if a mutable "reservation context" can be provided that exposes that as opposed to just general error handling shutting down the entire worker. In our SDKs we put in extra work to ensure that errors don't shutdown workers immediately, and now we're not only having a single error shutdown a worker, but we're exposing it to users as if it's normal error handling.

all-sdk/autotuning.md
Sushisource: Fixed is good

all-sdk/autotuning.md
cretz: Because it blurs the line between who holds what info. If you want to provide contextual info on invocation of reserve/release, ok, but to have a slot supplier and our system maintain counts and such seems like there's a bit of redudancy and duplication. If given the option between potentially useful and simple initial interface, I think we should err on the side of the latter initially.

all-sdk/autotuning.md
cretz: I think utilities that help code reuse may make sense, but having every interface implement something only a few need just to show a metric that some may not have seems unnecessary. Just makes for larger implementation surface.

Quinn-With-Two-Ns: I think metric emitting should be owned by the slot supplier itself not based on what interfaces it implements

all-sdk/autotuning.md
cretz: Not that easy. We are now going to clone/serialize every slot in use across the FFI boundary for every lang side that calls it. And to be thread-safe, even if not across a lang boundary, we're probably going to have to copy a snapshot of this info upon request. I do see how it is useful info, I'm just worried if we're assuming people need more data than they need today. Even with middleground, if you got rid of the vectors of slot infos, you'd just have the same counts a user could easily keep track of.

all-sdk/autotuning.md
Sushisource: Yeah, I agree. I just didn't want to have this get cluttered with the details of that.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: :+1: FWIW, I may still support an error returning from this if we think that's better ergonomics, but we should just log (and backoff?) or something. With Java, Python, and .NET we're going to have to specify thrown exception handling anyways regardless of whether it's on the interface.

all-sdk/autotuning.md
Sushisource: >We shouldn't need to track it outside of the slot supplier.

We must: "used slots" ~= "in-flight tasks". In flight tasks must be tracked for obvious reasons. These are stored today as your maps of `workflow id -> state about current task/workflow/etc` and `task token -> state about activity task` etc.

These methods would provide a read-only view into some lens on that tracked data.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,431 @@
...
+  /// task is provided. Will not be called for an already-cached workflow who is receiving a new task.
+  /// 
+  /// Because the number of available slots must be <= the number of workflows cached, if this returns false
+  /// when there are no idle workflows in the cache (IE: All other outstanding slots are in use), we will buffer the
+  /// task and wait for another to complete so we can evict it and make room for the new one.
+  fn can_allow_workflow(&self, slots_info: &WorkflowSlotsInfo, new_task: &WorkflowSlotInfo) -> bool;
+  /// Called when deciding what workflow should be evicted from the cache. May return the run id of a workflow to evict.
+  /// If None, or an invalid run id is returned, the default LRU eviction / strategy is used. The id returned must
+  /// be one of the workflows provided in `evictable` (the ones for which there is no actively processing WFT).
+  fn eviction_hint(&self, evictable: &[WorkflowSlotInfo]) -> Option<&str>;
```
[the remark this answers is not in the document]
Sushisource: Yeah, the point exactly is to allow them to control the caching "algorithm" without having to actually store the data.

They could pick LRU, or they could pick smallest history first like in @Quinn-With-Two-Ns 's example, or whatever they want.

They can simply return `None` here and we'll use LRU

all-sdk/autotuning.md
Sushisource: Right, but you mean it's cancelled by the caller, right? That explains some of the confusion here since in Rust you'd just drop the future and therefore never see the result.

That's why I didn't have a bool. If the _implementer_ would return false, IMO they should just keep blocking. So, even in Go/.NET with a stop channel or cancel token, we could also have this return void since in those situations if the caller cancels the call before it's finished, they should also just ignore the result.

all-sdk/autotuning.md
cretz: Sure in flight tasks must be tracked by us, I just question whether the user needs to be able to access them to make their decisions. And of course there is the cost of copying that info upon access (or otherwise synchronizing access).

Also arguably this is general purpose per-worker stat info request unrelated to this project unless we think users of this project need this general purpose info. I wonder if we should have a concept of "get status" on workers and pass the worker that is asking as part of each call.

all-sdk/autotuning.md
Quinn-With-Two-Ns: Ignoring the result is a race condition since it is possible to receive a slot even if tried to cancel the call. You need to know if it suceeded/failed

all-sdk/autotuning.md
Sushisource: Note that the "default implementations" section does make some reference to those.

all-sdk/autotuning.md
Quinn-With-Two-Ns: So Go and Java currently have no such map or info explicitly tracked by the SDK. I wouldn't want to provide that info from the SDK unless there was a strong user ask and there was no other way to implement the ask. I agree with Chad that this seems unrelated to auto tuning and maybe should be a separate proposal

Sushisource: It must exist at least partially in some form somewhere, otherwise how are the workflow/activity contexts supplying the various bits of information they do?

I agree that this could be some separate generic worker info call.

So, anyway if this is hard to provide in Go/Java (I can't imagine it is really, since I'm not thinking that a huge amount of stuff is actually going to end up in these structs, activity/wf type is the primary initial motivation). 

That said, that can also argue the other way that it's easy for the implementer to track themselves. So, that in mind sure, I'll take this out and we can bring it back later if it becomes important.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Quinn-With-Two-Ns: So I am advocating to maintain compatibility 

>Today if, before I make a wft completed call to server, I reserve a slot for eager activity, does worker_task_slots_available show it as no longer available?

worker_task_slots_available  would show as available

all-sdk/autotuning.md
Sushisource: > I think metric emitting should be owned by the slot supplier itself not based on what interfaces it implements

The supplier is free to emit whatever it likes - there just happen to be a couple known ones that we already emit, and I don't see why we wouldn't want to always emit those under a consistent name if the user happens to be able to provide it.

all-sdk/autotuning.md
Sushisource: Why not? It's a nice thing to have automatically get a consistent name if you provide it, same argument as with total cache size. I'm gonna close this comment and the one on the API in favor of the discussion about max cache size since that's got more comments in it and they're the same discussion really.

all-sdk/autotuning.md
Sushisource: I'm OK with that but curious to hear why you think so.

I made it this way because I liked the idea of the user explicitly combining the workflow things into one piece and providing that, since it makes it clear they are somewhat related to each other, but it's not the most important reason.

all-sdk/autotuning.md
Quinn-With-Two-Ns: >It must exist at least partially in some form somewhere, otherwise how are the workflow/activity contexts supplying the various bits of information they do?

It doesn't exist in any centralized map

all-sdk/autotuning.md
Sushisource: I was thinking that, but, I figured since you're waiting on that call to resolve you would either have seen it resolve before you issued the cancel or not, and if it wasn't resolved before that means to the implementer that is *must* not count as having been reserved

But... I can see how that is difficult to guarantee all the time.

So anyway, tldr on all this yes I agree it needs to be like that in those langs. In Rust it can just return `()` and the future can be dropped if the reservation is no longer desired and that's easy to do race-free. I'll add comments to that effect.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Sushisource: Yeah I'm inclined to keep the errors and just log them as warnings. That way everything will at least end up in the right place logging wise. Users can still call explicit shutdown if they want.

Sushisource:
OK I resolved a bunch of comments that I addressed, feel free to start a new thread if there's something that wasn't. I just needed to reduce the visual clutter a bit.

Main bigger questions that are still open I think:

* Should the cache sizer be an actual cache (I'm pretty strongly inclined no here, but, we could make it a facade for one)
* Providing a context object to reserve/try reserve. Pretty sure we need one at least to start with with `task_queue` / `worker_id` etc, but, do we feel that we should also put a `shutdown()` on it, or anything else?
* Should the traits have optionally-implementable methods for the things that would emit known metrics like max cache size and total slots?

Also could set up a meeting for these, but I want to give any other people who are interested some time to add comments too

all-sdk/autotuning.md
Quinn-With-Two-Ns: > that we already emit

We don't already emit max slots in any SDK do we?

> see why we wouldn't want to always emit those under a consistent name if the user happens to be able to provide it.

 Consider I share a slot provider that implements this interface across 3 workers because I want to restrict the total number of slots used across these 3 workers bounded. Now since your atomically emitting the metric instead of letting the slot supplier do it you are emitting 3x the value. I think we should keep slot supplier metrics in the slot supplier and add utilities to help code reuse if we see any.

Sushisource: > We don't already emit max slots in any SDK do we?
Good point lol, we don't.

The example with sharing implementation confusion is legit. Fair enough, we can put these metrics behind functions a user is welcome to call if they provide the same functionality.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: > worker_task_slots_available would show as available

Ok, my mistake. I thought the simple act of reserving for eager today affected the metric. Then nevermind, I agree then reservation and usage can remain separate. (can consider thread about this resolved)

all-sdk/autotuning.md
cretz: I would like to not emit a metric that doesn't have a valid value. What would you emit for available slot metric for implementations that have no concept of available slots? No integer is acceptable here. Slot suppliers should own their metrics, and there is only one slot supplier we will provide where available slots makes sense (the default one).

all-sdk/autotuning.md
cretz: This may be language specific, but many may not care about cache size in their impl.

I'm ok keeping them together, but it becomes harder to justify the separate interfaces then if they must always both be implemented anyways. So now I'm wondering if we want to change the naming a bit from `SlotSupplier` to a more general purpose `Limiter`, `Tuner`, `Controller`, `TaskManager`, etc. So:

* `ActivitySlotSupplier` is instead `ActivityController` and has the slot supplier stuff
* `WorkflowSlotSupplierAndCacheManager` is instead `WorkflowController` and has slot supplier stuff and cache manager stuff

Basically this lifts the low-level notion of "slot supplying" and "cache managing" to a general purpose controller for limiting/controlling work. I think it will also be more clear to users to `controller = new ResourceBasedWorkerController()` + `myWorkerConfig.WorkflowController = controller`. They don't have to care what slots are or that the interface is limited to slot supplying only or whatever.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: > without having to actually store the data

But that means they have to have access to the cache to even know what's in there (or have their own cache of IDs).

> They could pick LRU

Can they? Do you call this interface every time the cache is _accessed_? Without full control of the caching accesses, one cannot do advanced implementation. I think we should either choose always-LRU or allow custom cache impls. I think it's too difficult to do this hybrid.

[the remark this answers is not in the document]
Sushisource: Ah yeah good point. Seemingly all we'd need to do though is add a `get` or `just_fetched` call, and that'd be sufficient. I don't want it to be a *real* cache implementation because actually storing things won't work well across the language boundary, plus just the intrinsic danger of them returning the wrong item, or losing things that shouldn't be lost, etc. With the facade / sizer even if they screw up we can fall back to something that still operates correctly.

all-sdk/autotuning.md
Sushisource: > but I think interceptors and data converters have equal if not more potential for introducing bugs.
This is true, but, also doesn't supply a reason to introduce _additional_ dangerous things. Those are necessary, but the user actually storing the cached data for us certainly isn't.

all-sdk/autotuning.md
Quinn-With-Two-Ns: Not a fan of combining slot and cache management. Slots are per worker, where the cache is per process in Go and per worker factory in Java. I would also like to leave the door open to how we map the cache to workers in the SDK ideally so we can have per worker cache in Go if desired.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: Right, if we don't want a real cache impl, let's not even let them choose which thing to evict. Let's keep that complication out.

Sushisource:
Resolutions to the above:

* Keep cache sizer to just the "should_allow" kind of function to start with
* Yes, provide task queue & build id (if set) initially
* No. Have implementers report the metric. Provide a helper for easy access to the well-defined metrics.

all-sdk/autotuning.md
```diff
@@ -93,77 +93,70 @@ trait SlotSupplier {
...
-  // ... Possibly also metric information
-}
-struct ActivitySlotsInfo {
-  used_slots: Vec<ActivitySlotInfo>,
+/// This trait lets implementors obtain other information that might be relevant to their decision on whether to hand
+/// out a slot. It's a trait rather than a struct because the most up-to-date information should be available even
+/// if waiting for some time in the blocking version of `reserve_slot`.
+pub trait SlotReservationContext<SK: SlotKind>: Send + Sync {
+  /// Returns information about currently in-use slots
+  fn used_slots(&self) -> &[SK::Info];
```
Quinn-With-Two-Ns, line 120: Shouldn't the slot supplier know this?

all-sdk/autotuning.md
```diff
@@ -93,77 +93,70 @@ trait SlotSupplier {
...
+  
+  // ... anything else users end up wanting here
 }
-struct LocalActivitySlotsInfo {
-  used_slots: Vec<LocalActivitySlotInfo>,
+
+pub enum SlotReleaseReason {
```
Quinn-With-Two-Ns, line 125: If the worker is shutting down is that an `Error` or `NeverUsed`?

all-sdk/autotuning.md
Sushisource: It *can* know it, but, I don't really see a reason to force the implementor to track it themselves when we already have to track it ourselves at least in some capacity as discussed here: https://github.com/temporalio/proposals/pull/88#discussion_r1502691203

Certainly in Core this made dramatically more sense to provide automatically because the `SlotSupplier` gets wrapped by a struct in Core that has to track all this stuff anyway, and I anticipate creating a similar wrapper in Go/Java.

If it turns out to be a huge pain to provide as I'm working through Java, it can be left off - but I imagine it's going to be pretty straightforward to do.

all-sdk/autotuning.md
Sushisource: I would say `NeverUsed`, since if it _is_ used, then we're waiting until that task completes before finishing shutdown. (or, if we're not, then it'd be an `Error`... but we should be waiting for tasks to finish)

all-sdk/autotuning.md
Quinn-With-Two-Ns: I think just because we can provided something doesn't mean we should. We can always add information based on user feadback, it is much harder to remove. My main concern is the adding the overhead of keeping track of all this data getting updated from potential 10,000s of goroutines in Go when a user doesn't care about the information.

cretz: Same, I want the lightest API surface needed to do an an acceptable job (may be only slot kind and task queue name atm)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,452 @@
...
+is provided. These guarantees can't be compile-time enforced across the language boundary, but are at least useful after
+runtime validation.
+
+```rust
+trait SlotSupplier {
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available. In languages with explicit cancel mechanisms, this should be cancellable and
+  /// return a boolean indicating whether a slot was actually obtained or not. In Rust, the future can simply
+  /// be dropped if the reservation is no longer desired.
+  async fn reserve_slot(&self, ctx: &dyn SlotReservationContext<Self::SlotKind>);
```
cretz, line 96: Can we add a cancellation mechanism to this? (I forget if Rust implicitly supports cancellation of async like Python, but this at least needs to be called out for other langs)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,452 @@
...
+pub struct RustToLangBridgeSlotSupplier {
+  // ... store callbacks etc
+}
+
+impl<T: SlotKind> SlotSupplier<SlotKind=T> for RustToLangBridgeSlotSupplier {
+  // ... call appropriate callbacks, convert data types, etc
+}
+```
+
+## A neat example
```
cretz, line 331: This is neat, many of our users keep asking how they can limit activities, so being able to acquire a zookeeper/etcd lock for a specific activity types is helpful. But we're gonna have to provide helpers if we want to advertise support for this, it's too complicated for most users.

all-sdk/autotuning.md
Sushisource: The comment addresses exactly that. There's no need for an explicit cancel in Rust.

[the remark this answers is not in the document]
Sushisource: This is what I originally had, and is what I have in Rust because internally I store a struct that auto-releases on drop.

As for user data though, if we're saying elsewhere we'd prefer to not have something until there's a demonstrated need for it, I think that's probably even more applicable here. I don't know what the immediate use case is for this, but I do have known use cases for the `used_slots` on context.

all-sdk/autotuning.md
Quinn-With-Two-Ns: Fine with me, might be worth commenting so we are consistent across SDKs.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Sushisource: Any language with real destructors, yeah, definitely.

all-sdk/autotuning.md
cretz: There's no question that the information may be valuable, it's who is responsible for it. But I don't have a strong opinion here.

all-sdk/autotuning.md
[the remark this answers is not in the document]
cretz: > As for user data though, if we're saying elsewhere we'd prefer to not have something until there's a demonstrated need for it, I think that's probably even more applicable here

Less applicable, because this is future proofing. We cannot add a return type to this easily later. The immediate use case would be like etcd lease info. But just an opaque object that we give back to them is nice.

all-sdk/autotuning.md
Quinn-With-Two-Ns: If users need to know information about what slots their slot provider handed out they should record that themselves since if they need to record any custom information they will have to handle that themselves anyway

Sushisource: Well, sure, but this isn't custom information - it's information that's extracted from the task which we'll be storing anyway. It's just making it available to them rather than asking them to do that work themselves. I don't think there's any strong reason they should record it versus us. It's just making their lives easier for something that's very likely to come up in many or most implementations.

all-sdk/autotuning.md
```diff
@@ -0,0 +1,452 @@
...
+  /// Tries to immediately reserve a slot, returning true if a slot is available.
+  fn try_reserve_slot(&self, ctx: &dyn SlotReservationContext<Self::SlotKind>) -> bool;
+
+  /// Marks a slot as actually now being used. This is separate from reserving one because the pollers need to
+  /// reserve a slot before they have actually obtained work from server. Once that task is obtained (and validated)
+  /// then the slot can actually be used to work on the task.
+  /// 
+  /// Users' implementation of this can choose to emit metrics, or otherwise leverage the information provided by the
+  /// `info` parameter to be better able to make future decisions about whether a slot should be handed out.
+  fn mark_slot_used(&self, info: <Self::SlotKind as SlotKind>::Info<'_>);
```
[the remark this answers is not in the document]
cretz: Hrmm, yeah I first thought the "neat example" blocked in this call, but I see it doesn't. It means slot reservation may be unnecessarily expensive for things that never actually use the slot. Basically it turns remote semaphores to be based on poll instead of usage. But this is just a product of how we obtain work from server. We should at least document that people should not block in "mark used" since we don't give them async to do so.

all-sdk/autotuning.md
Sushisource: The resource based one does need it. It needs to know how many slots are used since you need to know that you're not going to hand out more slots than there is room in the cache (or even just to define a configurable hard limit).

This is zero cost in Core, for example. The info is all tracked already. Fetching system resources like memory usage for that implementation is dramatically more expensive for example.

We can benchmark... but I feel that's a little odd here given there are virtually zero changes we do actually benchmark and I don't see any reason to believe this would add substantial overhead compared to anything else we do. I mean it's literally just stuffing some things (which are likely just references to objects that already exist elsewhere) in a collection, which isn't very expensive to do.

Quinn-With-Two-Ns: >The resource based one does need it. It needs to know how many slots are used since you need to know that you're not going to hand out more slots than there is room in the cache (or even just to define a configurable hard limit).

The resource based one just needs to know the number it doesn't need a whole map of every slot and details about it.

> but I feel that's a little odd here given there are virtually zero changes we do actually benchmark

I get for Core it is zero cost, but for Go and Java this is new. I thought in slack we already talked about benchmarking these changes so I didn't even think this was a new requirement? I assumed as part of benchmarking we would test the old SDK implementation with the new fixed slot supplier implementation?

Sushisource: Yeah, of course, but isolating providing this particular function will require it's own independent round of testing - which is fine, but further than we have gone before for any micro-benchmark kind of thing.

>The resource based one just needs to know the number it doesn't need a whole map of every slot and details about it.
Yep - but this is more generally useful and (I would bet quite readily) not meaningfully more costly.

In any case, none of this is really a hard blocker either way on this proposal. We need the context object and it'll have basic stuff like `get_task_queue`. Whether or not this specific function lives on it we can decide after seeing how much effort and overhead it costs, though I'm pretty sure those are both small.

all-sdk/autotuning.md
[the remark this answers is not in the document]
Sushisource: I can add it back. I'm just a little worried about how extensive of changes that might result in in Go/Java where there is no permit object like in Core. We'd have to track it everywhere like Core does, which ultimately is nice, but could be quite a lot of code churn.

I think any kind of uniquely identifying a particular slot/permit would have the same consequence. It's useful though, in the abstract.

@Quinn-With-Two-Ns what do you think?

all-sdk/autotuning.md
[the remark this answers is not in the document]
Sushisource: I feel pretty strongly that marking used can't ever block at all (and yeah I can mention that in the comment).

The reasoning is that it _happened_, and the implementer has no choice but to accept that fact. If that then means, later, that they take a long time to hand out the next permit that's acceptable, but they can't block on mark used because a) it gums things up, and b) they can't change anything anyway, it's too late.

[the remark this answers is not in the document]
cretz: Works for me (can consider this resolved but can leave discussion open for others if you want)

all-sdk/autotuning.md
```diff
@@ -0,0 +1,452 @@
...
+```rust
+trait SlotSupplier {
+  type SlotKind: SlotKind;
+  /// Blocks until a slot is available. In languages with explicit cancel mechanisms, this should be cancellable and
+  /// return a boolean indicating whether a slot was actually obtained or not. In Rust, the future can simply
+  /// be dropped if the reservation is no longer desired.
+  async fn reserve_slot(&self, ctx: &dyn SlotReservationContext<Self::SlotKind>);
+
+  /// Tries to immediately reserve a slot, returning true if a slot is available.
+  fn try_reserve_slot(&self, ctx: &dyn SlotReservationContext<Self::SlotKind>) -> bool;
```
cretz, line 99: If people are allowed to use remote tooling like zookeeper/etcd, this should be async

cretz (approved):
LGTM!

all-sdk/autotuning.md
```diff
@@ -0,0 +1,477 @@
...
+      self.lock_svc_client.release_lock();
+    }
+    self.per_worker_allowed.release();
+  }
+}
+```
+
+## Drawbacks
+
+* Gives users a big footgun. This is mitigated by the fact that most users are probably just going to use one of our
```
atihkin, line 401: Anything we can do to mitigate? For example, letting users fallback to a default implementation if something goes wrong?

Sushisource: Not really possible - there's not any way to distinguish between the user's implementation intentionally not handing out slots, or it not handing out slots because it's broken.

Quinn-With-Two-Ns (approved):

PR #89, opened 2024-04-23
Add Nexus APIs to the Temporal Go SDK

bergundy:
Part of the ongoing effort to integrate Nexus RPC into Temporal.

[Rendered](https://github.com/temporalio/proposals/blob/49b1fc6053d139b54b6f030249463715cd0d9570/nexus/sdk-go.md)

### DIFF
--- a/nexus/sdk-go.md
+++ b/nexus/sdk-go.md
@@ -238,37 +238,9 @@ SDK's selectors.
 > NOTE: In the future, as the Go SDK adds support for typed futures, we will add a strongly typed variant of this API.
 
 ```go
-package temporalnexus
-
-type Endpoint interface {
-	// contains filtered or unexported methods
-}
-
-// NewNamedEndpoint creates an with targeted by name as registered in the endpoint registry. If the endpoint is renamed
-// calls refering to the old name will fail.
-func NewNamedEndpoint(name string) Endpoint
-
-// START: Potential future
-
-// NewEndpointFromID creates an endpoint targeted by a stable endpoint ID from the endpoint registry.
-func NewEndpointFromID(id string) Endpoint
-
-// NewEndpointFromSelectors creates an endpoint that is resolved by applying the given selectors.
-// For example {"env" "prod", "region": "us-east-1"}.
-func NewEndpointFromSelectors(selectors map[string]string) Endpoint
-
-// END: Potential future
-```
-
-```go
 // NexusOperationOptions are options for starting a Nexus Operation from a Workflow.
 type NexusOperationOptions struct {
 	ScheduleToCloseTimeout time.Duration
-	// The endpoint to send this operation request to. Optional. If not provided the endpoint will be resolved in
-	// the worker based on the given worker options NexusEndpointResolver function.
-	// It is generally recommended to keep workflow code enviroment agnostic and only set this from a workflow in
-	// ambiguous cases.
-	Endpoint temporalnexus.Endpoint
 }
 
 // NexusOperationExecution is the result of [NexusOperationFuture.GetNexusOperationExecution].
@@ -288,9 +260,15 @@ type NexusOperationFuture interface {
 	GetNexusOperationExecution() Future
 }
 
-// ExecuteNexusOperation executes a Nexus Operation.
-// The operation argument can be a string, a [nexus.Operation] or a [nexus.OperationReference].
-func ExecuteNexusOperation(ctx Context, service string, operation any, input any, options NexusOperationOptions) NexusOperationFuture
+// NexusClient is a client for executing Nexus Operations from a workflow.
+type NexusClient interface {
+	// ExecuteOperation executes a Nexus Operation.
+	// The operation argument can be a string, a [nexus.Operation] or a [nexus.OperationReference].
+	ExecuteOperation(ctx Context, operation any, input any, options NexusOperationOptions) NexusOperationFuture
+}
+
+// Create a [NexusClient] from a service name and an endpoint name.
+func NewNexusClient(service, endpoint string) NexusClient
 ```
 
 **Usage**:
@@ -301,15 +279,12 @@ import (
 	"time"
 
 	"go.temporal.io/sdk/workflow"
-	"go.temporal.io/sdk/temporalnexus"
 )
 
 func MyCallerWorkflow(ctx workflow.Context) (MyOutput, error) {
-	fut := workflow.ExecuteNexusOperation(ctx, "payments", "start-transaction", MyInput{ID: "tx-deadbeef"}, workflow.NexusOperationOptions{
+	client := workflow.NewNexusClient("payments", "prod-payments")
+	fut := client.ExecuteOperation(ctx, "start-transaction", MyInput{ID: "tx-deadbeef"}, workflow.NexusOperationOptions{
 		ScheduleToCloseTimeout: time.Hour,
-		// Optional, if not provided the endpoint will be resolved in the worker based on the given worker
-		// options.
-		Endpoint: temporalnexus.NewNamedEndpoint("prod-payments"),
 	})
 	var exec workflow.NexusOperationExecution
 	_ = fut.GetNexusOperationExecution().Get(ctx, &exec)
@@ -321,22 +296,6 @@ func MyCallerWorkflow(ctx workflow.Context) (MyOutput, error) {
 
 > NOTE: To cancel a Nexus Operation, cancel the context used to execute it.
 
-### Endpoint Registration in the Worker
-
-To avoid injecting configuration into the workflow - which is inherently unsafe and could break determinism - the SDK
-provides a worker option to resolve a service to an endpoint.
-
-> NOTE: This could also be implemented in an interceptor but for consistency with all SDKs, including ones where
-> interceptors run in the sandboxed workflow environment, this functionality has been extracted out to the worker.
-
-```go
-opts := worker.Options{
-	NexusEndpointResolver: func(ctx workflow.Context, service string) temporalnexus.Endpoint {
-		return temporalnexus.NewNamedEndpoint("prod-" + service)
-	}
-}
-```
-
 ### Interceptors
 
 For now we'll only intercept outbound APIs from a workflow, extending the `WorkflowOutboundInterceptor` interface.
@@ -344,7 +303,7 @@ More interceptors are likely to come later.
 
 ```go
 type WorkflowOutboundInterceptor interface {
-	ScheduleNexusOperation(ctx Context, service, operation string, input any, options NexusOperationOptions) NexusOperationFuture
+	NewNexusClient(service, endpoint string) workflow.NexusClient
 	RequestCancelNexusOperation(ctx Context, service, operation, id string, options nexus.CancelOperationOptions) error
 }
 ```
--- a/nexus/sdk-go.md
+++ b/nexus/sdk-go.md
@@ -21,6 +21,9 @@ Temporal is leveraing the Nexus RPC protocol to facilitate calling across namesp
 This proposal discusses exposing Nexus concepts in the Temporal Go SDK, reusing definitions from the Nexus Go SDK.
 
 > NOTE: The proposed APIs will all be marked experimental until we get user feedback and we feel they've matured enough.
+>
+> Due to limitation in the SDK package structure, some of the types shown in this proposal are aliases to types defined
+> in the `internal` package. We define them inline here for brevity.
 
 ## Temporal Operations
 
@@ -193,10 +196,10 @@ opStartTransactionAlt4 := temporalnexus.NewWorkflowRunOperationWithOptions(
 	})
 ```
 
-### Register an Operation with a Worker
+### Register Operations As Part of a Service with a Worker
 
-We define a `NexusOperationRegistry` interface with a single method `RegisterNexusOperation` and embed it in the
-`Registry` interface:
+We define a `NexusServiceRegistry` interface with a single method `RegisterNexusService` and embed it in the `Registry`
+interface:
 
 ```go
 import "github.com/nexus-rpc/sdk-go/nexus"
@@ -204,21 +207,27 @@ import "github.com/nexus-rpc/sdk-go/nexus"
 Registry interface {
 	WorkflowRegistry
 	ActivityRegistry
-	NexusOperationRegistry // <-- New
+	NexusServiceRegistry // <-- New
 }
 
-NexusOperationRegistry interface {
-	// RegisterNexusOperation registers an operation with a worker. Panics if an operation with the same name has
-	// already been registered on this worker or if the worker has already been started. A worker will only poll for
-	// Nexus tasks if any operations are registered on it.
-	RegisterNexusOperation(op nexus.RegisterableOperation)
+NexusServiceRegistry interface {
+	// RegisterNexusService registers a service with a worker. Panics if a service with the same name has already
+	// been registered on this worker or if the worker has already been started. A worker will only poll for
+	// Nexus tasks if any services are registered on it.
+	RegisterNexusService(service *nexus.Service)
 }
 ```
 
 **Usage**:
 
 ```go
-myWorker.RegisterNexusOperation(myOperation)
+service := nexus.NewService("payments")
+err := service.Register(myOperation, myOtherOperation)
+if err != nil {
+	panic(err)
+}
+
+myWorker.RegisterNexusService(service)
 ```
 
 ## Starting an Operation from a Workflow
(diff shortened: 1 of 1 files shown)


nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+	WorkflowRegistry
+	ActivityRegistry
+	NexusOperationRegistry // <-- New
+}
+
+NexusOperationRegistry interface {
+	// RegisterNexusOperation registers an operation with a worker. Panics if an operation with the same name has
+	// already been registered on this worker or if the worker has already been started. A worker will only poll for
+	// Nexus tasks if any operations are registered on it.
+	RegisterNexusOperation(op nexus.RegisterableOperation)
```
mfateev, line 212: I recommend registering multiple Nexus endpoints with the same worker.

bergundy: I agree we'll want this but should we do it now or start simple? I anticipate this being the common case.

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+import (
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"go.temporal.io/sdk/client"
+)
+
+// NewWorkflowRunOperation map an operation to a workflow run.
+func NewWorkflowRunOperation[I, O any](
+	name string,
+	workflow func(internal.Context, I) (O, error),
+	getOptions func(context.Context, I, nexus.StartOperationOptions) (client.StartWorkflowOptions, error),
```
cretz, line 86: Can this param be nil?

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+) nexus.Operation[I, O]
+
+type WorkflowRunOperationOptions[I, O any] struct {
+	// Operation name.
+	Name string
+	// Workflow function to map this operation to. The operation input maps directly to workflow input.
+	// The workflow name is resolved as it would when using this function in client.ExecuteOperation.
+	// OptionsGetter must be provided when setting this option. Mutually exclusive with Handler.
+	Workflow func(internal.Context, I) (O, error)
```
cretz, line 95:
```suggestion
	Workflow func(workflow.Context, I) (O, error)
```

I think we may be able to avoid type aliases in the `temporalnexus` package, but understood if we cannot

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+	// Options for starting the workflow. Must be set if Workflow is set. Mutually exclusive with Handler.
+	OptionsGetter func(context.Context, I, nexus.StartOperationOptions) (client.StartWorkflowOptions, error)
```
cretz, line 97:
```suggestion
	GetStartWorkflowOptions func(context.Context, I, nexus.StartOperationOptions) (client.StartWorkflowOptions, error)
```

`GetX` is a bit clearer in Go than `Getter`. You can remove the "Start" and/or the "Workflow" part if you want though.

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+
+type WorkflowHandle[T any] interface {
+	ID() string
+	RunID() string
+}
+
+// ExecuteWorkflow starts a workflow run, linking the execution chain to a Nexus operation (subsequent runs started from
+// continue-as-new and retries).
+// Automatically propagates the callback and request ID from the nexus options to the workflow.
+func ExecuteWorkflow[I, O any, WF func(internal.Context, I) (O, error)](
```
cretz, line 113: The purpose of this is a bit confusing, can you clarify this is meant to be used in a handler?

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+		},
+	})
+
+// Alternative 3 - start a workflow with alternative inputs.
+opStartTransactionAlt3 := temporalnexus.NewWorkflowRunOperationWithOptions(
+	temporalnexus.WorkflowRunOperationOptions[MyInput, MyOutput]{
+		Name: "start-transaction",
+		Handler: func(ctx context.Context, input MyInput, opts nexus.StartOperationOptions) (temporalnexus.WorkflowHandle[MyOutput], error) {
+			// Workflows started with this API must take a single input and return single output.
+			return temporalnexus.ExecuteWorkflow(ctx, client.StartWorkflowOptions{
```
cretz, line 174: This doesn't match the signature of `ExecuteWorkflow` put above. Where's the `nexus.StartOperationOptions` parameter?

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/temporalnexus"
+)
+
+func MyHandlerWorkflow(workflow.Context, MyInput) (MyOutput, error)
+func MyHandlerWorkflowWithAlternativeInput(workflow.Context, MyWorkflowInput) (MyOutput, error)
+
+// Alternative 1 - shortest form, for workflows that have input and outputs that map 1:1 with the operation's I/O.
+opStartTransactionAlt1 := temporalnexus.NewWorkflowRunOperation(
```
cretz, line 147: To clarify, the only reason this exists is for people not wanting to make an options object? But we still make them make a full workflow start options callback? Did they really save much? Maybe instead you should have a `NewWorkflowRunOperationOptions` that accepts these fields instead of top-level overloads.

nexus/sdk-go.md
cretz: Can you clarify? Are you saying this should accept a collection instead of just requiring users to call it multiple times?

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+	// For synchronous operations, this will be resolved at the same as the containing [NexusOperationFuture]. For
+	// asynchronous operations, this future is resolved independently.
+	// If the operation is unsuccessful, this future will contain the same error as the [NexusOperationFuture].
+	// Use this method to extract the Operation ID of an asynchronous operation. OperationID will be empty for
+	// synchronous operations.
+	GetNexusOperationExecution() Future
+}
+
+// ExecuteNexusOperation executes a Nexus Operation.
+func ExecuteNexusOperation(ctx Context, service string, operation string, input any, options NexusOperationOptions) NexusOperationFuture
```
cretz, line 255:
```suggestion
func ExecuteNexusOperation(ctx Context, service string, operation any, input any, options NexusOperationOptions) NexusOperationFuture
```

I should be able to provide an operation I defined elsewhere

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+update, list workflows, etc...
+
+```go
+package temporalnexus
+
+import "github.com/nexus-rpc/sdk-go/nexus"
+
+func NewSyncOperation[I any, O any](
+	name string,
+	handler func(context.Context, client.Client, I, nexus.StartOperationOptions) (O, error),
```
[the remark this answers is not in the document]
bergundy: I don't think we should over complicate the API with future compat concerns.
I don't even think that `WorkflowQuery` is an option we want here, I'd make a `NewWorkflowQueryOperation` instead if we want to go that direction but in the previous proposal, you were the one advocating for a single `SyncOperation` concept and giving users a client.

I'm open to renaming this to `NewClientOperation` or `NewSyncClientOperation`.

nexus/sdk-go.md
```diff
@@ -0,0 +1,352 @@
...
+
+```go
+import (
+	"github.com/nexus-rpc/sdk-go/nexus"
+	"go.temporal.io/sdk/client"
+	"go.temporal.io/sdk/temporalnexus"
+)
+
+opGetStatus := temporalnexus.NewSyncOperation("get-status", func(ctx context.Context, c client.Client, id string, opts nexus.StartOperationOptions) (int, error) {
+	res, err := c.QueryWorkflow(ctx, id, "", "some-query", nil)
```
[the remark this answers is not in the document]
bergundy: Yes, operations have a limited pre-defined interface.

nexus/sdk-go.md
bergundy: No, you need at least a workflow ID to start a workflow.

nexus/sdk-go.md
bergundy: Sure, I'd prefer `GetOptions` then. I liked `Getter` because it follows the same form as `Handler` but I'm not attached to the name.

nexus/sdk-go.md
bergundy: Yes.

nexus/sdk-go.md
bergundy: Yes, that's the reason. I believe this will be the preferred way to expose workflows as operations so I made this shorthand form.

The reason I don't like the options struct is that you have to explicitly type the type params but yes a `NewWorkflowRunOperationOptions` also solves that. But keep in mind that even `NewWorkflowRunOperationOptions` may require overloads to support handler and (in the future) a result mapper.

nexus/sdk-go.md
bergundy: Oh I added that arg last minute and forgot to update the example.
Thanks for pointing it out.
I was debating taking the options from the context instead of forcing users to pass those in but I think the signature with the operation options is a better approach.

nexus/sdk-go.md
bergundy: Good callout, thanks!

nexus/sdk-go.md
bergundy: Looks like it's possible. Thanks!

bergundy: Fixed.

nexus/sdk-go.md
bergundy: Fixed.

nexus/sdk-go.md
bergundy: Fixed.

nexus/sdk-go.md
bergundy: Done.

nexus/sdk-go.md
bergundy: Changed to `GetOptions`.

nexus/sdk-go.md
bergundy: Here's a couple of alternative's I've posted in an internal thread:

The most flexible option is to accept a nexus.Handler in WorkerOptions and have the user construct it via:

```go
reg := &nexus.OperationRegistry{}
_ := reg.Register(op1, op2)
handler, _ = reg.NewHandler()
w := worker.New(client, "task-queue", worker.Options{NexusHandler: handler})
```

^ This allows users to customize the handler completely and intercept calls.

The second option is to accept the registry:

```go
reg := &nexus.OperationRegistry{}
_ := reg.Register(op1, op2)
w := worker.New(client, "task-queue", worker.Options{NexusOperationRegistry: reg})

```

Saves a line of code but loses some flexibility and we'll have to expose interceptors in the Go SDK.

With the first option, you get multiple services OOTB, you just need a "router" handler, which we can also add to the Nexus SDK.

Here's some more alternatives for registering multiple services:

```go
w.RegisterNexusService("service-name", reg)
// -- or --
w.RegisterNexusHandler(handler) // handle can contain a service "router"
// -- or --
w.RegisterNexusHandler("service-name", handler)
```

Using the handler concept gives the most flexibility.

Here are some relevant definitions for reference:

https://pkg.go.dev/github.com/nexus-rpc/sdk-go@v0.0.7/nexus#Handler
https://pkg.go.dev/github.com/nexus-rpc/sdk-go@v0.0.7/nexus#OperationRegistry
https://pkg.go.dev/github.com/nexus-rpc/sdk-go@v0.0.7/nexus#Operation
https://pkg.go.dev/github.com/nexus-rpc/sdk-go@v0.0.7/nexus#RegisterableOperation

mfateev: I like
```w.RegisterNexusService("service-name", reg)```
and
```w.RegisterNexusHandler("service-name", handler)```

Comment (bergundy) [reply nexus/sdk-go.md:214]: So you would want this to be the only option or should we _also_ support registering the individual operations?

Comment (bergundy) [reply nexus/sdk-go.md:214]: Something that I don't like about forcing specifying the service name is that it creates duplication.
You already have to tell the server to map a service to a namespace and task queue using the incoming service registry APIs.
Specifying the service name also complicates the service renaming process.

So I'd suggest having a way to register a handler or operations without forcing specifying the service name but still supporting backing multiple services with a single worker for these (what I assume are) less common use cases.

If we want a single API, the accepting a `Handler` is the most flexible option.

File: nexus/sdk-go.md (hunk 9 above)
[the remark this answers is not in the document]
Comment (cretz) [reply nexus/sdk-go.md:40]: > I don't think we should over complicate the API with future compat concerns.

I don't think it's overcomplicating. Can ignore most of my comment, but I would expect `NewSyncOperationWithOptions` to exist just like `NewWorkflowRunOperationWithOptions` does. (having said that, I wish there was only 1 call that took options always, but seems you're settled on 2).

File: nexus/sdk-go.md (hunk 2 above)
Comment (cretz) [reply nexus/sdk-go.md:87]: Can you add that this callback must be set by everyone in the docs? Usually people in Go don't expect to have to provide a workflow ID

File: nexus/sdk-go.md (hunk 7 above)
Comment (cretz) [reply nexus/sdk-go.md:149]: > The reason I don't like the options struct is that you have to explicitly type the type params

I don't think this is that bad of a thing in Go and I think we're adding extra methods do to this personal dislike. But I don't have a super strong opinion just so long as we have proper options-struct-based form we can grow on all of these.

File: nexus/sdk-go.md [hunk 11]
```diff
@@ -0,0 +1,355 @@
...
+
+type WorkflowRunOperationOptions[I, O any] struct {
+	// Operation name.
+	Name string
+	// Workflow function to map this operation to. The operation input maps directly to workflow input.
+	// The workflow name is resolved as it would when using this function in client.ExecuteOperation.
+	// GetOptions must be provided when setting this option. Mutually exclusive with Handler.
+	Workflow func(workflow.Context, I) (O, error)
+	// Options for starting the workflow. Must be set if Workflow is set. Mutually exclusive with Handler.
+	GetOptions func(context.Context, I, nexus.StartOperationOptions) (client.StartWorkflowOptions, error)
```
cretz, line 98: For others reading, the `Handler` term is commonly used in Go in places like `net/http.Server.Handler` and the `GetOptions` type of name is used in places like `crypto/tls.Config.GetCertificate` (even though it may feel inconsistent that it's not `GetHandler` or `OptionsGetter`)

nexus/sdk-go.md
cretz: To clarify, are we discussing accepting multiple different services? Or a single service but multiple different operations?

bergundy: Multiple services on the same task queue.

bergundy: I'm currently thinking we either keep `w.RegisterNexusOperation` assuming multiple services on the same task queue is going to be rarely used or drop this API completely to have a single registration method.

The most flexible approach is to allow registering a handler, either with `w.RegisterNexusHandler` or to pass in the handler in worker options.

For a single service registration would look like this:

```go
reg := &nexus.OperationRegistry{}
_ := reg.Register(op1, op2)
handler, _ = reg.NewHandler()
w.RegisterNexusHandler(handler)
```

For multiple services, we'd have:

```go
service1Reg := &nexus.OperationRegistry{}
_ = service1Reg.Register(op1, op2)
service1Handler, _ = reg.NewHandler()

service2Reg := &nexus.OperationRegistry{}
_ = service2Reg.Register(op3, op4)
service2Handler, _ = reg.NewHandler()

handler := temporalnexus.NewMuxHandler()
handler.RegisterService("svc1", service1Handler)
handler.RegisterService("svc2", service2Handler)

w.RegisterNexusHandler(handler)
```

mfateev: I don't think we want to support registering operations without associating them to a specific service.

bergundy:
Short update: We're going to add a "service" concept that's a grouping of operations and rename the current "service" concept to "endpoint".

This change has to start from the Nexus API spec and Go SDK and then come into this proposal.
I'm putting this one on hold until we have the prerequisites in place.

nexus/sdk-go.md
```diff
@@ -288,9 +260,15 @@ type NexusOperationFuture interface {
...
-func ExecuteNexusOperation(ctx Context, service string, operation any, input any, options NexusOperationOptions) NexusOperationFuture
+// NexusClient is a client for executing Nexus Operations from a workflow.
+type NexusClient interface {
+	// ExecuteOperation executes a Nexus Operation.
+	// The operation argument can be a string, a [nexus.Operation] or a [nexus.OperationReference].
+	ExecuteOperation(ctx Context, operation any, input any, options NexusOperationOptions) NexusOperationFuture
+}
+
+// Create a [NexusClient] from a service name and an endpoint name.
+func NewNexusClient(service, endpoint string) NexusClient
```
mfateev, line 271: Consider reversing the order to endpoint, service.

mfateev (approved):

nexus/sdk-go.md
```diff
@@ -0,0 +1,369 @@
...
+
+// NexusClient is a client for executing Nexus Operations from a workflow.
+type NexusClient interface {
+	// ExecuteOperation executes a Nexus Operation.
+	// The operation argument can be a string, a [nexus.Operation] or a [nexus.OperationReference].
+	ExecuteOperation(ctx Context, operation any, input any, options NexusOperationOptions) NexusOperationFuture
+}
+
+// Create a [NexusClient] from a service name and an endpoint name.
+func NewNexusClient(service, endpoint string) NexusClient
```
bergundy, line 271: Also considering adding context and an options struct to future proof this API.

nexus/sdk-go.md
bergundy: Hmm.. right, they'll need to if they want their operation to be idempotent. It should be a deterministic function from input to ID.

cretz (approved):
Approving now that registration is solved

PR #92, opened 2024-05-24
Ruby SDK - Phase 1 - Initial Comment Period

cretz:
# Summary

* This is the proposal for the first phase of the Ruby SDK.
* See it [rendered](https://github.com/cretz/temporal-proposals/blob/ruby-phase-1/ruby/sdk-phase-1.md).
* Much of this is inspired by the current [Temporal Ruby SDK](https://github.com/temporalio/sdk-ruby) (and its [proposals](https://github.com/temporalio/proposals/tree/master/ruby/previous) which were moved to a sub directory) and our other SDKs.

### DIFF
--- a/ruby/sdk-phase-1.md
+++ b/ruby/sdk-phase-1.md
@@ -53,6 +53,7 @@ These are the general requirements of the project
     library won't work with.
     * ❓ How hard do we want to work to avoid duplicate yard/RBS typing?
   * 💭 Why not Sorbet? Sorbet can come later or provided separately, but RBS seems to be the newer movement.
+  * 💭 Why have any typing? As a library we might as well help users where we can.
 * Everything is in the `Temporalio` module and released as a `temporalio` gem starting at 0.2.0
   * 💭 Why 0.2.0? 0.1.x versions already in use at https://rubygems.org/gems/temporalio.
 * Internal code will be in `Temporalio::Internal` module including bridge which is in `Temporalio::Internal::Bridge`
@@ -61,16 +62,16 @@ These are the general requirements of the project
 * Avoid cluttering the top-level `Temporalio` module with many small classes, but some are ok as we try to avoid a
   `Temporalio::Common` module
   * So `Temporalio::RetryPolicy` is preferred over `Temporalio::Common::RetryPolicy`.
-  * All search attribute things should be under `Temporalio::SearchAttribute` (e.g. `Temporalio::SearchAttribute::Key`)
+  * The search attribute collection is `Temporalio::SearchAttributes` and all things are under it, e.g.
+    `Temporalio::SearchAttributes::Key`.
+    * 💭 Why not `Temporalio::SearchAttribute` module with the collection underneath? The object has value on its own
+      so might as well make it a class.
+  * The base metric is `Temporalio::Metric`, and all metric things are under there, e.g. `Temporalio::Metric::Meter`.
+    * 💭 Why not `Temporalio::Metrics` module? The object has value on its own so might as well make it a class.
+   (e.g. `Temporalio::SearchAttribute::Key`)
     and all metric things under `Temporalio::Metric` (e.g. `Temporalio::Metric::Meter`).
-  * 💭 Why not top-level? Cluttering top-level module makes discoverability difficult and API docs unclear.
   * 💭 Why no `Temporalio::Common`? Ruby expects to have common things where they make sense instead of under a `Common`
     module.
-  * ❓ Do we want plural, e.g. `Temporalio::SearchAttributes` and `Temporalio::Metrics`? While there is no such thing as
-    a `SearchAttribute`, there is such thing as a `SearchAttributes` (i.e. search attributes collection). And there is
-    such thing as a `Metric` (i.e. base class for counter, gauge, and histogram). So it could be
-    `Temporalio::SearchAttributes` (the collection) and `Temporalio::Metric` (the base class), and they are not modules
-    but classes with the nested pieces underneath like we do elsewhere.
 * No compatibility with any existing Ruby SDK
   * Migration guides/docs may appear as needed.
   * 💭 Why? Need a clean slate to develop a high-quality SDK from first principles.
@@ -129,22 +130,27 @@ connection. The former has shortcut for creating the latter while it creates its
 ```ruby
 module Temporalio
   class Client
-    attr_reader :connection, :namespace
+    # This can be splatted into the constructor if one option needs to be
+    # changed. This is a Struct and not Data because we support Ruby 3.1.
+    Options = Struct.new(...)
 
     # Connect a client (i.e. Connection.new + Client.new)
-    def self.connect(target_host:, namespace: 'default', runtime: nil, ...)
+    def self.connect(target_host, namespace, runtime: nil, ...)
       ...
     end
 
-    # Create client with existing connection
-    def initialize(connection, namespace: 'default', ...)
+    # Create client with existing connection. This must be all keyword arguments
+    # and must match the Options struct.
+    def initialize(connection:, namespace:, ...)
       ...
     end
 
-    # Duped hash of keyword args of constructor
-    def options
+    # Duped options 
+    def dup_options
       ...
     end
+
+    # Several properties such as connection, namespace, workflow_service, etc
 
     # Returns workflow handle
     def start_workflow(
@@ -178,13 +184,15 @@ module Temporalio
 
   class Client
     class Connection
+      Options = Struct.new(...)
+
       # Perform connection
       def initialize(target_host:, runtime: nil, ...)
         ...
       end
 
-      # Duped hash of keyword args of constructor
-      def options
+      # Duped options 
+      def dup_options
         ...
       end
 
@@ -207,6 +215,10 @@ module Temporalio
 
       # Lots more
     end
+
+    class CloudOperationsClient
+      # Similar to client but for cloud ops
+    end
   end
 end
 ```
@@ -215,10 +227,12 @@ end
 * We expect most users and most samples to just use `Client.connect`.
   * 💭 Why not make users create a connection themselves instead of maintaining similar options sets across
     `Connection.initialize` and `Client.connect`? We have found most people only need a single client per connection.
-* Users that want raw service access need to use `client.connection.workflow_service` similar to .NET.
+* Users will have access to raw clients via `client.workflow_service` and similar
 * Workflow handles exist with all relevant calls.
-* Note that we prefer large sets of keyword arguments over hashes or options/config objects.
-  * 💭 Why? This is the Ruby preference and clearer to users.
+* Note that we use the kwargs constructor + options struct pattern so that a user can access duped options, change one,
+  and reconstruct via `Client.new(**my_new_options.to_h)`
+  * ❓Is there a better way to allow people to easily change an option? We didn't want a `with` pattern because, at
+    least for `Connection`, it requires a reconnect.
 * Must make clear in docs which things issue an RPC call (e.g. start workflow) vs ones that don't (e.g. get workflow
   handle).
 
@@ -235,17 +249,17 @@ class MyWorkflow < Temporalio::Workflow
   workflow_query_attr_reader :some_other_query
 
   def execute(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 
   workflow_signal
   def some_signal(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 
   workflow_query 'custom-query-name'
   def some_query(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 end
 ```
@@ -254,7 +268,7 @@ And client usage may look like:
 
 ```ruby
 # Connect client
-client = Client.connect('localhost:7233')
+client = Client.connect('localhost:7233', 'default')
 
 # Start workflow
 handle = client.start_workflow(
@@ -276,30 +290,38 @@ handle.signal(MyWorkflow.some_signal, 'some arg')
 
 Like other SDKs, the following classes will exist:
 
-* `Temporalio::Converter::DataConverter` - Accepts a payload converter (must be Ractor shareable), failure converter
+* `Temporalio::Converters::DataConverter` - Accepts a payload converter (must be Ractor shareable), failure converter
   (must be Ractor shareable), and a payload codec
-  * 💭 Why use `Temporalio::Converter` module instead of just `Temporalio::DataConverter`? There are many files that are
-    included in conversion and most regular users won't ever use them.
+  * 💭 Why use `Temporalio::Converters` module instead of just `Temporalio::DataConverter`? There are many files that
+    are included in conversion and most regular users won't ever use them so no need to clutter the top.
   * 💭 Why duplicate words instead of `Temporalio::Converter::Data`? `Temporalio::Converter` is not a base class and a
     `DataConverter` is a completely separate thing from, say, a `PayloadConverter`.
-  * ❓ Would we prefer `Temporalio::Conversion` as the module name?
-* `Temporalio::Converter::PayloadConverter` - Default composite payload implementation accepting a collection of
-  encoding payload converters accepted in constructor, with a default class attr that has a default set.
-  * Users are expected to extend this and override conversion methods as needed. Default instantiation of this is 0
-    encoding converters and a `NotImplementedError` if either is called.
-  * 💭 Why not have a separate base class instead of combining the composite one with the one to override? This was a
-    50/50 call, but most users will probably add an encoding impl instead of override at this level. This is admittedly
-    a slight deviation from other SDKs.
-* `Temporalio::Converter::PayloadConverter::Encoding` - Base unimplemented encoding converter
-  * ❓ Would we prefer `Temporalio::Converter::EncodingPayloadConverter`?
-* `Temporalio::Converter::PayloadConverter::JSON`, `Temporalio::Converter::PayloadConverter::JSONProto`, etc  - All
-  common encoding converters
-* `Temporalio::Converter::FailureConverter` - Default failure converter, can be overridden
-* `Temporalio::Converter::PayloadCodec` - Unimplemented base codec
-  * Important for implementers to ensure this is async capable if using async activities anywhere
-* `Temporalio::Converter::RawValue` - Can be used as way to move around unconverted payloads
+* `Temporalio::Converters::PayloadConverter` - Default base class for payload converters (only has two unimplemented
+  methods)
+  * Has a `default` class method with lazily created default that is composite of set of common encoding converters
+  * Note, Ruby has no type hinting, so there are no hints when converting from a payload to a value what you might want
+    to do beyond metadata on the payload itself (see the `JSONPlain` section next). We will try to make sure payload
+    conversion is done within the activity/workflow context of course.
+  * 💭 Why not make `Temporalio::Converters::PayloadConverter` a module and have
+    `Temporalio::Converters::PayloadConverter::Base` as the base class? This was a 50/50 call, but we are not taking the
+    pattern of `Base` around the repo for implementable classes.
+* `Temporalio::Converters::PayloadConverter::Composite` - Converter implementation that accepts a ollection of multiple
+  `Encoding` converters (see next top-level bullet)
+  * 💭 Why not just make the `PayloadConverter` be the composite impl and those overriding can just have no encoding
+    converters passed in? This was a 50/50 call, but it is clearer for users to have the base class just show them the
+    two `NotImplementedError` methods they must implement instead of the details of a composite converter.
+* `Temporalio::Converters::PayloadConverter::Encoding` - Base unimplemented encoding converter
+* `Temporalio::Converters::PayloadConverter::JSONPlain`, `Temporalio::Converter::PayloadConverter::JSONProtobuf`, etc -
+  All common encoding converters
+* `Temporalio::Converters::FailureConverter` - Default failure converter, can be overridden
+* `Temporalio::Converters::PayloadCodec` - Unimplemented base codec
+  * ❓How can we make this async capable? We need to be able to multiplex several calls to this concurrently, but we may
+    not be in an async-capable environment and/or don't know how implementations may want to multiplex, so we can't just
+    have a naive payloads-to-payloads implementation. We probably need to allow the payload to accept multiple "payload
+    sets" and return "payload sets", since each set may contain multiple payloads, but sets need to be independent.
+* `Temporalio::Converters::RawValue` - Can be used as way to move around unconverted payloads.
 
-#### Temporalio::Converter::PayloadConverter::JSON
+#### Temporalio::Converter::PayloadConverter::JSONPlain
 
 * Ruby has a `JSON` module in the standard library, but by default it's only primitives, arrays, hashes, etc and not
   classes.
@@ -355,10 +377,15 @@ end
   * ❓ Would we prefer activities as methods knowing they may not be able to be typed/referenced well? We can't really
     take the workflow signal-ish approach of those pseudo-decorators making class methods unless we say activities must
     be instance methods (since we would allow class methods as activities).
-* Activity classes are instantiated at registration time, not for each invocation.
-  * 💭 Why? Existing Temporal Ruby SDK does instantiate for each activity task, but this discourages use of shared
-    objects or state to be referenced. This does have the tradeoff our other SDKs have of requiring the function to be
-    thread safe.
+* If an activity class is provided to a worker, it is instantiated each attempt. But if an activity instance is provided
+  to a worker, execute is called on that instance.
+  * 💭 Why support every-attempt-instantiated activities? This is a reasonable expectation for Ruby things and is how
+    the current Temporal Ruby SDK operates. This allows users to use instance state without race concerns.
+  * 💭 Why support pre-instantiated activities? It is a very common need for SDKs to be able to support cross-attempt
+    state such as a DB client, and while Ruby users are unfortunately used to class-level/static attrs, it should not
+    be encouraged.
+  * ❓ Is it too confusing that the instance lifecycle is different depending on whether you pass in a class or an
+    instance?
 * `Temporalio::Activity::Context` is a thread/fiber-local context available to activities.
   * It has instance methods for info, heartbeat, etc.
   * It has a class method for `current` that returns from thread/fiber-local and class method shortcuts for all instance
@@ -392,7 +419,7 @@ module Temporalio
     end
 
     def initialize(
-      client,
+      client:,
       task_queue:,
       activities: [],
       activity_executors: nil,
@@ -417,21 +444,21 @@ end
 This phase only supports activity workers. Running a worker may look like:
 
 ```ruby
-require './my_activities_1.md'
-require './my_activities_2.md'
+require 'my_activities_1'
+require 'my_activities_2'
 
-client = Client.connect('localhost:7233')
+client = Client.connect('localhost:7233', 'default')
 
 # Create worker for task-queue-1
 worker1 = Temporalio::Worker.new(
-  client,
+  client: client,
   task_queue: 'task-queue-1',
   activities: [MyActivity1.new()]
 )
 
 # Create worker for task-queue-2
 worker2 = Temporalio::Worker.new(
-  client,
+  client: client,
   task_queue: 'task-queue-2',
   activities: [MyActivity2.new()]
 )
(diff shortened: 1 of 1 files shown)


ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+constraints, approach, and requirements changed enough to justify a clean slate and while many things may be leveraged,
+it never went beyond alpha anyways.
+
+Lots of approaches taken here are similar to recent Python and .NET SDKs.
+
+### Requirements
+
+These are the general requirements of the project
+
+* Ruby >= 3.1
```
bergundy, line 32: Wondering what we've done to validate this is acceptable.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    and all metric things under `Temporalio::Metric` (e.g. `Temporalio::Metric::Meter`).
+  * 💭 Why not top-level? Cluttering top-level module makes discoverability difficult and API docs unclear.
+  * 💭 Why no `Temporalio::Common`? Ruby expects to have common things where they make sense instead of under a `Common`
+    module.
+  * ❓ To we want plural, e.g. `Temporalio::SearchAttributes` and `Temporalio::Metrics`? While there is no such thing as
+    a `SearchAttribute`, there is such thing as a `SearchAttributes` (i.e. search attributes collection). And there is
+    such thing as a `Metric` (i.e. base class for counter, gauge, and histogram). So it could be
+    `Temporalio::SearchAttributes` (the collection) and `Temporalio::Metric` (the base class), and they are not modules
+    but classes with the nested pieces underneath like we do elsewhere.
+* No compatibility with any existing Ruby SDK
```
bergundy, line 74:
```suggestion
* No compatibility with _any_ existing Ruby SDK
```

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  * Client (technically connection) accepts a runtime and if unset it defaults to `default`
+* Initialization of the runtime class starts a Ruby thread that just runs the command loop from Rust.
+  * 💭 Why does a runtime need to eat a Ruby thread? The Rust side needs to have a way to make Ruby calls, and only Ruby
+    can start a Ruby thread. So on the Rust side there is essentially a loop with a channel waiting without holding the
+    GVL with things to run on the Ruby side.
+  * 💭 Why do this in a constructor? That is the Ruby way, a constructed object is ready for use.
+  * 💭 Why only a single thread? We expect callbacks to Ruby from Rust to be short (hopefully only a block invocation
+    that is a `Queue.push`) so we don't expect making all of these run serial is an issue. However if it does become an
+    issue, we can use multiple threads and a Rust approach which allows multiple channel consumers.
+* Users can create their own runtime with their own telemetry options and such and pass that to clients.
```
bergundy, line 120: I've never seen the need to create more than a single runtime in our SDKs. Seems like we can avoid this extra API surface to expose accepting a runtime and just have all constructs reference a singleton.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    raise NoMethodError
+  end
+end
+```
+
+And client usage may look like:
+
+```ruby
+# Connect client
+client = Client.connect('localhost:7233')
```
bergundy, line 257: namespace defaults to `default`?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  * ❓ Would we prefer `Temporalio::Conversion` as the module name?
+* `Temporalio::Converter::PayloadConverter` - Default composite payload implementation accepting a collection of
+  encoding payload converters accepted in constructor, with a default class attr that has a default set.
+  * Users are expected to extend this and override conversion methods as needed. Default instantiation of this is 0
+    encoding converters and a `NotImplementedError` if either is called.
+  * 💭 Why not have a separate base class instead of combining the composite one with the one to override? This was a
+    50/50 call, but most users will probably add an encoding impl instead of override at this level. This is admittedly
+    a slight deviation from other SDKs.
+* `Temporalio::Converter::PayloadConverter::Encoding` - Base unimplemented encoding converter
+  * ❓ Would we prefer `Temporalio::Converter::EncodingPayloadConverter`?
```
bergundy, line 294: Don't know what the Ruby standard is here but as you know in our other SDKs we've chosen the longer name.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    be instance methods (since we would allow class methods as activities).
+* Activity classes are instantiated at registration time, not for each invocation.
+  * 💭 Why? Existing Temporal Ruby SDK does instantiate for each activity task, but this discourages use of shared
+    objects or state to be referenced. This does have the tradeoff our other SDKs have of requiring the function to be
+    thread safe.
+* `Temporalio::Activity::Context` is a thread/fiber-local context available to activities.
+  * It has instance methods for info, heartbeat, etc.
+  * It has a class method for `current` that returns from thread/fiber-local and class method shortcuts for all instance
+    methods.
+    * 💭 Why class methods equivalents? `Temporalio::Activity::Context.heartbeat` is easier to read than
```
bergundy, line 366: In other SDKs, `heartbeat`, `log`, `info`, etc... is on the `activity` module.
Why not do that here too? E.g:

```
Temporalio::Activity::Context.current

Temporalio::Activity::heartbeat
Temporalio::Activity::info
Temporalio::Activity::log
```

bergundy (approved):

ruby/sdk-phase-1.md
josh-berry: All older versions of Ruby are EOL; I don't think we need to support EOL software: https://www.ruby-lang.org/en/downloads/

cretz: Correct: https://en.wikipedia.org/wiki/History_of_Ruby. But all of this is still pending user feedback and so we'll likely be surveying the community on how acceptable they consider this.

ruby/sdk-phase-1.md
cretz: They will reference a lazy global which is basically a singleton. This is deemed acceptable in other core-based SDKs and so should be here too, no need to make Ruby special.

ruby/sdk-phase-1.md
cretz: Yes

ruby/sdk-phase-1.md
cretz: The Ruby standard here is to keep it shorter if already in the module name. Ruby is the only language I can think of that does this commonly. While you can require a file, there is no "importing" a module, it must always be qualified (but people often alias it).

ruby/sdk-phase-1.md
cretz: Because `Temporalio::Activity` is the class that is being extended by the activity implementation and we don't want people to think that this is only usable in that extension nor do we want to clutter the inherited class with methods like these when they are part of the context.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  common encoding converters
+* `Temporalio::Converter::FailureConverter` - Default failure converter, can be overridden
+* `Temporalio::Converter::PayloadCodec` - Unimplemented base codec
+  * Important for implementers to ensure this is async capable if using async activities anywhere
+* `Temporalio::Converter::RawValue` - Can be used as way to move around unconverted payloads
+
+#### Temporalio::Converter::PayloadConverter::JSON
+
+* Ruby has a `JSON` module in the standard library, but by default it's only primitives, arrays, hashes, etc and not
+  classes.
```
antlai-temporal, line 305: What about JSON Protobuf, I'm assuming not supported...

Also, what are the implications for cross-language integration. For example, an activity in python that returns a json serialized object called from a Ruby workflow. Is the workaround to manually create the ruby object from json?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+      normal there, and in Python we only offered package-level methods because that's normal there.
+    * 💭 Why allow access to an instance? Some users want to pass the activity context to another thread.
+  * 💭 Why `Temporalio::Activity::Context` instead of just `Temporalio::Activity` or putting helpers on base class or
+    something? It is a common requirement in our SDKs to be able to access the activity context from the environment.
+    Also, it is a common requirement to be able to do that from utility methods outside of the explicit activity call.
+* Activity can choose its executor, default is `:default`.
+  * See worker section for more information on executors.
+* Cancellation is reported to the executor.
+  * For thread pool executor (the default), this raises a cancellation exception via `Thread.raise`.
+  * For fiber executor, this is done via `Fiber.raise`.
```
antlai-temporal, line 377: Is there a way to abstract the low level implementation (fiber vs thread pool) from how the cancellation is reported to the activity code. Ideally, we would like to change the implementation without modifying application code...

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+* Activities are defined as classes.
+  * 💭 Why classes instead of methods/functions like every other SDK? There are tradeoffs to both approaches. As
+    classes, they are more easily referenced, a bit more easily typed in RBS, and metadata can be applied more easily at
+    the class level. This comes at a cost of not being able to easily share state across classes, but users can provide
+    something in the constructor multiple activities use. Open to considering activity methods instead.
+  * ❓ Would we prefer activities as methods knowing they may not be able to be typed/referenced well? We can't really
+    take the workflow signal-ish approach of those pseudo-decorators making class methods unless we say activities must
+    be instance methods (since we would allow class methods as activities).
+* Activity classes are instantiated at registration time, not for each invocation.
+  * 💭 Why? Existing Temporal Ruby SDK does instantiate for each activity task, but this discourages use of shared
```
antlai-temporal, line 359: The temptation for programmers is to add state (in the activity object) that lingers across multiple activity invocations, something really fragile... Is there a way to freeze that activity object so that it can only be used in a pure functional style?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  end
+end
+```
+
+* Activities are defined as classes.
+  * 💭 Why classes instead of methods/functions like every other SDK? There are tradeoffs to both approaches. As
+    classes, they are more easily referenced, a bit more easily typed in RBS, and metadata can be applied more easily at
+    the class level. This comes at a cost of not being able to easily share state across classes, but users can provide
+    something in the constructor multiple activities use. Open to considering activity methods instead.
+  * ❓ Would we prefer activities as methods knowing they may not be able to be typed/referenced well? We can't really
```
Sushisource, line 355: I think classes makes good sense for Ruby. The typing is good and Ruby is all about the OOP nonsense. The biggest downside is users thinking that storing class state is somehow magically going to be synchronized across workers or something, but I'm not hugely concerned about that.

ruby/sdk-phase-1.md
Sushisource: Yeah, that's my concern too. I think instantiating them per-invocation is probably safer, and sharing things like dependencies could be done w/ either static fields or simply by passing them in each time.

ruby/sdk-phase-1.md
Sushisource: +1 - would be nice to just have `Temporalio::Activity::Context.raise`

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    end
+  end
+end
+```
+
+* Activities are defined as classes.
+  * 💭 Why classes instead of methods/functions like every other SDK? There are tradeoffs to both approaches. As
+    classes, they are more easily referenced, a bit more easily typed in RBS, and metadata can be applied more easily at
+    the class level. This comes at a cost of not being able to easily share state across classes, but users can provide
+    something in the constructor multiple activities use. Open to considering activity methods instead.
```
josh-berry, line 354: This wouldn't be for an MVP, but could we eventually provide both if we wanted to? It doesn't seem like there's an obvious trade-off winner here.

Another plus in favor of functions is they are lighter-weight to define; you don't have to make a class just to wrap what is effectively a function anyway.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+* Like other SDK methods, `run` is async-capable.
+* `activity_executors` can be a hash of `Temporalio::Worker::ActivityExecutor`s to run certain types of activities.
+  * Default is `{:default: Temporalio::Worker::ActivityExecutor::ThreadPool.new(max_concurrent_activities), :fiber: Temporalio::Worker::ActivityExecutor::Fiber.DEFAULT}`
+  * Activity can choose its executor, default is `:default`.
+  * Registration time will fail if activity references an executor that does not exist.
+  * Worker run will fail immediately before even starting if it is not run with a fiber scheduler present but any
+    activities reference a `Fiber` executor.
+    * 💭 Why wait until worker run to fail? Workers can be created outside of an async environment, it's just important
+      that they are run within an async environment if there are async activities.
+    * ❓ Or should we just enforce this at worker instantiation time too?
```
Sushisource, line 428: I think that's best if doable

Sushisource (approved):
:gem:

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    classes, they are more easily referenced, a bit more easily typed in RBS, and metadata can be applied more easily at
+    the class level. This comes at a cost of not being able to easily share state across classes, but users can provide
+    something in the constructor multiple activities use. Open to considering activity methods instead.
+  * ❓ Would we prefer activities as methods knowing they may not be able to be typed/referenced well? We can't really
+    take the workflow signal-ish approach of those pseudo-decorators making class methods unless we say activities must
+    be instance methods (since we would allow class methods as activities).
+* Activity classes are instantiated at registration time, not for each invocation.
+  * 💭 Why? Existing Temporal Ruby SDK does instantiate for each activity task, but this discourages use of shared
+    objects or state to be referenced. This does have the tradeoff our other SDKs have of requiring the function to be
+    thread safe.
```
josh-berry, line 361: So activities are essentially singletons? Bleh, that seems harder to reason about and rather counter-intuitive to anyone new to Temporal. Is this a common pattern in Ruby?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    running multiple blocking things at once without external libraries or forcing threads on users.
+* Like other SDK methods, `run` is async-capable.
+* `activity_executors` can be a hash of `Temporalio::Worker::ActivityExecutor`s to run certain types of activities.
+  * Default is `{:default: Temporalio::Worker::ActivityExecutor::ThreadPool.new(max_concurrent_activities), :fiber: Temporalio::Worker::ActivityExecutor::Fiber.DEFAULT}`
+  * Activity can choose its executor, default is `:default`.
+  * Registration time will fail if activity references an executor that does not exist.
+  * Worker run will fail immediately before even starting if it is not run with a fiber scheduler present but any
+    activities reference a `Fiber` executor.
+    * 💭 Why wait until worker run to fail? Workers can be created outside of an async environment, it's just important
+      that they are run within an async environment if there are async activities.
```
josh-berry, line 427: Are we also thinking about workers that have a combination of sync and async activities?

ruby/sdk-phase-1.md
cretz: > What about JSON Protobuf, I'm assuming not supported...
JSON protobuf will be supported out of the box

> Also, what are the implications for cross-language integration. For example, an activity in python that returns a json serialized object called from a Ruby workflow. Is the workaround to manually create the ruby object from json?

Yes, that will be a Ruby hash (just like it would be a JS object). If you want to convert it to a Ruby class you would have to do that yourself. We can consider more advanced JSON-to-class conversion mechanisms later, but by default Ruby doesn't really have any without the "addition" approach which is not very cross-language compatible.

ruby/sdk-phase-1.md
cretz: > Is there a way to abstract the low level implementation (fiber vs thread pool) from how the cancellation is reported to the activity code

It is abstracted from a user POV. This discusses what _our_ SDK code will do when these happen. From a user POV it's just an exception that occurs.

ruby/sdk-phase-1.md
cretz: Technically in .NET if you use DI w/ scoped activities it does instantiate them per activity task, but overall in our SDKs you are allowed to share object/outside-of-function state across activity invocations without using globals. I am a hesitant to make Ruby the special case of forcing globals to share state.

But we can probably have ways to let the user choose. For instance, we can accept two ways of providing an activity: providing an instance and providing a class. I wouldn't want the activity class form to instantiate per attempt _by default_ because it'd be confusing to have whether you did instance vs class to have different lifetimes. But we could support some setting on the activity class to say they want it per attempt and hope they have no constructor parameters. Or we'd have to devise some DI because in general it is difficult for Temporal to control lifetimes of user activity objects due to statefulness needs.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    modules with internal things.
+* Avoid cluttering the top-level `Temporalio` module with many small classes, but some are ok as we try to avoid a
+  `Temporalio::Common` module
+  * So `Temporalio::RetryPolicy` is preferred over `Temporalio::Common::RetryPolicy`.
+  * All search attribute things should be under `Temporalio::SearchAttribute` (e.g. `Temporalio::SearchAttribute::Key`)
+    and all metric things under `Temporalio::Metric` (e.g. `Temporalio::Metric::Meter`).
+  * 💭 Why not top-level? Cluttering top-level module makes discoverability difficult and API docs unclear.
+  * 💭 Why no `Temporalio::Common`? Ruby expects to have common things where they make sense instead of under a `Common`
+    module.
+  * ❓ To we want plural, e.g. `Temporalio::SearchAttributes` and `Temporalio::Metrics`? While there is no such thing as
```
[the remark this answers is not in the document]
cretz: But in the case of `SearchAttributes` (plural) and `Metric` (singular), they may in fact be classes (the former being the search attribute collection and the latter being the base metric class).

ruby/sdk-phase-1.md
cretz: It is doable, it just means you have to instantiate the worker in the async context as well

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  * 💭 Why? Our dependencies as a library have transitive effects on our users, so we have a responsibility to be
+    lightweight.
+* Must be both thread-blocking/thread-safe and async-capable
+  * PoC shows that this is most easily done with Rust block/callback to `Queue` `push`/`pop` since in modern Ruby
+    `Queue.pop` does not block fibers and many other things don't either. Therefore many SDK features like clients and
+    codecs don't have to distinguish between sync/async.
+  * Activities need to work in both ways and _do_ have to make a distinction (discussed later).
+  * 💭 Why async-capable? Core and all SDKs are very parallel and so we want to give that benefit to users if they want
+    it.
+  * 💭 Why thread-blocking? That's the most common way Ruby is used.
```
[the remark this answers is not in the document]
cretz: > Does either sync or async depend on the other? Or are the implementations independent?

Not following what is meant by "depend on the other", but they are independent for the most part. Basically it's just the difference of starting a new thread or a new fiber when invoking the activity, not much else changes.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+* Obvious general repo/project rules apply.
+  * Linting and auto-formatting (Rubocop check/autocorrect but have to customize Rubocop a lot because its defaults are
+    very dogmatic).
+  * Type checking (Steep).
+  * Deployed to RubyGems as separate gems for pure source, x64 linux, x64 darwin, x64 windows (i.e. mingw-ucrt), aarch64
+    linux (i.e. arm64), and arm64 darwin.
+  * CI against supported platforms on oldest and newest Ruby versions supported.
+  * MIT licensed.
+  * Full docs as README.
+  * No need for ruby.temporal.io, https://www.rubydoc.info/gems/temporalio is acceptable.
```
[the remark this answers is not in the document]
cretz: Sure, though we should do the same for https://java.temporal.io/ too (nothing there, should redirect to https://www.javadoc.io/doc/io.temporal) and https://go.temporal.io/ too (redirects to temporal.io, should redirect to https://pkg.go.dev/go.temporal.io/sdk maybe, though there is https://pkg.go.dev/go.temporal.io/api and https://pkg.go.dev/go.temporal.io/server too). Want to start the effort there?

ruby/sdk-phase-1.md
cretz: We can technically do this, and could technically provide activities as classes/interfaces/structs in every other SDK but we have tried to stick with a single way that is best with the language. I think we should only do one here.

ruby/sdk-phase-1.md
cretz: Up to a user, they own the lifetime of these objects not us. They could instantiate different ones for different workers or share across workers. But yes, they live for the life of the worker which makes them long-lived usually (same with something like a database client). There is not a single common patterns for this kind of thing in Ruby that I am aware of. The current Temporal Ruby SDK does instantiate per activity task and didn't take into account the need to share things like a database client which is common.

ruby/sdk-phase-1.md
cretz: Yes, this may be common (and is common in Python). This is just saying _if_ any are async we make sure we're in an async environment. Sync activities will always work.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+These are the general requirements of the project
+
+* Ruby >= 3.1
+  * 💭 Why not older versions? They are all EOL'd and we need Ractors and other things that come with 3.x.
+* Limited dependencies
+  * This means we will try to make `google-protobuf` the only dependency. We should not depend on `async` or anything
+    else if we can help it.
+  * 💭 Why? Our dependencies as a library have transitive effects on our users, so we have a responsibility to be
+    lightweight.
+* Must be both thread-blocking/thread-safe and async-capable
```
dandavison, line 39: What do you mean exactly by "async-capable"? (The term is used in a few places)

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
Sushisource: Yeah, I think you maybe misread me, I said because they *are* classes

ruby/sdk-phase-1.md
Sushisource: Ah, cool that makes sense, that wasn't immediately clear to me.

ruby/sdk-phase-1.md
cretz: This can definitely be argued, but I think enough of our SDKs default to "default" and some prod clusters do use "default" that yeah it's not a big deal.

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: I did, sorry!

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+      ...
+    end
+  end
+end
+```
+
+* Following the API of the current Temporal Ruby SDK, workers can be run individually, but we will also provide a class
+  method to run several at once.
+  * 💭 Why is something needed to run multiple workers for the user? Because in Ruby they don't have easy ways of
+    running multiple blocking things at once without external libraries or forcing threads on users.
```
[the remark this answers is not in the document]
cretz: In most of our SDK languages there are better ways to run multiple workers at once without eating a thread per worker. JS is always single threaded, Python has `asyncio.gather`, Go has goroutines, and .NET has tasks. In Java we have a `WorkerFactory` that lets you run multiple workers at once because it lacks good language-level reuse-thread capabilities. Ruby has the same problem. So I was thinking a `WorkerPool` or similar that can run many workers in a single blocking run call.

ruby/sdk-phase-1.md
cretz: In Ruby there are threads and fibers (basically coroutines). By async-capable I mean fiber-capable. So you can use https://github.com/socketry/async and, say, use the client to wait on 100 workflow results without using a thread each. But many Ruby users are totally fine using a thread each and we need to support that too.

This is very much like Python, we just chose not to provide a synchronous version of the Python client (though we still could if we wanted).

ruby/sdk-phase-1.md
drewhoskins-temporal, line 46: Context related to this topic, ruby @ Stripe was single-threaded by default.  Multithreaded carve-outs did happen but were relatively unusual.
This meant some libraries weren't designed to be thread-safe and used things like class variables.
Default was therefore to have many workflow pollers but only one activity poller per process, which was inefficient, but we added a config flag folks could turn on to make their workers run multiple activity pollers.  I really wanted to make the multi-poller version the default but didn't do so before I left.

cretz: This is why we plan to also support single-threaded/coroutine-based implementations. It's important to us at Temporal that people be able to do a lot of concurrent things with the SDK in just a single thread. Granted the current design is queue-pop-based which makes it work with fiber-based async. I wasn't planning on supporting other async approaches (e.g. event machine) but if there is enough desire for pluggable-non-fiber-based async, we can look into that.

> Multithreaded carve-outs did happen but were relatively unusual.

These carve-outs were likely required for to use the existing Ruby SDKs.

drewhoskins-temporal: On the carve-outs, I'm referring to the "pre-temporal world."  Later, with the old SDK, we simply limited to one activity poller, which fixed any race conditions.  It was a cautious approach, as most customers probably would have been fine, but there was FUD.

ruby/sdk-phase-1.md
cretz: Not sure the emphasis is needed

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+    lightweight.
+* Must be both thread-blocking/thread-safe and async-capable
+  * PoC shows that this is most easily done with Rust block/callback to `Queue` `push`/`pop` since in modern Ruby
+    `Queue.pop` does not block fibers and many other things don't either. Therefore many SDK features like clients and
+    codecs don't have to distinguish between sync/async.
+  * Activities need to work in both ways and _do_ have to make a distinction (discussed later).
+  * 💭 Why async-capable? Core and all SDKs are very parallel and so we want to give that benefit to users if they want
+    it.
+  * 💭 Why thread-blocking? That's the most common way Ruby is used.
+* By as type-safe as possible with RBS
```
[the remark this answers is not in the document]
cretz: Fixed

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: Fixed

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
josh-berry: Sure, will make a note for later (I don't think this is urgent)

ruby/sdk-phase-1.md
josh-berry: Yeah if we want to pick a single way, then I'm in favor of doing the most-flexible thing. Long as we have room to add the other way later if we learn that users are really opinionated about it.

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
josh-berry: > Not following what is meant by "depend on the other",

Is sync implemented in terms of—or does it reuse code written for—async, or vice versa? I think you've answered my first question here though.

For my second question, I'm still trying to understand what is MVP-shaped and what isn't—e.g. would we be comfortable doing a first/early release with one but not the other?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,439 @@
...
+  * MIT licensed.
+  * Full docs as README.
+  * No need for ruby.temporal.io, https://www.rubydoc.info/gems/temporalio is acceptable.
+  * Full integration test suite.
+
+### Basic Features
+
+By the very first version, the Ruby SDK should have feature parity with every other SDK (except in cases where we have
+decided other features in other SDKs will never be ported). Here's a non-exhaustive list of some features that includes
+that are not discussed later:
```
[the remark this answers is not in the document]
cretz: Fixed

ruby/sdk-phase-1.md
cretz: We are going to update this to require the namespace. The default of `default` is unreasonable for cloud and many others.

ruby/sdk-phase-1.md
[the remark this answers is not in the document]
cretz: Sorry, missed this question. This is integral behavior and the same code will work async/sync so there is no real separating it.

ruby/sdk-phase-1.md
josh-berry: Yeah I'm fine with it being a singleton/lazy global under the hood. I just don't want us to design ourselves into a corner.

cretz:
I have updated the proposal with the following:

* Clarify why we need typing at all
* Use `Temporalio::SearchAttributes` as both the collection and class under which all nested search attribute things reside
* Use `Temporalio::Metric` as both the base metric and class under which all nested metric things reside
* Clarified use of `Options` structs on client/connection
* Clarified there will be a `CloudOperationsClient`
* Made `namespace` a required field instead of defaulting to `default`
* Changed `Temporalio::Converter` module to `Temporalio::Converters`
* Changed `Temporalio::Converters::PayloadConverter` to be a simple base class instead of having it impl composite, and created `Temporalio::Converters::PayloadConverter::Composite`
* Updated activity lifecycle to say we will instantiate activities on each attempt if registered as a class, or we will just call method on existing instance if registered as an instance
* Clarified that activity executors can validate activities at registration time instead of runtime, which means that workers have to be _created_ in an async context if they have any async-context-needed activities (as opposed to just run in an async context)

cretz:
Closing this PR in favor of a new one for a fresh round of comments. Will link back to this one for posterity, but unaware of any unresolved issues here. Anyone here can feel free to comment on the new one.

PR #93, opened 2024-07-10
Ruby phase 1

cretz:
# Summary

* This is the proposal for the first phase of the Ruby SDK. This is the second discussion of this proposal, see #92 for past discussion.
* See it [rendered](https://github.com/cretz/temporal-proposals/blob/ruby-phase-1/ruby/sdk-phase-1.md).
* Much of this is inspired by the current [Temporal Ruby SDK](https://github.com/temporalio/sdk-ruby) (and its [proposals](https://github.com/temporalio/proposals/tree/master/ruby/previous) which were moved to a sub directory) and our other SDKs.

For the public, please also join us on `#ruby-sdk` in [Slack](https://t.mp/slack).

### DIFF
--- a/ruby/sdk-phase-1.md
+++ b/ruby/sdk-phase-1.md
@@ -53,6 +53,7 @@ These are the general requirements of the project
     library won't work with.
     * ❓ How hard do we want to work to avoid duplicate yard/RBS typing?
   * 💭 Why not Sorbet? Sorbet can come later or provided separately, but RBS seems to be the newer movement.
+  * 💭 Why have any typing? As a library we might as well help users where we can.
 * Everything is in the `Temporalio` module and released as a `temporalio` gem starting at 0.2.0
   * 💭 Why 0.2.0? 0.1.x versions already in use at https://rubygems.org/gems/temporalio.
 * Internal code will be in `Temporalio::Internal` module including bridge which is in `Temporalio::Internal::Bridge`
@@ -61,16 +62,16 @@ These are the general requirements of the project
 * Avoid cluttering the top-level `Temporalio` module with many small classes, but some are ok as we try to avoid a
   `Temporalio::Common` module
   * So `Temporalio::RetryPolicy` is preferred over `Temporalio::Common::RetryPolicy`.
-  * All search attribute things should be under `Temporalio::SearchAttribute` (e.g. `Temporalio::SearchAttribute::Key`)
+  * The search attribute collection is `Temporalio::SearchAttributes` and all things are under it, e.g.
+    `Temporalio::SearchAttributes::Key`.
+    * 💭 Why not `Temporalio::SearchAttribute` module with the collection underneath? The object has value on its own
+      so might as well make it a class.
+  * The base metric is `Temporalio::Metric`, and all metric things are under there, e.g. `Temporalio::Metric::Meter`.
+    * 💭 Why not `Temporalio::Metrics` module? The object has value on its own so might as well make it a class.
+   (e.g. `Temporalio::SearchAttribute::Key`)
     and all metric things under `Temporalio::Metric` (e.g. `Temporalio::Metric::Meter`).
-  * 💭 Why not top-level? Cluttering top-level module makes discoverability difficult and API docs unclear.
   * 💭 Why no `Temporalio::Common`? Ruby expects to have common things where they make sense instead of under a `Common`
     module.
-  * ❓ Do we want plural, e.g. `Temporalio::SearchAttributes` and `Temporalio::Metrics`? While there is no such thing as
-    a `SearchAttribute`, there is such thing as a `SearchAttributes` (i.e. search attributes collection). And there is
-    such thing as a `Metric` (i.e. base class for counter, gauge, and histogram). So it could be
-    `Temporalio::SearchAttributes` (the collection) and `Temporalio::Metric` (the base class), and they are not modules
-    but classes with the nested pieces underneath like we do elsewhere.
 * No compatibility with any existing Ruby SDK
   * Migration guides/docs may appear as needed.
   * 💭 Why? Need a clean slate to develop a high-quality SDK from first principles.
@@ -129,22 +130,27 @@ connection. The former has shortcut for creating the latter while it creates its
 ```ruby
 module Temporalio
   class Client
-    attr_reader :connection, :namespace
+    # This can be splatted into the constructor if one option needs to be
+    # changed. This is a Struct and not Data because we support Ruby 3.1.
+    Options = Struct.new(...)
 
     # Connect a client (i.e. Connection.new + Client.new)
-    def self.connect(target_host:, namespace: 'default', runtime: nil, ...)
+    def self.connect(target_host, namespace, runtime: nil, ...)
       ...
     end
 
-    # Create client with existing connection
-    def initialize(connection, namespace: 'default', ...)
+    # Create client with existing connection. This must be all keyword arguments
+    # and must match the Options struct.
+    def initialize(connection:, namespace:, ...)
       ...
     end
 
-    # Duped hash of keyword args of constructor
-    def options
+    # Duped options 
+    def dup_options
       ...
     end
+
+    # Several properties such as connection, namespace, workflow_service, etc
 
     # Returns workflow handle
     def start_workflow(
@@ -178,13 +184,15 @@ module Temporalio
 
   class Client
     class Connection
+      Options = Struct.new(...)
+
       # Perform connection
       def initialize(target_host:, runtime: nil, ...)
         ...
       end
 
-      # Duped hash of keyword args of constructor
-      def options
+      # Duped options 
+      def dup_options
         ...
       end
 
@@ -207,6 +215,10 @@ module Temporalio
 
       # Lots more
     end
+
+    class CloudOperationsClient
+      # Similar to client but for cloud ops
+    end
   end
 end
 ```
@@ -215,10 +227,12 @@ end
 * We expect most users and most samples to just use `Client.connect`.
   * 💭 Why not make users create a connection themselves instead of maintaining similar options sets across
     `Connection.initialize` and `Client.connect`? We have found most people only need a single client per connection.
-* Users that want raw service access need to use `client.connection.workflow_service` similar to .NET.
+* Users will have access to raw clients via `client.workflow_service` and similar
 * Workflow handles exist with all relevant calls.
-* Note that we prefer large sets of keyword arguments over hashes or options/config objects.
-  * 💭 Why? This is the Ruby preference and clearer to users.
+* Note that we use the kwargs constructor + options struct pattern so that a user can access duped options, change one,
+  and reconstruct via `Client.new(**my_new_options.to_h)`
+  * ❓Is there a better way to allow people to easily change an option? We didn't want a `with` pattern because, at
+    least for `Connection`, it requires a reconnect.
 * Must make clear in docs which things issue an RPC call (e.g. start workflow) vs ones that don't (e.g. get workflow
   handle).
 
@@ -235,17 +249,17 @@ class MyWorkflow < Temporalio::Workflow
   workflow_query_attr_reader :some_other_query
 
   def execute(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 
   workflow_signal
   def some_signal(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 
   workflow_query 'custom-query-name'
   def some_query(arg)
-    raise NoMethodError
+    raise NotImplementedError
   end
 end
 ```
@@ -254,7 +268,7 @@ And client usage may look like:
 
 ```ruby
 # Connect client
-client = Client.connect('localhost:7233')
+client = Client.connect('localhost:7233', 'default')
 
 # Start workflow
 handle = client.start_workflow(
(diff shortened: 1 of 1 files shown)


ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+
+By the very first version, the Ruby SDK should have feature parity with every other SDK (except in cases where we have
+decided other features in other SDKs will never be ported). Here's a non-exhaustive list of features we should include
+that are not discussed later:
+
+* Interceptors and an OpenTelemetry implementation
+* Custom metrics and optional metrics buffer for custom processing
+* Typed search attributes, memos, headers, etc
+* API key, RPC metadata, HTTP CONNECT proxy support, etc
+* Loggers with contextual information
```
cretz, line 103: This will likely be our own logger abstraction

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+  * 💭 Why thread-blocking? That's the most common way Ruby is used.
+* Be as type-safe as possible with RBS
+  * All SDKs leverage generics to try to catch mistyped arguments for, say, calling a workflow/activity and we should
+    too.
+  * It is acknowledged that RBS and Ruby type hinting in general is rough at this time, but we must do our best.
+  * We may be able to leverage https://github.com/AaronC81/sord or similar to not have to duplicate docs and sig files,
+    but in general Ruby typing right now is a mess and we likely will hit bugs or advanced cases a yard/RBS combination
+    library won't work with.
+    * ❓ How hard do we want to work to avoid duplicate yard/RBS typing?
+  * 💭 Why not Sorbet? Sorbet can come later or provided separately, but RBS seems to be the newer movement.
```
alexevanczuk, line 55: Hey @cretz , the org where I work is very excited about the official Ruby SDK for temporal!

Just wanted to drop a note that we use Sorbet heavily and it seems like it's the much more popular typing system in the community (despite the introduction of RBS). It won't affect our usage much either way (although first-class sorbet typing would make it slightly easier to integrate), but if you were interested in exploring/chatting more about sorbet, just wanted to offer to jump on a synchronous call to talk through concerns/tradeoffs. I've added sorbet to many gems and applications and would be very happy to chat through things.

Thanks for prioritizing this work!!

cretz: Thanks! Yes, I definitely believe RBS is weaker than Sorbet and Steep is not a great tool. We considered no typing at all and that still may be an option. But during PoC, RBS was acceptable for this limited use case, but I expect most users to treat this as an untyped library, including Sorbet users. So I definitely agree it won't affect most people's usage either way. At this point with such low RBS adoption, it essentially just becomes a check on ourselves (we did this successfully in the last incarnation at https://github.com/temporalio/sdk-ruby/tree/v0.1.1).

We were just wanting to avoid maintaining Sorbet annotations in repo. But we would totally encourage external ones at https://github.com/Shopify/rbi-central or https://github.com/sorbet/sorbet-typed or wherever.

Would definitely like to chat on this further. If easier, can come join us in `#ruby-sdk` in [Slack](https://t.mp/slack). The pinned message in that channel also has a link to a survey w/ a request at the end to setup a meeting to discuss Ruby if that's preferred.

alexevanczuk: I filled out the survey, thanks!

I commented in that channel in case you wanted to chat through Sorbet more :) 

Thanks again!

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+    end
+  end
+end
+```
+
+* Activities are defined as classes.
+  * 💭 Why classes instead of methods/functions like every other SDK? There are tradeoffs to both approaches. As
+    classes, they are more easily referenced, a bit more easily typed in RBS, and metadata can be applied more easily at
+    the class level. This comes at a cost of not being able to easily share state across classes, but users can provide
+    something in the constructor multiple activities use. Open to considering activity methods instead.
```
bdchauvette, line 376: You might've already discussed  this internally, but a middle ground here might be supporting any object that responds to `#call` rather than `#execute`.

Using `#call` would potentially allow simple activities to be implemented as a `Proc` or `Lambda`, while still allowing people to use custom classes for more complex scenarios.

In the spirit of "there's more than one way to do it", this would give SDK users multiple options for defining activities, and they could choose whichever one is most appropriate for their circumstances:

```ruby
module Activities
  # Constant procs
  SomeProcActivity = ->(params) {
    # ...
  }

  # Proc factory methods
  def self.another_proc_activity(config)
    ->(params) {
      # can use config to customize proc
    }
  }

  # Maybe some kind of DSL?
  DSLActivity = Temporalio.activity(options) do |params|
    # ...
  end

  # Child classes for full control
  ChildClassActivity < Temporalio::Activity
    def initialize(config)
      # ...
    end

    def call(params)
      # ...
    end
  end
end
```

```ruby
# Activity registration using a hash, similar to TS API:
worker1 = Temporalio::Worker.new(
  # ...

  activities: {
    SomeProcActivity: Activities::SomeProcActivity,
    AnotherProcActivity: Activities.another_proc_activity(config),
    DSLActivity: Activities::DSLActivity,
    ChildClassActivity: Activities::ChildClassActivity,

    # Even inline activities, which is almost certainly not a good
    # idea for production code, but could be useful in some situations.
    InlineActivity: ->(params) {
      # ...
    }
  }

  # ...
)
```

And possibly a block-based API on the worker:

```ruby
worker.register_activity('MyActivity') do |params|
  # ...
end
```

As well as anything that responds to `#to_proc`:

```ruby
worker.register_activity('MyActivity', &:my_activity)
```

All of these could of course also be supported using the `#execute` method, but then there'd be a discrepancy between the SDK's `#execute` and the proc's `#call`.

---

Some similar APIs:
- the [Rack specification](https://github.com/rack/rack/blob/main/SPEC.rdoc), which makes heavy use of objects that respond to `#call`
- [`Logger#formatter`](https://ruby-doc.org/3.3.4/stdlibs/logger/Logger.html#attribute-i-formatter), which is documented as supported either a proc or an instance [`Logger::Formatter`](https://ruby-doc.org/3.3.4/stdlibs/logger/Logger/Formatter.html) (which responds to `#call`)
- AWS SDK uses [a `Callable` interface](https://github.com/search?q=repo%3Aaws%2Faws-sdk-ruby+%22%5Bcallable%5D%22&type=code) for various callback-like options in some services

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+* Note that we use the kwargs constructor + options struct pattern so that a user can access duped options, change one,
+  and reconstruct via `Client.new(**my_new_options.to_h)`
+  * ❓Is there a better way to allow people to easily change an option? We didn't want a `with` pattern because, at
+    least for `Connection`, it requires a reconnect.
+* Must make clear in docs which things issue an RPC call (e.g. start workflow) vs ones that don't (e.g. get workflow
+  handle).
+
+### Workflows in Client
+
+Early thoughts on workflows are that they'll be Ractors. For the purposes of this phase, all that may matter is how
```
bdchauvette, line 241: I left a comment about this in the Google form as well, but I'm not sure if Ractors would be the right move here. They're still very experimental and still have some fundamental limitations:

- https://rubykaigi.org/2023/presentations/ko1.html
  > Ractor was introduced in Ruby 3.0 as an experimental feature to enable parallel programming in Ruby. Now Ractor is not widely used for many reasons. For example, from the perspective of a Ractor implementer, Ractor does not have enough performance and quality. In this talk, we will explain which features are not enough, why they are difficult, and how to solve these difficulties.

- https://rubykaigi.org/2024/presentations/ko1.html

  > This talk presents recent updates to Ractor, which enables parallel and concurrent programming on Ruby.
  >
  > Ractor still lacks fundamental features. For example, we cannot use “require” method and “timeout” methods on non-main Ractors because of synchronization and implementation issues. We will discuss such problems and how to solve them. From a performance point of view, we have introduced the M:N thread scheduler in Ruby 3.3 and we will show the performance analysis with recent improvements.


They also still output scary warnings, even in Ruby 3.3:

```
irb(main):001> Ractor.new { puts RUBY_VERSION }
(irb):1: warning: Ractor is experimental, and the behavior may change in future versions of Ruby! Also there are many implementation issues.
3.3.4
=> #<Ractor:#2 (irb):1 terminated>
```

I definitely think that Ractors could be a good solution in a future world where Ractors are more polished, but IMO for now it would be safer to follow the [`coinbase/temporal-ruby`](https://github.com/coinbase/temporal-ruby) SDK and use Fibers as the basis for scaling workflow executions.

ruby/sdk-phase-1.md
cretz: > Using #call would potentially allow simple activities to be implemented as a Proc or Lambda, while still allowing people to use custom classes for more complex scenarios.

A couple of important aspects we want for activities are 1) they are referenceable without calling (e.g. `Workflow.execute_activity(MyActivityClass, ...`), and 2) people can use existing things (e.g. mixins) and there is not much magic.

> In the spirit of "there's more than one way to do it", this would give SDK users multiple options for defining activities, 

We actually try to avoid this in SDKs where we can at declaration time. Sure some SDKs do have a couple of ways, but in general SDKs say "activities are defined as X" without many values for X and without more magic than we have to have. And remember there is the referenceability concern.

ruby/sdk-phase-1.md
cretz: One of the biggest challenges in SDK development is sandboxing for determinism. There are two main impediments to determinism: state isolation and restrict API use (e.g. disk IO). Ignoring the latter, if we can get the former it may be worth it. JS is lucky in that it can spin up a new V8 instance. But it is a nightmare for Go and Java and .NET and other devs to remember that they cannot share state. We expended a lot of effort making a Python sandbox just to get this state isolation so that globals don't pollute.

We will have to evaluate performance to see if this can work, and we can provide a way to opt-out of Ractors (for the most part none of the Ractor stuff would be visible to users unless they wanted to do bad things like expose something shareable).

> Fibers as the basis for scaling workflow executions.

We will definitely still be using fibers (we will have our own fiber scheduler since we need deterministic coroutines), but if we can get state isolation by default, it may be worth the performance penalty. This is more of a phase 2 thing though, workflows aren't going to be done now.

bdchauvette: Got it, that makes sense! If it's only for workflow-internal things, I think Ractors sound great.

But if the goal is to isolate global state even between activities, I could see that being very challenging with Ractors. For example, I'm not sure if ActiveRecord would would play nicely with Ractors, because (AFAIK) the connection pool wouldn't be able to be shared between Ractors.

> Interestingly enough, deterministic isolated Temporal workflows is like one of the only good use cases I have seen for Ractors, heh.

Same, TBH 😅 

I _want_ to use them, but haven't really found a use case yet, especially given their current limitations :(

cretz: > But if the goal is to isolate global state even between activities, I could see that being very challenging with Ractors

Definitely not. Activities will not have anything to do with Ractors at all. Activities are just normal Ruby code where you can do what you want (we will allow different executors so people can choose whether to run fiber/async vs threaded).

ruby/sdk-phase-1.md
bdchauvette: > A couple of important aspects we want for activities are 1) they are easily referenceable by a caller without calling (e.g. Workflow.execute_activity(MyActivityClass, ..., w/ possible generic typing support w/ RBS/Sorbet), and 2) people can use existing things (e.g. mixins) and there is not much magic.

Got it. If I'm understanding them correctly, those all make sense to me as goals / requirements 👍

For (1), is that referring to the ability to call an activity with something like `MyActivity.execute` rather than `workflow.execute_activity(MyActivity)`? If so, then using a method other than `#call` definitely makes sense to me. IIUC, that'll help prevent users from trying to execute activity classes that haven't been wired up to Temporal.

> We actually try to avoid this in SDKs where we can at declaration time. Sure some SDKs do have a couple of ways, but in general SDKs say "activities are defined as X" without many values for X and without more magic than we have to have. And remember there is the referenceability concern.

I totally get this, and I'm _personally_ not a huge fan of how Ruby has adopted Perl's TIMTOWTDI philosophy. But I also think a lot of Rubyists definitely view that as one of Ruby's strengths, so just wanted to put in a plug for that way of thinking 😅

That being said, I don't think anyone will avoid using Temporal just because they can't implement activities as procs, so it's probably not worth the effort to implement (& support!) that level of flexibility.

ruby/sdk-phase-1.md
bdchauvette: Sounds great! Thanks for helping me understand!

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+decided other features in other SDKs will never be ported). Here's a non-exhaustive list of features we should include
+that are not discussed later:
+
+* Interceptors and an OpenTelemetry implementation
+* Custom metrics and optional metrics buffer for custom processing
+* Typed search attributes, memos, headers, etc
+* API key, RPC metadata, HTTP CONNECT proxy support, etc
+* Loggers with contextual information
+* Full error hierarchy with `Temporalio::Error` as the root
+* ❓ Are there any Ruby libraries so widely adopted they might as well be in the standard library and therefore deserve
```
bdchauvette, line 105: ActiveSupport is probably the big contender here, but TBH I don't know how much the Temporal would really benefit from having it available.

Optional support for things like `ActiveSupport::Duration` might be nice for things like `start_delay`, but that could be a pain to implement (& maintain) alongside normal `Numeric` arguments.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+
+* There are many things missing here, but the idea is clear.
+* We expect most users and most samples to just use `Client.connect`.
+  * 💭 Why not make users create a connection themselves instead of maintaining similar options sets across
+    `Connection.initialize` and `Client.connect`? We have found most people only need a single client per connection.
+* Users will have access to raw clients via `client.workflow_service` and similar
+* Workflow handles exist with all relevant calls.
+* Note that we use the kwargs constructor + options struct pattern so that a user can access duped options, change one,
+  and reconstruct via `Client.new(**my_new_options.to_h)`
+  * ❓Is there a better way to allow people to easily change an option? We didn't want a `with` pattern because, at
```
bdchauvette, line 234: One very common pattern in Ruby is to have config methods (including some initializers) accept an optional block that yields the config object for modification:

```ruby
class Client
  def initialize(**options)
    @config = Configuration.new(**options)
    yield @config if block_given?

    # ...
  end
end
```

```ruby
client = Client.new do |config|
  config.namespace = "my-namespace"
end
```

I think that would feel very natural for many Rubyists, but might be a pretty large departure from how the other SDKs are configured?

ruby/sdk-phase-1.md
cretz: :+1: We definitely plan to have samples showing integrations, but not sure we need an explicit integration for this. Whatever integrations we do have would be optional/additional because it is really important that we have as few dependencies as possible in the primary SDK because they our dependency constraints become users' transitive constraints. I believe we can have only 1 dependency (protobuf) at this time. `temporalio-opentelemetry` is the only additional gem I can think we'll have at first.

ruby/sdk-phase-1.md
cretz: For client connect and initialization (and some others) we are wanting keyword arguments since the vast majority are optional and we're not really wanting multiple ways to configure the client (e.g. allow both block and kwargs). In this case it's more like `Client.connect("localhost:7233", "my-namespace")` which is a shortcut for `Client.new(Connection.new("localhost:7233"), "my-namespace")`. There is a PR coming either here in a couple of hours or early tomorrow that will show this.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+      `Temporalio::Activity::Context.current.heartbeat`. In .NET we made them get current instance first because that's
+      normal there, and in Python we only offered package-level methods because that's normal there.
+    * 💭 Why allow access to an instance? Some users want to pass the activity context to another thread.
+  * 💭 Why `Temporalio::Activity::Context` instead of just `Temporalio::Activity` or putting helpers on base class or
+    something? It is a common requirement in our SDKs to be able to access the activity context from the environment.
+    Also, it is a common requirement to be able to do that from utility methods outside of the explicit activity call.
+* Activity can choose its executor, default is `:default`.
+  * See worker section for more information on executors.
+* Cancellation is reported to the executor.
+  * For thread pool executor (the default), this raises a cancellation exception via `Thread.raise`.
```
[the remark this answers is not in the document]
cretz: Yes, this was a topic of discussion in Python too where this kind of thing is even less common (we have to make a C call to https://docs.python.org/3/c-api/init.html#c.PyThreadState_SetAsyncExc, it's not even in the standard library). We can provide a way to opt-out of this (and we will provide a "shield" way to ensure certain code runs), but we've found it's way more dangerous to let runaway activities continue to run that cannot get interrupted. If there is some other model in the standard library for propagating cancellation to threaded tasks, we'd like to use it.

I will make a note to make sure we use `handle_interrupt` for our exception only and make sure to have tests demonstrate other forms of interrupt.

ruby/sdk-phase-1.md
cretz: > is that referring to the ability to call an activity with something like `MyActivity.execute` rather than `workflow.execute_activity(MyActivity)`?

No, we definitely want the latter. We learned from other SDKs that it is best to not make it look like any local invocation.

> they can't implement activities as procs

Maybe we can have sugar for this, e.g. `class ProcActivity < Temporalio::Activity` and a class method on `Temporalio::Activity` that creates it. But it is the same type of sugar a user can write if we make sure that we support activities as classes.

ruby/sdk-phase-1.md
cretz: Here is the intentionally-limited PR: https://github.com/temporalio/sdk-ruby/pull/155. Would love feedback there, I sprinkled some PR comments throughout.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+  * The search attribute collection is `Temporalio::SearchAttributes` and all things are under it, e.g.
+    `Temporalio::SearchAttributes::Key`.
+    * 💭 Why not `Temporalio::SearchAttribute` module with the collection underneath? The object has value on its own
+      so might as well make it a class.
+  * The base metric is `Temporalio::Metric`, and all metric things are under there, e.g. `Temporalio::Metric::Meter`.
+    * 💭 Why not `Temporalio::Metrics` module? The object has value on its own so might as well make it a class.
+   (e.g. `Temporalio::SearchAttribute::Key`)
+    and all metric things under `Temporalio::Metric` (e.g. `Temporalio::Metric::Meter`).
+  * 💭 Why no `Temporalio::Common`? Ruby expects to have common things where they make sense instead of under a `Common`
+    module.
```
cretz, line 74: We may have to walk this back a bit. It is unreasonable to have `Temporalio::WorkflowIDReusePolicy`, `Temporalio::WorkflowIDConflictPolicy`, etc cluttering the top-level. However, we don't want `RetryPolicy` to be in `Common`. So I guess we may need some kinda hybrid rule that says use the `Common` module for _small_ or _lesser used_ common classes?

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+    be encouraged.
+  * ❓ Is it too confusing that the instance lifecycle is different depending on whether you pass in a class or an
+    instance?
+* `Temporalio::Activity::Context` is a thread/fiber-local context available to activities.
+  * It has instance methods for info, heartbeat, etc.
+  * It has a class method for `current` that returns from thread/fiber-local and class method shortcuts for all instance
+    methods.
+    * 💭 Why class methods equivalents? `Temporalio::Activity::Context.heartbeat` is easier to read than
+      `Temporalio::Activity::Context.current.heartbeat`. In .NET we made them get current instance first because that's
+      normal there, and in Python we only offered package-level methods because that's normal there.
```
cretz, line 395:
```suggestion
  * It has a class method for `current` that returns from thread/fiber-local.
```

I think we are going to walk back this decision for also having class methods. I think it's not really common in Ruby to define class and instance methods of the same name, and `Singleton` and others show that people are ok calling `.current` or something.

ruby/sdk-phase-1.md
```diff
@@ -0,0 +1,496 @@
...
+  * 💭 Why only a single thread? We expect callbacks to Ruby from Rust to be short (hopefully only a block invocation
+    that is a `Queue.push`) so we don't expect making all of these run serial is an issue. However if it does become an
+    issue, we can use multiple threads and a Rust approach which allows multiple channel consumers.
+* Users can create their own runtime with their own telemetry options and such and pass that to clients.
+  * 💭 Why not a singleton? Technically users may want different runtimes, though hopefully not ever. Still, tests do
+    and it doesn't make sense in many cases to have a singleton that a user creates/initializes.
+
+## Clients
+
+`Temporalio::Client` is a namespace-specific client and `Temporalio::Client::Connection` is a namespace-independent
```
mjameswh, line 127: > `Temporalio::Client::Connection` is a namespace-independent connection

Are you still holding on to this statement, given the semantic change introduced by Cloud's API Keys? I'm just curious, not arguing either way.

mjameswh (approved):
Looks good to me.

ruby/sdk-phase-1.md
cretz: Yes. `Connection` is the Rust `Client` and the `CloudOperationsClient` here (as in other SDKs) will leverage the `Connection`. Can see this separation in practice already at `main` of https://github.com/temporalio/sdk-ruby (and in Python SDK and .NET SDK).

cretz:
Merged since all comments settled and we have long since started implementation.

PR #94, opened 2024-08-15
External client configuration proposal

cretz:
# Summary

* This is the proposal for external client configuration (i.e. config files and profiles)
* See it [rendered](https://github.com/cretz/temporal-proposals/blob/external-client-configuration/all-sdk/external-client-configuration.md)
* This is the first round of review
* Like other proposals, all feedback welcome. If the PR discussion gets too large or the review round comes to an end w/ more review needed, it will be closed and a new one will be created referencing this one.

### DIFF
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -24,22 +24,33 @@ What are the reasons for building this?
 
 ### Goals
 
-* Common, well-specified configuration format usable by Temporal tooling for client options
+* Common, well-specified/type-safe configuration format usable by Temporal tooling for client options and usable
+  manually by advanced users
 * Implementation in all Temporal SDKs for reading and writing of this configuration
 * All Temporal CLI/SDKs updated to have easy client creation using this configuration
 * Common on-disk location in every platform for reading/writing this format
 * Ability to provide environment variables to represent values in the configuration
 * Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
 * 0 new dependencies in SDKs
+  * 💭 Why?
+    * Dependencies in our SDKs subject our users to these dependencies and their version constraints, so we should avoid
+      this as much as possible.
+  * ❓ Can we just have an "extension" that has new dependencies?
+    * Yes, but ideally users don't have to add dependencies on extension libraries just to get this functionality.
 * Profile names within the configuration
+  * 💭 Why?
+    * It is common in configurations like this to want to be able to share a single file with many configurations.
+
+### Future Goals
+
+None identified at this time. The goals are simple enough for the MVP and the non-goals are ones we never want.
 
 ### Non-goals
 
 * Common format for declarative workers, schedules, etc. This is only for client options.
-* File-based hierarchy. We don't need many levels of file merging for these options at this time.
-* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
-* Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
-  of course easy).
+* File-based hierarchy. We don't need many levels of file merging for these options.
+* Use external configuration by default in SDKs (it is default in CLI and samples). Users need to opt-in to this for
+  compatibility and clarity reasons (but it is of course easy).
 
 ### High-level Usage Examples
 
@@ -70,7 +81,7 @@ you could do this in environment variables for mTLS auth:
     export TEMPORAL_TLS_CLIENT_CERT_PATH=path/to/my/client.pem
     export TEMPORAL_TLS_CLIENT_KEY_PATH=path/to/my/client.key
 
-Now running the same code again unmodified will run against cloud. You can do ths same thing with config instead of
+Now running the same code again unmodified will run against cloud. You can do this same thing with config instead of
 environment variables:
 
     temporal config set --key address --value my-ns.a1b2c.tmprl.cloud:7233
@@ -97,15 +108,35 @@ Similar to AWS tooling and other tools, you can have profiles. So you can have:
 
 For dev and then in production, this might exist:
 
-    temporal config set --profile prod --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
-    temporal config set --profile prod --key namespace --value my-dev-ns.a1b2c
-    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/dev-cert.pem
-    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/dev-cert.key
+    temporal config set --profile prod --key address --value my-prod-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile prod --key namespace --value my-prod-ns.a1b2c
+    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/prod-cert.pem
+    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/prod-cert.key
 
 Or alternatively it could be a fixed file that that creates or environment variables or whatever.
 
 Now a simple setting of `TEMPORAL_PROFILE` environment to `dev` or `prod` will switch connectivity (or `--profile` on
 CLI or the string in the SDK if you don't want to use environment variable).
+
+#### Manually Working with Configuration
+
+If a user wanted to hand-write a configuration, they can have `my-file.json` like:
+
+```json
+{
+  "profiles": {
+    "default": {
+      "address": "my-host:7233",
+      "namespace": "my-namespace"
+    }
+  }
+}
+```
+
+* This is the same format used by the CLI and the CLI can even be used to edit this file.
+* `TEMPORAL_CONFIG_FILE` env var can point to this file or the config file can be manually set in the CLI/SDK.
+* There is a type safe contract this file conforms to for anyone wanting to work with it programmatically.
+* All SDKs accept the type safe model this file represents if users would rather provide config that way.
 
 #### Possible UI Creation Flow
 
@@ -159,8 +190,8 @@ whether this is done by default can be discussed, but the idea is the same.
       * `namespace` - Client namespace.
       * `apiKey` - Client API key.
       * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
-        upon the client and may change. 💭 Why? We regret not making TLS the default in the past. If TLS is an object,
-        it can have the following possible fields:
+        upon other settings here. 💭 Why? We regret not making TLS the default in the past, so we will make it default
+        if `apiKey` is present. If TLS is an object, it can have the following possible fields:
         * `clientCertPath` - File path to mTLS client cert. Mutually exclusive with `clientCertData`.
         * `clientCertData` - Cert data. Mutually exclusive with `clientCertPath`.
         * `clientKeyPath` - File path to mTLS client key. Mutually exclusive with `clientKeyData`.
@@ -253,6 +284,26 @@ default is assumed.
   * There's no obvious env var approach to array of strings, and it can't be completely comma-delimited because header
     values often contain commas. Would rather not force users to have an index in the env var name. Should we just not
     support gRPC meta from env var until we think this through?
+
+### Loading Configuration
+
+Configuration is loaded in the following manner:
+
+* Try to load profile from configuration file
+  * Only if file loading is enabled
+  * Can If user provides specific config file, use that, otherwise use the default location
+  * If user specifies the profile, use that, otherwise use the default
+* Try to overwrite with specific configuration environment variables
+  * Only if env loading is enabled
+  * If user specifies the profile, use that, otherwise use the default
+  * If the `default` profile is being loaded, try the profile-specific environment variable first, e.g.
+    `TEMPORAL_DEFAULT_ADDRESS` is used over `TEMPORAL_ADDRESS` if it's present.
+  * The `default` env vars are only for the default profile, meaning a profile of `foo` cannot use `TEMPORAL_ADDRESS`,
+    only `TEMPORAL_FOO_ADDRESS` if present.
+
+That is it, nothing more complicated. When the configuration is provided to the SDKs/CLI, they apply their normal logic
+which includes defaults for things that can be defaulted, or errors if they expect something that is not provided (this
+is SDK specific).
 
 ## Implementation
 
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -1,0 +1,340 @@
+# External Client Configuration
+
+Notes for this proposal:
+
+* Everything is subject to change both during proposal time and afterwards during implementation.
+* For each place a decision is made, a "💭 Why X? ..." aside is present explaining why.
+* For each place an open question exists, a "❓ X?" exists to ask it.
+
+## Overview
+
+There are many ways to access Temporal programmatically including CLI, SDK, and other components that leverage those
+such as samples and Terraform providers.
+
+Temporal needs a way for users to provide client configuration without hardcoding in CLI arguments or SDK options. This
+will allow code to seamlessly switch between profiles such as between a dev-server and cloud. The same configuration
+must work across all SDKs/CLIs.
+
+### Goals
+
+* Common, well-specified configuration format usable by Temporal tooling for client options
+* Implementation in all Temporal SDKs for reading and writing of this configuration
+* All Temporal CLI/SDKs updated to have easy client creation using this configuration
+* Common on-disk location in every platform for reading/writing this format
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
+* Profile names within the configuration
+
+### Non-goals
+
+* Common format for declarative workers, schedules, etc. This is only for client options.
+* File-based hierarchy. We don't need many levels of file merging for these options at this time.
+* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
+* Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
+  of course easy).
+
+## Specification
+
+### Values and File Format
+
+* The format is a JSON object with the following fields:
+  * `profiles` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
+      separately-cased profile names that are equal case-insensitively.
+      * `address` - Client address, aka gRPC "host:port".
+      * `namespace` - Client namespace.
+      * `apiKey` - Client API key.
+      * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
+        upon the client and may change. 💭 Why? We regret not making TLS the default in the past. If TLS is an object,
+        it can have the following possible fields:
+        * `clientCertPath` - File path to mTLS client cert. Mutually exclusive with `clientCertData`.
+        * `clientCertData` - Cert data. Mutually exclusive with `clientCertPath`.
+        * `clientKeyPath` - File path to mTLS client key. Mutually exclusive with `clientKeyData`.
+        * `clientKeyData` - Key data. Mutually exclusive with `clientKeyPath`.
+        * `serverCaCertPath` - File path to server CA cert. Mutually exclusive with `serverCaCertData`.
+        * `serverCaCertData` - CA cert data. Mutually exclusive with `serverCaCertPath`.
+        * `serverName` - Override SNI name when connecting to server.
+        * `disableHostVerification` - Boolean to disable verifying TLS server host. May not be available in all Temporal
+          clients. 💭 Why? Not all SDKs offer this feature today.
+      * `codec` - Remote codec information to use for all encoding/decoding of payloads. May not be available in all
+        Temporal clients. 💭 Why? Not all SDKs allow use of a remote codec today.
+        * `endpoint` - Endpoint to the remote codec.
+        * `auth` - Authorization header for the remote codec.
+      * `grpcMeta` - Object representing HTTP headers. Values must be strings for now (can discuss arrays later).
+        ❓ Should we call this `headers`? We do in some cases and not others.
+  * 💭 Why JSON?
+    * Because YAML and TOML and others require a dependency in our SDKs. AWS for example has a homegrown INI-ish
+      implementation they have had to build in to every language.
+    * We accept that while JSON is technically human-authorable, it's not as easy as other formats. This is acceptable
+      since we will be providing tooling to mutate the config.
+    * ❓ Should we consider "json with comments" and some kind of very basic pre-parser in each language that strips
+      them? This could come later and not part of MVP.
+* This whole format will be defined in proto in the API repo in the `sdk` package.
+  * 💭 Why?
+    * Clear, centralized contract.
+    * Our SDKs already use proto dependency and this gives them stable JSON without requiring separate JSON dependency.
+    * Users can use the proto tooling to read/write configs programmatically since they are just proto JSON.
+* Do not validate that all keys are known in the file.
+  * 💭 Why?
+    * Originally this section was strict validation, but it became clear that we should allow newer-CLI-created config
+      files that may have newer keys to work with older SDKs.
+    * Similarly, it became clear that we're not validating environment variables this way, why config file keys?
+    * We want the tools that mutate this file to have strong validation on the keys being added, so hopefully this
+      avoids the concern of key typos.
+* All keys are optional in the file. Ideally in the final merged form, we'd always require `address` and we always
+  require `namespace` in namespace-specific clients (i.e. non cloud ops API), but it is unreasonable since we must fit
+  within SDK defaults.
+  * Originally this section did say `address` and `namespace` were required, but as SDK implementations came about, it
+    became clear that can't be done easily, so CLI/SDK defaults need to remain.
+* The default config file location is `~/.config/temporalio/temporal.json` for all platforms
+  * So in Windows this is often `C:\Users\MyUsername\.config\temporalio\temporal.json`
+  * ❓ Should we skip the subdir or make it `~/.config/temporal.json`? Unsure if there are any future files that would
+    go in here.
+  * Existing CLI used `~/.config/temporalio/temporal.yaml`
+    * ❓ Is this confusing for them? Are we concerned about them vs what is best moving forward?
+
+### Environment variables
+
+* Environment variables are in the form `TEMPORAL_<PROFILE_>_KEY_SUB_KEY_CAMEL_CASE`. If the profile part is missing,
+default is assumed.
+  * 💭 Why `TEMPORAL` prefix?
+    * Reasonable to differentiate.
+  * This means that `TEMPORAL_ADDRESS` and `TEMPORAL_DEFAULT_ADDRESS` are the same thing.
+  * This means that camel case parts are separated by underscores _except_ the profile name itself.
+    * So a config value of `profiles.myProfile.tls.clientCertPath` is `TEMPORAL_MYPROFILE_TLS_CLIENT_CERT_PATH`.
+    * 💭 Why?
+      * This is clearer for typing and it is unreasonable to try and break up user-supplied camel case.
+  * Also the profile may contain an underscore and retain that and match that first. So
+    `profiles.my_profile_tls.tls.clientCertPath` is `TEMPORAL_MY_PROFILE_TLS_TLS_CLIENT_CERT_PATH`.
+    * 💭 Why?
+      * It is unreasonable to elide any characters in the user-provided profile name.
+* Environment variables are case-insensitive.
+* The CLI has existing environment variables. Some of these work and some should be still supported but deprecated in
+  favor of the new forms. Below are the CLI env vars and how they apply:
+  * `TEMPORAL_ADDRESS` - works just fine.
+  * `TEMPORAL_NAMESPACE` - works just fine.
+  * `TEMPORAL_API_KEY` - works just fine.
+  * `TEMPORAL_TLS` - works just fine.
+  * `TEMPORAL_TLS_CERT`, `TEMPORAL_TLS_CERT_DATA`, and other mTLS client values - need to be deprecated and replaced
+    with newer forms (only slight changes here).
+    * 💭 Why not just have the `tls` config field be `certData` instead of `clientCertData` so this doesn't have to
+      change?
+      * Not being clear this is for mTLS client use is confusing to users. And we'd have to change the path one anyways
+        because the CLI is not suffixed with path.
+  * `TEMPORAL_TLS_X` for all other values - ones that don't match need to be deprecated and replaced with newer forms.
+  * `TEMPORAL_CODEC_ENDPOINT` - works just fine.
+  * `TEMPORAL_CODEC_AUTH` - works just fine.
+  * 💭 Why deprecate some instead of making all our clients support these variables that exist in CLI?
+    * It is more sane/reasonable to have a simple to understand config-field-to-env-var format than it is to have these
+      special environment variables be inconsistent outliers.
+    * For non-CLI uses (i.e. SDKs), it is unreasonable to burden them with past CLI choices.
+* There are some special environment variables respected by clients:
+  * `TEMPORAL_PROFILE` - the name of the profile to load if one is not provided at load time. The default is  `default`.
+  * `TEMPORAL_CONFIG_FILE` - the path to the config file. The default is `~/.config/temporalio/temporal.json`.
+* ❓ How do to grpc-meta as env var? Like for `My-Header: My-Value`
+  * Env var names are case-insensitive and don't use dashes, so we can't put header name in env var name. And we may not
+    want to canonicalize the header names even though most gRPC clients do it for you.
+  * There's no obvious env var approach to array of strings, and it can't be completely comma-delimited because header
+    values often contain commas. Would rather not force users to have an index in the env var name. Should we just not
+    support gRPC meta from env var until we think this through?
+
+## Implementation
+
+### CLI
+
+* CLI will accept `--profile`.
+  * This deprecates `--env`/`--env-file`
+    * 💭 Why deprecate?
+      * The `env` name is confusing with environment variables. Note we do not use the term "environment" anywhere in
+        this document except for environment variables.
+      * The `env` setting uses YAML files in `~/.config/temporalio/temporal.yaml`.
+    * Error if both options are present, or in the case of default, if there are default files for both.
+      * 💭 Why error in the case of both default env and profile being present instead of merge?
+        * This is the new way and we should not try to support both at the same time, it can get confusing.
+  * The default is `default`, and this can also be set via `TEMPORAL_PROFILE` env var.
+* CLI will accept `--config-file`.
+  * Default is `~/.config/temporalio/temporal.json`, also can be set via `TEMPORAL_CONFIG_FILE` env var.
+* CLI will have a whole new set of `profile` commands that operate similar to `env` commands today.
+  * Will not go into detail in this proposal, but it's very similar.
+  * We will strictly validate keys when setting whereas we did not with `env`.
+* Unlike `env`/`--env`, this does not blindly apply any field as any CLI argument.
+  * 💭 Why?
+    * We want strict validation of config values. The idea that you can set a `workflow-id` as part of this profile does
+      not make sense. These are client options only, declarative options for other things can be discussed separately.
+* CLI should defer to the Go SDK for loading/using the config/profile.
+
+### SDKs
+
+SDKs must support loading client configuration from optional profile name and optional config name. This needs to
+include a way to make sure that user-defaults can be set.
+
+Possible language-specific ideas are below. Note, they are not exact and the naming is all subject to change, they are
+simply ideas for discussion. Many of the approaches below are based on several rewritings after seeing how they don't
+work in some SDKs. The overall approach is to just make client and connection options loadable from configuration
+without overthinking shortcuts or merging or hierarchies.
+
+#### Go SDK Idea
+
+Some decisions here are explained here even though they apply to later SDKs. 
+
+* New function in `client`: `LoadConfig(LoadConfigOptions) (*proto.Config, error)`.
+  * This is mostly a helper if people want it.
+* `LoadConfigOptions` has the following:
+  * `ConfigFile string` - Override the file to use.
+  * `DisableFile bool` - Disable reading from file.
+  * `DisableEnv bool` - Disable reading from env vars.
+* New function in `client`: `LoadOptionsFromConfig(string profile, LoadConfigOptions) (Options, error)`.
+  * 💭 Why require profile, isn't there a default of `default`?
+    * Yes, and maybe that's what empty string is too, but in general Go doesn't have good parameter defaults and we
+      don't want a differently named overload just for this.
+* New function in `client`: `NewOptionsFromConfig(*proto.ConfigProfile) (Options, error)`.
+  * 💭 Why is this needed if `LoadOptionsFromConfig` exists?
+    * People need a way to provide the config to load from instead of always assuming it can be loaded.
+  * 💭 Why not have the config proto to use in the `LoadConfigOptions` instead?
+    * It gets to be a bit of a merge game if you have a `BaseConfig` there, and it's a bit confusing if you have
+      `PreloadedConfig` there which implies all other options don't matter.
+* 💭 Why not just have the load config options on the client options and let dialing load?
+  * People need to adjust options _after_ loaded (overwrite defaults, etc).
+* 💭 Why not have `(*Options).LoadFromConfig` helper method on the options instead?
+  * The merging logic gets far too complicated to know what to overwrite and what not to. It's better to force the user
+    to start from config and handle merging themselves if they need that.
+* Go SDK _does_ support codec settings because it supports remote codecs.
+  * If the remote codec setting is set, the data converter option will be set.
+  * 💭 Why not have a `PayloadCodecFromConfig` type of thing?
+    * This is easy enough for users to do with the `LoadConfig` helper and the remote codec objects.
+    * Most won't need it because regular `LoadOptionsFromConfig` creates it with default payload conversion. Only those
+      that need customized payload conversion will need to create the remote codec themselves.
+* Options like grpc-meta sets header provider, API key sets credentials, etc.
+  * 💭 Why make a user that just wants to _add_ a header on top of config have to use `LoadConfig` and do it the hard
+    way?
+    * There's no clean way to support single header adding on top of existing ones in the Go SDK today.
+
+Simplest example:
+
+```go
+options, err := client.LoadOptionsFromConfig("default")
+if err != nil {
+  return err
+}
+cl, err := client.Dial(options)
+```
+
+#### Java SDK Idea
+
+* New static `ServiceStubsOptions.loadConfig()` and
+  `ServiceStubsOptions.loadConfig(@Nullable String configFile, boolean disableFile, boolean disableEnv)` that return the
+  full protobuf `Config` object.
+  * ❓Any better place to put this? It's a proto utility, not a service stub utility.
+* New static `WorkflowServiceStubsOptions.loadBuilderFromConfig()` and
+  `WorkflowServiceStubsOptions.loadBuilderFromConfig(@Nullable String profile, @Nullable configFile, ` +
+  `boolean disableFile, boolean disableEnv)` and
+  `WorkflowServiceStubsOptions.newBuilderFromConfig(ConfigProfile profile)`.
+  * 💭 Why not have these loaders as instance methods on the builder?
+    * Same reason as Go, because the merging logic of knowing what to do with existing data is too complicated and it's
+      best for the user to do that as needed.
+* Need the same for operator and cloud service stubs options.
+* Need same methods for `WorkflowClientOptions`.
+  * Unfortunately with how Java works, there are separate client options, and they too can be in config.
+  * ❓ Any alternative suggestions here? We could have some kind of shortcut to do multiple steps here.
+* ❓ Does Java have a concept of a remote codec?
+
+Simplest example:
+
+```java
+var stubsOptions = WorkflowServiceStubsOptions.loadBuilderFromConfig().build();
+var stubs = WorkflowServiceStubs.newServiceStubs(stubsOptions);
+var clientOptions = WorkflowClientOptions.loadBuilderFromConfig().build();
+var client = WorkflowClient.newInstance(stubs, clientOptions);
+```
+
+#### TypeScript SDK Idea
+
+* New `client` module function `loadConfig(options?)` where the `options` type is
+  `{ configFile?: string, disableFile?: bool, disableEnv?: bool }` and it returns a `proto.Config`.
+* New `client` module function `loadOptionsFromConfig(options?)` where the optional `options` are
+  either `proto.ConfigProfile` or `{ profile?: string, configFile?: string, disableFile?: bool, disableEnv?: bool }` and
+  it returns a `ConnectionOptions & ClientOptions`.
+  * Could also have it return `{ connectionOptions: ConnectionOptions, clientOptions: ClientOptions }` if that makes
+    more sense than a union.
+* Consider a new async `Client.connect` static method that accepts `ConnectionOptions & ClientOptions` and is a
+  shortcut to the two-step process today.
+* TypeScript SDK _does not_ support codec settings and will error if seen in config.
+  * 💭 Why error?
+    * Because a user with a profile configured with a codec should expect that codec to be used, not silently ignored.
+
+Simplest example:
+
+```typescript
+const options = loadOptionsFromConfig();
+const connection = Connection.connect(options);
+const client = new Client({ connection, ...options });
+
+// Or maybe:
+
+const { connectionOptions, clientOptions } = loadOptionsFromConfig();
+const connection = Connection.connect(connectionOptions);
+const client = new Client({ connection, ...clientOptions });
+```
+
+#### Python SDK Idea
+
+* In `service` module, new function `load_config` with kwarg parameters of `config_file: Optional[str] = None`,
+  `disable_file: bool = False`, `disable_env: bool = False` and it returns a `proto.Config`.
+* In `service` module, new static method `ConnectConfig.load_from_config` with positional parameter of
+  `profile: str = 'default'` and kwarg parameters of `config_file: Optional[str] = None`, `disable_file: bool = False`,
+  and `disable_env: bool = False` and it returns a `ConnectConfig`.
+* In `service` module, new static method `ConnectConfig.from_config` that accepts a single positional parameter of
+  `proto.ConfigProfile`.
+* In `client` module, new static methods for `ClientConfig.load_from_config` and `ClientConfig.from_config` same as
+  above two for service module.
+* In `client` module, new `TypedDict` class called `ClientConnectConfig` that extends `ClientConfig` but adds all the
+  other parameters for `connect`.
+  * On that class, new static methods for `load_from_config` and `from_config` that are like the others.
+  * This config can mutated and splatted as the args to `Client.connect`.
+    * 💭 Why not a separate `connect` call purely for config?
+      * Many users want to set other options after config loads, and like with other SDKs adding config loading at the
+        same time has option setting has many merge problems, so we need them to be separate steps.
+* Like TypeScript, Python SDK _does not_ support codec settings and will error if seen in config.
+
+Simplest example:
+
+```python
+config = ClientConnectConfig.load_from_config()
+client = await Client.connect(**config)
+```
+
+#### .NET SDK Idea
+
+* New static methods on `TemporalConnectionOptions` for `LoadConfig()` and
+  `LoadConfig(string? configFile, bool disableFile, bool disableEnv)` and it returns `proto.Config`.
+  * ❓Like Java, any better place to put this? It's a proto utility, not an options utility.
+* New static methods on `TemporalConnectionOptions` for `LoadFromConfig()`,
+  `LoadFromConfig(string profile = "default", string? configFile, bool disableFile, bool disableEnv)`.
+  and `LoadFromConfig(proto.ConfigProfile)` that all return `TemporalConnectionOptions`.
+* Same static methods on `TemporalClientConnectOptions`.
+* Same static methods on `TemporalClientOptions`.
+* Like TypeScript, .NET SDK _does not_ support codec settings and will error if seen in config.
+
+Simples example:
+
+```csharp
+var options = TemporalClientConnectOptions.LoadFromConfig();
+var client = await TemporalClient.ConnectAsync(options);
+```
+
+### Samples
+
+* All samples will be updated to load from config and default to local dev server w/ default namespace.
+  * In samples we try to avoid shared code. In .NET and Python, there is no default target host, and in Ruby there will
+    not even be a default namespace. So every sample in these languages is going to have to not only have the
+    load-from-config code, but also conditionally apply the target host as `localhost:7233` if not set in external
+    config.
+
+### UI
+
+* The UI could offer, say on API key creation, the ability to download a config file for use.
+  * Since we require all profiles in a single file in the default place, the downloaded file would be just including the
+    `default` profile and the instructions would tell them where to place it or they could provide this config file
+    in any client at connection option time.
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -15,6 +15,13 @@ Temporal needs a way for users to provide client configuration without hardcodin
 will allow code to seamlessly switch between profiles such as between a dev-server and cloud. The same configuration
 must work across all SDKs/CLIs.
 
+### Why?
+
+What are the reasons for building this?
+
+* Users want to switch environments without changing their code, but today they cannot.
+* Our samples should be able to be run on multiple environments including cloud, but today they are localhost only.
+
 ### Goals
 
 * Common, well-specified configuration format usable by Temporal tooling for client options
@@ -34,6 +41,111 @@ must work across all SDKs/CLIs.
 * Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
   of course easy).
 
+### High-level Usage Examples
+
+#### Simple Switch from Local to Cloud
+
+Similar to `temporal env` today, you can use the CLI to manage configurations with `temporal config`.
+
+For example, say you had this Go code for starting a workflow:
+
+```go
+options, err := client.LoadOptionsFromConfig("default")
+if err != nil {
+  return err
+}
+cl, err := client.Dial(options)
+run, err := cl.ExecuteWorkflow(
+  ctx,
+  client.StartWorkflowOptions{ID: "my-id", TaskQueue: "my-task-queue"},
+  MyWorkflow,
+)
+```
+
+By default this only works on `localhost:7233` with namespace `default` when you run it. But if you were a cloud user,
+you could do this in environment variables for mTLS auth:
+
+    export TEMPORAL_ADDRESS=my-ns.a1b2c.tmprl.cloud:7233
+    export TEMPORAL_NAMESPACE=my-ns.a1b2c
+    export TEMPORAL_TLS_CLIENT_CERT_PATH=path/to/my/client.pem
+    export TEMPORAL_TLS_CLIENT_KEY_PATH=path/to/my/client.key
+
+Now running the same code again unmodified will run against cloud. You can do ths same thing with config instead of
+environment variables:
+
+    temporal config set --key address --value my-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --key namespace --value my-ns.a1b2c
+    temporal config set --key tls.clientCertPath --value path/to/my/cert.pem
+    temporal config set --key tls.clientKeyPath --value path/to/my/cert.key
+
+Now running the same code again unmodified will run against cloud. Or in a Temporal future with API keys and a
+known/fixed API endpoint:
+
+    temporal config set --key cloudApiKey --value my-api-key
+
+And that would be all that is needed (but that is a future scenario, note `cloudApiKey` is not a defined key today, but
+`apiKey` is, but you have to put in the `address` and `namespace` still).
+
+#### Profile Switching
+
+Similar to AWS tooling and other tools, you can have profiles. So you can have:
+
+    temporal config set --profile dev --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile dev --key namespace --value my-dev-ns.a1b2c
+    temporal config set --profile dev --key tls.clientCertPath --value path/to/my/dev-cert.pem
+    temporal config set --profile dev --key tls.clientKeyPath --value path/to/my/dev-cert.key
+
+For dev and then in production, this might exist:
+
+    temporal config set --profile prod --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile prod --key namespace --value my-dev-ns.a1b2c
+    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/dev-cert.pem
+    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/dev-cert.key
+
+Or alternatively it could be a fixed file that that creates or environment variables or whatever.
+
+Now a simple setting of `TEMPORAL_PROFILE` environment to `dev` or `prod` will switch connectivity (or `--profile` on
+CLI or the string in the SDK if you don't want to use environment variable).
+
+#### Possible UI Creation Flow
+
+When a user creates an API key, the completion screen can provide a download of the config file that will have the
+address, namespace, and API key set. So it can say something like:
+
+```
+API key is created, make sure you copy it here because it is not stored:
+
+    <some long api key string>
+
+Alternatively, you can download it as a configuration file:
+
+    <some button or something that downloads a file name temporalio.json>
+
+You can place this configuration file in the default place used by CLIs/SDKs:
+
+* macOS - `$HOME/Library/Application Support/temporalio/temporal.json`
+* Windows - `%AppData%/temporalio/temporal.json` (often `C:\Users\<myuser>\AppData\Roaming\temporalio\temporal.json`)
+* Linux - `$HOME/.config/temporalio/temporal.json`
+
+This will be automatically consumed by the SDKs/CLI. This file can also be put anywhere else and `TEMPORAL_CONFIG_FILE`
+environment variable can be set to point to it.
+```
+
+#### Possible CLI Creation Flow
+
+Assuming cloud support is present in `temporal` CLI, you can have the normal certificate generation steps, e.g.
+
+```
+temporal cloud gen ca ...
+temporal cloud gen leaf ...
+temporal cloud namespace add-accepted-client-ca ...
+```
+
+And that last step can by default set a `temporal-cloud` profile in the configuration with the address, namespace, and
+mTLS certificates. So now all you have to do to use cloud in CLI is either `--profile temporal-cloud` or set
+`TEMPORAL_PROFILE=temporal-cloud` environment variable. Granted you may want better certificate management than that and
+whether this is done by default can be discussed, but the idea is the same.
+
 ## Specification
 
 ### Values and File Format
@@ -43,7 +155,7 @@ must work across all SDKs/CLIs.
     * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
       for environment variables which are case insensitive. It is a validation error to have a config file with two
       separately-cased profile names that are equal case-insensitively.
-      * `address` - Client address, aka gRPC "host:port".
+      * `address` - Client address, aka gRPC "host:port". This cannot be a URL, it must be `host:port`.
       * `namespace` - Client namespace.
       * `apiKey` - Client API key.
       * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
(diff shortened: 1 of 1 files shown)


all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+    * We want the tools that mutate this file to have strong validation on the keys being added, so hopefully this
+      avoids the concern of key typos.
+* All keys are optional in the file. Ideally in the final merged form, we'd always require `address` and we always
+  require `namespace` in namespace-specific clients (i.e. non cloud ops API), but it is unreasonable since we must fit
+  within SDK defaults.
+  * Originally this section did say `address` and `namespace` were required, but as SDK implementations came about, it
+    became clear that can't be done easily, so CLI/SDK defaults need to remain.
+* The default config file location is `~/.config/temporalio/temporal.json` for all platforms
+  * So in Windows this is often `C:\Users\MyUsername\.config\temporalio\temporal.json`
+  * ❓ Should we skip the subdir or make it `~/.config/temporal.json`? Unsure if there are any future files that would
```
cretz, line 93:
```suggestion
  * ❓ Should we skip the subdir and make it `~/.config/temporal.json`? Unsure if there are any future files that would
```

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+* Environment variables are in the form `TEMPORAL_<PROFILE_>_KEY_SUB_KEY_CAMEL_CASE`. If the profile part is missing,
+default is assumed.
+  * 💭 Why `TEMPORAL` prefix?
+    * Reasonable to differentiate.
+  * This means that `TEMPORAL_ADDRESS` and `TEMPORAL_DEFAULT_ADDRESS` are the same thing.
+  * This means that camel case parts are separated by underscores _except_ the profile name itself.
+    * So a config value of `profiles.myProfile.tls.clientCertPath` is `TEMPORAL_MYPROFILE_TLS_CLIENT_CERT_PATH`.
+    * 💭 Why?
+      * This is clearer for typing and it is unreasonable to try and break up user-supplied camel case.
+  * Also the profile may contain an underscore and retain that and match that first. So
```
cretz, line 109: We should also limit the characters a profile name can have where we can on mutation (e.g. in the CLI) and say profiles with invalid characters in the name will fail when used in any client.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+  of course easy).
+
+## Specification
+
+### Values and File Format
+
+* The format is a JSON object with the following fields:
+  * `profiles` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
```
dandavison, line 44:
```suggestion
      for environment variables which are case insensitive on Windows. It is a validation error to have a config file with two
```

dandavison, line 44:
```suggestion
      for environment variables which are case insensitive on Windows. It is a validation error to have a config file with two
```

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+  * `TEMPORAL_CODEC_ENDPOINT` - works just fine.
+  * `TEMPORAL_CODEC_AUTH` - works just fine.
+  * 💭 Why deprecate some instead of making all our clients support these variables that exist in CLI?
+    * It is more sane/reasonable to have a simple to understand config-field-to-env-var format than it is to have these
+      special environment variables be inconsistent outliers.
+    * For non-CLI uses (i.e. SDKs), it is unreasonable to burden them with past CLI choices.
+* There are some special environment variables respected by clients:
+  * `TEMPORAL_PROFILE` - the name of the profile to load if one is not provided at load time. The default is  `default`.
+  * `TEMPORAL_CONFIG_FILE` - the path to the config file. The default is `~/.config/temporalio/temporal.json`.
+* ❓ How do to grpc-meta as env var? Like for `My-Header: My-Value`
```
cretz, line 136:
```suggestion
* ❓ How to do grpc-meta as env var? Like for `My-Header: My-Value`
```

all-sdk/external-client-configuration.md
Sushisource: Shouldn't this be `%APPDATA%` on windows? I think that's a bit more conventional.

Otherwise, yes I would definitely keep the `temporalio` subdir

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+      * `apiKey` - Client API key.
+      * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
+        upon the client and may change. 💭 Why? We regret not making TLS the default in the past. If TLS is an object,
+        it can have the following possible fields:
+        * `clientCertPath` - File path to mTLS client cert. Mutually exclusive with `clientCertData`.
+        * `clientCertData` - Cert data. Mutually exclusive with `clientCertPath`.
+        * `clientKeyPath` - File path to mTLS client key. Mutually exclusive with `clientKeyData`.
+        * `clientKeyData` - Key data. Mutually exclusive with `clientKeyPath`.
+        * `serverCaCertPath` - File path to server CA cert. Mutually exclusive with `serverCaCertData`.
+        * `serverCaCertData` - CA cert data. Mutually exclusive with `serverCaCertPath`.
```
dandavison, line 57: Camel case is getting awkward and hard to predict here (people might think it's "CA"). Is there a strong reason for it? Did you consider `server-ca-cert-data`, `api-key` and `client-cert-path`?

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+  `LoadConfig(string? configFile, bool disableFile, bool disableEnv)` and it returns `proto.Config`.
+  * ❓Like Java, any better place to put this? It's a proto utility, not an options utility.
+* New static methods on `TemporalConnectionOptions` for `LoadFromConfig()`,
+  `LoadFromConfig(string profile = "default", string? configFile, bool disableFile, bool disableEnv)`.
+  and `LoadFromConfig(proto.ConfigProfile)` that all return `TemporalConnectionOptions`.
+* Same static methods on `TemporalClientConnectOptions`.
+* Same static methods on `TemporalClientOptions`.
+* Like TypeScript, .NET SDK _does not_ support codec settings and will error if seen in config.
+
+Simples example:
```
cretz, line 320:
```suggestion
Simplest example:
```

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+        * `serverCaCertData` - CA cert data. Mutually exclusive with `serverCaCertPath`.
+        * `serverName` - Override SNI name when connecting to server.
+        * `disableHostVerification` - Boolean to disable verifying TLS server host. May not be available in all Temporal
+          clients. 💭 Why? Not all SDKs offer this feature today.
+      * `codec` - Remote codec information to use for all encoding/decoding of payloads. May not be available in all
+        Temporal clients. 💭 Why? Not all SDKs allow use of a remote codec today.
+        * `endpoint` - Endpoint to the remote codec.
+        * `auth` - Authorization header for the remote codec.
+      * `grpcMeta` - Object representing HTTP headers. Values must be strings for now (can discuss arrays later).
+        ❓ Should we call this `headers`? We do in some cases and not others.
```
Sushisource, line 66: :shrug: I kinda lean towards headers since it's maybe slightly more what people are used to, but, meta is more accurate

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+  * 💭 Why JSON?
+    * Because YAML and TOML and others require a dependency in our SDKs. AWS for example has a homegrown INI-ish
+      implementation they have had to build in to every language.
```
Sushisource, line 69: It might be worth seeing if all of our SDKs actually already include some YAML/TOML dep transitively. If we do, I'd say let's go for one of those, but, JSON is also fine.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+    * We accept that while JSON is technically human-authorable, it's not as easy as other formats. This is acceptable
+      since we will be providing tooling to mutate the config.
+    * ❓ Should we consider "json with comments" and some kind of very basic pre-parser in each language that strips
+      them? This could come later and not part of MVP.
```
Sushisource, line 73: I think so because then we can provide a template. We could also provide a JSON schema def

all-sdk/external-client-configuration.md
cretz: For our use case, environment variables are considered case insensitive everywhere (see the next section).

all-sdk/external-client-configuration.md
cretz: This is proper JSON formatting of key names, but we can use hyphens if we prefer them. They are translated to language idiomatic forms except for the CLI but yeah we don't want to have CLI translating from one to the other. I don't mind hyphen-based.

cretz: Yes, this is proper JSON formatting of key names. And especially for our case, since it will be proto JSON, there is no real way to do hyphens well. Most users won't use the file directly, but if we are concerned with how this looks in the CLI, maybe the CLI can translate to hyphens in either direction? It is important that any user with any lang can just read this in as proto JSON and immediately have a well-typed object to mutate, and we can't get that with hyphens.

all-sdk/external-client-configuration.md
cretz: > Shouldn't this be %APPDATA% on windows? I think that's a bit more conventional.

Yes, I do think that is better and is what https://pkg.go.dev/os#UserConfigDir does even though CLI today used https://pkg.go.dev/os#UserHomeDir because it had to be compatible with `tctl`. I will change.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+    * For non-CLI uses (i.e. SDKs), it is unreasonable to burden them with past CLI choices.
+* There are some special environment variables respected by clients:
+  * `TEMPORAL_PROFILE` - the name of the profile to load if one is not provided at load time. The default is  `default`.
+  * `TEMPORAL_CONFIG_FILE` - the path to the config file. The default is `~/.config/temporalio/temporal.json`.
+* ❓ How do to grpc-meta as env var? Like for `My-Header: My-Value`
+  * Env var names are case-insensitive and don't use dashes, so we can't put header name in env var name. And we may not
+    want to canonicalize the header names even though most gRPC clients do it for you.
+  * There's no obvious env var approach to array of strings, and it can't be completely comma-delimited because header
+    values often contain commas. Would rather not force users to have an index in the env var name. Should we just not
+    support gRPC meta from env var until we think this through?
```
Sushisource, line 141: Could just pass an escaped json object? I'm fine w/ delaying that until later.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+
+### SDKs
+
+SDKs must support loading client configuration from optional profile name and optional config name. This needs to
+include a way to make sure that user-defaults can be set.
+
+Possible language-specific ideas are below. Note, they are not exact and the naming is all subject to change, they are
+simply ideas for discussion. Many of the approaches below are based on several rewritings after seeing how they don't
+work in some SDKs. The overall approach is to just make client and connection options loadable from configuration
+without overthinking shortcuts or merging or hierarchies.
```
Sushisource, line 176: Not mentioned here but the actual loading and parsing of the file in Core SDKs should be done by Core I think, so that we can centralize:

* Logic of default profile / file location
* Merging of env vars
* Read/write the files

Since it's already a proto core can just return that after loading / merging.

all-sdk/external-client-configuration.md
Sushisource: Well, only Java and Core - see my other comment about having loading happen through Core. Pretty easy to do in Core.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+## Specification
+
+### Values and File Format
+
+* The format is a JSON object with the following fields:
+  * `profiles` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
+      separately-cased profile names that are equal case-insensitively.
+      * `address` - Client address, aka gRPC "host:port".
```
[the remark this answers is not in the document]
cretz: I don't think it should include a URI scheme, it should be the same `host:port` format accepted by all SDKs/CLI IMO. There are some cases like in Go where you can set `dns:///` for custom resolver support and such, but I am afraid to make general statements on the general contract concerning that.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+        * `clientCertPath` - File path to mTLS client cert. Mutually exclusive with `clientCertData`.
+        * `clientCertData` - Cert data. Mutually exclusive with `clientCertPath`.
+        * `clientKeyPath` - File path to mTLS client key. Mutually exclusive with `clientKeyData`.
+        * `clientKeyData` - Key data. Mutually exclusive with `clientKeyPath`.
+        * `serverCaCertPath` - File path to server CA cert. Mutually exclusive with `serverCaCertData`.
+        * `serverCaCertData` - CA cert data. Mutually exclusive with `serverCaCertPath`.
+        * `serverName` - Override SNI name when connecting to server.
+        * `disableHostVerification` - Boolean to disable verifying TLS server host. May not be available in all Temporal
+          clients. 💭 Why? Not all SDKs offer this feature today.
+      * `codec` - Remote codec information to use for all encoding/decoding of payloads. May not be available in all
```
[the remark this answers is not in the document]
cretz: We have chosen to call this `codec` in CLI args which this is (kinda) modeled off of. I think if we want to call this external codec argument "remote codec" we should do it there too.

all-sdk/external-client-configuration.md
cretz: I am leaning there as well, but I would need to suggest a CLI update for that.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
Sushisource: Then let's just explicitly say it doesn't have one I think

all-sdk/external-client-configuration.md
cretz: I can say I know multiple don't. In fact, it's even worse, not all SDKs even have a JSON parser. The only thing they all have is protobuf (which includes protobuf JSON). Here's my understanding:

* Go - Proto yes, JSON stdlib, YAML yes transitively (stupid testify), TOML no
* Java - Proto yes, JSON no (in serviceclient where we need it), YAML no, TOML no
* TypeScript - Proto yes, JSON stdlib, YAML no, TOML no
* Python - Proto yes, JSON stdlib, YAML doubtful, TOML doubtful
* .NET - Proto yes, JSON stdlib, YAML no, TOML no

Sure we can have Core bake in these things for the last 3, but Java is still hurt here, and I think proto JSON works well here.

all-sdk/external-client-configuration.md
cretz: Yeah, can come later as I don't expect anyone to hand edit these anymore than they do AWS config files. We'd have to hand-write the JSON comment remover.

all-sdk/external-client-configuration.md
cretz: Yes, that can work but is a bit rough, might be better to just put header name in env var, treat env var underscores as hyphens in header name, and lowercase them all since gRPC and HTTP/2 do that anyways.

all-sdk/external-client-configuration.md
cretz: Makes sense, but I don't think there's a writing step to centralize, and I still expect users to be able to use the protos themselves if they want. And I think people want to manipulate the results, so I think we should centralize all of those things in Core, but not the conversion of the proto result to the client options, that should remain in lang.

Sushisource: Yeah agree

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,340 @@
...
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
+* Profile names within the configuration
+
+### Non-goals
+
+* Common format for declarative workers, schedules, etc. This is only for client options.
+* File-based hierarchy. We don't need many levels of file merging for these options at this time.
+* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
```
lorensr, line 33: I'd prioritize authoring DX / YAML over 0 deps

cretz: It's really bad for libraries to force a dependency on all users of a library just for an optional side thing like this, that's really rough DX (we already suffer it so much with proto versions and that's an actual needed dependency).

all-sdk/external-client-configuration.md
cretz: Updated

all-sdk/external-client-configuration.md
cretz: Added outstanding question for this

all-sdk/external-client-configuration.md
cretz: Fixed

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: Clarified

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+    temporal config set --profile dev --key namespace --value my-dev-ns.a1b2c
+    temporal config set --profile dev --key tls.clientCertPath --value path/to/my/dev-cert.pem
+    temporal config set --profile dev --key tls.clientKeyPath --value path/to/my/dev-cert.key
+
+For dev and then in production, this might exist:
+
+    temporal config set --profile prod --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile prod --key namespace --value my-dev-ns.a1b2c
+    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/dev-cert.pem
+    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/dev-cert.key
```
cretz, line 103:
```suggestion
    temporal config set --profile prod --key address --value my-prod-ns.a1b2c.tmprl.cloud:7233
    temporal config set --profile prod --key namespace --value my-prod-ns.a1b2c
    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/prod-cert.pem
    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/prod-cert.key
```

josh-berry (commented):
Sorry this took some time to get to.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+### Goals
+
+* Common, well-specified configuration format usable by Temporal tooling for client options
+* Implementation in all Temporal SDKs for reading and writing of this configuration
+* All Temporal CLI/SDKs updated to have easy client creation using this configuration
+* Common on-disk location in every platform for reading/writing this format
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
+* Profile names within the configuration
```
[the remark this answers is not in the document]
cretz: > Why do we need this? 

Multiple profiles in the same configuration makes it easier to edit from CLI and have a single config to ship around with multiple profiles if you want.

> Can we not achieve the same thing in a simpler fashion by providing a way to set the configuration file path from the environment (for example)?

We offer this _too_, but the default is a single place.

> The only value I can think of for having them is in the CLI itself, where it's convenient to be able to refer to different environments by short names rather than full config file paths. But I'm not sure if that use case needs to bleed into the SDKs.

CLI and SDKs need to share the same approach here. A single config file that contains multiple profiles helps all clients IMO.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+### High-level Usage Examples
+
+#### Simple Switch from Local to Cloud
+
+Similar to `temporal env` today, you can use the CLI to manage configurations with `temporal config`.
+
+For example, say you had this Go code for starting a workflow:
+
+```go
+options, err := client.LoadOptionsFromConfig("default")
```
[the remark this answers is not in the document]
cretz: You can't make optional parameters in Go. This was discussed above:

* 💭 Why require profile, isn't there a default of `default`?
  * Yes, and maybe that's what empty string is too, but in general Go doesn't have good parameter defaults and we don't want a differently named overload just for this.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+    temporal config set --key address --value my-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --key namespace --value my-ns.a1b2c
+    temporal config set --key tls.clientCertPath --value path/to/my/cert.pem
+    temporal config set --key tls.clientKeyPath --value path/to/my/cert.key
+
+Now running the same code again unmodified will run against cloud. Or in a Temporal future with API keys and a
+known/fixed API endpoint:
+
+    temporal config set --key cloudApiKey --value my-api-key
```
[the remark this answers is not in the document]
cretz: > yuck, I do not like having cloud and non-cloud config keys

This does not exist today FYI. But it's an example because this is not just a cloud config key, this is saying "default to all the other cloud things".

> I'm wary of even providing "buillt-in" defaults for cloud because I think they will have more confidence/comfort in being able to see exactly what's different.

A single known API domain and that TLS defaults to true are reasonable cloud defaults.

> If we must have different API keys because of talking to the control plane vs. data plane (or whatever)

Again this does not exist today, and that's not why we need different API keys. The reason we want different _config_ values is because there are things you can assume about cloud that you may not be able to assume about others. Regardless, this is just a sample of something we could do in the future. In the meantime, you can use `apiKey` (and always will be able to, even for cloud).

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+And that would be all that is needed (but that is a future scenario, note `cloudApiKey` is not a defined key today, but
+`apiKey` is, but you have to put in the `address` and `namespace` still).
```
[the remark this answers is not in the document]
cretz: Ok, will delete it

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+  * `profiles` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
+      separately-cased profile names that are equal case-insensitively.
+      * `address` - Client address, aka gRPC "host:port". This cannot be a URL, it must be `host:port`.
+      * `namespace` - Client namespace.
+      * `apiKey` - Client API key.
+      * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
+        upon the client and may change. 💭 Why? We regret not making TLS the default in the past. If TLS is an object,
+        it can have the following possible fields:
```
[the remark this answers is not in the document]
cretz: > yikes. I do not like this idea of inconsistent defaults between clients. This seems like a big footgun and security issue waiting to happen.

Yeah, I should have worded this better. I mean it's dependent upon the other config values. We have chosen to make TLS defaulted to false everywhere _unless_ API key is provided and it's true everywhere. It's not per client. It's a bigger footgun we're already seeing when people forget to set `--tls` when they use API keys.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+    * Clear, centralized contract.
+    * Our SDKs already use proto dependency and this gives them stable JSON without requiring separate JSON dependency.
+    * Users can use the proto tooling to read/write configs programmatically since they are just proto JSON.
+* Do not validate that all keys are known in the file.
+  * 💭 Why?
+    * Originally this section was strict validation, but it became clear that we should allow newer-CLI-created config
+      files that may have newer keys to work with older SDKs.
+    * Similarly, it became clear that we're not validating environment variables this way, why config file keys?
+    * We want the tools that mutate this file to have strong validation on the keys being added, so hopefully this
+      avoids the concern of key typos.
```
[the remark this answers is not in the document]
cretz: > I don't think we can assume users will use only our tools to mutate config.

It's not a strong assumption, it's just one of the multiple reasons for lack of strict key validation on config load

> What are the consequences of a silent failure (e.g. due to mistyped key), and are we sure that we're not going to add some key in future that might have significant consequences if missing / if a default is inadvertently used?

Depends on the key.

> Also, we should consider providing optional strict validation even if we decide not to do it by default, so users can at least discover problems in a straightforward way.

This adds surface area to loading APIs. Can you suggest the APIs for how this looks? Should we add an option to every way that a config is loaded that fails on unrecognized key just for this use case? Are we sure this feature is worth it on MVP?

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+* All keys are optional in the file. Ideally in the final merged form, we'd always require `address` and we always
+  require `namespace` in namespace-specific clients (i.e. non cloud ops API), but it is unreasonable since we must fit
+  within SDK defaults.
+  * Originally this section did say `address` and `namespace` were required, but as SDK implementations came about, it
+    became clear that can't be done easily, so CLI/SDK defaults need to remain.
+* The default config file location is `<app-config-dir>/temporalio/temporal.json` for all platforms
+  * This is as defined by https://pkg.go.dev/os#UserConfigDir which, as of this writing, is the logic defined in docs
+    and in code at https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/file.go;l=528.
+  * Existing CLI used `~/.config/temporalio/temporal.yaml`
+    * ❓ Is this confusing for them? Are we concerned about them vs what is best moving forward?
```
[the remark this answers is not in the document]
cretz: I don't think there's much value here in specifying advanced migration here. I think a documented migration is fine. This is only a few keys, not some large thing that needs some kind of transform CLI call. I think we can add something like this if it's needed, but we need to me more cautious adding features to an initial version that doesn't even exist.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+### Environment variables
+
+* Environment variables are in the form `TEMPORAL_<PROFILE_>_KEY_SUB_KEY_CAMEL_CASE`. If the profile part is missing,
+default is assumed.
```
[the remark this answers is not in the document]
cretz: I don't think we should restrict underscores, but I put a question about this later about some limits we should put on names. I cannot think of a way that it would collide today, though maybe future values can. I don't think it's worth harming experience for this collision.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+  * 💭 Why `TEMPORAL` prefix?
+    * Reasonable to differentiate.
+  * This means that `TEMPORAL_ADDRESS` and `TEMPORAL_DEFAULT_ADDRESS` are the same thing.
```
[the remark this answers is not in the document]
cretz: IMO the profile specific env var is preferred over the default. I can add a blurb.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+      change?
+      * Not being clear this is for mTLS client use is confusing to users. And we'd have to change the path one anyways
+        because the CLI is not suffixed with path.
+  * `TEMPORAL_TLS_X` for all other values - ones that don't match need to be deprecated and replaced with newer forms.
+  * `TEMPORAL_CODEC_ENDPOINT` - works just fine.
+  * `TEMPORAL_CODEC_AUTH` - works just fine.
+  * 💭 Why deprecate some instead of making all our clients support these variables that exist in CLI?
+    * It is more sane/reasonable to have a simple to understand config-field-to-env-var format than it is to have these
+      special environment variables be inconsistent outliers.
+    * For non-CLI uses (i.e. SDKs), it is unreasonable to burden them with past CLI choices.
```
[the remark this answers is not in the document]
cretz: It is better to be inconsistent with the CLI than let a CLI's inconsistencies affect new code in SDKs IMO. Are you saying we should make all SDKs work with old CLI env vars?

> I think maintaining different variants of different env variables will be confusing and frustrating for users

The older ones should probably not be documented. But unless you suggest burdening all SDKs ever with some just-went-GA-CLI env vars, I don't see a better solution. Can you suggest your solution?

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+* ❓ How to do grpc-meta as env var? Like for `My-Header: My-Value`
+  * Env var names are case-insensitive and don't use dashes, so we can't put header name in env var name. And we may not
+    want to canonicalize the header names even though most gRPC clients do it for you.
+  * There's no obvious env var approach to array of strings, and it can't be completely comma-delimited because header
+    values often contain commas. Would rather not force users to have an index in the env var name. Should we just not
+    support gRPC meta from env var until we think this through?
+
+## Implementation
+
+### CLI
```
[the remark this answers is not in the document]
cretz: Correct, this was discussed above (and if we happen to have any other clients that use their own made-up/inconsistent env vars, but luckily I think it's only CLI)

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+```
+
+And that last step can by default set a `temporal-cloud` profile in the configuration with the address, namespace, and
+mTLS certificates. So now all you have to do to use cloud in CLI is either `--profile temporal-cloud` or set
+`TEMPORAL_PROFILE=temporal-cloud` environment variable. Granted you may want better certificate management than that and
+whether this is done by default can be discussed, but the idea is the same.
+
+## Specification
+
+### Values and File Format
```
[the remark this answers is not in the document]
cretz: Ok. We are starting to bloat this proposal, but I expected it to be the case as separate reviewers need their own additional sections before approval.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+var stubsOptions = WorkflowServiceStubsOptions.loadBuilderFromConfig().build();
+var stubs = WorkflowServiceStubs.newServiceStubs(stubsOptions);
+var clientOptions = WorkflowClientOptions.loadBuilderFromConfig().build();
+var client = WorkflowClient.newInstance(stubs, clientOptions);
+```
+
+#### TypeScript SDK Idea
+
+* New `client` module function `loadConfig(options?)` where the `options` type is
+  `{ configFile?: string, disableFile?: bool, disableEnv?: bool }` and it returns a `proto.Config`.
```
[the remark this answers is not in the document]
cretz: Not sure I agree even in TS. I like having to explicit disable defaults anywhere.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+SDKs must support loading client configuration from optional profile name and optional config name. This needs to
+include a way to make sure that user-defaults can be set.
+
+Possible language-specific ideas are below. Note, they are not exact and the naming is all subject to change, they are
+simply ideas for discussion. Many of the approaches below are based on several rewritings after seeing how they don't
+work in some SDKs. The overall approach is to just make client and connection options loadable from configuration
+without overthinking shortcuts or merging or hierarchies.
+
+#### Go SDK Idea
```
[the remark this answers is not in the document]
cretz: This is unnecessarily expanding the scope at first IMO...

> newConfigWithDefaults()

That's just creating a new Go config

> mergeConfig()

We have explicitly avoided any merging logic exposed to users because we do not support merging of configs because earlier versions of this proposal tried and it became too complicated

> readFromEnv()

This is what loading while disabling file does

> readFromFile()

This is what loading while disabling env does. And regardless, this is why we chose proto JSON. So people can use proto reading, writing, merging, etc.

> "build the config yourself"

Yes, this is one of the motivators behind proto JSON. You have a typesafe representation of the config you can build yourself or do anything you want with.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+* Users want to switch environments without changing their code, but today they cannot.
+* Our samples should be able to be run on multiple environments including cloud, but today they are localhost only.
+
+### Goals
+
+* Common, well-specified configuration format usable by Temporal tooling for client options
+* Implementation in all Temporal SDKs for reading and writing of this configuration
+* All Temporal CLI/SDKs updated to have easy client creation using this configuration
+* Common on-disk location in every platform for reading/writing this format
```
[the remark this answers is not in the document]
drewhoskins-temporal: Some of these goals seem more must-have than others.  Can we categorize them into P0/P1/P2?

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
```
drewhoskins-temporal, line 33: This is broad and seems like it precludes using any prebuilt configuration packages, which seems like it should probably be part of the discussion.  What are you really trying to avoid, how critical is that (P0/P1/P2), and why?
Can there be layering where we have an extension that has dependencies?

all-sdk/external-client-configuration.md
drewhoskins-temporal: One of the scenarios Candace highlighted is for folks who are migrating production and dev environment code to cloud.  (We should align on whether that's a goal (P0/P1/P2)).

For such a use case, we might want to provide--and/or allow users to provide--extensions that allow configuration in whatever their system is, whether yaml or otherwise.
Also, if it's opt in as you say below, I don't see taking dependencies as a P0 problem.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+* Profile names within the configuration
+
+### Non-goals
+
+* Common format for declarative workers, schedules, etc. This is only for client options.
+* File-based hierarchy. We don't need many levels of file merging for these options at this time.
+* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
+* Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
```
drewhoskins-temporal, line 41: This is for the SDK, right?  Our samples would use it by default?

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+)
+```
+
+By default this only works on `localhost:7233` with namespace `default` when you run it. But if you were a cloud user,
+you could do this in environment variables for mTLS auth:
+
+    export TEMPORAL_ADDRESS=my-ns.a1b2c.tmprl.cloud:7233
+    export TEMPORAL_NAMESPACE=my-ns.a1b2c
+    export TEMPORAL_TLS_CLIENT_CERT_PATH=path/to/my/client.pem
+    export TEMPORAL_TLS_CLIENT_KEY_PATH=path/to/my/client.key
```
drewhoskins-temporal, line 71: I'm confused how these relate to "default"

all-sdk/external-client-configuration.md
drewhoskins-temporal: In any case, please reword this.  What "extreme focus on human authoring of this format" means is in the eye of the beholder.  Might be better to put something about human authoring features as a goal with a lower priority.  Then we can align on degrees of importance rather than stark goal/non-goal status.

all-sdk/external-client-configuration.md
drewhoskins-temporal: Could users get sniped by the default lack of isolation by using these bash commands or the commands below?  (I know I personally have.)  Would rather see a file whose impacts are scoped by default rather than the equivalent of global variables.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+`apiKey` is, but you have to put in the `address` and `namespace` still).
+
+#### Profile Switching
+
+Similar to AWS tooling and other tools, you can have profiles. So you can have:
+
+    temporal config set --profile dev --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile dev --key namespace --value my-dev-ns.a1b2c
+    temporal config set --profile dev --key tls.clientCertPath --value path/to/my/dev-cert.pem
+    temporal config set --profile dev --key tls.clientKeyPath --value path/to/my/dev-cert.key
```
drewhoskins-temporal, line 96: It would be nicer if users weren't required to run imperative commands to do config for customers who lack such hooks.  I.e. they could just declare the values statically and then the SDK can parse that.

I also see a lot of "it worked on my box" bugs with these sorts of imperative configs where somebody has already run script A and then runs script B, which works fine.  Whereas another dev checks out the code and only runs script B and it doesn't work.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+Temporal needs a way for users to provide client configuration without hardcoding in CLI arguments or SDK options. This
+will allow code to seamlessly switch between profiles such as between a dev-server and cloud. The same configuration
+must work across all SDKs/CLIs.
+
+### Why?
+
+What are the reasons for building this?
+
+* Users want to switch environments without changing their code, but today they cannot.
+* Our samples should be able to be run on multiple environments including cloud, but today they are localhost only.
```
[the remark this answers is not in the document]
cretz: I suspect because it's an easier transition for them. If we have an environment we promote for easy use, it seems reasonable to make samples run there easily. Also as I understand it, for many "digital natives" (or saas-first or cloud-first or whatever you'd like to call them), they'd go straight to cloud without a local step as they do with many other products they use. Also, running a sample against cloud allows them to confirm they have their cloud namespace setup properly.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: I don't think these can be separate as they apply to the system as a whole. I don't think there's a clear split point with these goals that can be done in the future. The MVP is meant to be a very simple approach to accomplishing the goals, and everything else can be done in the future. The only place where I think work can be split is by language. Otherwise, these goals all amount to basically one feature.

all-sdk/external-client-configuration.md
cretz: > Also, if it's opt in as you say below, I don't see taking dependencies as a P0 problem.

I think you're confusing opt-in from a caller POV and opt-in from a dependency POV. Someone can depend on our library and still have to opt-in to calling something on it, but just the act of depending on the library brings in dependencies. Dependencies are not opt-in, only the use of the call.

> Might be better to put something about human authoring features as a goal with a lower priority. Then we can align on degrees of importance rather than stark goal/non-goal status.

The ideal of the non-goal was to be a never priority. For this project there just happened to be no "future goals" like there are in other projects because meeting the goals is so simple as to not have any. Can I have priority infinite?

all-sdk/external-client-configuration.md
cretz: > What are you really trying to avoid, how critical is that (P0/P1/P2), and why?

Trying to avoid any new dependencies in our libraries that then forces uses to depend on them transitively and locks them into our version constraints. We have seen this issue with the libraries we do have to depend on (e.g. protobuf) where by using our libraries, we are forcing our users' apps to be bound by version constraints we have. Libraries like ours should avoid dependencies they don't need because they are transitively infectious to a user's entire application.

>  Can there be layering where we have an extension that has dependencies?

Yes, and this is supported in all SDKs. In Python it's just an "extra" you have to enable, but for all others these extensions with extra dependencies are entirely separate libraries they have to depend on. The tradeoff is that they now have to add a separate dependency just for configuration support which I was hoping to avoid. Ideally this is as easy as possible for users to consume, and adding a separate library for them to depend on (or force an extra upon install) is not as easy as not having to do it. To compare with AWS SDK, their configuration format they write parsers for in every language, but we don't have to go that far.

I will add some clarity to the document explaining this. I think the burden is on justifying the addition of a dependency, not maintaining the lack of one.

all-sdk/external-client-configuration.md
cretz: Right. I will clarify. The samples will be written to opt-in to configuration. Also, the CLI will use external configuration by default.

all-sdk/external-client-configuration.md
cretz: > I'm confused how these relate to "default"

As @josh-berry requested, I will add a section clarifying how configuration is read from environment variables and the file. In this case, `TEMPORAL_ADDRESS` is the env var equivalent of `TEMPORAL_DEFAULT_ADDRESS` which is the equivalent of setting the `address` for profile `default` which is the equivalent of providing `--address` on the CLI or setting it in the SDK client options.

> Could users get sniped by the default lack of isolation by using these bash commands or the commands below? Would rather see a file whose impacts are scoped by default rather than the equivalent of global variables.

This is your choice as a user, and we even allow env vars to be disabled when loading in an SDK, but we can't just eschew environment variables for this use case because people may set them accidentally.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,454 @@
...
+
+By default this only works on `localhost:7233` with namespace `default` when you run it. But if you were a cloud user,
+you could do this in environment variables for mTLS auth:
+
+    export TEMPORAL_ADDRESS=my-ns.a1b2c.tmprl.cloud:7233
+    export TEMPORAL_NAMESPACE=my-ns.a1b2c
+    export TEMPORAL_TLS_CLIENT_CERT_PATH=path/to/my/client.pem
+    export TEMPORAL_TLS_CLIENT_KEY_PATH=path/to/my/client.key
+
+Now running the same code again unmodified will run against cloud. You can do ths same thing with config instead of
```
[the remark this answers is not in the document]
cretz: Fixed.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
josh-berry: > It is better to be inconsistent with the CLI than let a CLI's inconsistencies affect new code in SDKs IMO. Are you saying we should make all SDKs work with old CLI env vars?

No I'm not.

> Can you suggest your solution?

Your solution is fine but incomplete—all I am saying is we need to deprecate the old variables in CLI so CLI is consistent with everything else.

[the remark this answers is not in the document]
cretz: > all I am saying is we need to deprecate the old variables in CLI so CLI is consistent with everything else.

I agree. The line you are commenting on is specifically under a bullet explaining why we are deprecating them.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
josh-berry: I don't agree this is expanding scope—you have just said we already have 4/5 of the things I'm asking for but presented in a different (IMO less composable) form. And the 5th, `mergeConfig()`, is something we will have to implement internally anyway if we want consistent behavior when an env variable overrides a file, or when a file overrides built-in defaults. It just needs to be exposed.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
josh-berry: Will close this in favor of the comment above.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
josh-berry: > > What are the consequences of a silent failure (e.g. due to mistyped key), and are we sure that we're not going to add some key in future that might have significant consequences if missing / if a default is inadvertently used?
> 
> Depends on the key. It's hard to speculate what an older SDK not recognizing a newer config key is because we can't predict what future keys will be added here.

Right, so we can't know if an unknown key is safe to ignore or not, which makes me think that strict validation might actually be quite important. I will preemptively say that I understand we can't protect users against every misconfiguration, but protecting against a mis-typed field name, or recognizing and flagging some unknown config that might be critical, both feel like pretty common issues that can easily be surfaced/avoided.

Also, adding strict validation and relaxing it later is much easier than skipping it now and adding it later. We're much less likely to get complaints about error messages that don't happen.

> > Also, we should consider providing optional strict validation even if we decide not to do it by default, so users can at least discover problems in a straightforward way.
> 
> This adds surface area to loading APIs. Can you suggest the APIs for how this looks? Should we add an option to every way that a config is loaded that fails on unrecognized key just for this use case? Are we sure this feature is worth it on MVP?

Sure, for each `LoadConfig()` function, accept an optional `strict` flag that defaults to `true` (or `DisableStrict` flag that defaults to `false`, if you must, for languages that don't support defaulting to `true`). Though I think I may have talked myself into adding strict validation as part of the MVP, and removing it later if users complain, in which case no new API surface is necessary.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
josh-berry: Yeah I have no opinion on what form the migration should take, merely that it should exist.

all-sdk/external-client-configuration.md
```diff
@@ -254,6 +285,26 @@ default is assumed.
...
+  * Only if file loading is enabled
+  * Can If user provides specific config file, use that, otherwise use the default location
+  * If user specifies the profile, use that, otherwise use the default
+* Try to overwrite with specific configuration environment variables
+  * Only if env loading is enabled
+  * If user specifies the profile, use that, otherwise use the default
+  * If the `default` profile is being loaded, try the profile-specific environment variable first, e.g.
+    `TEMPORAL_DEFAULT_ADDRESS` is used over `TEMPORAL_ADDRESS` if it's present.
+  * The `default` env vars are only for the default profile, meaning a profile of `foo` cannot use `TEMPORAL_ADDRESS`,
+    only `TEMPORAL_FOO_ADDRESS` if present.
```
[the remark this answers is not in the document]
cretz: LGTM, will incorporate on my next commit

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: >> CLI and SDKs need to share the same approach here.
> Why

Because the same configuration needs to work in all Temporal clients (SDKs and CLIs).

> The CLI has some use cases/scenarios that don't apply to the (other) SDKs, like being used for operational/admin tasks, possibly in multiple environments.

Sure, and Java has scenarios that don't apply to Python and so on. But this configuration should apply to both. I think AWS CLI vs SDK is a decent model here. I think consistency should be the default and we need a really good reason to be inconsistent.

> I am struggling to come up with a "multi-profile use case" that can't also easily be handled with multi-file. Can you provide an example?

Sure, technically everything you can do with a single file you can do with multi file, so you will struggle if you're trying to see a situation where we _must_ force multi file. Same goes with any config that has separate sections. For instance, you could say you're struggling to come up with a use case why the `tls` config option can't also easily be handled with multi-file. Why not have the TLS file separate, wouldn't this make downloading it and sharing it easier?

The reason we have nested keys in a configuration is for convenience of not having to ship around multiple files or deal with multiple files. You can load up an entire file at once and mutate multiple profiles. You can onboard a new dev by giving a single file with all profiles in it pointing to different clusters. It is easier to glance at all configuration in a file. The entire configuration for use in an SDK can be one set of bytes (some users may not use files, they may provide the bytes some other way). Etc. The same reasons you may want more than one profile in a file extend to wanting more than one key in a file.

I think justifying not using a separate file for separate keys (whether those keys are profiles or just nested options like `tls`) is more important. AWS and others put all profiles in a single file and I think we should too. When Temporal says "configuration" it should be a single thing, whether that's a proto object, a set of bytes in memory, a set of bytes on disk, etc. To have separate profiles be separate objects, separate sets of bytes, separate disk files, etc is a bit rough.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: > Are you saying that setting cloudApiKey (or any other cloud-specific setting) would hypothetically auto-magically say "you're running against cloud now" and change other config values (e.g. apiKey) automatically?

Yes

> I hate it when libraries do that.

I mean we could have just removed api key and had authentication headers. But we recognize shortcuts are important for ergonomics. We may in the future be able to accept a single value from a user to run on cloud (whether in the CLI, SDK, config, etc) and we should consider it.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: That the TypeScript SDK doesn't have it is not emblematic of whether being able to disable a default behavior via a disable flag is idiomatic. And definitely not good enough justification to make TypeScript unique in the option to disable default things while loading.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: > you have just said we already have 4/5 of the things I'm asking for

I was afraid of the "etc" there and other potential things that may be requested before an initial version is delivered

> And the 5th, mergeConfig(), is something we will have to implement internally anyway if we want consistent behavior when an env variable overrides a file, or when a file overrides built-in defaults. It just needs to be exposed.

The initial version did have advanced merging logic where you could merge two configs, but this was hard to reason about. Specific variables override specific variables, this is not the same as merging though. With configuration options, especially in programmatic form, there is not a difference between unset and false. So how do you know if someone turned off TLS? Can you write an example implementation of `mergeConfig` using Go structures or proto objects? You may find it is challenging because merging programmatic objects is not the same as overwriting specifically set values that are specifically provided. The difference is subtle but important IMO. This is why we require that you create a new config from the configuration instead of "apply" configuration to an existing object.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: I guess it depends on if you want backwards and forwards compatibility with configuration. Most configuration formats I'm aware of are not strict in this way (including AWS), but I don't have a strong enough opinion to continue to resist this, so it will be added.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: :+1: I think defining how the configuration is documented (and/or migrations, tutorials, samples, etc) is out of scope for this proposal and part of project planning instead.

all-sdk/external-client-configuration.md
dandavison: TOML is becoming very popular among respected projects for configuration. To make sure we've considered all possibilities, let's add vendoring TOML parsers as something to consider. We'd have to keep up with any security patches, but otherwise these should be stable.

The most important thing here is DX when using the config files. And we've got to assume that new settings will be added to these config files over time, as we add new product features. As mentioned elsewhere in the proposal, the lack of comments in JSON is extremely undesirable. And of course, the comma-requirements result in very bad DX for anyone who is not using an autoformatting editor-extension for JSON.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,505 @@
...
+  * 💭 Why?
+    * It is common in configurations like this to want to be able to share a single file with many configurations.
+
+### Future Goals
+
+None identified at this time. The goals are simple enough for the MVP and the non-goals are ones we never want.
+
+### Non-goals
+
+* Common format for declarative workers, schedules, etc. This is only for client options.
```
dandavison, line 50: Let's go slowly here. In the future, if we want to support declarative configuration of other aspects of Temporal, then we will not want multiple config files. The natural design decision at this point is to say

> We're currently designing for client configuration only. But the design we've chosen allows the possibility that other top-level keys will be added in the future.

In which case client config should be under its own top-level key, allowing other top-level keys in the future for all the things we're deliberately not designing or thinking about today: settings, schedules, etc

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
dandavison: Agree with @cretz.

- We may want to make it possible for someone to check out the repo in a VSCode cloud workspace and run samples. (E.g. because they like to work that way, or they are working on a non-conventional development environment such as a tablet or phone).
- We may want to create a playground app where users can run and edit our samples.
- An organization may be testing out their cloud namespace and want to use samples for that.
- An organization may be onboarding Temporal users to cloud namespaces and want them to try running samples against their cloud namespace.

all-sdk/external-client-configuration.md
cretz: :+1: Vendoring is an option.

Another challenge is type-safe contracts. With JSON we have a form of type-safe contract across every SDK - protobuf. It supports nesting, documentation, merging, serialization/deserialization, unhandled field support, requires 0 new dependencies, and there's a precedent w/ our usage of it in similar ways (e.g. workflow history). With TOML, the models will have to be separately maintained in every language. Obviously totally doable, just a series of tradeoffs.

Also, I think it's worth discussing how much we expect configuration to be hand-edited. I don't think many people hand-edit AWS SDK/CLI configuration (I have admittedly been using them as a reference here because there's not a lot of SDK configurations out there). Why do we think more people will hand-edit our SDK/CLI config vs that SDK/CLI config? Are we making blind assumptions here? Also, I do know that vscode and other similar tools have no problem with hand-editing JSON (with comments), is their hadn-edited configuration story unacceptable for us?

all-sdk/external-client-configuration.md
cretz: I think declarative resources should be tackled separately. Having said that, we can probably add non-client things in the future. But I'm not sure we need to qualify client things. The reason this is CLI/SDK configuration and not client configuration is because you need these things to basically do anything with the CLI/SDK. From a user POV it's just "config". If we want to add things that do not apply to clients (e.g. worker configuration) that may make sense, but I still think these general options can be top level.

I think asking all users to add this `client.` prefix for our own personal future proofing is a bit rough. IMO it is a worse developer experience to make users use `TEMPORAL_CLIENT_ADDRESS` and `client.address` and what not over `TEMPORAL_ADDRESS` and `address`. These latter have been in use for years already. I don't thin we should harm developer experience here over a perceived future.

all-sdk/external-client-configuration.md
dandavison: > Also, I do know that vscode and other similar tools have no problem with hand-editing JSON (with comments), is their hadn-edited configuration story unacceptable for us?

IMO JSON works as a choice for VSCode because they wrapped the parser to allow comments and tolerate trailing commas, which we could do, but that's more technical hassle than vendoring a TOML parser. (Also in VSCode they know that many users will have an autoformatter extension such as `prettier` installed, which makes the experience even better).

all-sdk/external-client-configuration.md
dandavison: > I think declarative resources should be tackled separately. Having said that, we can probably add non-client things in the future.

The proposed file name is `<app-config-dir>/temporalio/temporal.json`, so this file clearly must include all temporal config. And that's a good thing! We don't want multiple files.

> But I'm not sure we need to qualify client things.

In the proposed schema, all the keys related to client config are already nested below `profiles`, and `name`. There are currently 6 of those keys: `address`, `namespace`, `api_key`, `tls`, `codec`, `grpcMeta`.

But 6 is just the number in the first iteration of the proposal; it's bound to grow over time, and the result will be that the other sections we add to the file will be their own root nodes, swimming oddly among a sea of client config details. If we look at file formats such as TOML and TOML-ish things like `~/.gitconfig`, there are no top-level keys.

I can only see advantage and no disadvantage in grouping the client config under its own key (e.g. `clientConfig`), especially since we're already two levels in. Can you list the disadvantages you see with that?

all-sdk/external-client-configuration.md
dandavison: As I suggested above, we should consider vendoring also, so it would be up to the SDK/CLI whether they wish to vendor or use a dependency.

To make it concrete, here's an example of what it might look like in TOML. Not supporting multiple profiles, since whether or not we want that is a separate question that we're discussing. As I've said above, I think there's a strong argument for single-profile per file.

```toml
[client]
address = "localhost:7233"
namespace = "default"
api_key = "<my-key>"
codec = { auth = "<my-headers>", endpoint = "https://my-codec-server.com" }

[client.tls]
client_cert_path = "/path/to/certs/client/cert.pem"
client_key_path = "/path/to/certs/client/key.pem"
server_ca_cert_path = "/path/to/certs/ca/cert.pem"

[client.grpc_meta]
my_grpc_header_1 = "some-value"
my_grpc_header_2 = "some-value"
```

all-sdk/external-client-configuration.md
dandavison: If hyphens don't work well then let's consider underscores. This is of course subjective and up to preference, but it's a design decision we have to make. I don't usually see camel case in config files -- can you give examples?

> Yes, this is proper JSON formatting of key names.

Not sure what you meant by "proper".

all-sdk/external-client-configuration.md
cretz: > will be that the other sections we add to the file will be their own root nodes, swimming oddly among a sea of client config details. 

But there are no other sections today. I don't agree we should harm developer experience today for some may-never-happen unrealized future.

> Can you list the disadvantages you see with that?

I mentioned this in the last paragraph above:

> I think asking all users to add this `client.` prefix for our own personal future proofing is a bit rough. IMO it is a worse developer experience to make users use `TEMPORAL_CLIENT_ADDRESS` and `client.address` and what not over `TEMPORAL_ADDRESS` and `address`. These latter have been in use for years already. I don't think we should harm developer experience here over a perceived future that may not come about.

I don't want everyone to qualify these things as client-specific config just because there is technically the possibility of some non-client future.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: > Regarding development contexts, consider an organization with a large monorepo containing many different projects that make use of Temporal. The different projects in the monorepo will want the ability to commit independent Temporal config files in their own subdirectories of the monorepo.

I do not agree with this statement. These configurations are not code configurations they are environment configurations. I can't imagine any reason to commit Temporal config files any more than AWS CLI/SDK config files. In fact, I believe it is important that these configuration values remain environment config only.

I would expect the dozens of configuration options that a user can set in the SDK to remain set in the SDK. Notice I didn't add all of the client options as configuration options. I only added ones that would be environment specific and have value not being in code or committed. Overrides and defaults can still be part of code. This configuration is only meant for non-committed data that changes from one environment to the next.

all-sdk/external-client-configuration.md
cretz: > that's more technical hassle than vendoring a TOML parser

I am not sure this is true

all-sdk/external-client-configuration.md
cretz: > Not sure what you meant by "proper".

JSON formatting, specifically of protobuf in our case but as a best practice for JSON in general, expects camel case key names.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
dandavison: Yes, I think this is the key thing we need to establish first. If this configuration file is just configuring the client connection, then it's just environment and wouldn't typically be committed.

The thing is that the proposal currently has the file name as `.../temporal.json` or `.../temporal.toml`. And that raises the following questions

1. Will Temporal want to support file-based configuration of things other than client connections?
2. If so, should the client connection be a stanza within a broader config file?

I understand that you'd prefer not to think about unknown things in the future, but if we're naming the file `temporal.{json,toml}` then unfortunately we don't have that option, since we'd be taking the name that a future general Temporal config file would use. (If we were to name it `temporal-client-config.{json,toml}`, then we could perhaps design things so that in the future it can also be supplied via a stanza within a more general `temporal.{json,toml}`).

Temporal aims to be a very general application development framework, and already has a fairly large product surface area. It would be a very bold statement at this point in history to say "Temporal won't ever want any file-based config beyond client connection configuration."

Concretely, what possibilities can we think of right now? Any individual point I list we'll probably be able to come up with criticisms of (i.e. why we might think it's not needed in file-based config), but the point is that there are already possibilities, and those will only increase as time passes; it seems likely to me that the name `temporal.{json,config}` will be wanted in the future for config that includes non-client-connection things.

- dev-server
  - start-up flags: always enable $experimental_feature via dynamic config flags
  - enable labs / experimental features in UI
  - file-based persistence config


- telemetry
  - configure collectors uniformly across all SDKs 
  - temporal developers may want to configure the same collector for server and SDK-side traces to obtain views of the entire workflow lifecycle
  - set SDK logging options so that they apply uniformly across CLI and any SDK


- CLI
  - dark mode vs light mode
  - enable/disable color/hyperlinks/decorations
  - hyperlink format (e.g. support creation of hyperlinks by the CLI that route to a different UI for viewing workflows) 


- Application gRPC logic?
  - I'm not sure about this -- it seems useful for development, but dangerous/inappropriate to risk having dev defaults applied in production.
  - Perhaps users would want to globally set certain "policy"-type options that are used in multiple gRPCs APIs.
    - E.g. workflow ID reuse/conflict policies? (I know I personally always want `TERMINATE_IF_RUNNING` when working on samples, but at the same time can't actually commit it to my samples).

all-sdk/external-client-configuration.md
dandavison: Well I was certainly about to respond with "Citation?" :) I do see that in Google's style guide and some others, which I'd forgotten. But there are multiple entire language communities that don't use camel case for anything -- there's no blanket statement to be made about JSON keys.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: > Will Temporal want to support file-based configuration of things other than client connections?

Not to my knowledge. But maybe the better way to ask is "Will Temporal want to support file-based configuration of things other than environment-specific configuration?" because I think the environment-specific configuration is so minimal it can remain top-level. I think the answer to that question is "no".

> If so, should the client connection be a stanza within a broader config file?

I don't believe there is any concept of broader config for this use case. This is basically a file that can be used as an alternative to environment variables.

> It would be a very bold statement at this point in history to say "Temporal won't ever want any file-based config beyond client connection configuration."

I don't think it's too bold to say "Temporal won't ever want to reuse this file/env-based config for anything beyond environment configuration (which is only client so far)".

> dev-server

I can't think of anything about the dev server that belongs in this configuration (nor can I for any other CLI arg that is not already included in this doc). This is not a general purpose configuration, this is per-environment configuration. Basically anything you might use env vars for.

> telemetry

This I can see being environment configuration and I think it could live right alongside the current environment configuration. In fact, unlike the client options that are opt-in to taking from environment configuration, I can see making these implicitly loaded (there are caveats about reading from disk by default in libraries, but can be discussed).

> CLI

Not this kind of configuration IMO. Not only do we not want any CLI-specific configuration here IMO, but CLI args don't always qualify, including ones that affect output.

> Application gRPC logic?

I don't want to confuse that in this kind of configuration. We do have ways to set a lot of options, but those options are not part of what I would consider this configuration for these environment specific things. Sure one can say "but we want to have different conflict policy per environment", well sure anyone can technically say "we want to have different <option> per environment" and they are more than welcome to, but I don't think that goes in _this_ config.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,505 @@
...
+* Common on-disk location in every platform for reading/writing this format
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
+  * 💭 Why?
+    * Dependencies in our SDKs subject our users to these dependencies and their version constraints, so we should avoid
+      this as much as possible.
+  * ❓ Can we just have an "extension" that has new dependencies?
+    * Yes, but ideally users don't have to add dependencies on extension libraries just to get this functionality.
+* Profile names within the configuration
```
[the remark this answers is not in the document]
cretz: The goal is "single file configuration", I just worded it differently here

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
dandavison: > This is basically a file that can be used as an alternative to environment variables.

> This is not a general purpose configuration, this is per-environment configuration.

> Basically anything you might use env vars for

> Not only do we not want any CLI-specific configuration here IMO

> I don't want to confuse that in this kind of configuration. We do have ways to set a lot of options, but those options are not part of what I would consider this configuration for these environment specific things.

> this file/env-based config

> I don't think that goes in this config.

In all of the above statements you're repeating an assertion -- but it's precisely that assertion that I'm questioning, so repeating the assertion doesn't add anything. The logic here (and I do think I made it clear above) is that

1. **Premise**: if Temporal will want to use file-based config for non-client-connection things in the future, then we cannot have connection-only config squatting on the name `temporal.json`

2. Therefore, we can only use that name if Temporal will not seek to use file-based config for non-client-connection-things in the future.

So we need to establish whether we'll want to use file-based config for things other than client connection. That's why I'm encouraging others to think about this / brainstorm what uses we might have for file-based config in the future. I thought about it a bit and have given some points above suggesting that the answer might be "yes, we'll want to have file-based config for things other than client connection".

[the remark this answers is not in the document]
cretz: > if Temporal will want to use file-based config for non-client-connection things in the future, then we cannot have connection-only config squatting on the name temporal.json

I contend Temporal does not want to use file-based config for non-_environment_ configuration, and _environment_ configuration is currently all client config and in the future probably also all client config or almost surely the vast majority (maybe one or two runtime config options).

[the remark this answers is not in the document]
dandavison: Right, now we're making progress! So (correct me if I'm getting you wrong), you're basically using "environment config" with the following meaning:

> The SDKs and CLI all require certain resources to be specified (e.g. server address), together with things like encryption settings and credentials for those resources. The key point is that these _environment_ choices (kind of) don't change the behavior of the SDK/CLI -- they're about choosing from among a range of possible resource providers.

> This is in contrast to other configuration settings that change the behavior of the SDKs and CLIs.

And I agree, this is a distinction we should make.

I've given several reasons above to think that *non-environment* config is something we will want, whether sooner or later: I've given concrete suggestions for what those might be today, and I've also appealed to the fact that Temporal has a large product surface area, ambitions, and a lot of time in which it will exist in the future hence is likely to end up wanting this at some point.

So unless we can identify reasons why we in fact do not think we'll ever want non-environment config, the challenge here is to design file-based configuration in a way that supports the following requirements (but not limited to these):

1. It is possible to define non-environment config.
1. It is possible to add new top-level non-environment config sections in the future.
1. It is possible to define multiple alternative environment profiles.
1. It is possible to use env vars to supply values for all environment config values where an env var makes sense.

all-sdk/external-client-configuration.md
cretz: Even beyond best practice, in this case I'm specifically proposing using a proto model to define the contract for the configuration, which has strict JSON expectations: https://protobuf.dev/programming-guides/proto3/#json.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
drewhoskins-temporal: Thanks, I definitely buy some of these.  (Though, others are written from our perspective rather than the user's perspective)  The meta-point I want to make is that we should incorporate some of these user needs into our goals to make sure we understand priority and are solving their complete needs.  (i.e. Let's not just discuss it in a thread but build it into the doc.)

Here are some possible stories in addition to the ones y'all provided.

* A user is figuring out how to use Temporal copies code from these samples which uses these client configs into their production code and expects it to work well.
* A user has a multi-region setup and wants to vary some of their client connection config depending on the region.
* User is migrating from self-hosted to cloud and wants to dynamically flip between the two environments, e.g. using a feature flag.

We should err on the side of writing more stories down and then, if we don't want to cater to them, saying so explicitly.  This will make sure we're aligned.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: > I've given concrete suggestions for what those might be today

But I think they should be rejected for config except for environment-specific suggestions, no matter how far in the future non-environmental config is suggested. It doesn't mean other avenues for non-environment things can't be explored, I just don't think we need to account for them as part of this config which should be environment config only.

I think we just have to decide if this config is environment-config only or not. I propose so, forever into the future.

all-sdk/external-client-configuration.md
drewhoskins-temporal: Yes, the advantage of  configuration files is that they simply need to be present--nothing needs to be explicitly run beforehand.  Thus we eliminate a class of bugs where the script didn't get run.

all-sdk/external-client-configuration.md
drewhoskins-temporal: > Trying to avoid any new dependencies in our libraries that then forces uses to depend on them transitively and locks them into our version constraints. We have seen this issue with the libraries we do have to depend on (e.g. protobuf) where by using our libraries, we are forcing our users' apps to be bound by version constraints we have. Libraries like ours should avoid dependencies they don't need because they are transitively infectious to a user's entire application.

This a good flag--let's state _that_ as the goal rather than the implementation choice of not taking any dependencies at all.   For example, we could rely on older versions of certain technologies so we're not forcing people to upgrade.

Also, I also if we can make dependencies dynamic and only loaded, for example, when a certain file is present.

+1 that the ability to comment is worth listing among our goals and deciding whether it's a goal.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
dandavison: > I just don't think we need to account for them as part of this config which should be environment config only.

Right, I just don't think we can use the name `temporal.{json,toml}` if it's never going to have general config in it. So maybe

- `temporal-profiles.{json,toml}`
- `temporal-client-config.{json,toml}`


Alternatively, perhaps we should be less strict and call it `temporal.toml` and allow that it might acquire non-environment config in the future, as other top-level keys that are siblings of the `profiles` key.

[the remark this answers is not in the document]
dandavison: To be clear: I don't really like the idea of multiple files. I think users will think of it all as "config", and will be annoyed that Temporal has "multiple different config files", and so it will make the product seem more complicated. So, even though we might like to avoid a difficult design process at this stage, that might not be the right decision for the product.

all-sdk/external-client-configuration.md
cretz: I don't think the ability to comment should be a goal.

> This a good flag--let's state that as the goal rather than the implementation choice of not taking any dependencies at all. For example, we could rely on older versions of certain technologies so we're not forcing people to upgrade. Or we could provide a fallback plan so that most users get nice things.

No, version constraints go in all directions for compatibility because a newer version may break compatibility in a major version. There is no such thing as "depending on every version ever into the future", you have to have some constraint. If we say "I depend on toml 1.x", when "toml 2.x" comes about our users come complaining to us. And you can't say "toml x.x" because newer ones may break compat. And you can't just upgrade to "toml 2.x" when released because many users don't want to upgrade. These are not hypothetical, we've seen all of these complaints and situations in practice.

It is important for us not to impose constraints on our users unless we absolutely have to. I suppose I can change "0 dependencies" to "0 external library version constraints imposed on our users" if the former is unacceptable as a stated goal, but they are saying the same things.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: We can change the default filename, no problem. I have no strong preference on which alternative there is chosen. I suspect people won't really use it directly anyways very much because if they want to hand-edit a file they might put it somewhere more reasonable and provide that filename when loading or as an env var. Or they'll use the CLI. Similar to AWS config.

I would be concerned however if we need to change the env var names or property names or the CLI commands or anything. I do think from a user POV we still need to treat this as "the config" and not inject client in those places.

cretz:
Closing in favor of a new round

PR #95, opened 2024-09-05
External client configuration proposal

cretz:
# Summary

* This is the proposal for external client configuration (i.e. config files and profiles)
* See it [rendered](https://github.com/cretz/temporal-proposals/blob/external-client-configuration/all-sdk/external-client-configuration.md)
* This is the second round of review (see #94 for the first)
* Like other proposals, all feedback welcome. If the PR discussion gets too large or the review round comes to an end w/ more review needed, it will be closed and a new one will be created referencing this one.

Notable changes since first round:

* Added more usage examples including making configuration files more prominent
* Removed the concept of profile-specific environment variables
* Changed the default config filename to make it clear it's not for all config that may ever happen, just for clients
* Switched file format to TOML
* Changed SDKs to use "extension" approach since they'll need a new TOML dependency (there are some outstanding questions here regarding whether to use Rust Core)
* Added env var support for gRPC headers

### DIFF
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -259,6 +259,14 @@ client = await Client.connect(namespace="my-specific-namespace", **config)
     * After discussion, it was decided future non-client use of this config (which may never occur) can be sub keys of
       this file instead of having a separate file or separate config for every separate use. And client is the common
       enough initial (and maybe only) use case to get the top-level without burdening users.
+* Unlike all other ways to provide settings, all file paths inside configuration files must be absolute or loading
+  should fail, no relative paths
+  * 💭 Why?
+    * We don't want to make paths relative to the config file, because people load this config into a common structure
+      and different meanings for different-sourced values (CLI vs env var vs config file vs programmatic) is confusing
+      and can't be represented well
+    * We're replacing `env set` with `config set` and the former supports relative to the calling user, so it'd be
+      confusing to change that
 
 ### Environment variables
 
@@ -344,7 +352,7 @@ is SDK specific).
         * This is the new way and we should not try to support both at the same time, it can get confusing.
   * The default is `default`, and this can also be set via `TEMPORAL_PROFILE` env var.
 * CLI will accept `--config-file`.
-  * Default is `~/.config/temporalio/temporal.toml`, also can be set via `TEMPORAL_CONFIG_FILE` env var.
+  * Has default specified before, also can be set via `TEMPORAL_CONFIG_FILE` env var.
 * CLI will have a whole new set of `config` commands that operate similar to `env` commands today.
   * Will not go into detail in this proposal, but it's very similar.
   * We will strictly validate keys when setting whereas we did not with `env`.
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -24,22 +24,33 @@ What are the reasons for building this?
 
 ### Goals
 
-* Common, well-specified configuration format usable by Temporal tooling for client options
+* Common, well-specified/type-safe configuration format usable by Temporal tooling for client options and usable
+  manually by advanced users
 * Implementation in all Temporal SDKs for reading and writing of this configuration
 * All Temporal CLI/SDKs updated to have easy client creation using this configuration
 * Common on-disk location in every platform for reading/writing this format
 * Ability to provide environment variables to represent values in the configuration
 * Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
 * 0 new dependencies in SDKs
+  * 💭 Why?
+    * Dependencies in our SDKs subject our users to these dependencies and their version constraints, so we should avoid
+      this as much as possible.
+  * ❓ Can we just have an "extension" that has new dependencies?
+    * Yes, but ideally users don't have to add dependencies on extension libraries just to get this functionality.
 * Profile names within the configuration
+  * 💭 Why?
+    * It is common in configurations like this to want to be able to share a single file with many configurations.
+
+### Future Goals
+
+None identified at this time. The goals are simple enough for the MVP and the non-goals are ones we never want.
 
 ### Non-goals
 
 * Common format for declarative workers, schedules, etc. This is only for client options.
-* File-based hierarchy. We don't need many levels of file merging for these options at this time.
-* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
-* Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
-  of course easy).
+* File-based hierarchy. We don't need many levels of file merging for these options.
+* Use external configuration by default in SDKs (it is default in CLI and samples). Users need to opt-in to this for
+  compatibility and clarity reasons (but it is of course easy).
 
 ### High-level Usage Examples
 
@@ -70,7 +81,7 @@ you could do this in environment variables for mTLS auth:
     export TEMPORAL_TLS_CLIENT_CERT_PATH=path/to/my/client.pem
     export TEMPORAL_TLS_CLIENT_KEY_PATH=path/to/my/client.key
 
-Now running the same code again unmodified will run against cloud. You can do ths same thing with config instead of
+Now running the same code again unmodified will run against cloud. You can do this same thing with config instead of
 environment variables:
 
     temporal config set --key address --value my-ns.a1b2c.tmprl.cloud:7233
@@ -97,15 +108,35 @@ Similar to AWS tooling and other tools, you can have profiles. So you can have:
 
 For dev and then in production, this might exist:
 
-    temporal config set --profile prod --key address --value my-dev-ns.a1b2c.tmprl.cloud:7233
-    temporal config set --profile prod --key namespace --value my-dev-ns.a1b2c
-    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/dev-cert.pem
-    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/dev-cert.key
+    temporal config set --profile prod --key address --value my-prod-ns.a1b2c.tmprl.cloud:7233
+    temporal config set --profile prod --key namespace --value my-prod-ns.a1b2c
+    temporal config set --profile prod --key tls.clientCertPath --value path/to/my/prod-cert.pem
+    temporal config set --profile prod --key tls.clientKeyPath --value path/to/my/prod-cert.key
 
 Or alternatively it could be a fixed file that that creates or environment variables or whatever.
 
 Now a simple setting of `TEMPORAL_PROFILE` environment to `dev` or `prod` will switch connectivity (or `--profile` on
 CLI or the string in the SDK if you don't want to use environment variable).
+
+#### Manually Working with Configuration
+
+If a user wanted to hand-write a configuration, they can have `my-file.json` like:
+
+```json
+{
+  "profiles": {
+    "default": {
+      "address": "my-host:7233",
+      "namespace": "my-namespace"
+    }
+  }
+}
+```
+
+* This is the same format used by the CLI and the CLI can even be used to edit this file.
+* `TEMPORAL_CONFIG_FILE` env var can point to this file or the config file can be manually set in the CLI/SDK.
+* There is a type safe contract this file conforms to for anyone wanting to work with it programmatically.
+* All SDKs accept the type safe model this file represents if users would rather provide config that way.
 
 #### Possible UI Creation Flow
 
@@ -159,8 +190,8 @@ whether this is done by default can be discussed, but the idea is the same.
       * `namespace` - Client namespace.
       * `apiKey` - Client API key.
       * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
-        upon the client and may change. 💭 Why? We regret not making TLS the default in the past. If TLS is an object,
-        it can have the following possible fields:
+        upon other settings here. 💭 Why? We regret not making TLS the default in the past, so we will make it default
+        if `apiKey` is present. If TLS is an object, it can have the following possible fields:
         * `clientCertPath` - File path to mTLS client cert. Mutually exclusive with `clientCertData`.
         * `clientCertData` - Cert data. Mutually exclusive with `clientCertPath`.
         * `clientKeyPath` - File path to mTLS client key. Mutually exclusive with `clientKeyData`.
@@ -253,6 +284,26 @@ default is assumed.
   * There's no obvious env var approach to array of strings, and it can't be completely comma-delimited because header
     values often contain commas. Would rather not force users to have an index in the env var name. Should we just not
     support gRPC meta from env var until we think this through?
+
+### Loading Configuration
+
+Configuration is loaded in the following manner:
+
+* Try to load profile from configuration file
+  * Only if file loading is enabled
+  * Can If user provides specific config file, use that, otherwise use the default location
+  * If user specifies the profile, use that, otherwise use the default
+* Try to overwrite with specific configuration environment variables
+  * Only if env loading is enabled
+  * If user specifies the profile, use that, otherwise use the default
+  * If the `default` profile is being loaded, try the profile-specific environment variable first, e.g.
+    `TEMPORAL_DEFAULT_ADDRESS` is used over `TEMPORAL_ADDRESS` if it's present.
+  * The `default` env vars are only for the default profile, meaning a profile of `foo` cannot use `TEMPORAL_ADDRESS`,
+    only `TEMPORAL_FOO_ADDRESS` if present.
+
+That is it, nothing more complicated. When the configuration is provided to the SDKs/CLI, they apply their normal logic
+which includes defaults for things that can be defaulted, or errors if they expect something that is not provided (this
+is SDK specific).
 
 ## Implementation
 
--- a/all-sdk/external-client-configuration.md
+++ b/all-sdk/external-client-configuration.md
@@ -1,0 +1,340 @@
+# External Client Configuration
+
+Notes for this proposal:
+
+* Everything is subject to change both during proposal time and afterwards during implementation.
+* For each place a decision is made, a "💭 Why X? ..." aside is present explaining why.
+* For each place an open question exists, a "❓ X?" exists to ask it.
+
+## Overview
+
+There are many ways to access Temporal programmatically including CLI, SDK, and other components that leverage those
+such as samples and Terraform providers.
+
+Temporal needs a way for users to provide client configuration without hardcoding in CLI arguments or SDK options. This
+will allow code to seamlessly switch between profiles such as between a dev-server and cloud. The same configuration
+must work across all SDKs/CLIs.
+
+### Goals
+
+* Common, well-specified configuration format usable by Temporal tooling for client options
+* Implementation in all Temporal SDKs for reading and writing of this configuration
+* All Temporal CLI/SDKs updated to have easy client creation using this configuration
+* Common on-disk location in every platform for reading/writing this format
+* Ability to provide environment variables to represent values in the configuration
+* Well-specified hierarchy for on-disk overridable by environment variables overridable by in-CLI/SDK values
+* 0 new dependencies in SDKs
+* Profile names within the configuration
+
+### Non-goals
+
+* Common format for declarative workers, schedules, etc. This is only for client options.
+* File-based hierarchy. We don't need many levels of file merging for these options at this time.
+* Extreme focus on human authoring of this format. This is because it'd likely go against the "0 new dependencies" rule.
+* Use external configuration by default. Users need to opt-in to this for compatibility and clarity reasons (but it is
+  of course easy).
+
+## Specification
+
+### Values and File Format
+
+* The format is a JSON object with the following fields:
+  * `profiles` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
+      separately-cased profile names that are equal case-insensitively.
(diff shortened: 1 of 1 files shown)


all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,540 @@
...
+    <some long api key string>
+
+Alternatively, you can download it as a configuration file:
+
+    <some button or something that downloads a file name temporalio-client.json>
+
+You can place this configuration file in the default place used by CLIs/SDKs:
+
+* macOS - `$HOME/Library/Application Support/temporalio/temporal-client.json`
+* Windows - `%AppData%/temporalio/temporal.json` (often `C:\Users\<myuser>\AppData\Roaming\temporalio\temporal-client.json`)
```
cretz, line 170:
```suggestion
* Windows - `%AppData%/temporalio/temporal-client.json` (often `C:\Users\<myuser>\AppData\Roaming\temporalio\temporal-client.json`)
```

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,540 @@
...
+      files that may have newer keys to work with older SDKs.
+    * Similarly, it became clear that we're not validating environment variables this way, why config file keys?
+    * We want the tools that mutate this file to have strong validation on the keys being added, so hopefully this
+      avoids the concern of key typos.
+* All keys are optional in the file. Ideally in the final merged form, we'd always require `address` and we always
+  require `namespace` in namespace-specific clients (i.e. non cloud ops API), but it is unreasonable since we must fit
+  within SDK defaults.
+  * Originally this section did say `address` and `namespace` were required, but as SDK implementations came about, it
+    became clear that can't be done easily, so CLI/SDK defaults need to remain.
+* The default config file location is `<app-config-dir>/temporalio/temporal.json` for all platforms
```
cretz, line 281:
```suggestion
* The default config file location is `<app-config-dir>/temporalio/temporal-client.json` for all platforms
```

cretz:
New updates based on team discussion (will also update PR description):

* Switched file format to TOML
* Changed SDKs to use "extension" approach since they'll need a new TOML dependency (there are some outstanding questions here regarding whether to use Rust Core)
* Added env var support for gRPC headers

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+```toml
+[profile.temporal-cloud]
+address = "my-dev-ns.a1b2c.tmprl.cloud:7233"
+tls.client_cert_path = "path/to/my/dev-cert.pem"
+tls.client_key_path = "path/to/my/dev-cert.pem"
+```
+
+And then still use CLI like
+
+    temporal workflow list --namespace my-specific-namespace
```
mjameswh, line 185: Uh? Why is this implicitly using the `temporal-cloud` profile?

Did you mean the profile above to be named `default` instead? Or add a `--profile temporal-cloud` argument to this command and the one below?

mjameswh: Also, would this work on Cloud _today_? or only once we get global endpoints?

If not, I think we should use a non-cloud example, or clearly indicate that this is pending future Cloud changes.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+client = await Client.connect(namespace="my-specific-namespace", **config)
+```
+
+## Specification
+
+### Values and File Format
+
+* The format is a TOML object with the following fields:
+  * `profile` - Object with keys as profile names. The default profile should be named `default`.
+    * `<name>` - Profile object with configuration values. Profile names are case-insensitive. 💭 Why? They are needed
```
mjameswh, line 205: No longer true:
~They are needed for environment variables which are case insensitive~

Still, I think case-insensitive profile names make sense. Could want to reopen the env variable avenue later, and that would be a likely source of error for users anyway.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+      for environment variables which are case insensitive. It is a validation error to have a config file with two
+      separately-cased profile names that are equal case-insensitively.
+      * `address` - Client address, aka gRPC "host:port". This cannot be a URL, it must be `host:port`.
```
mjameswh, line 208: Let's be specific about IPv6 address, e.g.: `[ipv6address]:port`

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+      * `namespace` - Client namespace.
+      * `api_key` - Client API key.
+      * `tls` - A boolean (true is same as empty object) _or_ an object. Note the default for this value is dependent
+        upon other settings here. 💭 Why? We regret not making TLS the default in the past, so we will make it default
+        if `api_key` is present. If TLS is an object, it can have the following possible fields:
+        * `client_cert_path` - File path to mTLS client cert. Mutually exclusive with `client_cert_data`.
```
mjameswh, line 214: For all path options:
 - If relative, evaluates relative to the config file?
 - Is there anything particular that we should specify about paths on Windows, to avoid cross-languages incoherencies? Should we be worried about WSL and similar?

Don't necessarily need to write this just here, but please put it somewhere so that it survives the review.

all-sdk/external-client-configuration.md
cretz: > Did you mean the profile above to be named default instead?

Yes, sorry, when translating to TOML in https://github.com/temporalio/proposals/pull/95/commits/b8cf8febcb7c68f4a54415e7b111428994cfc8eb I accidentally change the profile name, needs to be `default`, will fix

cretz: > Also, would this work on Cloud today? or only once we get global endpoints?

Yes for data plane. Clients are just addresses and namespaces and auth. What shared endpoints will provide is a way for cloud ops API to use this as well.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+```csharp
+var options = ClientConfig.LoadClientConnectOptions();
+var client = await TemporalClient.ConnectAsync(options);
+```
+
+### Samples
+
+* All samples will be updated to load from config and default to local dev server w/ default namespace.
+  * In samples we try to avoid shared code. In .NET and Python, there is no default target host, and in Ruby there will
+    not even be a default namespace. So every sample in these languages is going to have to not only have the
```
mjameswh, line 517: > So every sample in these languages is going to have to not only have the load-from-config code, but also conditionally apply the target host as `localhost:7233` if not set in external config.

🤔 That kind of defeats on of the purpose of the load-from-config utility function, and would have little demonstrative value in samples as most "sample-level" users would, in practice, either fully adopt the config file approach, or fully specify their client config programmatically.

I think we should revisit the question on either load-from-config should fill in a default address and namespace. I totally understand why that's tricky in these languages, but I'm not sure the current proposal is the most reasonable compromise.

all-sdk/external-client-configuration.md
cretz: > Still, I think case-insensitive profile names make sense

I'm not really looking to protect from typos here. I don't think case sensitivity of profile name is any more painful than case sensitivity of config keys (we require all lowercase) or env var names (we require all upper) or CLI flag names or whatever. If we need to relax case sensitivity later we can, but I think it's normal for people to expect it.

> Are we ok accepting any valid TOML key as profile name? even quoted keys?

I am ok with this for the most part, or at least I can't see any obvious reason why not. But I admittedly did not cover advanced validation of things in this proposal. I tend to want to accept loosely and have the usage fail if needed as opposed to have lots of matching validation in lots of different languages.

all-sdk/external-client-configuration.md
cretz: > If relative, evaluates relative to the config file?

Good question. I generally like relative to mean current working directory, but I understand that's confusing for external configs. I will clarify that this should be relative to config file.

> Is there anything particular that we should specify about paths on Windows, to avoid cross-languages incoherencies? Should we be worried about WSL and similar?

Nope, not that I can think of

all-sdk/external-client-configuration.md
cretz: > I think we should revisit the question on either load-from-config should fill in a default address and namespace.

There is no default address and namespace in some SDKs is the point. We have _removed_ this default as we've become more mature knowing that the older defaults (which we cannot change) are not as useful in deployed code. But for those SDKs w/out a default address/namespace there _is_ a default address/namespace in samples, so they must set it. But that's only for samples, we shouldn't affect non-sample code w/ sample-only defaults.

all-sdk/external-client-configuration.md
mjameswh: > Looking back, I will probably back off validation here and just say "address passed to the underlying client, we encourage host:port".

I disagree. `temporal config set` should validate for what is expected to be the correct format. Not doing so will make this a common source of errors for new users for a benefit that only matters for a very tiny percentage of "expert users", who could anyway still provide "out-of-norm" values by editing the config file by hand.

cretz: So what do you believe the validation for `address` should be if I want to have Go allow a type of value that, say, Python doesn't support, such as a custom resolver? To validate this value properly means to assume how an SDK uses the value or assume that all of our SDKs share a common validation of this value which is not accurate.

all-sdk/external-client-configuration.md
Sushisource: Why not just make it required for the config to specify these things (even for langs where the in-code options allow a default)? It sure would make the config file more readable by default, and can still be overridden with env var.

cretz: Because some people just want to configure, say, the remote codec with a config file or env var. Or there is nothing in config and only address is set by env var and namespace is expected to be set by user. Or config file only sets address and namespace is set via env var. All config keys are optional (as is even the presence of the config file). It's only when the client options are used does whether a value being present matter.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,526 @@
...
+    * We have to take a TOML dependency which is not fair to put on users that don't need this functionality.
+* Needs `ClientConfig` interface (and other interfaces as needed).
+* Needs `loadClientConfig(options?)` where the `options` type is
+  `{ configFile?: string, disableFile?: bool, disableEnv?: bool }` and it returns a `ClientConfig`.
+* Needs `loadClientOptionsFromConfig(options?)` where the optional `options` are
+  either `ClientConfig` or `{ profile?: string, configFile?: string, disableFile?: bool, disableEnv?: bool }` and
+  it returns a `ConnectionOptions & ClientOptions`.
+  * Could also have it return `{ connectionOptions: ConnectionOptions, clientOptions: ClientOptions }` if that makes
+    more sense than a union.
+* TypeScript SDK _does not_ support codec settings and will error if seen in config.
```
mjameswh, line 452: > Because a user with a profile configured with a codec should expect that codec to be used, not silently ignored.

I agree with your argument, but I'm not totally comfortable with the consequence.

To error in such cases implies that a user who has codecs AND is using an SDK that doesn't have remote codec support would have to create two distinct profiles, one for use with CLI with remote codec configured, and the other for use with his SDK, without remote codec configured. That's very inconvenient.

Not saying that's wrong, but to be coherent with this decision, we will need to move forward with implementing remote codecs in all SDKs. Not difficult, but IIRC, there were doubts about either that's a good idea.

Otherwise, we may find ourselves needing to add an extra flag to `LoadClientOptions` in all SDKs that don't have support for remote codecs, to intentionally ignore remote codec settings from config file/environment (i.e. just turn off that specific error). That's ugly. I'd definitely rather not get down to this.

all-sdk/external-client-configuration.md
cretz: There are three ways this path setting can be set: config file, env var, and programmatically. After some thought, it is confusing to have different ways of specifying the same value have different relative bases, so I think it makes the most sense to be all relative to the current working directory. I will add a point clarifying this.

all-sdk/external-client-configuration.md
cretz: :+1: I agree we'd need to support remote codecs in each SDK to get the benefit of this setting in every SDK. In the meantime, erroring is the clearest approach I think.

cretz:
Added updates for:

* Config file being `temporal.toml` knowing future non-client config can be in a subkey
* Core-based SDKs will use Rust Core if they can (TS is still a bit up in the air)

all-sdk/external-client-configuration.md
mjameswh: Validation in CLI should be based on a least-common-denominator format that we guarantee to be supported in every SDKs to support.

IMO, that should be:

```
ipv4:port
hotname:port
[ipv6]:port
```

Anything more than that should not be accepted as input on `temporal config set`. That way, we promote using things that will work in all SDKs.

However, I don't think we need the _SDK side_ to validate that. So if a user edit his config file and manually puts a value of another form, the SDK will just pass that to the grpc library, and that may work or that may not.

all-sdk/external-client-configuration.md
cretz: Treating these config keys different based on where they are set is confusing IMO. We have a want in our SDKs to be able to load the environment config as an object to be viewed. We want that object to have the _exact_ values the user set, not some resolved form (just in case they want to see what was there or want to save or whatever). How are you going to have different rules for the same config key from different places and still give a full config object with overrides?

all-sdk/external-client-configuration.md
```diff
@@ -441,6 +449,9 @@ var client = WorkflowClient.newInstance(stubs, clientOptions);
 * New package and module for `envconfig`.
   * 💭 Why not just in the SDK?
     * We have to take a TOML dependency which is not fair to put on users that don't need this functionality.
+  * ❓ What about "native connection"?
+    * Python and .NET and Ruby will use Rust Core to load this, so TypeScript may have to have two forms too, up to
```
mjameswh, line 453: TS will only have a lang-side TOML parser. The profile feature has enough value for people to want that in client-only scenarios, so a pure-JS implementation is a must. From that point, it would make no sense to expose both a JS and a Native variant of the "load config" API.

all-sdk/external-client-configuration.md
```diff
@@ -337,7 +344,7 @@ is SDK specific).
         * This is the new way and we should not try to support both at the same time, it can get confusing.
   * The default is `default`, and this can also be set via `TEMPORAL_PROFILE` env var.
 * CLI will accept `--config-file`.
-  * Default is `~/.config/temporalio/temporal-client.toml`, also can be set via `TEMPORAL_CONFIG_FILE` env var.
+  * Default is `~/.config/temporalio/temporal.toml`, also can be set via `TEMPORAL_CONFIG_FILE` env var.
```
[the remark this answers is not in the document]
cretz: This is kinda specified above at:

> The default config file location is `<app-config-dir>/temporalio/temporal.toml` for all platforms. [...] This is as defined by https://pkg.go.dev/os#UserConfigDir which, as of this writing, is the logic defined in docs and in code at https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/file.go;l=528.

And there's a hint before in one of the flows as sample instructions to users:

* macOS - `$HOME/Library/Application Support/temporalio/temporal.toml`
* Windows - `%AppData%/temporalio/temporal.toml` (often `C:\Users\<myuser>\AppData\Roaming\temporalio\temporal.toml`)
* Linux - `$HOME/.config/temporalio/temporal.toml`

But I can take that part and make it clearer here if needed.

all-sdk/external-client-configuration.md
josh-berry: I am strongly in favor of relative paths in the config file being relative to the config file, and not the user's PWD. I think the extra implementation complexity of resolving the path differently depending on whether it comes from the config file or the environment is acceptable.

Putting a relative path in a config file, and then interpreting that path relative to the user's PWD at runtime, would be extremely surprising and confusing to most config file users IMO. This is not how most config-file-using apps I am familiar with behave.

It also makes redistribution of config files much harder, because you cannot include supplementary files alongside them without having some special logic somewhere to rewrite paths etc. You can no longer just zip up a config file and corresponding TLS keys etc. and send it to someone, you have to have the recipient rewrite the paths etc. to work reliably.

josh-berry: As a compromise, I propose _requiring_ absolute paths and rejecting relative paths as invalid config. We can then revisit this point later.

I do agree that users should have programmatic access to the as-written values in the env and config file. However, I think my concern about relative paths is serious enough that it does need to be addressed one way or another.

cretz: > I propose requiring absolute paths and rejecting relative paths as invalid config. We can then revisit this point later.

I assume you are not proposing this for `--tls-cert-path` that already exists on CLI today? We already support relative paths in `TEMPORAL_TLS_CERT`, are you proposing when we deprecate this in favor of `TEMPORAL_TLS_CLIENT_CERT_PATH` we remove the ability to use relative paths there? Also, are you proposing when we deprecate `temporal env --key tls-cert-path --value my/relative/path.cert` that it's replacement of `temporal config --key tls.client-cert-path --value my/relative/path.cert` won't work anymore?

josh-berry: I am proposing that, in the config, we require absolute paths. I am not proposing any other changes (certainly no changes to the behavior of the existing env variables or flags).

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
mjameswh: Yeah, it's clearly indicated _elsewhere_, and I would normally not be picky on this kind of small incoherencies.

But this specific line sounds like it is authoritative, so being incomplete in this case is more confusing.

TBH, I'd just remove this line, or formulate it as an example rather than a prescriptive statement.

all-sdk/external-client-configuration.md
cretz: > I am proposing that, in the config, we require absolute paths

Ah, ok I misread "As a compromise, I propose requiring absolute paths and rejecting relative paths as invalid config" as all config, but you mean config _file_. I think we can mostly do this.

So, concerning the last part of my comment above, can you clarify what you expect the behavior of `temporal config --key tls.client-cert-path --value my/relative/path.cert` to be that writes to config file (compared to the deprecated form of `temporal env --key tls-cert-path --value my/relative/path.cert`)? Before, our env config supported relative paths, but now our new one does not, which I assume is an intentional regression?

josh-berry: Yep, I think we would have to expand the path to its absolute form before writing to the config file, unless we want to simply fail saying the path is invalid/must be absolute.

cretz: Ok, I will update this doc clarifying that config file _loading_ should fail if the path is not absolute (for now) and maybe config file writing from CLI can resolve the path.

all-sdk/external-client-configuration.md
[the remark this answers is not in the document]
cretz: Removed the part saying what the default is here under the CLI section hoping people will look at the general specification section

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,551 @@
...
+      * `codec` - Remote codec information to use for all encoding/decoding of payloads. May not be available in all
+        Temporal clients. 💭 Why? Not all SDKs allow use of a remote codec today.
+        * `endpoint` - Endpoint to the remote codec.
+        * `auth` - Authorization header for the remote codec.
+      * `grpc_meta` - Object representing HTTP headers. Values must be strings for now (can discuss arrays later).
+        ❓ Should we call this `headers`? We do in some cases and not others.
+  * 💭 Why TOML?
+    * Original document was JSON, but team discussion decided that the benefit to humans of TOML outweighed the concerns
+      of making SDKs have extensions with TOML support.
+* For settings that accept files above, they are relative to the current working directory.
```
josh-berry, line 232: Need to clarify that the config file must hold absolute paths and the CLI will always write absolute paths.

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,551 @@
...
+* The default config file location is `<app-config-dir>/temporalio/temporal.toml` for all platforms.
+  * This is as defined by https://pkg.go.dev/os#UserConfigDir which, as of this writing, is the logic defined in docs
+    and in code at https://cs.opensource.google/go/go/+/refs/tags/go1.23.0:src/os/file.go;l=528.
+  * Existing CLI used `~/.config/temporalio/temporal.yaml`
+    * ❓ Is this confusing for them? Are we concerned about them vs what is best moving forward?
+  * 💭 Why `temporal.toml` instead of `temporal-client.toml`?
+    * After discussion, it was decided future non-client use of this config (which may never occur) can be sub keys of
+      this file instead of having a separate file or separate config for every separate use. And client is the common
+      enough initial (and maybe only) use case to get the top-level without burdening users.
+* Unlike all other ways to provide settings, all file paths inside configuration files must be absolute or loading
```
josh-berry, line 262: 👀👍

all-sdk/external-client-configuration.md
```diff
@@ -0,0 +1,551 @@
...
+  * `TEMPORAL_CONFIG_FILE` - the path to the config file. The default is `~/.config/temporalio/temporal-client.toml`.
+
+### Loading Configuration
+
+Configuration is loaded in the following manner:
+
+* Try to load profile from configuration file
+  * Only if file loading is enabled (which it is by default)
+  * If user provides specific config file, use that, otherwise use the default location
+  * If user specifies the profile, use that, otherwise use the default
```
josh-berry, line 332: Do we want to add a step in here somewhere with the config-file-specific absolute path validation?

josh-berry (approved):

all-sdk/external-client-configuration.md
cretz: Ah, I clarified this elsewhere but not here. Will do.

cretz:
No outstanding issues, merging

PR #96, opened 2024-10-24
Ruby phase 2 - Workflows

cretz:
# Summary

* This is the proposal for the second phase of the Ruby SDK about workflows. Please review and leave comments!
  * Can see the first phase [here](https://github.com/temporalio/proposals/blob/master/ruby/sdk-phase-1.md) or look at PRs #92 and #93 for discussion on it.
* See it [rendered](https://github.com/cretz/temporal-proposals/blob/ruby-phase-2/ruby/sdk-phase-2.md).
* Much of this is inspired by the previous [Temporal Ruby SDK](https://github.com/temporalio/sdk-ruby/tree/5b2429eb7e8cd1318c2f33149fccbb3ddacd533f) (and its [proposals](https://github.com/temporalio/proposals/tree/master/ruby/previous) which were moved to a sub directory) and our other SDKs.
* The design of workflows and their handlers is an early one and we welcome any suggestions for alternative designs here. Nothing is concrete.
* There are plenty of outstanding questions on the proposal, some of which need more implementation and performance measurements to decide.

For general discussion beyond comments here, please also join us on `#ruby-sdk` in [Slack](https://t.mp/slack).

### DIFF
--- a/ruby/sdk-phase-2.md
+++ b/ruby/sdk-phase-2.md
@@ -52,7 +52,7 @@ class CallGreetingServiceActivity < Temporalio::Activity::Definition
 end
 
 class GreetingWorkflow < Temporalio::Workflow::Definition
-  workflow_query_attr :language
+  workflow_query_attr_reader :language
 
   def initialize(name)
     @greetings = {
@@ -227,7 +227,7 @@ module Temporalio
         # was not enough value to have attr_accessor shortcut for update or
         # attr_writer shortcut for signals since both are not commonly just
         # getter/setter.
-        def self.workflow_query_attr(query_method)
+        def self.workflow_query_attr_reader(query_method)
 
         # Called in the workflow definition to change the name from the default.
         def self.workflow_name(name)
@@ -276,8 +276,8 @@ module Temporalio
     def self.query_handlers
     def self.update_handlers
 
-    def info
-    def current_update_info
+    def self.info
+    def self.current_update_info
 
     def self.search_attributes
     def self.upsert_search_attributes(*updates)
--- a/ruby/sdk-phase-2.md
+++ b/ruby/sdk-phase-2.md
@@ -1,0 +1,569 @@
+# Ruby - Phase 2
+
+This proposal is for implementing workflows in the Ruby SDK.
+
+See [phase 1](sdk-phase-1.md) for details on phase 1 which includes client and activities (granted some things may have
+changed since first authored).
+
+Notes for this proposal (like phase 1):
+
+* Everything is subject to change both during proposal time and afterwards during implementation.
+* For each place a decision is made, a "💭 Why X? ..." aside is present explaining why.
+* For each place an open question exists, a "❓ X?" exists to ask it.
+
+Here is a high-level overview of potentially controversial decisions:
+
+* `Temporalio::Workflow` is both the class to extend and the class with class methods for calling statically to do
+  things.
+  * The `workflow_` class methods on the `Temporalio::Workflow` can only be called at class definition time and cannot
+    be called on `Temporalio::Workflow` directly. These are for adjusting the definition.
+  * All other explicitly-defined class methods on `Temporalio::Workflow` are for use in workflow, can only be used at
+    workflow runtime, and must be called on the workflow class itself (e.g. it is a failure to call
+    `MyWorkflow.wait_condition`).
+* Class methods are dynamically made available for signals, queries, and updates to use from clients.
+  * So `MyWorkflow.my_signal` can be used in a client, and it just returns a signal definition, unlike the instance
+    form.
+* By default `TracePoint` will be enabled for workflow threads that will make sure users are not calling illegal calls,
+  but there is a way to disable for blocks of code.
+* `Ractor`s will be used for state isolation (TBD what is enabled by default).
+
+These points are explained below.
+
+## Workflow Definition/Usage
+
+Examples below are taken from
+[this Python sample](https://github.com/temporalio/samples-python/tree/main/message_passing/introduction), so not too
+many comments about general non-Ruby behavior are inline.
+
+Example definition:
+
+```ruby
+class CallGreetingServiceActivity < Temporalio::Activity
+  GREETINGS = {
+    arabic: 'مرحبا بالعالم',
+    chinese: '你好，世界',
+    english: 'Hello, world',
+    french: 'Bonjour, monde',
+    hindi: 'नमस्ते दुनिया',
+    portuguese: 'Olá mundo',
+    spanish: '¡Hola mundo'
+  }
+
+  def execute(language)
+    # Simulate a network call
+    sleep(0.2)
+    GREETINGS[language.to_sym]
+  end
+end
+
+class GreetingWorkflow < Temporalio::Workflow
+  workflow_query_attr :language
+
+  def initialize(name)
+    @greetings = {
+      chinese: '你好，世界',
+      english: 'Hello, world'
+    }
+    @language = :english
+    @mutex = Temporalio::Workflow::Mutex.new
+  end
+
+  def execute
+    Temporalio::Workflow.wait_condition do
+      @approved_for_release && Temporalio::Workflow.all_handlers_finished?
+    end
+    @greetings[@language]
+  end
+
+  workflow_query
+  def languages(input)
+    return [:chinese, :english] if input[:include_unsupported]
+    greetings.keys.sort
+  end
+
+  workflow_signal
+  def approve(input)
+    @approved_for_release = true
+    @approver_name = input[:name]
+  end
+
+  workflow_update
+  def set_language(language)
+    previous, @language = @language, language.to_sym
+    previous
+  end
+
+  workflow_update_validator(:set_language)
+  def validate_set_language(language)
+    raise "#{language} not supported" unless @greetings.include?(language.to_sym)
+  end
+
+  workflow_update
+  def set_language_using_activity(language)
+    language = language.to_sym
+    unless @greetings.include?(language)
+      @mutex.synchronize do
+        greeting = Temporalio::Workflow.execute_activity(
+          CallGreetingServiceActivity,
+          language,
+          start_to_close_timeout: 10
+        )
+        unless greeting
+          raise Temporalio::Error::ApplicationError, "Greeting service does not support #{language}"
+        end
+        @greetings[language] = greeting
+      end
+    end
+    previous, @language = @language, language
+    previous
+  end
+end
+```
+
+Example worker:
+
+```ruby
+client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
+worker = Temporalio::Worker.new(
+  client:,
+  task_queue: 'my-task-queue',
+  activities: [GreetingActivity]
+  workflows: [GreetingWorkflow]
+)
+worker.run(shutdown_signals: ['SIGINT'])
+```
+
+Example client usage:
+
+```ruby
+client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
+
+# Start workflow
+wf_handle = client.start_workflow(
+  GreetingWorkflow,
+  id: 'my-workflow-id',
+  task_queue: 'my-task-queue'
+)
+
+# Send query
+supported_languages = wf_handle.query(
+  GreetingWorkflow.languages,
+  { include_unsupported: false }
+)
+puts "Supported languages: #{supported_languages}"
+
+# Execute update
+previous_language = wf_handle.execute_update(
+  GreetingWorkflow.set_language,
+  :chinese
+)
+current_language = wf_handle.query(GreetingWorkflow.language)
+puts "Language changed: #{previous_language} -> #{current_language}"
+
+# Start update then wait for complete
+update_handle = wf_handle.start_update(
+  GreetingWorkflow.set_language_using_activity,
+  :arabic,
+  wait_for_stage: Temporalio::Client::WorkflowUpdateWaitStage::ACCEPTED
+)
+previous_language = update_handle.result
+current_language = wf_handle.query(GreetingWorkflow.language)
+puts "Language changed: #{previous_language} -> #{current_language}"
+
+# Send signal
+wf_handle.signal(
+  GreetingWorkflow.approve,
+  { name: 'some-approver-name' }
+)
+puts "Workflow result: #{wf_handle.result}"
+```
+
+Things to note about workflow definition:
+
+* Workflows must extend (directly or indirectly) from `Temporalio::Workflow`.
+  * It must implement `execute` which is the entry point.
+  * Workflow can customize its name via `workflow_name <name>` in the workflow somewhere.
+  * 💭 Why a class?
+    * This fits well with the Ruby mindset and many of our SDKs utilize classes to encapsulate workflows and their
+      handlers/state.
+* Workflow has several class methods that start with `workflow_` that can only be called during class definition time
+  not runtime. See next section for list.
+* When used in a client:
+  * The class of the workflow can be provided instead of the name, e.g. `my_client.start_workflow(MyWorkflowClass, ...)`
+  * The signal, query, or update can reference a no-arg class method of the same name as the instance method, e.g.
+    `my_wf_handle.signal(MyWorkflowClass.my_signal, ...)`, which returns a signal/query/update definition
+    * Are we ok with `respond_to_missing?` and `inherited` for class method access of these values? Is this too much
+      magic?
+* Workflow cannot be instantiated by users (i.e. outside of workflow environment)
+
+## Workflow API
+
+Approximate guess of how the `Workflow` class methods may look:
+
+```ruby
+module Temporalio
+  class Workflow
+
+    ########
+    # Class methods that can only be called during class definition time
+    ########
+
+    # Placed above handlers. See dynamic section for dynamic. Raw args means the
+    # handlers will be given their arguments as RawValue instead of converted.
+    #
+    # ❓ Notice raw_args. We could instead make type hints this way, e.g. via
+    # `args: [String, Integer]` or something and plumb that all the way through
+    # to the converter for a user's custom converter to use (same for activities
+    # and the workflow as a whole). If we did that we would also need to allow
+    # return type hints. Do we want to support this hacky form of type hints?
+    def self.workflow_signal(name: nil, dynamic: false, raw_args: false,
+                             unfinished_policy: HandlerUnfinishedPolicy::WARN_AND_ABANDON)
+    def self.workflow_query(name: nil, dynamic: false, raw_args: false)
+    def self.workflow_update(name: nil, dynamic: false, raw_args: false,
+                             unfinished_policy: HandlerUnfinishedPolicy::WARN_AND_ABANDON)
+    def self.workflow_update_validator(update_method)
+
+    # Exposes an attribute as a query. 💭 Why? We found this _very_ valuable in
+    # .NET. Think of it as attr_reader for workflows. It was also decided there
+    # was not enough value to have attr_accessor shortcut for update or
+    # attr_writer shortcut for signals since both are not commonly just
+    # getter/setter.
+    def self.workflow_query_attr(query_method)
+
+    # Called in the workflow definition to change the name from the default.
+    def self.workflow_name(name)
+
+    # Placed above initialize to get start arguments.
+    def self.workflow_init
+
+    # See dynamic section.
+    def self.workflow_dynamic
+
+    # Called in the workflow definition for initialize and execute to get
+    # RawValue arguments (useful for dynamic workflows).
+    def self.workflow_raw_args
+
+    ########
+    # Class methods that can only be called during runtime, but don't have to be
+    # in a workflow context
+    ########
+
+    # Whether in a workflow (can only be called on Workflow class, not child).
+    def self.in_workflow?
+
+    # Built, fixed workflow definition (can only be called on a class extending
+    # workflow, not workflow itself).
+    def self.definition
+
+    ########
+    # Class methods that can only be called during runtime in a workflow context
+    ########
+
+    # No start_activity because that can be in a future (see next section).
+    def self.execute_activity(activity, *args, **left_off_for_brevity)
+    def self.execute_local_activity(activity, *args, **left_off_for_brevity)
+
+    # Returns ChildWorkflowHandle.
+    def self.start_child_workflow(workflow, *args, **left_off_for_brevity)
+    # Returns actual result.
+    def self.execute_child_workflow(workflow, *args, **left_off_for_brevity)
+
+    # These are a custom hash class that only supports []= mutation and
+    # validates input as a signal/query/update definition. Basically they will
+    # be simple delegators to frozen hashes only updatable via index. Symbol
+    # keys or nil key for dynamic. All values are "definition" objects.
+    def self.signal_handlers
+    def self.query_handlers
+    def self.update_handlers
+
+    def self.search_attributes
+    def self.upsert_search_attributes(*updates)
+    def self.raw_memo
+    def self.memo_value(name)
+    def self.upsert_memo(*updates)
+
+    def self.external_workflow_handle(id, run_id: nil)
+
+    def self.patched(id)
+    def self.deprecate_patch(id)
+
+    def self.cancellation
+    # Just a shortcut to getting a future that fails on cancel. Good for
+    # Future.any_of use.
+    def self.cancellation_future
+    def self.logger
+    def self.now
+    def self.payload_converter
+    def self.random
+    def self.uuid
+    def self.wait_condition(&)
+
+    class ContinueAsNewError < Error
+      def initialize(*args, workflow: nil, **left_off_for_brevity)
+    end
+
+    class Unsafe
+      def replaying?
+      # See TracePoint later on why this is present.
+      def illegal_call_tracing_disabled(&)
+    end
+
+    # Other nested classes omitted for brevity
+  end
+end
+```
+
+Notes:
+
+* Notice lack of `timeout` on `wait_condition`, we will support Ruby `Timeout` via the fiber scheduler.
+* Notice lack of `sleep`, we will support Ruby `sleep` via the fiber scheduler.
+* A `cancellation` can be provided to every async thing, but by default it's the workflow-level one.
+* A workflow builds a `Workflow::Definition` object which contains metadata about the structure including a set of
+  `Workflow::Definition::Signal`s, `Workflow::Definition::Query`s, and `Workflow::Definition::Update`s built from the
+  handlers (but of course handlers can be added/removed at runtime via the `_handlers` hashes).
+* The `workflow_` methods above can only be called during class definition time not workflow runtime.
+* The runtime class methods above can only be called during runtime and can only be called on the `Workflow` class
+  itself. This means `Workflow.wait_condition` is ok, but `FooWorkflow.wait_condition` or `self.class.wait_condition`
+  is not.
+  * ❓ Is this acceptable? This is a consequence of us wanting to share workflow context class with workflow base
+    class. Basically at the top of these calls there's a `if self != Workflow` check.
+* `respond_to_missing?` + `inherited` will be leveraged such that all definition-time defined handlers will return their
+  definition if they are called as class methods. So `FooWorkflow.some_signal` will return a
+  `Workflow::Definition::Signal` if an instance method of `some_signal` exists marked as a `workflow_signal`.
+  * 💭 Why? It is helpful for callers/client-side to be able to reference something more explicit than a symbol, and it
+    also helps with making sure if the name is changed we use that name instead of the method name. Symbols can always
+    be used though.
+  * ❓ Is this acceptable or too much magic?
+  * ❓ Is there a better way to reference an interaction outside of the instance of the class that works well with Ruby
+    generic typing frameworks, e.g. Sorbet?
+    * We are trying to avoid the proxy model that Java takes because we have found it is too confusing to users to hide
+      that method invocation is actually, say, activity invocation.
+
+## Dynamic Workflows, Signals, Queries, and Updates
+
+* Support dynamic pieces similar to how we have done in Python and .NET.
+* Create a new `Temporalio::Converters::RawValue` object that is a pass-through object for raw values (still subject to
+  payload codecs), and a `payload_converter` is available in workflow/activity context to convert it.
+  * Handler methods will use `RawValue` for args if told to, similar with workflow `execute`/`initialize` if
+    `workflow_raw_args` is set.
+  * `RawValue` can be returned too.
+* To mark a workflow as dynamic, `workflow_dynamic` class method is called. Only one dynamic workflow can be present for
+  a worker. The workflow can retrieve its workflow type name via `Workflow.info`
+* Dynamic signals, queries, and updates are enabled via `dynamic: true` on the calls before the methods.
+  * Dynamic handlers first argument is the string name (it's not a symbol), and all successive arguments are the actual
+    arguments.
+  * Only one dynamic handler of a type (signal, query, or update) can be on a workflow.
+
+## Workflow Asynchrony
+
+Ruby does not have a built-in task/promise/future/etc concept. However it does have fibers and we can leverage those.
+
+### Custom Fiber Scheduler
+
+We will write a custom Fiber scheduler. This will provide the following benefits:
+
+* Can leverage `Timeout` for timeouts.
+* Can leverage `sleep` for sleeps.
+* Able to fail some IO calls.
+* Allows usage of most of https://github.com/socketry/async.
+  * We will have many tests confirming many behaviors here but we still get to avoid a runtime dependency.
+  * ❓ Why not just take a dependency on `async` gem? It does have a deterministic scheduler and does a decent job of
+    everything we want, but not only do we want to avoid transitive dependencies in our library (for versioning and
+    other reasons), we cannot rely on its logic into the future.
+
+### Futures
+
+Since Ruby does not have a concept of a task/promise/future, we have to create one. Users need all the common features
+of a future without us depending on a library. Here's a high-level view of what the class may look like:
+
+```ruby
+class Future
+  # Returns [Future, Setter].
+  def new_with_setter
+
+  # Returns future whose result is set to the first completed future's result or
+  # failed with the first future's failure.
+  def self.any_of(*futures)
+  # Returns future whose result is set to nil when all futures succeed or
+  # has its failure set to the first failure.
+  def self.all_of(*futures)
+
+  # Returns future whose result is set to the first future completed (so a
+  # future in a future).
+  def self.any_of_no_raise(*futures)
+  # Returns future whose result is set to nil when all futures complete
+  # regardless of whether any future failed.
+  def self.all_of_no_raise(*futures)
+
+  # Starts the provided block in the background. Block result will be future result,
+  # block exception will be future failure.
+  def initialize(&)
+
+  # True even if the result was set with nil
+  def result?
+  # Nil if not yet set _or_ set with nil
+  def result
+
+  def failure?
+  # Nil if not (yet) failed
+  def failure
+
+  # Raises on failure
+  def wait
+  # Does not raise on failure, returns
+  def wait_no_raise
+
+  class Setter
+    # Fails if this or failure are already set
+    def result=
+    # Fails if this or result are already set
+    def failure=
+  end
+end
+```
+
+Notes:
+
+* Cancellation can get confusing. We can either immediately cancel when waiting or we can let the futures cancel
+  themselves.
+  * How SDKs handle this today:
+    * Go - Multi wait (`Selector.Get`) does not cancel immediately, and neither does `Future.Get`
+    * Java - Multi wait (`Promise.anyOf`) does not cancel immediately, and neither does `Promise.get` (but there is
+      `Promise.cancellableGet`)
+    * TypeScript - Multi wait (`Promise.race`/`Promise.any`) does not cancel immediately, and neither does
+      `await promise`
+    * Python - Multi wait (`asyncio.wait` + `FIRST_COMPLETED`) does cancel immediately, and so does `await future`
+    * .NET - Multi wait (`Workflow.WhenAny`) does not cancel immediately, and neither does `await task`
+  * For Ruby we have chosen to match all non-Python and not interrupt on cancellation on any of the calls, but we do
+    offer a `Workflow.cancellation_future` that can be used in `any_of`.
+    * This has the added benefit that `async` library will continue to work well with our futures and vice-versa.
+* We have chosen not to have additional combinators (e.g. `then` or `compose`) because these are easy to compose on
+  their own and we can add later if we want, but we want a simple API at first.
+* Nothing in the "Workflow API" returns these futures because it is easy enough to wrap.
+* Unlike some other Ruby libraries (e.g `concurrent-ruby`), we do not wrap exceptions. That means, like our other SDKs,
+  if you need to know in a `all_of` scenario which future failed, you either check the futures or put more detail into
+  the thing raised from the future block.
+  * 💭 Why not wrap exceptions? Because certain exception types are treated differently by the outer workflow failure
+    handling logic to differentiate task failure from workflow failure, and we don't want to have special
+    unwrap-from-future-fail in there. It can get confusing for users to catch too.
+* We chose not to make whether `any_of` raises be configurable. There is a lot of confusion about what return value
+  and exception situation should be across the industry in different languages. We're taking a very simple approach
+  and if users struggle with `any_of` and needing to know which future succeeded/failed, they can be more specific in
+  their return/raise inside their future block.
+  * We chose to add an `any_of_no_raise` that does not raise.
+    * 💭 Why not just an arg? Because it would change return value which is a confusing thing to do just based on arg.
+    * 💭 Why not `any_of!`? Because it's not necessarily more dangerous needing `!`, just a different behavior.
+* We chose not to make whether `all_of` raises be configurable. There is a lot of confusion about what return value
+  and exception situation should be across the industry in different languages. We're taking a very simple approach
+  and if users struggle with `all_of` need to wait for other futures to succeed when one fails, they need to make sure
+  none fail (can return exception instead or something).
+  * We chose to add an `all_of_no_raise` that does not raise.
+    * 💭 Why not just an arg? Because we have `any_of_no_raise` and this is consistent with that.
+    * 💭 Why not `all_of!`? Because it's not necessarily more dangerous needing `!`, just a different behavior.
+
+Example of running many activities and waiting on all to complete:
+
+```ruby
+class ManyActivityWorkflow < Temporalio::Workflow
+  def execute(count)
+    Temporalio::Workflow::Future.all_of(count.times.map do
+      Future.new do
+        Workflow.execute_activity(MyActivity, 'some arg' start_to_close_timeout: 5)
+      end
+    end).wait
+  end
+end
+```
+
+### Task Running
+
+Temporal workflow tasks for the workflow worker portion are short-lived CPU-bound work. While these usually only take a
+few milliseconds to run, they still use a thread and so we need to give the user some control over this.
+
+* We will extract the thread pool from `Temporalio::Worker::ActivityExecutor::ThreadPool` to
+  `Temporalio::Worker::ThreadPool` (and have activity executor use it).
+* `Temporalio::Worker::WorkflowTaskExecutor` will exist with two implementations
+  * `Temporalio::Worker::WorkflowTaskExecutor::ThreadPool` that accepts a thread pool which defaults to max threads
+    unbounded.
+    * There is a `default` class method with a lazily created global default.
+    * ❓ Is this an acceptable default? Python uses `[Etc.nprocessors, 4].max` as the max threads, but we figure it can
+      be unbounded here and remain bounded by the Core `max_concurrent_workflow_tasks` (TODO: expose this, it was left
+      off in `main`).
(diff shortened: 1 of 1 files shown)


ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+* Everything is subject to change both during proposal time and afterwards during implementation.
+* For each place a decision is made, a "💭 Why X? ..." aside is present explaining why.
+* For each place an open question exists, a "❓ X?" exists to ask it.
+
+Here is a high-level overview of potentially controversial decisions:
+
+* `Temporalio::Workflow` is both the class to extend and the class with class methods for calling statically to do
+  things.
+  * The `workflow_` class methods on the `Temporalio::Workflow` can only be called at class definition time and cannot
+    be called on `Temporalio::Workflow` directly. These are for adjusting the definition.
```
dandavison, line 19: When a user writing client code writes something like

```ruby
supported_languages = wf_handle.query(
  GreetingWorkflow.<TAB>
```

will their IDE show them all the "framework" methods (including the unavailable methods starting `workflow_`). It is nice that in other languages the user's IDE only offers the methods from their own workflow interface. I think we would like to have the same be true for Ruby if possible.

cretz: No, I am afraid this is only available at runtime (via `method_missing`). Ruby does not allow you to access an instance method at the class level.

> I think we would like to have the same be true for Ruby if possible.

Completely agree, but I cannot think of a way to do this without some form of proxying which were trying to avoid for a few reasons (mostly because we'd have to lie about what the type actually is). And even if we did do proxying, Ruby is often not expressive enough to represent this well statically. The only way I could kinda see it is if we:

```
class Workflow
  # @return [self]
  def self.stub
    StubDelegator.new(self.class)
  end
end

# This if you want to keep accurate with typing
my_wf = MyWorkflow.stub
my_wf.update('my arg') # How to use handle and args

# This if you want better options
my_wf_handle.update(MyWorkflow.stub.update, 'my arg') # So now 'update' is not the right type, checkers complain
```

Open to suggestions here, but I fear Ruby just cannot statically express things to the IDE like this. Ruby developers are used to having to use symbols or runtime methods that are not in their IDE. Of course a user can provide YARD macros or Sorbet/RBI types to support this.

dandavison: You've probably already implicitly explained this, but in Python this is done by implementing static things like `@query` decorators as functions on a `workflow` module. Can you (re-)summarize why an approach like that doesn't work for Ruby?

cretz: In Python you can reference an _instance_ method via `Class.instance_method`, but in Ruby you cannot. You must have an instance of a class to reference an instance method on it. `FooWorkflow.my_query_method` will give a `NoMethodError`. You can do hacky things like `FooWorkflow.instance_method(:my_query_method)` to get an unbounde method, but you're into reflection there not IDE help.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+* We will extract the thread pool from `Temporalio::Worker::ActivityExecutor::ThreadPool` to
+  `Temporalio::Worker::ThreadPool` (and have activity executor use it).
+* `Temporalio::Worker::WorkflowTaskExecutor` will exist with two implementations
+  * `Temporalio::Worker::WorkflowTaskExecutor::ThreadPool` that accepts a thread pool which defaults to max threads
+    unbounded.
+    * There is a `default` class method with a lazily created global default.
+    * ❓ Is this an acceptable default? Python uses `[Etc.nprocessors, 4].max` as the max threads, but we figure it can
+      be unbounded here and remain bounded by the Core `max_concurrent_workflow_tasks` (TODO: expose this, it was left
+      off in `main`).
+    * ❓ Should the default share the same thread pool as activity executor?
```
Sushisource, line 496: I don't think so, specifically because of some of the workflow isolation stuff that may or may not need to happen in them. Seems easier and cleaner to keep them separate without much downside.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+
+* ❓ Determine performance impact with this present vs not
+
+### Ractors
+
+We will offer a way to run workflows in a `Ractor`. This will provide us protection against state mutation. The downside
+is that not only are Ractors not stable and may have bugs, but they also give off a warning. A workflow instance will be
+in a Ractor and Core activations will be communicated.
+
+* ❓ Should we enable this by default, disable this by default, or somehow force the user to make a choice
```
Sushisource, line 574: I don't think anyone cares about choosing between these two really. I think we need to figure out what has acceptable performance and correctness and do that. The warning / experimental status for Ractors seems like a nonstarter, though.

But, other than that, I can't really imagine why someone is coming along and deciding "Oh I really need to use Ractors and not TracePoint" for some reason. The only thing they care about is that the workflows run fast & errors get caught.

That is, of course, excepting some other weird interactions that aren't mentioned here or we don't know about.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+
+    # Other nested classes omitted for brevity
+  end
+end
+```
+
+Notes:
+
+* Notice lack of `timeout` on `wait_condition`, we will support Ruby `Timeout` via the fiber scheduler.
+* Notice lack of `sleep`, we will support Ruby `sleep` via the fiber scheduler.
```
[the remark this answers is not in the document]
cretz: Hrmm. Ruby explicitly delegates these two things to the scheduler: https://docs.ruby-lang.org/en/master/Fiber/Scheduler.html#method-i-timeout_after and https://docs.ruby-lang.org/en/master/Fiber/Scheduler.html#method-i-kernel_sleep.

The concern with erroring on these is that we are probably going to recommend use of the https://github.com/socketry/async library (and definitely heavily test for its use) because Ruby doesn't have good async primitives and that library leans on existing fibers. So users expect to be able to call https://socketry.github.io/async/source/Async/Task/index.html#Async::Task#with_timeout and https://socketry.github.io/async/source/Async/Task/index.html#Async::Task#sleep (now deprecated because they too encourage stdlib sleep).

This is the same type of thing with `Queue` where Ruby built in support for customized async inside it so people can use it without concern. I don't expect us to say "queue.pop confuses users and we should have our own queue" either.

In this case, I don't think these are considered "magic patched", I think it's expected by users that they work.

> If the main counter argument is that you can re-use some existing code, I know I've heard agreement from you and others before that that's very rare to actually be viable.

I think that's true elsewhere, but in Ruby, the lack of async primitives means we have to either write a ton of them, or allow use of the common `async` library built on the same fiber abstraction we're building on (almost all of that library's primitives will work based on research so far). But open to options here. I think we need to pick between:

* Allow/encourage use of https://github.com/socketry/async
* Write our own primitives (e.g. Mutex and notification and such) and disallow/discourage use of that library

We already have the basic future primitive, so technically the second option is viable, but the proposal currently is expecting the first. (pardon the long response)

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+* A `cancellation` can be provided to every async thing, but by default it's the workflow-level one.
+* A workflow builds a `Workflow::Definition` object which contains metadata about the structure including a set of
+  `Workflow::Definition::Signal`s, `Workflow::Definition::Query`s, and `Workflow::Definition::Update`s built from the
+  handlers (but of course handlers can be added/removed at runtime via the `_handlers` hashes).
+* The `workflow_` methods above can only be called during class definition time not workflow runtime.
+* The runtime class methods above can only be called during runtime and can only be called on the `Workflow` class
+  itself. This means `Workflow.wait_condition` is ok, but `FooWorkflow.wait_condition` or `self.class.wait_condition`
+  is not.
+  * ❓ Is this acceptable? This is a consequence of us wanting to share workflow context class with workflow base
+    class. Basically at the top of these calls there's a `if self != Workflow` check.
```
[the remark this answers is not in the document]
cretz: That is definitely a good question and subject to debate and is the reason I put "`Temporalio::Workflow` is both the class to extend and the class with class methods for calling statically to do things." at the top of the controversial decisions.

So basically you have these options:

### Option 1 - Together

```ruby
class MyWorkflow < Temporalio::Workflow
  def execute
    Temporalio::Workflow.wait_condition { false }
  end
end
```

### Option 2 - Context class named separate

```ruby
class MyWorkflow < Temporalio::Workflow
  def execute
    Temporalio::WorkflowContext.wait_condition { false }
  end
end
```

### Option 3 - Workflow class named separate

```ruby
class MyWorkflow < Temporalio::WorkflowImpl
  def execute
    Temporalio::Workflow.wait_condition { false }
  end
end
```

I am using `WorkflowImpl` here because there will be a `Temporalio::Workflow::Definition` that represents the static definition built from the impl (and signal, query, and definition classes, same as there is already `Temporalio::Activity::Definition`). Open to a better name, I just struggled to come up with one.

### Option 4 - Don't even require class extension

```ruby
class MyWorkflow
  def execute
    Temporalio::Workflow.wait_condition { false }
  end
end
```

You aren't buying _that_ much with a class extension, but it is clear and is common in Ruby and is what we do for activities.

Thoughts/preference? Any other/better options here? Mixins would suffer the same class-for-two-purposes reuse issue.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+* `respond_to_missing?` + `inherited` will be leveraged such that all definition-time defined handlers will return their
+  definition if they are called as class methods. So `FooWorkflow.some_signal` will return a
+  `Workflow::Definition::Signal` if an instance method of `some_signal` exists marked as a `workflow_signal`.
+  * 💭 Why? It is helpful for callers/client-side to be able to reference something more explicit than a symbol, and it
+    also helps with making sure if the name is changed we use that name instead of the method name. Symbols can always
+    be used though.
+  * ❓ Is this acceptable or too much magic?
```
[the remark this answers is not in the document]
cretz: Agreed. It is not very common in Ruby to have class methods and instance methods of the same name with different behavior, though it does happen here and there. But I think our use case uniquely justifies it.

The bigger struggle is type safety integration, but Ruby type safety is in such poor state I 1) cannot find a good way to support caller-side type safety in RBS/Sorbet (unless relying on something like [this](https://github.com/Shopify/tapioca?tab=readme-ov-file#writing-custom-dsl-compilers) which doesn't work in IDEs anyways), and 2) do not want to degrade ease of use by the majority of users using this as normal Ruby without type safety checks. Users can of course write their own signatures for these things, e.g. in RBS you might have:

```ruby
class MyWorkflow < Temporalio::Workflow
  def self.my_update: -> Temporalio::Workflow::Definition::Update[MyReturnType, MyArgType]

  def my_update: (MyArgType arg) -> MyReturnType
end
```

I am open to any/all designs here. Requirements for others reading:

* Workflow handlers for signal, query, and update must be easily implemented as instance methods
* Workflow handlers for signal, query, and update must be easily usable by clients without having an instance of the workflow (we don't want fake/proxy objects where method invocation does different things)

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+  * ❓ Is there a better way to reference an interaction outside of the instance of the class that works well with Ruby
+    generic typing frameworks, e.g. Sorbet?
+    * We are trying to avoid the proxy model that Java takes because we have found it is too confusing to users to hide
+      that method invocation is actually, say, activity invocation.
+
+## Dynamic Workflows, Signals, Queries, and Updates
+
+* Support dynamic pieces similar to how we have done in Python and .NET.
+* Create a new `Temporalio::Converters::RawValue` object that is a pass-through object for raw values (still subject to
+  payload codecs), and a `payload_converter` is available in workflow/activity context to convert it.
```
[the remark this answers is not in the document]
cretz: 👍 Python and .NET have these and they are very valuable (even for people not using dynamic, but just want to defer/control conversion for whatever reason)

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+  # Returns future whose result is set to nil when all futures succeed or
+  # has its failure set to the first failure.
+  def self.all_of(*futures)
+
+  # Returns future whose result is set to the first future completed (so a
+  # future in a future).
+  def self.any_of_no_raise(*futures)
+  # Returns future whose result is set to nil when all futures complete
+  # regardless of whether any future failed.
+  def self.all_of_no_raise(*futures)
```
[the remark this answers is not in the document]
cretz: > because what's the difference between `any_of_no_raise(...).wait` and `any_of_no_raise(...).wait_no_raise`?

No big difference if you have both of these calls together. But they have value independently (the first one returns a future the second you have chosen not to even be able to look at the failure) and have value as consistent shortcuts. I can't imagine anyone would call the latter combo, `wait_no_raise` will more often be used for people using singular futures they don't care about.

Most languages have forms of these. But I am open to better names though don't want to get too inconsistent with each other and don't want these to be arguments since they change the return type. Basic requirements I think are:

* Ability to wait for first of multiple futures or raise on first failure
* Ability to wait for first of multiple futures but do not raise on first failure
* Ability to wait on all of multiple futures or raise on first failure
* Ability to wait on all of multiple futures but do not raise on first failure
* Ability to wait on a future or raise on failure
* Ability to wait on a future and not raise on failure

This was my best attempt after many permutations to get those 6 methods in. Definitely open to clearer suggestions.

ruby/sdk-phase-2.md
cretz: There is a difference between thread pool and executor in this case. One benefit of reusing the thread pool (but still different executors of course) is that it is an "unbounded caching thread pool" which means the same threads could be reused across workflows and activities. The threads, once returned back to the pool, should have no side effects from its use. By having separate unbounded thread pools here, both pools will have idle threads that the other could be using unnecessarily. Definitely not an easy/hard thing to share vs keep separate because we will have lazily created global defaults for these (and already do for activity executor).

ruby/sdk-phase-2.md
cretz: > The warning / experimental status for Ractors seems like a nonstarter, though.

I am not sure this is true. The Ractor gives so much benefit to ensuring safe determinism, it's basically a language-supported sandbox. And if we don't enable them by default (or force the user to choose whether to use them for now), we never can enable them by default. And they are quite decent for now for limited uses from my understanding.

> I can't really imagine why someone is coming along and deciding "Oh I really need to use Ractors and not TracePoint" for some reason

This section is purely for Ractors, I think we will turn on TracePoint by default.

It is a very tough call here on whether to enable Ractors by default, because Ractors are a perfect fit for this use case and we would support turning them off, and if we have a non-Ractor default now, we will never be able to change to the safer Ractor default if we want to later.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
Sushisource: OK, yeah I can buy the argument about socketry using these under the hood.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
Sushisource: I like option 3. Clear separation, just as terse as option 1.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
Sushisource: Yeah I would then change the names a-la my suggestions - the most important part I think is just not to use `no_raise` in these names since it's not _really_ about not raising, it's about not short-circuiting in the event one fails.

ruby/sdk-phase-2.md
Sushisource: > we never can enable them by default.

So why is that true?

I think the warning isn't so much the problem as the content of the warning that basically says "these are buggy". If that's not _actually_ true, then cool, I think we can be fine with it, but, we should know that for sure.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Yeah, which is basically the same as option 2 just different on who gets the "Workflow" part. I will say I like "Context" for users more than I like "WorkflowImpl" for extenders, but I do like callers of workflow utilities to only need `Temporalio::Workflow` instead of `Temporalio::Workflow::Context` (and what drove my original approach). Either way, I am starting to lean towards separating what you extend from what you invoke. Any ideas for other names though besides either "Context" for workflow utilities or "WorkflowImpl" for the extension class?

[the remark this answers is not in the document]
Sushisource: Not really lol. Those are pretty obvious choices. Maybe `WorkflowDefinition` like I suggested instead of impl. I would err on making that longer since you type it way less.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: > Yeah I would then change the names a-la my suggestions 

Do you have a specific suggestion on names? So instead of:

* `any_of`
* `any_of_no_raise`
* `all_of`
* `all_of_no_raise`
* `wait`
* `wait_no_raise`

reading comment above, something like:

* `any_of`
* `any_of_success` - says success but includes failure?
* `try_all_of` - fails future if any fail?
* `all_of` - does not fail future if any fail, so one "x_of" doesn't ail the future and one does?
* `wait`
* `wait_no_raise`

Maybe something like `try` instead of `_no_raise`, e.g.

* `any_of`
* `try_any_of`
* `all_of`
* `try_all_of`
* `wait`
* `try_wait`

ruby/sdk-phase-2.md
cretz: > So why is that true?

Because you can't go from a less-restrictive default to a more-restrictive default without breaking people. Ractors prevent state sharing (e.g. a global is different in each Ractor). Also, things that are shared across Ractors must be marked as such via https://docs.ruby-lang.org/en/master/Ractor.html#method-c-make_shareable (which is an awesome feature and we already require this be done with payload and failure converters knowing they may be used inside sandbox). So turning on Ractors by default when they weren't before would break a lot of people that were, say, using their logger or OTel impl or other things people sometimes opt out of sandboxes for.

> I think the warning isn't so much the problem as the content of the warning that basically says "these are buggy". If that's not actually true, then cool, I think we can be fine with it, but, we should know that for sure.

We definitely have to test here, and they may be buggy for some advanced uses and maybe not our own. But even if that's the case, defaulting to experimental thing is rough, even though plenty of Ruby users do use Ractors today.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
Sushisource: I like the try names, but I think they should actually be the short-circuiting versions. IE: Existing `any_of` and `all_of` become the `try_xxx` versions, and the current `_no_raise` versions just become `any_of` and `all_of`. Since "try" implies "try until something breaks". This also matches the Rust naming scheme for these, FWIW.

I would probably keep `wait_no_raise` though rather than `try` there just because that's more obvious. And, I think, to my point, that is actually about raising or not where the other stuff is about short-circuiting.

ruby/sdk-phase-2.md
Sushisource: Ah. I see. But arguably they already need to not be doing all that stuff for proper workflows anyway?

I think, if we find that we're not running into implementation bugs, it can be an OK default.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+We will write a custom Fiber scheduler. This will provide the following benefits:
+
+* Can leverage `Timeout` for timeouts.
+* Can leverage `sleep` for sleeps.
+* Able to fail some IO calls.
+* Allows usage of most of https://github.com/socketry/async.
+  * We will have many tests confirming many behaviors here but we still get to avoid a runtime dependency.
+  * ❓ Why not just take a dependency on `async` gem? It does have a deterministic scheduler and does a decent job of
+    everything we want, but not only do we want to avoid transitive dependencies in our library (for versioning and
+    other reasons), we cannot rely on its logic into the future.
```
josh-berry, line 375: Is `async`s deterministic behavior actually part of its contract or no? (I'm thinking of how we got burned by Python's change in behavior for asyncio; I imagine you are too)

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Hrmm, to me "try" means failure is captured, e.g. .NET `int.Parse` vs `int.TryParse`. One fails as normal, another captures failure (same as the actual `try` statement in many languages).

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+
+* We will extract the thread pool from `Temporalio::Worker::ActivityExecutor::ThreadPool` to
+  `Temporalio::Worker::ThreadPool` (and have activity executor use it).
+* `Temporalio::Worker::WorkflowTaskExecutor` will exist with two implementations
+  * `Temporalio::Worker::WorkflowTaskExecutor::ThreadPool` that accepts a thread pool which defaults to max threads
+    unbounded.
+    * There is a `default` class method with a lazily created global default.
+    * ❓ Is this an acceptable default? Python uses `[Etc.nprocessors, 4].max` as the max threads, but we figure it can
+      be unbounded here and remain bounded by the Core `max_concurrent_workflow_tasks` (TODO: expose this, it was left
+      off in `main`).
```
josh-berry, line 495: Uhmmm, unbounded thread creation seems scary to me. I think I'm OK leaving it bounded by `max_concurrent_workflow_tasks` as long as it's actually designed that way intentionally and it's not just something that happens by accident (e.g. it should be explicit in our code/code comments that this is the intention).

ruby/sdk-phase-2.md
cretz: > Ah. I see. But arguably they already need to not be doing all that stuff for proper workflows anyway?

Usually, yes, but some things people accept they need to cheat (e.g. interceptor sending exception to sentry when not replaying, or a logger that writes to disk but is not the logger we offer)

> I think, if we find that we're not running into implementation bugs, it can be an OK default.

I'm leaning there too. But it will be super loud that we use them and how not to.

josh-berry: My vote is to force users to explicitly specify whether to use Ractors or not.

If we think we might want to change the default in the future, I think it's fine to require users to explicitly specify it now and remove the requirement later (so long as we're not asking them to do that for too many things). This enables us to migrate to the more restrictive default for future new users without breaking existing users.

We did something similar for Update IIRC.

josh-berry (commented):
general non-blocking comments, no need to wait for my approval.

ruby/sdk-phase-2.md
cretz: > Is asyncs deterministic behavior actually part of its contract or no?

Not explicitly, but most of the things it does are delegated to the underlying fiber scheduler. The other problem is not just that it is deterministic, but that it also doesn't make changes that would be incompatible (e.g. schedule two fibers in a different order).

Concerning determinism, they do say as a comment inside https://github.com/socketry/async/blob/72fae62658b68e85d53d4c6adf13d3b34108cfe8/lib/async/scheduler.rb#L428-L435 that they intentionally try to preserve determinism, but obviously a code comment is not a clear contract (and that's their scheduler, we'll have our own).

It's a tough call, because it's a full featured library with lots of utilities that leans on the deterministic fiber scheduler, and we surely don't want to recreate _every_ construct they have already created. But maybe we don't want to encourage use of this project? Java SDK gets along just fine with its primitive `Promise` support. We still want to lean on traditional Ruby things like queues, timeouts, and sleep I think, but we'd have to write our own mutexes, semaphores, etc.

ruby/sdk-phase-2.md
cretz: It is bounded by that from Core. But that is per worker and the thread pool is shared across workers for obvious benefits. So yes, Core ensures it cannot be more than the sum of max concurrent tasks across all workers using the thread pool.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: See https://github.com/temporalio/proposals/pull/96#discussion_r1817109910, maybe we don't want to encourage this `async` lib after all. It does put too much trust in their deterministic fiber scheduler use (even across versions).

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,575 @@
...
+```ruby
+class CallGreetingServiceActivity < Temporalio::Activity
+  GREETINGS = {
+    arabic: 'مرحبا بالعالم',
+    chinese: '你好，世界',
+    english: 'Hello, world',
+    french: 'Bonjour, monde',
+    hindi: 'नमस्ते दुनिया',
+    portuguese: 'Olá mundo',
+    spanish: '¡Hola mundo'
```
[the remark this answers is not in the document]
dandavison: Thanks! https://github.com/temporalio/samples-python/pull/149

I removed the exclamation marks from all of them after I realized I had it on the wrong end for Arabic... but apparently I missed the leading one in Spanish.

ruby/sdk-phase-2.md
cretz: Hrmm, no SDK requires extra options for running workflows. I think if we choose Ractors we can change the default to be less restrictive if we ever had to (though I don't suspect we ever would).

> We did something similar for Update IIRC.

This is not comparable I don't think because we _know_ we will provide a default in the future but it is not yet implemented. Here, we don't know, so this isn't necessarily temporary. I think we might as well decide now, because in 5 years I don't think our decision would change.

ruby/sdk-phase-2.md
josh-berry: Yeah, this would be a much easier call if determinism were an explicit design goal, but since it's not as such, I think it's probably safer to avoid the dependency for now. Perhaps we could align our own API surface with its API so we could switch later if there is user demand? That might also help reduce the cognitive overhead of learning Temporal if the user already knows `async`.

cretz: > I think it's probably safer to avoid the dependency for now

To clarify, we would never have a dependency, just that it would happen to work and we could recommend its use (and we'd test that it does in our tests). Having said that, I think I agree maybe we should not even recommend its use or discuss it.

> Perhaps we could align our own API surface with its API so we could switch later if there is user demand

It's too large. I think we can have `Future` (what they call `Task`) and `Mutex` manually implemented, and then decide whether to explicitly implement `sleep` and `Timeout` and `Queue` or allow reuse of the standard library ones (knowing in Ruby this isn't the same as leaning on a large set of async primitives, these are 3 very limited low-level common Ruby things).

I think I'm leaning towards no `async` library recommendation, `Future` + `Mutex` implemented, `sleep` + `Timeout` + `Queue` just work from the standard library.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
Sushisource: The `try` here is like "try them, but fail if any fails soon" whereas without it is "definitely do all the futures even if some of them fail".

But, I wouldn't cry about having it the other way around, I think that can work too w/ clear docs. I do think `wait` and `wait_no_raise` should stay that way though.

ruby/sdk-phase-2.md
cretz: We have confirmed that even for Sorbet users, custom DSL compilers don't run in IDEs by default anyways. I cannot see a way at this time to support typesafe workflow messages without a fake/proxy type of object which I am trying to avoid. Ruby users don't really expect DSLs to be well typed in their IDE.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Ok, after some internal discussion, I think I'm landing at something like:

```ruby
module Temporalio
  class Workflow
    class << self
      protected

      def workflow_name(name)
        raise 'TODO'
      end
    end

    module API
      def execute_activity(activity, blah1, blah2)
        raise 'TODO'
      end
    end

    extend API
  end
end
```

So basically you have protected class methods for adjusting the workflow definition, then an `API` set of module methods in the workflow class by default (`extend` in this way I think avoids YARD assuming they are part of the class, still investigating). This will allow others to write utilities for workflows using `Temporalio::Workflow::API` directly (or extending it themselves too). Doing a bit more investigation before accepting this approach.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Sure, so `wait`, `wait_no_raise` settled. Any other term we can think of for "swallow error" form of the others besides `try`? I will update proposal with the `try_` form, but definitely open to other words here.

cretz:
Updated with the following changes:

* Workflows now must extend from `Temporalio::Workflow::Definition`
* Activities now must extend from `Temporalio::Activity::Definition`
* Ractor-based task execution is now on by default
* Clarified that concurrency primitives in the standard library can be used
* We now _discourage_ use of the `async` library
* Changed names of no-failure `Future` class methods

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Fixed

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Clarified that we don't want to encourage `async` but the stdlib primitives that work in all concurrent environments (i.e. thread and fiber) can be used, but we will alias/wrap them.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: After discussion, proposal updated to require workflows extend from `Temporalio::Workflow::Definition` and use module functions on `Temporalio::Workflow`. As a consistency consequence, activities will now also have to extend from `Temporalio::Activity::Definition`.

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
cretz: Proposal updated to change `any_of_no_raise` and `all_of_no_raise` to `try_any_of` and `try_all_of` and clarify they do not raise when `wait`ed on.

ruby/sdk-phase-2.md
cretz: Proposal updated to Ractors by default with easy, well-documented way to opt out.

ruby/sdk-phase-2.md
cretz: Proposal updated to discourage use of `async` library and clarify that we do support stdlib things that work in all environments.

ruby/sdk-phase-2.md
josh-berry: Works for me.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+  def initialize(name)
+    @greetings = {
+      chinese: '你好，世界',
+      english: 'Hello, world'
+    }
+    @language = :english
+    @mutex = Temporalio::Workflow::Mutex.new
+  end
+
+  def execute
```
drewhoskins-temporal, line 66: nit: might want to show how args appear here.
(_should_ they appear here, or should they just be part of initialize?)
I think having both methods take them is a bit confusing, but I can buy the convenience of having `execute` take them.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+  end
+
+  workflow_update
+  def set_language(language)
+    previous, @language = @language, language.to_sym
+    previous
+  end
+
+  workflow_update_validator(:set_language)
+  def validate_set_language(language)
```
drewhoskins-temporal, line 92: What would be the experience if they don't keep the args in sync b/w validator and handler?  I'd like them to get reasonable feedback.
suggestion: This seems like one to see if sorbet/rbs can wrangle static typesafety via a generic, perhaps by putting in a nested class.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+
+```ruby
+client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
+worker = Temporalio::Worker.new(
+  client:,
+  task_queue: 'my-task-queue',
+  activities: [GreetingActivity],
+  workflows: [GreetingWorkflow]
+)
+worker.run(shutdown_signals: ['SIGINT'])
```
drewhoskins-temporal, line 128: Is `shutdown_signals` a choice that the user needs to make, or is there a good default?

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+wf_handle = client.start_workflow(
+  GreetingWorkflow,
+  id: 'my-workflow-id',
+  task_queue: 'my-task-queue'
+)
+
+# Send query
+supported_languages = wf_handle.query(
+  GreetingWorkflow.languages,
+  { include_unsupported: false }
```
drewhoskins-temporal, line 146: What's our static type safety story here?  Probably worth making sure that .rbi users can get good static validations..

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+worker.run(shutdown_signals: ['SIGINT'])
+```
+
+Example client usage:
+
+```ruby
+client = Temporalio::Client.connect('localhost:7233', 'my-namespace')
+
+# Start workflow
+wf_handle = client.start_workflow(
```
drewhoskins-temporal, line 137: re: subclassing `Workflow::Definition`, you mentioned that as a controversial decision.  Is the reason for that so that clients can refer to just the workflow here, without pulling in the definition which might be in a separate module?  If so, that seems like an okay tradeoff to me.
IIRC, coinbase/temporal ruby used the _instance_ to be the implementation specific stuff whereas the _class_ was used by the client.  I don't have a strong idea of how good that was, since Stripe chose not to isolate client concerns from definition concerns, so I don't know if we could have done better given coinbase SDK.  This was a usability win but ended up biting us after a few years after we got religion on packaging modules.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+
+```ruby
+module Temporalio
+  module Workflow
+
+    # Class users must extend for a workflow
+    class Definition
+      class << self
+        # Protected because they are only for this implementation
+        protected
```
drewhoskins-temporal, line 208: nit: You probably mean `private` as I don't think you need one instance accessing another instance's methods.
(`protected` is very niche in ruby, and `private` means what `protected` means in .NET)

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+        def self.workflow_update(name: nil, dynamic: false, raw_args: false,
+                                unfinished_policy: HandlerUnfinishedPolicy::WARN_AND_ABANDON)
+        def self.workflow_update_validator(update_method)
+
+        # Exposes an attribute as a query. 💭 Why? We found this _very_ valuable in
+        # .NET. Think of it as attr_reader for workflows. It was also decided there
+        # was not enough value to have attr_accessor shortcut for update or
+        # attr_writer shortcut for signals since both are not commonly just
+        # getter/setter.
+        def self.workflow_query_attr(query_method)
```
drewhoskins-temporal, line 230: This is cool.
nit: I wonder if `workflow_query_attr_reader` would more grokkable by the ruby community, despite the partial reduncancy of query vs reader.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+
+        # Called in the workflow definition to change the name from the default.
+        def self.workflow_name(name)
+
+        # Placed above initialize to get start arguments.
+        def self.workflow_init
```
drewhoskins-temporal, line 236: as mentioned above, I wonder if we can avoid this rigmarole since we're starting fresh.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+
+## Dynamic Workflows, Signals, Queries, and Updates
+
+* Support dynamic pieces similar to how we have done in Python and .NET.
+* Create a new `Temporalio::Converters::RawValue` object that is a pass-through object for raw values (still subject to
+  payload codecs), and a `payload_converter` is available in workflow/activity context to convert it.
+  * Handler methods will use `RawValue` for args if told to, similar with workflow `execute`/`initialize` if
+    `workflow_raw_args` is set.
+  * `RawValue` can be returned too.
+* To mark a workflow as dynamic, `workflow_dynamic` class method is called. Only one dynamic workflow can be present for
```
drewhoskins-temporal, line 358: I think this is a layering violation.  The workflow itself doesn't need to know whether it's dynamic; instead, it should be registered as dynamic on the worker, which also makes it easier to select which is the dynamic workflow for the worker.  Without that, if the user has two workflows registered as `dynamic_workflow` in the same codebase, it would barf even if only one is registered in the worker.
See https://github.com/coinbase/temporal-ruby/blob/f41efb7bb0c9e755a07382b94e00a4f31bcd33c1/lib/temporal/worker.rb#L97 for what we did in the coinbase SDK.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+  a worker. The workflow can retrieve its workflow type name via `Workflow.info`
+* Dynamic signals, queries, and updates are enabled via `dynamic: true` on the calls before the methods.
+  * Dynamic handlers first argument is the string name (it's not a symbol), and all successive arguments are the actual
+    arguments.
+  * Only one dynamic handler of a type (signal, query, or update) can be on a workflow.
+
+## Workflow Asynchrony
+
+Ruby does not have a built-in task/promise/future/etc concept. However it does have fibers and we can leverage those.
```
drewhoskins-temporal, line 367: This looks like this is out of date and seems to contradict your `Future` stuff below?

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+    * 💭 Why not just an arg? Because we have `try_any_of` and this is consistent with that.
+    * 💭 Why not `all_of!`? Because it's not necessarily more dangerous needing `!`, just a different behavior.
+
+Example of running many activities and waiting on all to complete:
+
+```ruby
+class ManyActivityWorkflow < Temporalio::Workflow
+  def execute(count)
+    futures = count.times.map do
+      Future.new do
```
drewhoskins-temporal, line 496: I found Temporal ruby's [implicit Futures](https://github.com/coinbase/temporal-ruby/blob/f41efb7bb0c9e755a07382b94e00a4f31bcd33c1/README.md?plain=1#L269-L270) weird, so this could work well.
However, please try to make sure that return types propagate through the future for static types.  Would be a shame to lose type safety for async.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+    # Simulate a network call
+    sleep(0.2)
+    GREETINGS[language.to_sym]
+  end
+end
+
+class GreetingWorkflow < Temporalio::Workflow::Definition
+  workflow_query_attr :language
+
+  def initialize(name)
```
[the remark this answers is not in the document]
cretz: I don't think it should be the default and we chose not to make it the default in .NET even though `[WorkflowInit]` was present since its inception. There is a `workflow_init` below to facilitate opting in just like every other SDK that allows it, not make Ruby some special thing here. Would also like not to eagerly validate that the constructor arguments match the execute arguments if not opted in.

ruby/sdk-phase-2.md
cretz: >  (should they appear here, or should they just be part of initialize?)

Here

>  I think having both methods take them is a bit confusing, but I can buy the convenience of having execute take them.

Like every other SDK

ruby/sdk-phase-2.md
cretz: This fails at register time like every other register-time validator in every other SDK

ruby/sdk-phase-2.md
cretz: This is a choice that a user needs to make. This is an application not a library and it's not neighborly for us to trap signals by default (or have any global side effects by default). This is already in `main` btw and was part of the activity PR.

ruby/sdk-phase-2.md
cretz: There is not a static type safety story at this time. It is still under investigation on the base way to do this.

ruby/sdk-phase-2.md
cretz: I do mean `protected` very intentionally and am aware of its quirks and its rareness in Ruby use (we don't use it elsewhere). These are for use by a workflow, e.g.

```ruby
class MyWorkflow < Temporalio::Workflow::Definition
  workflow_name 'my-custom-workflow-name'

  # ...
end
```

If `workflow_name` were private, this would fail. If `workflow_name` were public, then it will be easily invocable from the client side, but we only want signals, updates, and queries exposed as class methods client side (that return definitions), e.g. `my_handle.signal(MyWorkflow.my_signal, 'my-signal-arg')`. So in this case, protected makes sense. We considered putting these in a mixin instead, but it was unnecessary since we also want class extension anyways.

ruby/sdk-phase-2.md
cretz: :+1: Can add suffix

ruby/sdk-phase-2.md
cretz: Replied above, but I think this should be opt-in and was chosen that way in the last SDK even though we were starting fresh there too.

ruby/sdk-phase-2.md
cretz: > The workflow itself doesn't need to know whether it's dynamic

This is not true in any SDK nor should it be true here. Dynamic workflows/activities should be allowed to be in the same list at registration time with the rest of the workflows. We want authors to clearly mark and be aware that a workflow, activity, signal, query, or update is dynamic. While technically dynamic workflow and activity's signatures are no different, dynamic signal, query, and update signatures _are_ different and must opt in. So all dynamics in all forms are clearly author opt-in.

> Without that, if the user has two workflows registered as dynamic_workflow in the same codebase, it would barf even if only one is registered in the worker.

No, this is only checked at registration time, same as in every SDK with dynamics. You can have as many as you want implemented, you can only have one dynamic _per worker_. There are no globals here.

ruby/sdk-phase-2.md
cretz: This is accurate. Ruby does not have a built-in concept and we can leverage fibers. We add our own future later on top of fibers of course.

ruby/sdk-phase-2.md
cretz: Static typing is still under research, but our early research has shown it would not affect design (nor would we want it to for the majority of Ruby users that don't care about it and don't deserve to be affected by it), so we are not blocking this effort.

ruby/sdk-phase-2.md
```diff
@@ -0,0 +1,605 @@
...
+    # These are a custom hash class that only supports []= mutation and
+    # validates input as a signal/query/update definition. Basically they will
+    # be simple delegators to frozen hashes only updatable via index. Symbol
+    # keys or nil key for dynamic. All values are "definition" objects.
+    def self.signal_handlers
+    def self.query_handlers
+    def self.update_handlers
+
+    def info
+    def current_update_info
```
cretz, line 280:
```suggestion
    def self.info
    def self.current_update_info
```

ruby/sdk-phase-2.md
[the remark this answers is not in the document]
drewhoskins-temporal: TBC, what I'm proposing is just to require initialize to take start workflow args always.  That feels like the most normal and straightforward approach in any object-oriented setting, familiar to everyone.  Your approach will dramatically reduce the number of folks who take advantage.
I understand that from your perspective, the workflow init approach is "normal."  But stepping out of your Chad shoes, what we've done in other SDKs is pretty esoteric.  To most users, using the constructor is normal unless they happen to have used temporal in other languages, and even they will be plenty familiar with constructors.

ruby/sdk-phase-2.md
drewhoskins-temporal: We've often agreed that cross-SDK consistency is a secondary concern.

ruby/sdk-phase-2.md
drewhoskins-temporal: Going to comment here since that PR is in main.
I think this is important since every user will need to set up a `Worker`, and this will be part of the "getting started" scenario, so I have a few questions.
Looks like we default to `[]`.  If this is a choice that every developer needs to make, why is there a default?  What % of users do you think that default will be acceptable for?
Looking at your docs: https://github.com/temporalio/sdk-ruby/blob/67741fcbd9e354b9a8d03e7e923efe8844ccf1ca/temporalio/lib/temporalio/worker.rb#L82

I don't see any guidance.  It's typed as `String | Int` in rbs, and I don't see any validations.  How do users know what values are acceptable or get guidance on what's normal to pass in there?

cretz: >  If this is a choice that every developer needs to make, why is there a default?

It is not a choice every developer will make. This is a library and many developers will prefer to use a block body or the `Cancellation` token as a way to stop the worker(s). Signals is but one way to stop a worker, by no means the deafult and by no means one that everyone will want. These workers are just things that run how the user wants, we just provide helpers.

> How do users know what values are acceptable or get guidance on what's normal to pass in there in what situations and get feedback if they make a mistake?

How do they know that with `Signal.trap`? That's typed in https://github.com/ruby/rbs/blob/d9000d23c2c1d032a194848562806896c54fd0af/core/signal.rbs#L92-L93 and https://github.com/sorbet/sorbet/blob/0ab8f3c29260f8ecee31aa960de3692419e2d3c3/rbi/core/signal.rbi#L100-L113 as general item (I need to add `Symbol` there it seems). Why are we expecting Temporal to type signals better than RBS and Sorbet?

Sushisource (approved):

ruby/sdk-phase-2.md
cretz: Fixed

ruby/sdk-phase-2.md
cretz: Added

cretz:
Made two minor fixes, but otherwise no outstanding/unresolved things (or even updates) for a while. Merging.

PR #97, opened 2025-04-15
Temporal Nexus Python proposal

dandavison:
This is an early-stage design document for how Nexus functionality will be exposed in the Nexus Python SDK and in the Temporal Python SDK. The design is still evolving, and feedback is welcome.

The design proposed here is being prototyped in [this](https://github.com/temporalio/samples-python/pull/174) draft PR.

[Rendered markdown](https://github.com/dandavison/temporalio-proposals/blob/nexus-python/nexus/sdk-python.md)

### DIFF
--- a/nexus/sdk-python.md
+++ b/nexus/sdk-python.md
@@ -1,0 +1,330 @@
+# Python Nexus design proposal
+
+This is an early-stage design document for how Nexus functionality will be exposed in the Nexus Python SDK and in the Temporal Python SDK. The design
+is still evolving, and feedback is welcome.
+
+The design proposed here is being prototyped in this draft PR:
+[https://github.com/temporalio/samples-python/pull/174](https://github.com/temporalio/samples-python/pull/174)
+
+# Introduction
+
+A Nexus service is a collection of Nexus operations, each of which supports the four verbs `start`, `fetch_info`, `fetch_result`, and `cancel`.
+
+On the handler side of Nexus, users need to be able to:
+
+1. **Define a service interface**
+    
+    This will be consumed by caller code to create a type-safe client, and by the service implementation to verify that the interface has been implemented.
+    
+2. **Implement an operation in a service**
+
+This document focuses on these two areas (they’re the most complex aspects of Nexus SDK design).
+
+Note that this document discusses Temporal-specific Nexus at the end. Everything before that is intended to stand on its own, i.e. the part before the Temporal section is intended to make sense without any implication that Temporal is in the background.
+
+# 1. Defining a service interface
+
+The team writing the Nexus service will define the interface in Python (This Python definition will be the target of codegen from future Nexus service IDL).
+
+The interface will be defined in a dataclass-like style [thanks @cretz for this idea]:
+
+```python
+import nexusrpc
+
+@nexusrpc.service
+class MyNexusService:
+    echo: nexusrpc.Operation[EchoInput, EchoOutput]
+    hello: nexusrpc.Operation[HelloInput, HelloOutput]
+```
+
+Note that:
+
+- What the user actually writes here defines the operations as purely type-level constructs: the decorator creates `NexusOperation` instances at import time that can be referenced by the caller-side Nexus service client.
+
+The interface will be used by the caller as follows:
+
+```python
+from my_nexus_service import EchoInput, EchoOutput, MyNexusService
+
+@workflow.defn
+class EchoCallerWorkflow:
+    def __init__(self):
+        self.nexus_client = NexusClient(
+            MyNexusService,  # or string name "my-nexus-service",
+            "my-nexus-endpoint-name",
+            schedule_to_close_timeout=timedelta(seconds=30),
+        )
+
+    @workflow.run
+    async def run(self, message: str) -> EchoOutput:
+        return await self.nexus_client.execute_operation(
+            MyNexusService.echo,
+            EchoInput(message),
+        )
+
+```
+
+**[ALTERNATIVES CONSIDERED]** One might consider using a traditional (`typing.Protocol` / `abc.ABC`) interface (as Java does). However, while the caller-side Nexus service client might conform to those signatures, the handler-side service implementation would not (see below, but for at least two reasons: (1) operations are not represented by their start methods but rather by nullary factory functions returning an operation instance, and (2) even if the handler-side signature were that of the start method, it contains an option container from the request that would be extraneous in an interface definition).
+
+# 2. Implementing a Nexus operation
+
+Traditionally, in order to define an RPC/HTTP handler, you specify a function to act as the handler, and you arrange for a service or operation instance with a lifetime equal to that of the server, containing state created at server-start time, to be in-scope in the function body. For Python examples, see [gRPC,](https://github.com/grpc/grpc/blob/4e9357bca1408596663a218c0c608a4c0560a867/examples/python/route_guide/route_guide_server.py#L65) [xmlrpc [stdlib]](https://docs.python.org/3/library/xmlrpc.server.html#simplexmlrpcserver-example), [starlette/FastAPI](https://www.starlette.io/applications/#storing-state-on-the-app-instance), [Django class-based-views](https://docs.djangoproject.com/en/5.1/topics/class-based-views/). So, a Python developer writing an RPC/HTTP handler will expect to write something like
+
+```python
+# This code (from Python gRPC docs) is not part of the proposal
+class MyService:  # instantiated at server-start time
+    def my_handler(self, my_input, context) -> MyResult: ...
+```
+
+However, unlike traditional RPC/HTTP handlers, a Nexus operation represents a long-running execution and must support three additional verbs (`get_info`, `get_result`, `cancel`), and must support `start` returning both synchronously and asynchronously (including making that decision at request-handling time).
+
+## 2a. Fully manual class-based operation definition
+
+The “fully manual” way we will support this in Nexus Python is that a user will define a class for the service, and a class for each operation.
+
+**[ALTERNATIVES CONSIDERED]** None. Giving the user the option to define classes at both levels for maximum implementation flexibility is the natural choice for Python (as it was in Java).
+
+```python
+import nexusrpc.handler
+
+@nexusrpc.handler.service(interface=interface.MyNexusService, name="my-service")  # import-time check that interface was implemented
+class MyNexusService: # instantiated when instantiating worker
+    # User is free to define custom constructor
+
+    @nexusrpc.handler.operation(name="my-invalid-python-identifier")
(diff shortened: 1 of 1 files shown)


nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+
+The interface will be used by the caller as follows:
+
+```python
+from my_nexus_service import EchoInput, EchoOutput, MyNexusService
+
+@workflow.defn
+class EchoCallerWorkflow:
+    def __init__(self):
+        self.nexus_client = NexusClient(
```
cretz, line 52:
```suggestion
        self.nexus_client = workflow.NexusClient(
```

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+            MyNexusService,  # or string name "my-nexus-service",
+            "my-nexus-endpoint-name",
+            schedule_to_close_timeout=timedelta(seconds=30),
```
cretz, line 55: I think this needs to be per-operation kwarg. Java uniquely does some of these at the service/stub level because of how it proxies, but all other SDKs can do operation level.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+## 2a. Fully manual class-based operation definition
+
+The “fully manual” way we will support this in Nexus Python is that a user will define a class for the service, and a class for each operation.
+
+**[ALTERNATIVES CONSIDERED]** None. Giving the user the option to define classes at both levels for maximum implementation flexibility is the natural choice for Python (as it was in Java).
+
+```python
+import nexusrpc.handler
+
+@nexusrpc.handler.service(interface=interface.MyNexusService, name="my-service")  # import-time check that interface was implemented
```
cretz, line 90: Can we use a different name for the decorator besides service? I know being in `handler` resolves the ambiguity, but I still think we should avoid ambiguous names across the two modules of the nexusrpc project. Also, similar to operation below, when we talk about "service" we are talking about the contract, not ambiguously talking about the contract or implementation. Maybe `service_impl` or `service_handler` or something.

cretz, line 90: Not sure `interface` is a good name here. Maybe it doesn't need to be a kwarg at all?

cretz, line 90: The `name` parameter here seems like something that should be on the `service` side not the impl side. It is needed by callers too so it is part of the contract.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+class MyNexusService: # instantiated when instantiating worker
+    # User is free to define custom constructor
+
+    @nexusrpc.handler.operation(name="my-invalid-python-identifier")
+    def echo(self) -> nexusrpc.handler.Operation[EchoInput, EchoOutput]:  # typing.Protocol interface
```
cretz, line 95: IMO this should be called an operation handler and not just an operation (both decorator and this return class name). When we talk about "operations" we are talking about the contract, not ambiguously talking about the contract or implementation.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+        # Factory called when instantiating worker
+        return EchoOperation()  # User can pass anything to operation constructor
+
+class EchoOperation: # interface declaration optional; see below
```
cretz, line 99: Hrmm, I wonder if we should have this extend an `OperationHandler` ABC that has generics for proper composability. And there can be things like `SyncOperationHandler` ABC that extends this and has an `execute` function that just returns the type.

So this class would change to something like:

```python
class EchoOperationHandler(SyncOperationHandler[EchoInput, EchoOutput]):
    def execute(self, input: EchoInput, options: StartOperationOptions) -> EchoOutput:
        return EchoOutput(input.message)
```

This makes it clearer at type checking time whether it is valid. I do think we should also have an `AsyncioOperationHandler` and `SyncAsyncioOperationHandler`. I think it is confusing to have so many runtime options, especially as extensions and helpers are created.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+
+```python
+worker = Worker(
+    client,
+    task_queue=task_queue,
+    workflows=[HelloWorkflow],
+    # User has defined custom service and operation constructors,
+    # passing the service instance or the db client through
+    # to give verb handlers access to this db client
+    nexus_services=[MyNexusService(db_client=MyDBClient.connect(...))],
```
cretz, line 127: Same concerns here with reusing the name service to mean service impl. It can be confusing to a user whether they should pass in the service or the service (the sentence is ambiguous on purpose to convey confusion when discussing Nexus).

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+await worker.run()
+```
+
+An operation must implement the `nexusrpc.handler.Operation` interface:
+
+```python
+class Operation(Protocol, Generic[I, O]):
+    # start either returns the result synchronously, or returns an operation token. Which path is
+    # taken may be decided at operation handling time.
+    async def start(self, input: I, options: StartOperationOptions) -> Union[O, StartOperationResult]: ...
```
cretz, line 138: Asyncio should not only not be required, it should be discouraged as the default IMO. An `AsyncioOperationHandler` can exist for those that choose to opt-in to it, but it's not a good default IMO. Same throughout the rest of this proposal. I think it may be better to define the entire proposal without `async def` anywhere and have an "Asyncio" section showing how to do things with asyncio if you so choose.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+
+```python
+# This code (from Python gRPC docs) is not part of the proposal
+class MyService:  # instantiated at server-start time
+    def my_handler(self, my_input, context) -> MyResult: ...
+```
+
+However, unlike traditional RPC/HTTP handlers, a Nexus operation represents a long-running execution and must support three additional verbs (`get_info`, `get_result`, `cancel`), and must support `start` returning both synchronously and asynchronously (including making that decision at request-handling time).
+
+## 2a. Fully manual class-based operation definition
```
cretz, line 81: Can you show an example of defining a Nexus service dynamically at runtime? We took care in Java to only make the annotation/decorator-based forms be sugar over the manual definitions that could be created (for the services, operations, service impls, and operation handlers). You don't _have_ to use the annotation/decorator-based forms if you need to build at runtime dynamically.

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+    async def cancel(self, token: str, options: CancelOperationOptions) -> None: ...
+
+```
+
+These signatures most closely follow Go in that the primary input (start payload or token) is always pulled out as a separate parameter, and options structs contain everything relevant from the Nexus request.
+
+- `StartOperationOptions`
+    
+    ```python
+    @dataclass
```
cretz, line 150: Pedantic, but recommend frozen here

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+**[ALTERNATIVES CONSIDERED]** None. Giving the user the option to define classes at both levels for maximum implementation flexibility is the natural choice for Python (as it was in Java).
+
+```python
+import nexusrpc.handler
+
+@nexusrpc.handler.service(interface=interface.MyNexusService, name="my-service")  # import-time check that interface was implemented
+class MyNexusService: # instantiated when instantiating worker
+    # User is free to define custom constructor
+
+    @nexusrpc.handler.operation(name="my-invalid-python-identifier")
```
cretz, line 94: Like services, names of operations are part of the contract and therefore need to be on the contract side not the impl side

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+- The decorator propagates type annotations on the start method defined on the service to become run-time type annotation objects on the methods defined on the operation [[nexus-sdk PR](https://github.com/nexus-rpc/sdk-python/blob/e318efeb5b615a7dab916af1c5f01fe97ffff841/src/nexusrpc/handler.py#L294), [sdk-python PR](https://github.com/temporalio/sdk-python/blob/08430098aac4d92c3744b1245c5585819abadad3/temporalio/nexus/handler.py#L155-L169)]. This allows, e.g. a Nexus worker to instantiate the correct Python type from payloads received over the wire.
+
+**[ALTERNATIVES CONSIDERED]**
```
cretz, line 248: Not sure this is an alternative. Isn't what is below effectively what `@nexusrpc.handler.sync_operation` is sugar for?

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+
+The proposal here follows the  [2b. Convenience utilities for shorthand operation definition](#2b-convenience-utilities-for-shorthand-operation-definition) `@nexusrpc.handler.sync_operation` decorator above:
+
+```python
+@nexusrpc.handler.service(interface=interface.MyNexusService)
+class MyNexusService:
+    @temporalio.nexus.handler.workflow_operation
+    async def hello(
+        self, input: HelloInput, options: nexusrpc.handler.StartOperationOptions
+    ) -> nexusrpc.handler.StartOperationResult[HelloOutput]:
```
cretz, line 291: Could have a more specific, clearer subtype to return here, e.g. `temporalio.nexus.handler.WorkflowOperationHandler[HelloOutput]`

nexus/sdk-python.md
```diff
@@ -0,0 +1,330 @@
...
+
+It’s the “always-async operation backed by a Temporal execution primitive” that we’re discussing in this section. We’re going to provide a convenience shorthand, analogous to the one above for defining sync operations, that implements a workflow-backed operation in an opinionated way. For users who want an unopinionated interaction with workflows and other Temporal primitives, a Temporal client will be available via a `ContextVar` in the vanilla `start` methods and they can do whatever they want with it, as long as they return their result synchronously (this has a `10s` limit). In contrast, our utility will do the right thing w.r.t. e.g. request-ID-based deduplication, and will be the only way to implement an arbitrary-duration async operation in a Temporal Nexus service.
+
+The proposal here follows the  [2b. Convenience utilities for shorthand operation definition](#2b-convenience-utilities-for-shorthand-operation-definition) `@nexusrpc.handler.sync_operation` decorator above:
+
+```python
+@nexusrpc.handler.service(interface=interface.MyNexusService)
+class MyNexusService:
+    @temporalio.nexus.handler.workflow_operation
```
cretz, line 288: Safe to assume this is just shorthand for something like:

```python
def hello(self) -> WorkflowOperationHandler[HelloInput, HelloOutput]:
    return temporalio.nexus.handler.WorkflowOperationHandler(HelloWorkflow.run)
```

And I could do that myself if I didn't want the decorator (e.g. maybe doing something a bit more dynamic)?

dandavison:
This is being reviewed internally currently.

PR #98, opened 2025-04-25
CLI Extensions

cretz:
# Summary

This is a proposal for supporting extensions in the [CLI](https://github.com/temporalio/cli).

[**RENDERED**](https://github.com/cretz/temporal-proposals/blob/cli-extensibility/cli/cli-extensions.md)

Please leave comments/feedback. If the comment list grows unwieldy and/or we've resolved most but still have more to say, this PR can be closed and another will be opened. All alternative options/suggestions welcome. Requests for clarifications or justification on anything welcome.

Note, some things are missing by intention because they require more research, but we didn't want to hold up the ability to review.

As part of this, all legacy CLI proposals for `tctl` were moved to a `previous` folder since they aren't weren't for what is currently termed the "CLI".


cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+### Shortcuts for Common Tasks
+
+A custom extension can add something like
+
+    temporal mycompany do-company-thing
+
+And that can do anything programmatically and also get the benefit of having access to the Temporal client. So it could
+access some company resource (e.g. a database) and send an update to a workflow with some value.
+
+### Contributions
```
yuandrew, line 31: Will all contributions/extensions be in their own repos and have an extension discovery platform, or will we have some extensions live in the CLI repo (either temporal authored or external that meet a certain bar/requirement)?

cretz: All extensions will be external, even Temporal authored ones (unless we decide to make them built-in commands). We have things like https://temporal.io/code-exchange to help build a marketplace of these as needed. I need to change the title of this section to "Extension Ecosystem" since they are not contributions in the traditional sense.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+
+* Structs and Cobra flag sets for root-level flags and Temporal client flags
+* Ability to create `*slog.Logger` from root-level flags
+* Ability to dial a Temporal client from Temporal client flags
+* Utility to create payloads from raw data
+
+Other items can be added as needed. Extensions can/should also use the Go SDK as needed.
+
+There is not a plan to expose a printer at this time, so extensions will have to handle the possible `output` enumerates
+of `text`, `json`, `jsonl`, and `none` themselves. This is because the current CLI printer has too many quirks and is
```
Sushisource, line 148: My feeling is it's very likely they just _won't_

That's not really the end of the world or anything but I just want to acknowledge that's the likely outcome if we don't start with a helper.

Alternatively we could require them to always do `jsonl`, and just transform that into the desired format for them. I don't have a strong feeling on it, but it's an option.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+
+* `temporal-foo-bar_baz-qux`
+* `temporal-foo-bar_baz`
+* `temporal-foo`
+
+And the first one found will be executed with all arguments. And since underscores and dashes are the same,
+`temporal-foo-bar_baz` executable is called for both `temporal foo bar-baz qux` and `temporal foo bar_baz qux` though
+help text only shows the former.
+
+Built-in commands cannot be overridden by extensions, but subcommands can be added to existing commands.
```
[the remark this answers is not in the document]
cretz: I'm not too concerned about this as this is something a user installs. Being concerned that a user put `temporal-workflow-strt` `PATH` would be like being concerned that a user put `temporl` on their `PATH` and `temporl workflow start` could be hijacked. A user putting a binary on their path slightly misnamed from what they intend I think has the same concerns for top-level as sub-command level. I doubt similar concerns exist for `git-commt` executables, so not sure we need to take extra precautions here.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+
+Running `temporal help <command sequence>` is treated as `temporal <command sequence> --help`. So the same style of
+lookup is performed for extensions.
+
+### Invocation and Flags
+
+Temporal CLI supports arbitrarily ordered flags. So `temporal --address foo workflow start`,
+`temporal workflow --address foo start`, and `temporal workflow start --address foo` are all the same. It is still a
+subject of active research on whether flags of _parent_ commands can still be placed before the extension command in the
+CLI. Regardless, flags of the extension command _must_ come after the extension command because CLI does not know
```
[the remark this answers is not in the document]
cretz: The reason I mentioned the parent-flag-before-command (I should have called it built-in-flag-before-command) was more about how hard our library is going to make performing partial flag parsing on not-found commands. But I think it will work.

cli/cli-extensions.md
cretz: > That's not really the end of the world or anything but I just want to acknowledge that's the likely outcome if we don't start with a helper.

We don't have to block the whole project by starting with a helper if we can followup shortly thereafter. I don't want to hold up this project on designing a quality printer abstraction. So maybe I can change this to "we may expose a printer shortly after initial implementation"?

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+handle.
+
+Invocation of the extension is done as a subprocess. Stdin, stdout, stderr, exit code, etc are all handled by the
+extension and just relayed through the `temporal` process as is. It is an area of active research on whether interrupt
+signals can be ignored by the parent `temporal` process to be handled by the subprocess, though that is the goal.
+
+### Helper Library
+
+Built-in commands leverage several helpers to be consistent with the CLI ecosystem. Extension commands should be able to
+do the same, within reason, so a helper library will be made available. Originally it was thought that such a helper
```
dandavison, line 115: Can you outline how users who don't want to use Go will proceed? I expect Python will be a common language to want to implement extensions in.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+There is not a plan to expose a printer at this time, so extensions will have to handle the possible `output` enumerates
+of `text`, `json`, `jsonl`, and `none` themselves. This is because the current CLI printer has too many quirks and is
+not high quality enough for exposure. It is possible in the future a good printer abstraction can be exposed.
+
+It is an area of active research whether built-in commands will leverage this package or whether both will leverage
+common code independently.
+
+It is an area of active research whether the CLI's YAML-based code generation functionality will be made available to
+extensions.
+
```
dandavison, line 156: How will errors from extensions be presented? (Will there be any effort to make this have a consistent appearance, e.g. will the real CLI intercept and re-present stderr and/or exit codes from user executables?)

cli/cli-extensions.md
Sushisource: Yeah, I didn't say to hold it up. Just wanted to bring it up.

cli/cli-extensions.md
cretz: Like kubectl (https://kubernetes.io/docs/tasks/extend-kubectl/kubectl-plugins/#using-the-command-line-runtime-package), they will not get a helper library at this time. I can clarify here if needed that there are no plans for any helper libraries for CLI extensions beyond the language the CLI is written in. I actually don't expect Python will be a common language to implement extensions in (but Python may be a common language to write general CLIs, just not as extensions to the Temporal CLI in Go I don't think anymore than it is for `kubectl`).

cli/cli-extensions.md
cretz: There is no effort to make this any of the output consistent at this time. Extensions are free to do with stdin, stdout, stderr, and exit codes whatever they want.

cli/cli-extensions.md
dandavison: Python's a bit too slow to be a good language to write general CLIs in, but in this context where you know you're making a network call that disadvantage doesn't apply. It's fine if Go is the only offering initially but I don't think plugin authors should have to care which language the CLI is written in. And since plugins are often written by people to "scratch an itch" (often for their own development environments), I think it will be common to choose a more convenient language than Go.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+_does_ list all extensions by traversing the `PATH` and getting every `temporal-`-prefixed executable. There is no short
+description for any of these extensions, they are simply shown as available external commands. All extensions on the
+`PATH` are shown as they are, even if they are multiple commands deep. For example `temporal-foo-bar` and
+`temporal-workflow-dosomething-somethingelse` are shown as `foo bar` and `workflow dosomething somethingelse` in the
+list respectively, not as just `foo` or implied as part of the `workflow` command.
+
+Extensions on built-in commands are not shown as part of the parent command's help.
+
+Running `temporal help <command sequence>` is treated as `temporal <command sequence> --help`. So the same style of
+lookup is performed for extensions.
```
dandavison, line 88: Having a `temporal plugin` subcommand could be worth considering for discoverability (e.g. `temporal plugin list`)

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+These are just usage examples for the purposes of explaining extensions, this may not be what `temporal cloud` extension
+looks like when built.
+
+## Runtime Behavior
+
+### Lookup
+
+Extensions are any executables on the `PATH` with `temporal-` prefix. Extension executable names use `-` for each
+subcommand and `_` for each dash in the command (but still can be called with an underscore). This is similar to
+`kubectl` behavior.
```
dandavison, line 61: While there is precedent with kubectl and git, I wonder whether we should choose a slightly less common prefix such as `temporalcli-` or `temporal-cli-` or `-temporal-`. I've surveyed one temporal user, and they already have the following executables named `temporal-` (found by issuing `temporal-<TAB>`):

```
cli(main) temporal-
temporal-cancel-all
temporal-delete-all
temporal-generate-help-output
temporal-github-actions-delete-runs
temporal-otel-collector
temporal-python-release.sh
temporal-python-test-failures
temporal-samples-traffic
temporal-server
temporal-server-test-failures
temporal-slack-search
temporal-terminate-all
temporal-typescript-link
temporal-workflow-from-url
temporal-workflow-list-ids
temporal-workflow-start
```

cli/cli-extensions.md
cretz: You are more than welcome to write it in any language. If you care about convenience, you may choose bash. But if you want a helper library, it's only reasonable for us to maintain it in Go at this time (and for our use, the `temporal-cloud` CLI, we'll of course only use Go). Maybe we can revisit if the desire is large enough for Temporal CLI extensions in Python, though I suspect like similar Go CLIs there won't be much desire.

cli/cli-extensions.md
cretz: I don't think we should call these "plugins", but regardless, I found the `git` approach a bit simpler than a full, separate command for listing.

cli/cli-extensions.md
cretz: I expect the dashes to be like spaces on a command line. Looking at that list, I'd expect those to be CLI extensions same as I would for other extensible CLIs. All of those all now magically work with spaces instead of dashes with the exception of `temporal-server` and `temporal-workflow-start`. I don't think we should alter the simplicity of extension names based on this user's list.

cli/cli-extensions.md
cretz: :+1: Is anything holding up the proposal from your POV?

cli/cli-extensions.md
dandavison: The advantage of a subcommand is that when a user does `temporal --help`, they'll see a subcommand with a name like "extension", alerting them to the fact that extensions are a possibilty. Not blocking, but worth considering for discoverability.

cli/cli-extensions.md
Quinn-With-Two-Ns: Personally I am fine only providing helpers for Go at this time.

Quinn-With-Two-Ns (approved):

cli/cli-extensions.md
cretz: We can add a "To see all extensions, pass --all" to the end of `--help` if it's important that default help shows that extensions are thing, but I am hoping to avoid a new sub command at least not until there is more to do that just list commands.

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+* `temporal-foo-bar_baz-qux`
+* `temporal-foo-bar_baz`
+* `temporal-foo`
+
+And the first one found will be executed with all arguments. And since underscores and dashes are the same,
+`temporal-foo-bar_baz` executable is called for both `temporal foo bar-baz qux` and `temporal foo bar_baz qux` though
+help text only shows the former.
+
+Built-in commands cannot be overridden by extensions, but subcommands can be added to existing commands.
+
```
dandavison, line 75: It might make sense to reserve some names as unavailable for extensions, for example `temporal nexus`, `temporal config` etc?

Beyond that, are you going to impose any restrictions on valid names? (File systems are very permissive) We obviously want to support user's local language; I think that typically means allowing non-utf8 even (?).

cli/cli-extensions.md
```diff
@@ -0,0 +1,195 @@
...
+example, `temporal-foo` _should_ handle root-level flags like `--output`. However, this isn't a _must_ requirement. See
+the helper library section for some helpers.
+
+Currently, the only root-level flag that the parent `temporal` process respects when calling an extension is
+`--command-timeout` (even though it is still passed along). All other root-level flags are up to the extension to
+handle.
+
+Invocation of the extension is done as a subprocess. Stdin, stdout, stderr, exit code, etc are all handled by the
+extension and just relayed through the `temporal` process as is. It is an area of active research on whether interrupt
+signals can be ignored by the parent `temporal` process to be handled by the subprocess, though that is the goal.
```
dandavison, line 110: exec is an option on non-Windows platforms -- that way signals, exit code etc all work exactly as expected without any work. But of course subprocess has to be supported for Windows, so there may not be a justification for the two paths.

cli/cli-extensions.md
cretz: > It might make sense to reserve some names as unavailable for extensions

Yes we can reserve some, though we won't predict them all. I will make a blanket statement like "Built-in commands always take precedent so if a newer Temporal CLI has a built-in command that once was an extension, the command will be chosen and the extension name will have to be changed". This is a normal state of affairs for CLI extensions, and in many ways it can be a good thing because some extensions could be contributions upstreamed to become built-in.

cli/cli-extensions.md
cretz: Yes, we may take this approach as this is what `kubectl` does: https://github.com/kubernetes/kubectl/blob/285ed6ce48d8ba81d430845306c70533170cab58/pkg/cmd/cmd.go#L239-L259. The problem is that I don't think we can do anything as a wrapping process anymore, like enforce `command-timeout`, if we use this form of process replacement. But it still may be worth it.

Sushisource (approved):
Makes sense, nothing blocking.

I think making some reference to the possible security concerns in the proposal might be worthwhile, even if we aren't going to do anything about it at first. Git may not do anything, but, some tools do and I see why.

cretz:
(merging this, company is moving forward on it)

PR #104, opened 2026-01-12
Kotlin SDK API Proposal

mfateev:
#### WARNING

The Kotlin SDK is an experiment in using LLMs for both design and implementation. 

## Summary

This proposal defines the public API for the Temporal Kotlin SDK, providing an idiomatic Kotlin experience using coroutines and suspend functions.

**Formatted reading:** https://github.com/mfateev/proposals/tree/kotlin-api-review/kotlin

Start at the [README](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/README.md) and click "Next" at the bottom of each document to navigate through all 27 files in order.

### Key Features

- Coroutine-based workflows with `suspend fun`
- Type-safe activity and workflow execution via method references (no stubs)
- Native Kotlin Duration support (`30.seconds`)
- KOptions data classes for configuration
- Full interoperability with Java SDK

### Documentation Structure

| Section | Description |
|---------|-------------|
| [Kotlin Idioms](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/kotlin-idioms.md) | Duration, null safety, coroutine patterns |
| [Configuration](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/configuration/README.md) | KOptions, data conversion, interceptors |
| [Workflows](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/workflows/README.md) | Definition, signals, queries, updates, child workflows |
| [Activities](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/activities/README.md) | Regular and local activities |
| [Client](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/client/README.md) | KWorkflowClient, handles, advanced operations |
| [Worker](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/worker/README.md) | KWorkerFactory, KWorker setup |
| [Testing](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/testing.md) | KTestWorkflowEnvironment, time skipping |
| [Migration](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/migration.md) | Java SDK migration guide |
| [API Parity](https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/api-parity.md) | Java SDK comparison |

### DIFF
--- a/kotlin/worker/README.md
+++ b/kotlin/worker/README.md
@@ -17,28 +17,29 @@ The Kotlin SDK provides `KWorkerFactory` and `KWorker` which automatically enabl
 ### Basic Worker Setup
 
 ```kotlin
-val service = WorkflowServiceStubs.newLocalServiceStubs()
-val client = KWorkflowClient(service)
+val client = KClient.connect(KClientOptions(target = "localhost:7233"))
 
-// KWorkerFactory automatically enables Kotlin coroutine support
-val factory = KWorkerFactory(client)
-val worker = factory.newWorker("task-queue")
+// Create worker with workflows and activities at construction time
+val worker = KWorker(
+    client,
+    KWorkerOptions(
+        taskQueue = "task-queue",
+        workflows = listOf(MyWorkflowImpl::class),
+        activities = listOf(MyActivitiesImpl())
+    )
+)
 
-// Register workflows and activities
-worker.registerWorkflowImplementationTypes(MyWorkflowImpl::class)
-worker.registerActivitiesImplementations(MyActivitiesImpl())
-
-// Start the worker
-factory.start()
+// Run the worker (blocks until shutdown or fatal error)
+worker.run()
 ```
 
 ### Key Components
 
 | Component | Purpose |
 |-----------|---------|
-| `KWorkerFactory` | Creates workers with automatic coroutine support |
-| `KWorker` | Registers Kotlin workflows and suspend activities |
-| `KotlinPlugin` | Plugin for Java main apps (used automatically by KWorkerFactory) |
+| `KWorker` | Creates and runs workers with automatic coroutine support |
+| `KWorkerOptions` | Configuration including workflows and activities |
+| `KotlinJavaWorkerPlugin` | Plugin for Java main apps using Kotlin workflows |
 
 ## Related
 
--- a/kotlin/worker/setup.md
+++ b/kotlin/worker/setup.md
@@ -24,8 +24,8 @@ val worker = KWorker(
     )
 )
 
-// Start the worker
-worker.start()
+// Run the worker (blocks until shutdown signal or fatal error)
+worker.run()
 ```
 
 ## KWorker API
@@ -45,9 +45,30 @@ class KWorker(
     /** The underlying Java Worker for interop scenarios */
     val worker: Worker
 
+    /**
+     * Run the worker, blocking until a shutdown signal is received or a fatal error occurs.
+     * This is the recommended way to run a worker.
+     *
+     * Unlike start()/shutdown(), this method propagates fatal worker-runtime errors
+     * rather than silently swallowing them.
+     *
+     * @throws WorkerException if a fatal error occurs during worker execution
+     */
+    suspend fun run()
+
+    /**
+     * Start the worker without blocking. Use shutdown() to stop.
+     * Prefer run() for most use cases as it properly propagates fatal errors.
+     */
     fun start()
+
+    /** Gracefully shut down the worker, allowing in-flight tasks to complete. */
     fun shutdown()
+
+    /** Immediately shut down the worker, cancelling in-flight tasks. */
     fun shutdownNow()
+
+    /** Wait for the worker to terminate after shutdown is called. */
     suspend fun awaitTermination(timeout: Duration)
 }
 ```
@@ -73,7 +94,7 @@ data class KWorkerOptions(
 
 ## Multiple Workers
 
-For multiple task queues, create multiple workers:
+For multiple task queues, create multiple workers and run them concurrently:
 
 ```kotlin
 val orderWorker = KWorker(
@@ -94,9 +115,27 @@ val notificationWorker = KWorker(
     )
 )
 
-// Start all workers
+// Run all workers concurrently - blocks until shutdown or fatal error
+coroutineScope {
+    launch { orderWorker.run() }
+    launch { notificationWorker.run() }
+}
+```
+
+For advanced scenarios where you need manual control:
+
+```kotlin
+// Start workers without blocking
 orderWorker.start()
 notificationWorker.start()
+
+// ... do other work ...
+
+// Graceful shutdown
+orderWorker.shutdown()
+notificationWorker.shutdown()
+orderWorker.awaitTermination(30.seconds)
+notificationWorker.awaitTermination(30.seconds)
 ```
 
 ## KotlinJavaWorkerPlugin (For Java Main)
@@ -139,7 +178,7 @@ val worker = KWorker(
 )
 
 // Both run on the same worker - execution model is per-workflow-instance
-worker.start()
+worker.run()  // Blocks until shutdown or fatal error
 ```
 
 ## Related


kotlin/activities/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+```kotlin
+// Type-safe method reference
+val greeting = KWorkflow.executeActivity(
+    GreetingActivities::composeGreeting,
+    KActivityOptions(startToCloseTimeout = 30.seconds),
+    "Hello", "World"
+)
+
+// String-based (for cross-language interop)
```
cretz: Does cross-language interop in your mind include Java? Or would you expect an activity defined in Java to use the existing stub/proxy based approach even inside a Kotlin workflow?

kotlin/activities/definition.md
```diff
@@ -0,0 +1,171 @@
...
+    ): R
+
+    // 2 arguments
+    suspend fun <T, A1, A2, R> executeActivity(
+        activity: KFunction3<T, A1, A2, R>,
+        options: KActivityOptions,
+        arg1: A1, arg2: A2
+    ): R
+
+    // ... up to 6 arguments
```
cretz: In newer SDKs that don't have advanced generics for arbitrary arity, we have often just supported overloads with 0 or 1 parameter since we recommend no more than 1, but _technically_ harmless to support multiple here except that we're encouraging multiple.

kotlin/api-parity.md
```diff
@@ -0,0 +1,164 @@
...
+| `Workflow.sleep(Duration)` | Use `kotlinx.coroutines.delay()` - intercepted by dispatcher |
+| `Workflow.newTimer(Duration)` | Use `async { delay(duration) }` for racing timers |
+| `Workflow.wrap(Exception)` | Kotlin has no checked exceptions - not needed |
+| `Activity.wrap(Throwable)` | Kotlin has no checked exceptions - not needed |
+| `Workflow.newCancellationScope(...)` | Use Kotlin's `coroutineScope { }` with structured concurrency |
+| `Workflow.newDetachedCancellationScope(...)` | Use `supervisorScope { }` or launch in parent scope |
+| `Workflow.newWorkflowLock()` | Not needed - cooperative coroutines don't have true concurrency |
+| `Workflow.newWorkflowSemaphore(...)` | Use structured concurrency patterns (e.g., `chunked().map { async }.awaitAll()`) |
+| `Workflow.newQueue(...)` / `newWorkflowQueue(...)` | Use `mutableListOf` + `awaitCondition` - no concurrent access in suspend model |
+| `Workflow.newPromise()` / `newFailedPromise(...)` | Use `CompletableDeferred<T>` from kotlinx.coroutines |
```
cretz: Technically none of these are needed in Java either and can be written with wait conditions, but we have them anyways in Java (but not usually in other languages)

kotlin/client/workflow-client.md
```diff
@@ -0,0 +1,125 @@
...
+    ): R
+
+    // Overloads for 2-6 arguments...
+
+    /**
+     * Get a typed handle for an existing workflow by ID.
+     * Use this to signal, query, or get results from a workflow started elsewhere.
+     */
+    inline fun <reified T> getWorkflowHandle(workflowId: String): KWorkflowHandle<T>
+    inline fun <reified T> getWorkflowHandle(workflowId: String, runId: String): KWorkflowHandle<T>
```
cretz: Arguably one should be able to provide reified `R` too if such an optional approach is allowed, but meh

kotlin/client/workflow-handle.md
```diff
@@ -0,0 +1,292 @@
...
+
+// Workflow metadata
+val description = handle.describe()
+println("Workflow ID: ${handle.workflowId}, Run ID: ${handle.runId}")
+println("Status: ${description.status}")
+```
+
+## Architectural Note: Classes vs Interfaces
+
+All handle types (`KWorkflowHandle`, `KTypedWorkflowHandle`, `KUpdateHandle`, `WorkflowHandle`, `KChildWorkflowHandle`) are **classes** rather than interfaces. This design choice enables:
```
cretz: Mentioned before, but it is confusing to have:

* `KWorkflowHandle` - half-typed (just not result)
* `KTypedWorkflowHandle` - typed with result
* `WorkflowHandle` - untyped, named as if it'll be used from Java

I recommend:

* `KWorkflowHandle<T> : KWorkflowHandleUntyped` (extends only needed if reasonable)
* `KWorkflowHandleWithResult<T, R> : KWorkflowHandle<T>` (extends only needed if reasonable)
* `KWorkflowHandleUntyped`

(this is similar to what we do in .NET, but since JVM doesn't Java reification even though Kotlin fakes it in user land, they have to be named different)

kotlin/client/workflow-handle.md
```diff
@@ -0,0 +1,292 @@
...
+    suspend fun <R> executeUpdate(method: KSuspendFunction1<T, R>): R
+    suspend fun <R, A1> executeUpdate(method: KSuspendFunction2<T, A1, R>, arg: A1): R
+    suspend fun <R, A1, A2> executeUpdate(method: KSuspendFunction3<T, A1, A2, R>, arg1: A1, arg2: A2): R
+    // ... up to 6 arguments
+
+    // Updates - start and return handle without waiting for completion
+    suspend fun <R> startUpdate(
+        method: KSuspendFunction1<T, R>,
+        waitForStage: WorkflowUpdateStage = WorkflowUpdateStage.ACCEPTED,
+        updateId: String? = null
```
cretz: These should be in an options class if we want to be consistent (we can sugar out to these if needed)

kotlin/kotlin-idioms.md
```diff
@@ -0,0 +1,197 @@
...
+    suspend fun validateOrder(order: Order): Boolean
+
+    @ActivityMethod
+    suspend fun chargePayment(order: Order): PaymentResult
+}
+```
+
+> **Note:** Query methods are never `suspend` because queries must return immediately without blocking.
+
+## Coroutines and Concurrency
```
cretz: Do we have full control over the event loop including running the `awaitCondition` stuff when/how we want in that loop, or is this somehow just sugar on top of the Java one (sorry, could dig into the PoC, but being lazy)

kotlin/client/workflow-client.md
```diff
@@ -0,0 +1,125 @@
...
+## Creating a Client
+
+```kotlin
+val service = WorkflowServiceStubs.newLocalServiceStubs()
+
+// Create KWorkflowClient with DSL configuration
+val client = KWorkflowClient(service) {
+    setNamespace("default")
+    setDataConverter(myConverter)
+}
```
cretz: In newer SDKs, we found _requiring_ these separate steps unnecessary. Would recommend a static call on `KWorkflowClient` called `connect` that accepts an option set that is the union of service stub options and client options.

kotlin/worker/README.md
```diff
@@ -0,0 +1,51 @@
...
+// KWorkerFactory automatically enables Kotlin coroutine support
+val factory = KWorkerFactory(client)
+val worker = factory.newWorker("task-queue")
+
+// Register workflows and activities
+worker.registerWorkflowImplementationTypes(MyWorkflowImpl::class)
+worker.registerActivitiesImplementations(MyActivitiesImpl())
+
+// Start the worker
+factory.start()
```
cretz: We should discourage the `start` pattern IMO in favor of a `run` one. We have learned that `start`+`shutdown` can swallow fatal worker-runtime errors (which are of course very rare, but technically can occur).

GreyTeardrop (commented):
So excited for Kotlin-first Temporal SDK!

kotlin/kotlin-idioms.md
mfateev: https://github.com/temporalio/sdk-java/pull/2751

kotlin/worker/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+### Basic Worker Setup
+
+```kotlin
+val service = WorkflowServiceStubs.newLocalServiceStubs()
+val client = KWorkflowClient(service)
+
+// KWorkerFactory automatically enables Kotlin coroutine support
+val factory = KWorkerFactory(client)
+val worker = factory.newWorker("task-queue")
```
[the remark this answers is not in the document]
mfateev: Java is the only SDK that doesn't rely on global variables to share cache and thread pool. It uses factory for this. If we decide that it is not a good pattern we can go with a simpler user experience

mfateev:
Added https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/open-questions.md

kotlin/activities/definition.md
```diff
@@ -0,0 +1,171 @@
...
+The typed activity API uses direct method references - no stub creation needed. This approach:
+- Provides full compile-time type safety for arguments and return types
+- Allows different options (timeouts, retry policies) per activity call
+- Works with both Kotlin `suspend` and Java non-suspend activity interfaces
+- Similar to TypeScript and Python SDK patterns
+
+```kotlin
+// Define activity interface
+@ActivityInterface
+interface GreetingActivities {
```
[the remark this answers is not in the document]
mfateev: Agree, I've added this option to the proposal.

https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/open-questions.md

kotlin/activities/definition.md
```diff
@@ -0,0 +1,171 @@
...
+## Activity Execution API
+
+The `KWorkflow` object provides type-safe overloads using `KFunction` types:
+
+```kotlin
+object KWorkflow {
+    // 1 argument
+    suspend fun <T, A1, R> executeActivity(
+        activity: KFunction2<T, A1, R>,
+        options: KActivityOptions,
```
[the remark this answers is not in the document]
mfateev: Listed options at https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/open-questions.md

kotlin/activities/implementation.md
```diff
@@ -0,0 +1,199 @@
...
+Activity interfaces can have both suspend and non-suspend methods. The worker handles both automatically.
+
+## Defining Activities
+
+```kotlin
+@ActivityInterface
+interface OrderActivities {
+    // Suspend method - uses coroutines
+    @ActivityMethod
+    suspend fun chargePayment(order: Order): PaymentResult
```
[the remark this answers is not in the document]
mfateev: Added to the spec that the default parameters are not allowed.

kotlin/activities/implementation.md
```diff
@@ -0,0 +1,199 @@
...
+        .build())
+    .execute("chargePayment", PaymentResult.class, order);
+```
+
+This mirrors how Java workflows call Kotlin workflows - using string names for cross-language interop.
+
+## Registering Activities
+
+```kotlin
+worker.registerActivitiesImplementations(OrderActivitiesImpl(paymentService))
```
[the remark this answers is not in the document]
mfateev: Sorry, I don't understand the question.

kotlin/activities/implementation.md
```diff
@@ -0,0 +1,199 @@
...
+interface KActivityContext {
+    val info: KActivityInfo
+    fun heartbeat(details: Any? = null)  // No-op for local activities
+    fun <T> heartbeatDetails(detailsClass: Class<T>): T?  // Empty for local activities
+    val taskToken: ByteArray  // Throws for local activities
+    fun doNotCompleteOnReturn()  // Throws for local activities
+    val isDoNotCompleteOnReturn: Boolean
+    fun logger(): Logger
+    fun logger(name: String): Logger
+    fun logger(clazz: Class<*>): Logger
```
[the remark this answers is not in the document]
mfateev: Good catch. Java only has Workflow.getLogger. So, no need to have this at the activity context.

kotlin/activities/implementation.md
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Reified extension for easier Kotlin usage
+inline fun <reified T> KActivityContext.heartbeatDetails(): T?
```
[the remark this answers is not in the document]
mfateev: Java SDK supports only a single value for heartbeat details.

kotlin/activities/implementation.md
```diff
@@ -0,0 +1,199 @@
...
+
+    // Timing information
+    val scheduledTime: Instant?
+    val startedTime: Instant?
+    val scheduleToCloseTimeout: Duration?
+    val startToCloseTimeout: Duration?
+    val heartbeatTimeout: Duration?
+
+    // Heartbeat state
+    val heartbeatDetails: Any?
```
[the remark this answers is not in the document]
mfateev: Removed

kotlin/api-parity.md
```diff
@@ -0,0 +1,164 @@
...
+| `Workflow.getCurrentDetails()` | `KWorkflow.currentDetails` |
+| `Workflow.getMetricsScope()` | `KWorkflow.metricsScope` |
+
+### Side Effects & Utilities
+
+| Java SDK API | Kotlin SDK |
+|--------------|------------|
+| `Workflow.sideEffect(...)` | `KWorkflow.sideEffect()` |
+| `Workflow.mutableSideEffect(...)` | `KWorkflow.mutableSideEffect()` |
+| `Workflow.getVersion(...)` | `KWorkflow.version()` |
```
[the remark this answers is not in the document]
mfateev: **PENDING.** Need clarification: Is a full patching API (Option D) preferred? The current design has `KWorkflow.version()` as a simple side-effecting call. Should we instead provide a more structured patching API similar to newer SDKs?

kotlin/client/workflow-client.md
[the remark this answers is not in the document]
mfateev: **Accepted as proposed.** Renamed to `workflowHandle()` - more Kotlin idiomatic, no get/new prefix.

mfateev: **Accepted as proposed.** Added overload to specify result type:
```kotlin
// One type param - must specify result type when calling result()
val handle = client.workflowHandle<OrderWorkflow>(workflowId)
val result = handle.result<OrderResult>()

// Two type params - result type known
val handle = client.workflowHandle<OrderWorkflow, OrderResult>(workflowId)
val result = handle.result()  // Type already known
```

kotlin/client/workflow-client.md
```diff
@@ -0,0 +1,125 @@
...
+`KWorkflowClient` provides Kotlin-specific APIs with suspend functions for starting and interacting with workflows:
+
+```kotlin
+/**
+ * Kotlin workflow client providing suspend functions and type-safe workflow APIs.
+ *
+ * @param service The WorkflowServiceStubs to connect to
+ * @param options DSL builder for WorkflowClientOptions
+ */
+class KWorkflowClient(
```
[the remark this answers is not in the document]
mfateev: **Accepted as proposed.** The documents now consistently use the signature pattern where options come before args: `executeWorkflow(method, options, args)`. This is consistent with Java SDK conventions and more natural for Kotlin's named parameter usage.

kotlin/client/workflow-client.md
mfateev: **Accepted as proposed.** Updated to use `KWorkflowClient.connect(options)` pattern following .NET SDK conventions. This is cleaner and more explicit than the Java SDK's multi-step construction pattern.

kotlin/client/workflow-handle.md
mfateev: **Accepted as proposed.** Updated type hierarchy to:
- `KWorkflowHandleUntyped` - base untyped handle
- `KWorkflowHandle<T>` - typed handle (knows workflow interface)
- `KWorkflowHandleWithResult<T, R>` - typed handle with known result type

This matches the type hierarchy pattern from other SDKs while providing Kotlin-idiomatic naming.

kotlin/client/workflow-handle.md
mfateev: **Accepted as proposed.** Added `KUpdateOptions` and `KStartUpdateOptions` data classes to the workflow-handle.md documentation:

```kotlin
data class KUpdateOptions(
    val resultTimeout: Duration? = null,
    // Experimental
    @Experimental val links: List<Link>? = null
)

data class KStartUpdateOptions(
    val waitForStage: UpdateWaitPolicy,  // Required
    val resultTimeout: Duration? = null,
    @Experimental val links: List<Link>? = null
)
```

kotlin/worker/README.md
mfateev: **Accepted with modification.** Updated the documentation to recommend `factory.run()` (which blocks until shutdown) as the primary pattern, rather than discouraging `start()`. The documentation now shows:

```kotlin
// Recommended: blocks until worker shutdown
factory.run()

// Alternative for programmatic shutdown
factory.start()
// ... later ...
factory.shutdown()
```

This follows the pattern from other SDKs where `run()` is the simpler/preferred approach for most use cases.

kotlin/configuration/koptions.md
```diff
@@ -0,0 +1,296 @@
...
+    val scheduleToStartTimeout: Duration? = null,
+    val heartbeatTimeout: Duration? = null,
+    val taskQueue: String? = null,
+    val retryOptions: KRetryOptions? = null,
+    val cancellationType: ActivityCancellationType? = null,  // Java default: TRY_CANCEL
+    val disableEagerExecution: Boolean = false,
+    // Experimental
+    @Experimental val summary: String? = null,
+    @Experimental val priority: Priority? = null
+) {
```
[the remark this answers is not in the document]
mfateev: Added to open questions at https://github.com/mfateev/proposals/blob/kotlin-api-review/kotlin/open-questions.md#data-classes-vs-builderdsl-for-optionsconfig-classes

kotlin/open-questions.md
```diff
@@ -0,0 +1,611 @@
...
+@ActivityMethod
+suspend fun processOrder(params: ProcessOrderParams)
+```
+
+### Rationale
+
+1. **Replay Safety** - If default values change between deployments, replayed workflows behave differently, violating determinism
+2. **Serialization Ambiguity** - Unclear whether the caller serializes defaults or the worker applies them
+3. **Cross-Language Compatibility** - Other languages calling the activity/workflow don't know about Kotlin defaults
+4. **SDK Consistency** - Python SDK explicitly disallows default parameters; Go/Java don't have them
```
cretz: While I agree with the decision to disallow because it can be easily walked back if we change our minds, it may be worth noting that IIRC, Python, .NET, and Ruby all support default parameters. Here is a Python test confirming it: https://github.com/temporalio/sdk-python/blob/993de6d0e9b42bb01f24edfdb46e0795b00debcf/tests/worker/test_workflow.py#L3502-L3545.

kotlin/open-questions.md
```diff
@@ -0,0 +1,611 @@
...
+- Simple API, no wrapper classes
+- Familiar pattern
+
+**Cons:**
+- No compile-time type safety
+- Wrong argument types/count only caught at runtime
+
+---
+
+### Option B: Direct Overloads (0-7 Arguments)
```
cretz: Or "0-1 Argument(s)", meaning any > 1 argument, which Temporal discourages, can fall back to untyped. This is what we do in some newer SDKs.

kotlin/open-questions.md
mfateev: Good point about other SDKs supporting defaults. Given we're moving to 0-1 parameters (with >1 falling back to untyped), supporting a default for a single parameter is straightforward and aligns with Python/.NET/Ruby. Updated the docs to allow defaults for the 0-1 argument case.

kotlin/open-questions.md
```diff
@@ -0,0 +1,611 @@
...
+
+- Reduces boilerplate (no separate interface file)
+- More similar to Python SDK experience
+- Kotlin-only feature, no Java SDK changes required
+
+### Trade-offs
+
+- Different from Java SDK convention
+- Activity/workflow type names derived from method/class names (convention-based)
+- Respects `@ActivityMethod(name = "...")` and `@WorkflowMethod(name = "...")` annotations if present
```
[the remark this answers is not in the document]
mfateev: While this approach makes sense for newer SDKs starting fresh, requiring duplicate annotations would break backward compatibility with existing Java activities/workflows and feel unnatural to Java SDK users. Since Kotlin SDK builds on Java SDK, we'll maintain the current convention where annotations on the interface are sufficient.

kotlin/activities/implementation.md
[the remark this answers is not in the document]
mfateev: Agreed on heartbeat infallible. For cancellation:
- Suspend activities: Standard Kotlin coroutine cancellation
- Non-suspend activities: `KActivity.cancellationFuture()` returns `CompletableFuture<CancellationDetails>` which supports polling (`isDone`), blocking (`get`), or callback notification (`thenAccept`) - similar flexibility to Python's approach.

kotlin/api-parity.md
```diff
@@ -0,0 +1,164 @@
...
+
+This document describes the intentional differences between Java and Kotlin SDK APIs, and identifies remaining gaps.
+
+## APIs Not Needed in Kotlin
+
+The following Java SDK APIs are **not needed** in the Kotlin SDK due to language differences:
+
+| Java SDK API | Reason Not Needed |
+|--------------|-------------------|
+| `Workflow.sleep(Duration)` | Use `kotlinx.coroutines.delay()` - intercepted by dispatcher |
```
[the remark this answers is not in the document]
mfateev: Agreed. Added `KWorkflow.delay(duration)` as a simple alternative to stdlib `delay()`, plus `KWorkflow.delay(duration, summary)` for advanced users who need timer summaries.

kotlin/client/workflow-client.md
```diff
@@ -0,0 +1,149 @@
+# Workflow Client
```
cretz: Forgot to mention this, but in newer SDKs, we found just having one big client that is a workflow client + schedule client (w/ features to make an async activity completion handle) is a bit cleaner from an options POV. This will also help when standalone activity client and Nexus operation client come about. No need to change though if we don't want, there is also value in matching what Java does (though you will duplicate a lot of these client options each time).

kotlin/open-questions.md
cretz: API simplicity (and overload count) was my main driver in preferring just falling back to untyped, but in some SDKs (namely Python), we actually do support arbitrary multi-param typed as a different overload. Would be fine if we did that here too. And really, with how .NET is lambda expressions, it supports multi-param typed as well. So overall, yeah, ok w/ a typed multi-param, and having 0 or 1 be even simpler forms of that.

mfateev: Thanks for confirming. We'll go with the KArgs approach (Option C) - provides full type safety for all arities while keeping 0-1 argument cases simple. The `kargs()` wrapper for 2+ args is a small price for compile-time type checking.

kotlin/client/workflow-client.md
mfateev: Agreed. We'll use a unified `KClient` matching the Python/.NET pattern - single client covering workflows, schedules, async activity completion, and ready for future Nexus support. Cleaner options and less duplication.

kotlin/activities/implementation.md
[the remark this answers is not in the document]
mfateev: Agreed. Removed `KActivity.logger()` and `KWorkflow.logger()` from the API. The SDK will automatically populate MDC with activity/workflow context (workflowId, runId, activityType, taskQueue, etc.), and users can use standard SLF4J or their preferred logging framework.

This matches the Java SDK approach and avoids introducing a custom logging API.

kotlin/kotlin-idioms.md
mfateev: Kotlin SDK is implemented on top of Java SDK - it doesn't have its own event loop. The `awaitCondition` functionality will be built on top of the new `Async.await()` method being added to Java SDK (PR #2751), which returns a `Promise` for non-blocking condition waiting.

```kotlin
// Kotlin awaitCondition suspends the coroutine
KWorkflow.awaitCondition { condition }

// Implemented using Java's Async.await() under the hood
```

cretz (approved):
From my POV, this proposal is definitely good enough from a proposal POV (and is much more complete than past SDK lang proposals). So can merge whenever IMO, but may want more feedback.

Note, in all SDKs, the implementation deviates from the proposal as we learn things, so we've never expected/required 100% proposal accuracy. Also, not sure if it is part of the proposal to determine where this will live (I prefer https://github.com/temporalio/sdk-java/tree/master/temporal-kotlin myself where it can be versioned/published alongside the Java SDK dependency and even use `internal` packages from it if we wanted).

PR #107, opened 2026-04-13
Add ToolRegistry + AgenticSession proposal

lex00:
Design document for `ToolRegistry` and `AgenticSession` — cross-SDK LLM tool-calling primitives shipping as contrib modules in all six Temporal SDKs.

**ToolRegistry** maps tool names to JSON Schema definitions and dispatches LLM tool calls. **AgenticSession** wraps a tool loop with crash-safe heartbeating so activity retries resume mid-conversation rather than restart.

PRs are open:
- temporalio/sdk-go#2292
- temporalio/sdk-python#1435
- temporalio/sdk-typescript#2008
- temporalio/sdk-java#2839
- temporalio/sdk-ruby#417
- temporalio/sdk-dotnet#641


donald-pinckney:
Hi @lex00, thanks for submitting this proposal and concept around a tool registry and Temporal-provided tool loop! We have some planning to do on our side with regards to agentic loop support, and we'd be happy to get back to you in the near future (next week-ish) about this!

One question I would have at the moment: out of curiosity, why did go for the activity + heartbeating route, as opposed to running the agentic loop in the workflow (and LLM calls and some tools in small activities)?

lex00:
The existing `openai_agents`, `google_adk_agents`, `langgraph`, and `ai-sdk` integrations already run each LLM call as a separate Temporal activity; that pattern is well-served. ToolRegistry targets the complement: Go, Java, Ruby, .NET, and direct Anthropic/OpenAI usage without a framework.

**Zero workflow code required**

The primary driver was adoption. A user with an existing workflow calling a `run_code_review` activity should be able to make it crash-safe and mid-conversation resumable without learning workflow determinism constraints or restructuring around a multi-activity loop. One activity, drop in, done.

**Session continuity for long-running use cases**

The framework plugins rely on provider-side session IDs. For human-in-the-loop scenarios where a tool handler signals a workflow and waits hours for approval, those sessions expire. Storing conversation state in the heartbeat rather than a provider session means the conversation survives both crashes and provider-side expiry. A rejection comes back to the model as a tool result, so it can revise its next proposal rather than restart from scratch.

**Heartbeat timing**

The checkpoint is written before each LLM turn, not after. Conversation history is the source of truth; re-issuing a turn on retry is safe because the model produces the same (or equivalent) response given the same history.

**The tradeoff**

Individual tool calls are not visible in Temporal's event history. The conversation is a black box from the workflow's perspective. The workflow-native approach gives per-call visibility and per-call retry policies, which are real advantages. For the use cases in scope (direct API usage, fast tools, sessions of moderate length), turn-level checkpointing covers the durability need without that granularity.

toolregistry/README.md
```diff
@@ -0,0 +1,577 @@
...
+await run_tool_loop(
+    provider="anthropic",          # or "openai"
+    system="You are a code reviewer. Call flag_issue for each problem you find.",
+    prompt=prompt,
+    tools=tools,
+)
+return results
+
+# Crash-safe session
+async with agentic_session() as session:
```
donald-pinckney: You've said that this is implemented with hearbeating, but usually heartbeating is part of an activity that is executing. Is `agentic_session()` starting a workflow, and a activity in the workflow, and then running the contents of the context manager inside that activity?

lex00: Hi @donald-pinckney , thanks for taking a look — and apologies for the wall of text last time.

Yes, it runs inside an activity you define; agentic_session() itself doesn't start one. 

I've updated the README snippets to show the activity wrapper and added a couple of diagrams.

lex00:
The proposal's testing-snippets section now uses `ResponseBuilder` for Python and TypeScript and `MockResponse` for Go/Java/Ruby/.NET, matching each SDK's actual code. The split isn't called out anywhere in the proposal text or per-SDK READMEs.

Python's case is forced: the factory returns `dict` (Anthropic wire format), so naming it `ResponseBuilder` is honest. The other 4 (Go/Java/Ruby/.NET) follow each language's convention of co-naming type and factory.

TypeScript is the only SDK where the choice was free: `MockResponse` is a typed interface, so the factory could be called either `MockResponse` (with static methods) or `ResponseBuilder` (separate class). The current code chose `ResponseBuilder`.

Two questions:

1. Was the TypeScript choice intentional, or accidental drift from Go/Java/Ruby/.NET? If intentional, a one-line note in the proposal would help cross-SDK doc readers. If accidental, TypeScript could rename to `MockResponse` for cross-SDK alignment with one PR.
2. Independent of (1): worth a short paragraph in the proposal noting that Python's factory shape differs by language necessity?

lex00:
FYI from a cross-SDK read pass. 5 observations, ranked by severity. Happy to dig into any of them, file separate issues, or skip them all.

**1. Anthropic loop termination doesn't match canonical `while stop_reason == "tool_use"` pattern.** All 6 SDKs use `if not tool_calls or stop_reason == "end_turn": exit` instead of `if stop_reason != "tool_use": exit`. Equivalent in the happy path. Edge case: `stop_reason == "max_tokens"` with a partially-truncated tool_use block. Current code would dispatch the partial JSON args. Aligning with the Anthropic docs' canonical pattern (linked under "The agentic loop") would bail cleanly. One-line change per SDK.

**2. Heartbeat-payload size guard absent in all 6 SDKs.** "Known Limitations" mentions the ~2 MiB / ~30-60 turn limit. None of the SDKs warn or truncate as the payload approaches it. Three options: (a) document the limit more prominently and stop. (b) add a soft warning per SDK at e.g. 1 MiB. (c) hard guard with ApplicationFailure. Open question. Depends on whether compaction is the user's job or the SDKs'.

**3. Python `AnthropicProvider` constructs the client per-call.** `_providers.py` ~line 131-132. Other 5 SDKs build the LLM client once in the provider constructor. Python builds a fresh `anthropic.Anthropic()` (and HTTP pool) on every `run_tool_loop` invocation when no `client=` is passed. Cross-SDK inconsistency, low impact for typical single-loop activities.

**4. sdk-ruby `tool_registry` has no dedicated test suite.** Other 5 SDKs ship substantial test coverage.

**5. .NET `SessionCheckpoint.Version` relies on field-default initializer.** Other 5 SDKs explicitly set `version: 1` at heartbeat-write time. The default works today.

Naming convention question (`MockResponse` vs `ResponseBuilder`) was already raised in a separate comment above.

donald-pinckney:
Hi @lex00, wanted to say thanks for brining this proposal up! There are definitely interesting ideas here, but we aren't wanting to commit to this heartbeating-based direction currently. I'm going to close this PR (and the associated SDK PRs) for now.

If you want to keep iterating on the design, I think it would work fine for you work on this in a standalone repo, since I don't think they require SDK library internals.

PR #111, opened 2026-09-17
Update README.md to redirect folks to new space

karasowles:
Adding a note about archiving the proposals repo in favor of new feature request discussions space

- Origin issue: 

# Summary

Brief explanation of the feature.

# Basic example

If the proposal involves a new or changed API, include a basic code example.
Omit this section if it's not applicable.

# Motivation

Why are we doing this? What use cases does it support? What is the expected
outcome?

Please focus on explaining the motivation so that if this RFC is not accepted,
the motivation could be used to develop alternative solutions. In other words,
enumerate the constraints you are trying to solve without coupling them too
closely to the solution you have in mind.

# Detailed design

This is the bulk of the RFC. Explain the design in enough detail for somebody
familiar with Temporal to understand, and for somebody familiar with the
implementation to implement. This should get into specifics and corner-cases,
and include examples of how the feature is used. Any new terminology should be
defined here.

# Drawbacks

Why should we *not* do this? Please consider:

- implementation cost, both in term of code size and complexity
- whether the proposed feature can be implemented in user space
- the impact on teaching people Temporal
- integration of this feature with other existing and planned features
- cost of migrating existing applications

There are tradeoffs to choosing any path. Attempt to identify them here.

# Alternatives

What other designs have been considered? What is the impact of not doing this?

# Adoption strategy

If we implement this proposal, how will existing Temporal developers adopt it? Is
this a breaking change? Can we write a codemod? Should we coordinate with
other projects or libraries?

# How we teach this

What names and terminology work best for these concepts and why? How is this
idea best presented? As a continuation of existing Temporal patterns?

Would the acceptance of this proposal mean the Temporal documentation must be
re-organized or altered? Does it change how Temporal is taught to new developers
at any level?

How should this feature be taught to existing Temporal developers?

# Unresolved questions

Optional, but suggested for first drafts. What parts of the design are still
TBD?


README.md
```diff
@@ -1,4 +1,10 @@
-# [Temporal](https://temporal.io) proposals
+> [!IMPORTANT]
+> We're archiving this proposals repo, and opening a new GitHub Discussions space for community feature requests at https://github.com/temporalio/temporal/discussions. We'd love to hear from you over there.
```
yuandrew: Should this be past tense? I worry future readers would assume the transition is still happening

yuandrew (approved):

README.md
karasowles: good point I agree

